PR #2, opened 2020-05-26
Adding Cypress.io for Unit testing and Integration tests

pakb:
URL to state (and vice-versa) is functional for zoom level

We need some deployment to see how things are working, might be our next step?

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,13 +32,14 @@ npm install
 
 | command | what it does |
 |----|----|
-| `npm run serve` | Compiles and hot-reloads for development. Will serve the project under `http://localhost:8080` (or the next available port if `8080` is already used, see console output) |
+| `npm run serve` | Compiles and hot-reloads for development. Will serve the project under `http://localhost:8080` (or the next available port if `8080` is already used, see console output). You can change port number by using env variable `PORT` (for example `PORT=9999 npm run serve`) |
 | `npm run build` | Compiles and minifies for production |
 | `npm run lint` | Lints and fixes files | 
-| `npm run test:unit` | Runs unit tests from cypress (equivalent to `npm run cypress:run`) |
+| `npm run test:unit` | Runs unit tests from cypress (equivalent to `npm run cypress:run`). |
 | `npm run cypress:open` | Opens up the cypress app that lets you run tests with Chrome (or Firefox, but support is still in beta) |
 | `npm run cypress:run` | Runs all cypress tests headless, outputs results in the console |
 
+All script commands starting a webserver or using one (`serve` and all things related to cypress) will determine port to use by looking env variable `PORT`. If not present, will fallback to default port `8080`.
 
 ### Customize configuration
 See [Configuration Reference](https://cli.vuejs.org/config/).
--- a/README.md
+++ b/README.md
@@ -15,31 +15,26 @@ This module can then be imported in the store module (see below)
  As there can be only one instance of a Vuex's store per app, the store module is there for that. It as the responsibility to instantiate Vuex, and add any module related state data to the store.
 See [its README.md](src/modules/store/README.md) for more details.
 
+### Testing
+
+Unit and integration testing is done with Cypress.io. All things related to tests are in the `/cypress` folder, and more specifically in the folder `/cypress/integration`. See [TESTING.md](cypress/TESTING.md) for more documentation on testing in this project.
+
 ## Project setup
 ```
 npm install
 ```
 
-### Compiles and hot-reloads for development
-```
-npm run serve
-```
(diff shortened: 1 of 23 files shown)


README.md
```diff
@@ -15,31 +15,26 @@ This module can then be imported in the store module (see below)
...
-```
-npm run build
-```
+| command | what it does |
+|----|----|
+| `npm run serve` | Compiles and hot-reloads for development. Will serve the project under `http://localhost:8080` (or the next available port if `8080` is already used, see console output) |
+| `npm run build` | Compiles and minifies for production |
+| `npm run lint` | Lints and fixes files | 
+| `npm run test:unit` | Runs unit tests from cypress (equivalent to `npm run cypress:run`) |
+| `npm run cypress:open` | Opens up the cypress app that lets you run tests with Chrome |
```
boecklic, line 35: ... (FF in beta mode)

cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:8080",
```
boecklic, line 2: if `npm run serve` has to use a port above `8080` (bc. `8080` is already used), this fails... is it possible to somehow link cypress to the test server to automatically find the correct port?

src/modules/store/plugins/vuex-url-search-params.js
```diff
@@ -0,0 +1,225 @@
+import _ from 'lodash'
+
+import { isNumber } from "@/numberUtils";
+
+// Persist and rehydrate the `Vuex` store via url search params (window.location.search)
```
boecklic, line 5: a more extensive comment/doc about how this persistance / rehydration works (conceptaully) would be helpful (since it's a central part of the app).

cypress.json
pakb: we could force the port to `8080` when we launch the server for testing, and fail if the port is already used

pakb:
> choosing vue/vuex would actually be one as well, the first one...

wouldn't https://github.com/geoadmin/doc-guidelines/blob/master/adr/2020_04_06_next_gen_viewer.md be right that? I could make a link to it.
I'll definitely add something about Cypress, and also insert some lines about Selenium and such in the `md` file for Cypress

boecklic:
> > choosing vue/vuex would actually be one as well, the first one...
> 
> wouldn't https://github.com/geoadmin/doc-guidelines/blob/master/adr/2020_04_06_next_gen_viewer.md be right that? I could make a link to it.
True, link is good

cypress.json
boecklic: so a fully 'auto-port-selection' for both serving and testing is not possible?

package.json
```diff
@@ -6,7 +6,7 @@
     "serve": "vue-cli-service serve",
     "build": "vue-cli-service build",
     "test:unit": "cypress run",
-    "test:e2e": "start-server-and-test serve http://localhost:8080 cypress:run",
+    "test:headless": "start-server-and-test serve http://localhost:8080 cypress:run",
```
boecklic, line 9: Is it possible to read `:8080` port from `ENV`? like this it could easily be overwritten if `8080` is not available...

pakb: should be possible with `PORT` variable, for instance `PORT=9999 npm run test:headless`

pakb: but I wasn't able to do a full auto port pick. If `PORT` or default port are being used, tests will try to use what was given while `npm run serve` will choose a new port. This will fail tests most likely...

boecklic: Hmm, serving is done at the port specified with `PORT=9999`, but the testing still seems to wait for readyness at `8080`:
```

> web-mapviewer@0.1.0 test:headless /home/ltboc/src/web-mapviewer
> start-server-and-test serve http://localhost:8080 cypress:run

1: starting server using command "npm run serve"
and when url "http://localhost:8080" is responding with HTTP status code 200
running tests using command "npm run cypress:run"


> web-mapviewer@0.1.0 serve /home/ltboc/src/web-mapviewer
> vue-cli-service serve

 INFO  Starting development server...
98% after emitting CopyPlugin

 DONE  Compiled successfully in 3426ms                                                                                                                                                              4:18:07 PM


  App running at:
  - Local:   http://localhost:9999/ 
  - Network: http://192.168.1.128:9999/

  Note that the development build is not optimized.
  To create a production build, run npm run build.

^C
```

pakb: I runs test on port 9999 for me if I do `PORT=9999 npm run test:headless` but I haven't tried to set the env variable for the session globally, is that what you did?

boecklic: Nope, just like you... for the app it is using port `9999` (see "App running at", but at the top, it's saying "... and when url 'http://localhost:8080'" it's still mentioning `8080`)

PR #3, opened 2020-06-02
add github workflow config to automatically tag branches

boecklic:
add github workflow config to automatically tag branches according to semver, see https://github.com/marketplace/actions/github-tag-bump


pakb (approved):
doc on that added by https://github.com/geoadmin/web-mapviewer/pull/9

PR #4, opened 2020-06-11
Deploy with CI on successful builds

pakb:


pakb:
I was planning to deploy branches on int on demand (with a CLI call from the dev that wants it), and deploy automatically anything merge on `master` on int.
For now I have activated branch deploy on int, but for testing purposes. I will remove it before merging the PR.

pakb:
Something we just discovered while publishing the app on S3, we have to configure `publicPath` (VueCLI config) for each build, otherwise the app will try to get its JS and CSS from the root of the bucket, even though it was deployed as a branch...

boecklic:
> Something we just discovered while publishing the app on S3, we have to configure `publicPath` (VueCLI config) for each build, otherwise the app will try to get its JS and CSS from the root of the bucket, even though it was deployed as a branch...

Ok, I see, makes somehow sense...

README.md
```diff
@@ -41,5 +41,9 @@ npm install
 
 All script commands starting a webserver or using one (`serve` and all things related to cypress) will determine port to use by looking env variable `PORT`. If not present, will fallback to default port `8080`.
 
+### What about `package-lock.json` file?
+
+The CI uses this file to ensure it will not stubble upon a minor version of a library that breaks the app. So this file needs to be versioned, and kept up to date (each time a new library or version of a library is added to `package.json`, `npm install` will update `package-lock.json` accordingly).
```
boecklic, line 46: :+1: I was never sure what the best policy is here...

pakb:
Ok, I think I'm done with this CI, here's the current status :
- Pushes event on `master` triggers a deploy on the root of the int bucket with production ready app.
- Pushes event on `develop` triggers a deploy on the root of the dev bucket with unminified sources
- PR creation and update trigger a deploy of a subfolder on the dev bucket with the git branch name as folder name (we might want to think about a cleaning mechanism on branch delete)

I still need to check that the terraform modification I've made really trigger the CI when Pushes event are made on `develop` and `master`, but for that, I'll have to merge this PR 😉

procrastinatio:
@pakb Depending on how you did it, it will trigger two builds, one for the branch, the other for the PR.

pakb:
Build logs URL can take a little while (about a day) to have their signing working as stated on the lib repo issues

davidoesch:
One day for the logs? That is too long . We need to be able deploy within minutes — possible?

pakb:
only one build was triggered by the merge, and it was successfully deployed on the root of the dev bucket, yay 🎉

boecklic:
The question is if the logs itself are useful, or more reports as done e.g. for `mf-geoadmin`: https://eu-central-1.console.aws.amazon.com/codesuite/codebuild/974517877189/projects/mf-geoadmin3/build/mf-geoadmin3%3A3522b31d-cb7a-43d2-ae57-cf9a4aeba882/reports?region=eu-central-1&reports-meta=%7B%22f%22%3A%7B%7D%2C%22s%22%3A%7B%22property%22%3A%22status%22%7D%2C%22n%22%3A20%2C%22i%22%3A0%7D like this you get more or less detailed info about each test.... but I guess this is not possible?

PR #5, opened 2020-07-30
Unit test deploy script inputs

pakb:


pakb:
> 👍 testing also the deploy script!

Not all of it sadly, it was too cumbersome to mock up external modules (like `fileUtils.js`) as it's run through a bash call. If we want to thoroughly test this deploy script, we want to find a way to run it without requiring a `node` call so that we can stub and mock up `require` and `import` statements.

boecklic:
> Not all of it sadly, it was too cumbersome to mock up external modules (like `fileUtils.js`) as it's run through a bash call. If we want to thoroughly test this deploy script, we want to find a way to run it without requiring a `node` call so that we can stub and mock up `require` and `import` statements.

dumb question: how do you execute a JS script other than by using `node`?

pakb:
meaning, it's not simply inserted into Cypress environment (iFrame) and run from there, Cypress makes bash call to run the scripts.
Otherwise Cypress would have been fine stubbing anything with Sinon.
But I couldn't find a way to mock up part of this node script this way. Maybe we should not use Cypress for this but another test runner more fitted for Node, I wanted to "keep it simple" but that was awfully hard 😉

boecklic:
> meaning, it's not simply inserted into Cypress environment (iFrame) and run from there, Cypress makes bash call to run the scripts.
> Otherwise Cypress would have been fine stubbing anything with Sinon.
> But I couldn't find a way to mock up part of this node script this way. Maybe we should not use Cypress for this but another test runner more fitted for Node, I wanted to "keep it simple" but that was awfully hard wink

Keeping things simple is very hard!! That's why people then choose the easy way, which over time turns to be the complicated one!

Anyhow, I'd say with live with the already good solution for now, if we figure that a better testing coverage is necessary, we can still adapt.

PR #7, opened 2020-08-21
Adding i18n managment

pakb:


README.md
```diff
@@ -41,6 +41,7 @@ npm install
 | `npm run cypress:open` | Opens up the cypress app that lets you run tests with Chrome (or Firefox, but support is still in beta) |
 | `npm run cypress:run` | Runs all cypress tests headless, outputs results in the console |
 | `npm run deploy:#target#` | Target can be `dev`, `int` or `prod`. Build the app and deploys it on the target S3 Bucket. You need to have an AWS profile that has writing rights on the bucket. If you need to use another profile than the default one, use `AWS_PROFILE=another_profile_name npm run deploy:#target#`. For more information on what the deploy script does, [see below](#what-does-the-deploy-script-do). |
+| `npm run update:translations` | Update translation files according to our Google Spreadsheet. Requires a valid Google API Key as env variable `GOOGLE_API_KEY`. So Should be run like `GOOGLE_API_KEY=xxx npm run update:translations`. A key can be found in our `gopass`. |
```
ltshb, line 44: Would be good to use `summon` here in order to automate the API key read from gopass:
```
summon -p gopass npm run update:translation
```
Which should work (needs to be tested)

ltshb (changes requested):
Looks good just a small improvement about the `npm run update:translation`

README.md
pakb: `summon` is not installed by default on many device (mine included) and if I run a `summon -p gopass npm run update:translation` it fails because I have to specify where gopass is.

I used to have called `gopass` straight with `gopass show ...` in the script, but this requires most of the time the password to use the SSL certificate, and this interaction was lost by NPM resulting in a fail.

pakb: I've added a section in README.md about how to use `summon` while updating translations

ltshb (approved):
:+1:

PR #8, opened 2020-08-26
Little review on the current documentation

pakb:


README.md
```diff
@@ -96,6 +96,7 @@ Translations can then be updated with
 ```bash
 summon -p gopass npm run update:translations
 ```
+The file `secrets.yml` will tell `gopass` which keys to get from the store.
```
ltshb, line 99: Actually `secrets.yml` is used by summon not gopass. It does the mapping between the gopass key and the environment variable that should be exported

ltshb (changes requested):
Looks good just a small changes about `secrets.yml` and some suggestions.

For new comer (like me) it would be good to have a link to `Vue` and `Vuex` in the `Project structure`  chapter.

In the `Project structure` a Folder tree image of the structure would help to understand it.

In chapter `Tooling for translation update` I would specify the gopass store name: `infra-gopass-bgdi`

pakb:
Should be good, I've added what you suggested, thanks!

ltshb (approved):
:+1:

PR #12, opened 2020-09-14
Add header and search bar

pakb:


davidoesch:
Can’t wait to see this  - thx pascal. Nice 👍 start into the week

davidoesch (commented):
Any link to test ?

pakb:
http://web-mapviewer.dev.bgdi.ch/ltbtp_add_header_and_search/index.html

it's still WIP but here is where it's deployed anytime I commit something

davidoesch:
Thx

Regarding bg selector : we need to stick to the existing system

davidoesch:
For the 
- orientation icon we need to keep when rotated 
 — fullscreen: I like . 

- header height is bigger than current application- try to keep it smaller

pakb:
> Regarding bg selector : we need to stick to the existing system
I can do a bottom button like the current instalment, and then hide it (on the side of the screen) when we move the map so that we clear up the space. How does that sound like?

> * orientation icon we need to keep when rotated
Do yo mean the compass button (localisation and heading)?

> * header height is bigger than current application- try to keep it smaller
I'll try to shrink it as much as possible, but I wouldn't want to sacrifice the search bar's input height too much. IMHO it is better to have a nice input height so that it is easier to click it with "big clumsy fingers" on mobile

davidoesch:
> > Regarding bg selector : we need to stick to the existing system
> 
> I can do a bottom button like the current instalment, and then hide it (on the side of the screen) when we move the map so that we clear up the space. How does that sound like?
> 
Ok
> > * orientation icon we need to keep when rotated
> 
> Do yo mean the compass button (localisation and heading)?
> 
Yes and the geolocation button at the same time
> > * header height is bigger than current application- try to keep it smaller
> 
> I'll try to shrink it as much as possible, but I wouldn't want to sacrifice the search bar's input height too much. IMHO it is better to have a nice input height so that it is easier to click it with "big clumsy fingers" on mobile

Current height in my-geoadmin4 works. No complains so far. Save space . Keep as well the same font and size

davidoesch:
Search 
- results are different
- font size and spacing are different
Try to keep it the same - as well the css


5AD13428-399C-4148-A4EB-199280D01398

davidoesch:
https://github.com/geoadmin/web-mapviewer/pull/12#issuecomment-705675884 not implemented

pakb:
> as well the css

So you expect me to copy cat the whole CSS without making any change, even in mobile? I thought we agreed last month that the current UX in mobile is pretty poor and needs some changes. Please clarify your position.

> * magnifying glass (which we can later on have as toggle for routing -this is why we had this icon- it’s actually the only icon we use)

What do you mean by routing?

pakb:
@davidoesch Could you give me your opinion on how the menu bar looks like with a desktop? Should I add the links on the right past big screen? (>1200px)

davidoesch:
Full screen on/off: exactly like this . Perfect.
Me happy- you as well  I hope

pakb:
Ok, let's end the commit train there for now, let's review what has been done so far, merge it if good, and then open smaller PR to address issues that are still present with this big "draft" (and each issue addressed separately)

boecklic (approved):
Excellent work! I really like the clean Architecture of the app, with clear separation of state, actions and presentation. This will be a solid basis for any future stuff that's put on top!
To keep the dev speed and in accordance with @davidoesch this is good to merge. Polishing will be done at a later stage.

PR #17, opened 2020-10-30
Adding background selector buttons at the bottom

pakb:
https://jira.swisstopo.ch/browse/BGDIINF_SB-1085

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1085_switch_bg_buttons/index.html)


pakb:
@davidoesch what's your opinion on this implementation? I tried to stick to geoadmin's style this time 😉

davidoesch:
I like, 
what is missing
- [ ] void layer (as you mentioned the VT layer will be added later as soon as we have it in the backend)
- [ ] smaller border for the available layers (to discriminate from the button itself)
- [ ] red highlight of the current active layer

image

pakb:
@davidoesch when using the void layer, should the background of the map have some kind of grid or are we letting the background be totally empty?

davidoesch:
- [ ] Size Button too big on an iphone 7 image below left , the existing one worked perfectly and was discrete enough(image below right
- [ ] color button border: use the same black as on the right  in the current version
- [ ] button image: always use the aerial image to fill button as in the current version, independet wht is active. reason: good contrast of button
- [ ] order of bg selection: use the existing order (swissimage, color,bw,void) as on the right

image

pakb:
> button image: always use the aerial image to fill button as in the current version, independet wht is active. reason: good contrast of button

I don't understand this comment, could you elaborate on what you mean?

>  order of bg selection: use the existing order (swissimage, color,bw,void) as on the right

I can only choose where the void layer is, the others are ordered as they are given by the layers config (I want to avoid any specific code if possible with that, so that it is totally configurable from the backend)

davidoesch:
Button image:
It’s the aerial image in the Main  button. The main button is always the same . Not dependent of the current background state.

Order:
Then this has to be configured in the backend. Can you do so or tell me where to open the corresponding ticket?

pakb:
rebased on https://github.com/geoadmin/web-mapviewer/pull/19
will need to be reviewed merge first before doing this one

pakb:
> color button border: use the same black as on the right in the current version

Will be addressed automatically later when we chose our color palet. It currently uses `$black` variable from Bootstrap, as soon as we theme it to our liking, the button will also change.

Otherwise all feedback have been taken into account, @davidoesch could you give it another look?

davidoesch:
> Otherwise all feedback have been taken into account, @davidoesch could you give it another look?

> I've also hidden the BG button when the users goes fullscreen (with a quick touch on mobile), what's your opinion on that?
I like, go for it


- [ ] The image in the icon of color and bg are not that crisp. They are blurry-  . In the mf-geoadmin3 , they yellow road is better visible (so a zoomlevel more in the button , and feature is more clear)

A detail: in the currentversion when switiching on desktop- there is a mouseover effect. IMHO not needed, a NTH

pakb:
@davidoesch I've reduced the image used as background to fit the button size, so that it is a bit more crisp. I've also added a little animation on click/touch.
Are we good to go?

davidoesch:
Background: still not that crispy: we can ask @pauloamado to have the same quality as before. I test it  on android iphone and desktop--still kinda blury

davidoesch:
the rest is perfect ( even if the button occupy now more space on the screen)

boecklic (approved):
Looks good. If I understand correctly you kicked out cypress for the e2e tests?

pakb:
no we still use it, but through the Vue integration provided by `@vue/cli-plugin-e2e-cypress`

PR #18, opened 2020-10-30
[Proof of concept] Adding swipeable tooltip tray

pakb:
This is an experiment to see if a swipeable UI element, as the one that shows the tooltip on the new Swisstopo app, is a realistic approach on a web app.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1325_swipable_element/index.html)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -6,6 +6,7 @@
       <SearchModule />
     </MenuModule>
     <ToolboxModule />
+    <TooltipModule />
     <I18nModule />
   </div>
 </template>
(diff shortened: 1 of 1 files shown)


davidoesch:
with mf-geoadmin3 we had to takle the issue, that on small screens the open tooltip covered the map and was only hard to move

with your approached it can be solved. BUT: those tooltips are long so we need to be able to scroll in the swiped element itself -- does this make sense?

what is missing on mobile

- [ ] close button X on the right to close ticket
- [ ] tap function to slide the area up ( two states , as in swisstopo app)

PR #21, opened 2020-11-18
Adding geolocation button

pakb:
[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1437_geolocation/index.html)

### DIFF
--- a/src/api/layers.api.js
+++ b/src/api/layers.api.js
@@ -105,6 +105,13 @@ export class WMTSLayer extends Layer {
         return `${this.baseURL}1.0.0/${this.id}/default/${timestamp}/3857/{z}/{x}/{y}.${this.format}`;
     }
 
+    /**
+     * Resolve the {x-y} notation used in WMTS URLs and outputs all possible URLs
+     *
+     * Example : `"https://wmts{1-3}.geo.admin.ch"` will outputs `[ "https://wmts1.geo.admin.ch", "https://wmts3.geo.admin.ch", "https://wmts3.geo.admin.ch" ]`
+     *
+     * @returns {Array<String>} all possible backend URLs for this layer
+     */
     getURLs() {
         const mainURL = this.getURL();
         const urls = [];
--- a/src/api/layers.api.js
+++ b/src/api/layers.api.js
@@ -44,17 +44,23 @@ export class Layer {
      * @param {String} id
      * @param {Number} opacity
      * @param {Boolean} isBackground
+     * @param {String} baseURL
      */
     constructor(name= '',
                 type= null,
                 id= '',
                 opacity = 1.0,
-                isBackground = false) {
+                isBackground = false,
+                baseURL = null) {
         this.name = name;
         this.type = type;
         this.id = id;
         this.opacity = opacity;
         this.isBackground = isBackground;
+        this.baseURL = baseURL;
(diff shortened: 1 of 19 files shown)


davidoesch:
shall I give feedback? ready?

pakb:
The first geolocation and error handling is implemented, but I can't get the "tracking" to work yet.
You can already give a look at the UI element and how it react to not have the Geolocation authorization and such, tracking will come later today or tomorrow.

davidoesch:
- [ ] when hitting the gelocation button: zoom level is z=14.5 , expected result 15.5 ( corresponds to mf-geoadmin3 and gmaps app: user actually gets at this scale where he stands) or is there another reason? Even if there is a weak gsp signal (huge red circle ) we stay at this z

davidoesch:
otherwise I tested so far on iphone 7 with iOS 14.2 - : and it looks perfect. (not tested on other devices etc like desktop not on android etc.)

thx Patrick!

pakb:
oh right, I miss-read the zoom level of the current viewer (zoom LV95 : 8) which correspond to the 1:25'000 map, that's fixed. Tracking should now be operational, I'll wait for the rain to settle to try it out outside later today

src/modules/store/plugins/geolocation-management.plugin.js
```diff
@@ -0,0 +1,75 @@
...
+        if (mutation.type === 'setGeolocationActive') {
+            if (state.geolocation.active) {
+                navigator.geolocation.getCurrentPosition(
+                    position => {
+                        // if geoloc was previously denied, we clear the flag
+                        if (state.geolocation.denied) {
+                            store.dispatch('setGeolocationDenied', false);
+                        }
+                        // we center the view on the position of the user
+                        store.dispatch('setCenter', readPositionEpsg3857(position));
```
pakb, line 55: that is unnecessary, I should remove it (it's already done by `handlePositionAndDispatchToStore`

davidoesch:
and: I like your idea to have the button with a transparency

src/api/layers.api.js
```diff
@@ -94,19 +102,39 @@ export class WMTSLayer extends Layer {
         if (this.timeSeriesConfig) {
             timestamp = this.timeSeriesConfig.currentTimestamp
         }
-        return `${WMTS_BASE_URL}1.0.0/${this.id}/default/${timestamp}/3857/{z}/{x}/{y}.${this.format}`;
+        return `${this.baseURL}1.0.0/${this.id}/default/${timestamp}/3857/{z}/{x}/{y}.${this.format}`;
+    }
+
+    getURLs() {
```
boecklic, line 108: maybe add a comment about what this function does and why it's used.

src/modules/store/plugins/geolocation-management.plugin.js
boecklic: So if I understand correctly, geolocation is relying on the same handling as when position is entered via URL params?

boecklic (approved):

davidoesch:
Maybe this helps:  There is an url Parameter called geolocation=true which activates geolocation

pakb:
@boecklic yes, to center the view on the user's geolocation we go through the same mechanism as if we add written a new location in the URL (all those `set where the map look` functions goes through the same endpoint in the store in the end, that's the goal)

PR #24, opened 2020-12-02
BGDIINF_SB-1497 update github tagging action

boecklic:
updated github action to  add semver tags to the latest version (related to https://github.com/geoadmin/doc-guidelines/pull/65)


pakb (approved):
I will add something about that in the doc with my next round of doc cleanup

PR #23, opened 2020-12-02
Adding support for all layers types from the BOD

pakb:
- [x] GeoJSON
- [x] simple WMTS
- [x] single tile WMS
- [x] support for WMS `gutter` option (tiled WMS)
- [x] Aggregate Layer

@ltclm @rebert if you know any kind of pattern I've not listed above, please add them so that I may know I've missed some stuff 😉 

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1460_support_for_all_layer_types/index.html)

PR #25, opened 2020-12-18
Add geolocation url param

pakb:
adding `geolocation=true` to the URL will activate geolocation right after app has started

this PR is based on the code of https://github.com/geoadmin/web-mapviewer/pull/23 so https://github.com/geoadmin/web-mapviewer/pull/23 needs to be merged first

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1461_geolocation_url_param/index.html?geolocation=true)


ltshb (commented):
lgtm:

General comments about frontend development (might require new Jira ticket for those): 

- it seems to me that we don't have code guideline for frontend javascript ? or at least I haven't found them.
- Do you use any formatting tool ? From the last two PRs plus this one I've noticed that sometimes you have 2 spaces indentation and sometimes 4 spaces ! I think we should have the indentation as well as the whole code style defined and then stick to it (like we do in backend python code). A formatting tool for javascript would be also a good idea. It seems to me also that the line length is not defined ?
- Do you use eslint or any linting tool ?

@boecklic what is your opinion on these comments ?

pakb:
That's a fair point, I'm preparing a PR with https://prettier.io/ (and a plugin made for `.vue` files). That will enforce some consistency throughout the code.
Although I will wait for this PR to be merged, then use `develop` as a base for that, because most files will be altered in some way (don't want to spend a day managing merge conflict... 😅 )

boecklic:
@pakb @ltshb full support for covering JS with coding guidelines and linter/formatter as well (and to do it soon), specially given the external devs joining in march.

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -9,7 +9,7 @@ import ImageWMS from "ol/source/ImageWMS"
 import TileWMS from "ol/source/TileWMS"
 import TileGrid from "ol/tilegrid/TileGrid"
 import addLayerToMapMixin from "./utils/addLayerToMap-mixins";
-import {TILEGRID_ORIGIN, TILEGRID_RESOLUTIONS, WMS_TILE_SIZE} from "../../../../config";
+import {TILEGRID_ORIGIN, TILEGRID_RESOLUTIONS, WMS_TILE_SIZE} from "@/config";
```
boecklic, line 12: :+1: for bringing the `@` notation back

boecklic (approved):

PR #27, opened 2021-01-14
Adding prettier.io to lint and reformat Vue code

pakb:
With support for `.vue` files
it is integrated to `es-lint` so it is run with `npm run lint` and is run each time the app is built


boecklic (approved):
:+1:

PR #28, opened 2021-01-28
Adding tooltip support

pakb:
[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1428_tooltip/index.html)

This does not cover the UI totally, but focuses on features under the hood. The swipeable element we've been talking about will be added in another PR later


src/api/features.api.js
```diff
@@ -60,7 +60,7 @@ export const identify = (layer, coordinate, mapExtent, screenWidth, screenHeight
     }
     axios
       .get(
-        `${API_BASE_URL}/rest/services/${layer.getTopicForIdentifyAndTooltipRequests()}/MapServer/identify`,
+        `${API_BASE_URL}rest/services/${layer.getTopicForIdentifyAndTooltipRequests()}/MapServer/identify`,
```
boecklic, line 63: just a general note (not related to this exact change): you might want to explicitly name the API you consider, so `CHSDI_BASE_URL` (or following the pattern you use below for what3words `CHSDI_API_BASE_URL`), this will make transition easier in the future when parts slowly move away from chsdi (as already happened e.g. with shortlink etc.)

boecklic (approved):
Nice! Made one comment

src/api/features.api.js
pakb: that's one way of looking at it, I was more thinking of "here is `https://api3.geo.admin.ch`, I do not care what is behind with routing, but that's my API base URL".
In the end, they will all be served under one roof, isn't it?
IMO, the least the front is aware of internal backend routing, the better

boecklic: in an ideal world yes ;-) I'm afraid we won't get completely away with this, but we can also adapt once we know more.

src/api/features.api.js
```diff
@@ -0,0 +1,185 @@
+import { API_BASE_URL } from '@/config'
+import axios from 'axios'
+
+/**
+ * Describe a feature from the backend (see {@link getFeature}) below
+ */
+export class Feature {
+  /**
+   * @param {WMSLayer|WMTSLayer|GeoJsonLayer|AggregateLayer} layer the layer in which this feature belongs
```
procrastinatio, line 9: Tooltip for WMS, WMTS and GeoJSON is OK. For aggregated layer (e.g. _ch.swisstopo-vd.geometa-gemeinde_) is it the fall?

src/api/features.api.js
```diff
@@ -0,0 +1,185 @@
...
+/**
+ * Loads a feature metadata and tooltip content from this two endpoint of the backend
+ *  - http://api3.geo.admin.ch/services/sdiservices.html#identify-features
+ *  - http://api3.geo.admin.ch/services/sdiservices.html#htmlpopup-resource
+ * @param {WMSLayer|WMTSLayer|GeoJsonLayer|AggregateLayer} layer the layer from which the feature is part of
+ * @param {String|Number} featureID the feature ID in the BGDI
+ * @param {String} lang the language for the HTML popup
+ * @return {Promise<Feature>}
+ */
+const getFeature = (layer, featureID, lang = 'en') => {
```
procrastinatio, line 111: Why '_en_'?

src/api/features.api.js
```diff
@@ -0,0 +1,185 @@
...
+      reject('Needs a valid feature ID')
+    }
+    // combining the two requests in one promise
+    const topic = layer.getTopicForIdentifyAndTooltipRequests()
+    const featureUrl = `${API_BASE_URL}rest/services/${topic}/MapServer/${layer.id}/${featureID}`
+    axios
+      .all([
+        axios.get(featureUrl, {
+          params: {
+            sr: 3857,
```
procrastinatio, line 127: Simple question, are all backend _EPSG:3857_? In _test-map.geo.admin.ch_ some reprojection were done in the client

src/api/features.api.js
```diff
@@ -0,0 +1,185 @@
...
+        let featureCoordinate = []
+        // if GeoJSON type is Point, we grab the coordinates
+        if (featureGeoJSONGeometry.type === 'Point') {
+          featureCoordinate = featureGeoJSONGeometry.coordinates
+        } else if (
+          featureGeoJSONGeometry.type === 'MultiPoint' &&
+          featureGeoJSONGeometry.coordinates.length === 1
+        ) {
+          // or if the GeoJSON type is MultiPoint, but there's only one point in the array, we grab it
+          featureCoordinate = featureGeoJSONGeometry.coordinates[0]
```
procrastinatio, line 159: Yuks!

procrastinatio (commented):
Very impressive. For all layers but the _aggregated_ one, it works perfectly.

src/api/features.api.js
pakb: that's what the backend gives us 😢

src/api/features.api.js
pakb: I have not tested all backend endpoint yet, but as for now everything is working with EPSG:3857 (with some modifications I made too, like this S3 tile identify thing)

src/api/features.api.js
pakb: Would you prefer "de"? It's a fallback, it should not happened as the code gives the current lang as a param

src/api/features.api.js
pakb: Ho right, I have not thought about this use case.

PR #29, opened 2021-02-05
adding a mouse position

procrastinatio:
Display mouse position in status bar with a projection switch

See https://jira.swisstopo.ch/browse/BGDIINF_SB-1436

[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1436_display_coordinates_in_statusbar/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersMousePosition.vue
+++ b/src/modules/map/components/openlayers/OpenLayersMousePosition.vue
@@ -1,14 +1,16 @@
 <template>
     <div id="mouse-position" ref="mousePosition">
-        <button v-for="proj in availableProjections" :key="proj" @click="changeProjection(proj)">
-            {{ proj }}
-        </button>
+        <select @change="changeProjection($event.target.value)">
+            <option v-for="proj in availableProjections" :key="proj" :value="proj">
+                {{ proj }}
+            </option>
+        </select>
     </div>
 </template>
 
 <script>
 import MousePosition from 'ol/control/MousePosition'
-import { createStringXY } from 'ol/coordinate'
+import { createStringXY, toStringXY, toStringHDMS } from 'ol/coordinate'
 import { get as getProjection } from 'ol/proj'
 
 export default {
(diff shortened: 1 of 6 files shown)


src/modules/map/components/openlayers/OpenLayersMousePosition.vue
```diff
@@ -0,0 +1,141 @@
...
+import { forward as LLtoMGRS, LLtoUTM } from '@/utils/militaryGridProjection'
+import { formatThousand } from '@/utils/numberUtils'
+
+export default {
+    inject: ['getMap'],
+    data: function () {
+        return {
+            mousePositionControl: null,
+            projection: 'EPSG:3857',
+            availableProjections: ['EPSG:3857', 'EPSG:2056', 'EPSG:21781', 'EPSG:4326'],
```
pakb, line 24: I don't think this is ever used?!

tests/e2e/specs/mouseposition.spec.js
```diff
@@ -0,0 +1,45 @@
+/// <reference types="cypress" />
+
+const mousePositionSelector = 'div#mouse-position > select'
+const olSelector = '.ol-viewport'
+const olMousePosition = '.ol-mouse-position'
```
pakb, line 5: For CSS selector within Cypress, please check this link : https://docs.cypress.io/guides/references/best-practices#Selecting-Elements
I would prefer that we use dedicated custom HTML attribute, like `data-cy="mouse-position"` so that it is not linked to something that might change later on when we redo the UI (it's the next step when we will have all features done, we will do a cycle or two of UI/UX refinement)

pakb (commented):
Welcome to the project ! 😄 
Here some minor comments/.changes on your PR. Great overall!

src/modules/map/components/openlayers/OpenLayersMousePosition.vue
procrastinatio: No

tests/e2e/specs/mouseposition.spec.js
procrastinatio: OK, for my code. But how do add `data-cy` to OL bits? You have to rely on something already existing, like `class` or `id`, do you ?

procrastinatio:
@pakb  OK. Corrected, squashed and reworded.

pakb (approved):

PR #31, opened 2021-02-15
Add layer management in menu

pakb:
switching to NPM 7 package-lock file format
replacing 'hand made' font awesome integration with lib

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1661/index.html)


ltshb (approved):
From what I can judge it looks good to me.

davidoesch:
hmm 
- [ ] we have a shift: 
https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1661/index.html#/map?lat=46.816337&lon=8.208096&z=6.977&layers=ch.kantone.cadastralwebmap-farbe 

image

- [ ] Move via drag & drop of layers would 

- [ ] info button is missing in the  layer  (maybe this is out scope for this PR?)

- [ ] copy button is missing in the  layer  eg in https://s.geo.admin.ch/7374e5d65d (maybe this is out scope for this PR?)

procrastinatio:
@davidoesch  Regarding Cadastralwebmap, it looks like a bug in OL with client-zoom at a higher scale.

procrastinatio:
@davidoesch  In fact, the WMTS/WMS/proxy should be configured to allow layers to display at zoom 0 in EPSG:3857...by setting the extent accordingly in EPSG:21781 or 2056 (we are global now)

davidoesch:
"WMTS/WMS/proxy should be configured " can you open the corresponding tikcet  @procrastinatio

pakb:
I don't remember if the info button is the one I heard "Oh forget this one, no one uses it" during one of our call or not, so I didn't do it this time, but if it is required to have it I'll add it through another PR.

Drag&drop is IMHO a nice to have feature that can be worked on later on. It's not listed in the feature list, so I assume this is a non-MVP feature (but a "future" one), I'll add it to this list.

A copy button? that's not listed anywhere in the feature list, it does not seem to show for every layers, what's the catch? What does it do? Simply duplicate the layer (that does not seem to be very useful or advised...)?

boecklic:
@davidoesch @pakb @procrastinatio a **merged PR** is not the place to discuss this kind of things, please. Pls have a quick call to discuss and then create tickets (either Bug or Feature).

PR #32, opened 2021-02-15
Lint JSDocs

pakb:
enforce line braking in JS Docs too
is based on https://github.com/geoadmin/web-mapviewer/pull/31 (so it will have to wait for #31 to be merged first)


ltshb (approved):
:+1:

PR #36, opened 2021-03-03
Transforming header red line into loading bar

pakb:
[Test link](https://web-mapviewer.dev.bgdi.ch/feature_BGDIINF_SB-1416/index.html)


pakb:
need to fix some e2e tests with this new 'app loading' approach

tests/e2e/fixtures/layers.fixture.json
```diff
@@ -0,0 +1,108 @@
...
+    "timestamps": [
+      "current"
+    ],
+    "hasLegend": true,
+    "label": "Public transport stops",
+    "type": "wmts",
+    "serverLayerName": "ch.bav.haltestellen-oev"
+  },
+  "ch.swisstopo.pixelkarte-farbe": {
+    "co`nfig3d": "ch.swisstopo.swisstlm3d-karte-farbe_3d",
```
boecklic, line 47: the '`' seems to be a typo?

tests/e2e/fixtures/layers.fixture.json
```diff
@@ -0,0 +1,108 @@
+{
+  "ch.bfe.windenergie-geschwindigkeit_h50": {
```
boecklic, line 2: I guess the layer selection here is not random, it might be helpful to add a comment why you chose these particular layers (cause they are probably typical representatives of a certain kind of layer?)

tests/e2e/support/commands.js
```diff
@@ -4,20 +4,38 @@
 // https://on.cypress.io/custom-commands
 // ***********************************************
 
+const addLayerFixtureAndIntercept = () => {
+    cy.intercept('**/rest/services/all/MapServer/layersConfig**', {
```
boecklic, line 8: :+1: for the mock, nice

boecklic (approved):

tests/e2e/fixtures/layers.fixture.json
pakb: that was to minimize changes in my tests, but I could surely embrace the "I make my layer myself" in the test, and rename them accordingly. It will make more sense

tests/e2e/fixtures/layers.fixture.json
pakb: it is, nice catch 👍

src/modules/menu/MenuModule.vue
```diff
@@ -1,46 +1,54 @@
...
-    .menu-tray {
-        position: fixed;
-        top: $headerHeight;
-        right: 0;
-        background: $white;
-        width: 95%;
-        max-width: 40rem;
+    .header-content {
+        display: flex;
+        height: $header-height;
```
ltshb, line 37: Just for my understanding, why changing the naming convention here ? is `$header-height` a javascript variable, Vue variable ? For variable I would prefer camelCase or snake_case as kebab case

src/modules/menu/components/header/HeaderLoadingBar.vue
```diff
@@ -0,0 +1,42 @@
...
+</template>
+
+<script>
+export default {}
+</script>
+
+<style lang="scss">
+@import 'node_modules/bootstrap/scss/bootstrap';
+@import 'src/scss/variables';
+$loading_element_width: 40rem;
```
ltshb, line 12: Why here using snake case ? and not kebab case like the others; eg. `$header-height` ?

ltshb (commented):
Why not leaving the red bar after loading ? I know that on the original viewer there is not red bar at the bottom of the header, but I find it would look cleaner to have always the red bar (like in other Federal headers) and having it transformed as load bar when loading. We have now the opportunity to add small improvements :-)

src/modules/menu/MenuModule.vue
pakb: that's a SCSS variable (Sass), and as everything in CSS is with this style of notation I've changed it to that. You can find all our variables in `src/scss/variables.scss`

src/modules/menu/components/header/HeaderLoadingBar.vue
pakb: my bad, I'll switch to `the-correct-notation-for-css`

tests/e2e/fixtures/layers.fixture.json
pakb: so now we have really fake layers with fake IDs for test, this way there's no confusion

pakb:
@ltshb that's a good question we could ask to @davidoesch 
I added this red line because it is part of the guidelines, but I also understand that it takes some "real estate" for not much

procrastinatio (approved):
Nothing to say on the PR _per se_.

But I would do sperate PR for libraries/modules updates, maybe something like _greenkeepers_ (or equivalent)?

pakb:
`package-lock.json` updates are an inherent tradeoff we have to live with, so that the CI can build with the exact same libs as the last `npm install` had.
In reviews, you can ignore `package-lock.json` altogether, only check if something changes in `package.json`

PR #39, opened 2021-03-12
Move loading bar at the top

pakb:
Also make it smaller, and with a new loading animation that goes ping pong red/transparent

[Test link](https://web-mapviewer.dev.bgdi.ch/fix_BGDIINF_SB-1416/index.html)

### DIFF
--- a/src/modules/menu/MenuModule.vue
+++ b/src/modules/menu/MenuModule.vue
@@ -1,6 +1,7 @@
 <template>
     <transition name="slide-up">
-        <div v-show="showHeader" class="header" :class="{ 'extra-height': showLoadingBar }">
+        <div v-show="showHeader" class="header">
+            <HeaderLoadingBar v-if="showLoadingBar" />
             <div class="header-content align-items-center p-1 flex-fill">
                 <SwissFlag class="swiss-flag ml-1 mr-2" />
                 <HeaderSwissConfederationText class="d-none d-sm-block" />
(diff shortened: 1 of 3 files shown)


boecklic:
Is there sth. like a minimum duration (like 0.5s), so that user doesn't have impression of 'flickering' (or even that sth is wrong)?

pakb:
The animation takes a little while to kick in, so we could see there's a grace period of 0.5 sec before you see this happen.
I could also add a little opacity easing to make it less aggressive

davidoesch:
Reference here is google drive. We copied the behavior

Open a google docs/ spreadsheet .
And copy post large amount of data
Loading bar appears 

Also: the mf-Geoadmin3 settings: they are ok and great: not one complaint but several kudos for the feature

so: stick if possible to the existing duration

pakb:
I copied the current animation, I just edited a bit the end of it, so that it would go red the whole screen and then go transparent in the same fashion, not just disappear and start again.
But time wise, it's the same as before

ltshb (commented):
I'm not a frontend expert, but would not make sense to put the `HeaderLoadingBar` component on top of the `header` class div in https://github.com/geoadmin/web-mapviewer/blob/8d92bd7f8bb3d1e6c810b64d6b06d659a723a8c5/src/modules/menu/MenuModule.vue#L13 ? This would make the code easier to understand when looking at `MenuModule` component.

pakb:
> I'm not a frontend expert, but would not make sense to put the `HeaderLoadingBar` component on top of the `header` class div 

Totally! It was there because back then it was at the bottom of the menu. But now that it is at the top I should move it accordingly

ltshb (approved):
:+1:

PR #40, opened 2021-03-16
BGDIINF_SB-1675 add ADR about URL param structure

boecklic:
This PR adds the decision about the new format for URL parameters.

### DIFF
--- a/adr/2021_03_16_url_param_structure.md
+++ b/adr/2021_03_16_url_param_structure.md
@@ -1,0 +1,100 @@
+# URL parameter structure
+
+> Status: accepted
+
+> Date: 16.03.2021
+
+## Context
+
+The mapviewer application is configured with several URL parameters. The current format for the layer configuration looks as follows (example for topic "Snow"):
+
+```
+...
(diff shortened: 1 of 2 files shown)


adr/2021_03_16_url_param_structure.md
```diff
@@ -0,0 +1,100 @@
...
+ch.bav.haltestellen-oev,f
+...
+```
+
+### layerID
+
+The `layerID` can be one of the following
+
+- an id from the catalog (e.g. `ch.swisstopo.schneeschuhwandern`)
+- an id from the catalog, parametrized with e.g. time (and possibly other parameters in the future): e.g. `ch.swisstopo.zeitreihen@time=18641231@height=100m` (note: this allows for further parametrization in the future like e.g. height)
```
pakb, line 80: if we enforce ISO compliance, maybe the example in the ADR should also use some ISO format (with `-`)

hansmannj (approved):
LGTM and represents what we have discussed :+1:

ltshb (approved):
:+1:

PR #41, opened 2021-03-17
Small fixes found while reading the code

fredj:


pakb:
nice catch, thanks! should I wait a little bit before merging this? (in case you find something else)

fredj:
@pakb I've pushed a last commit to remove a duplicated import, you can merge if you're happy with it

pakb (approved):

PR #45, opened 2021-03-22
Base branch for drawing

pakb:
Drawing mode is in the store at `state.drawing.mode`
The GeoJSON output is here : `state.drawing.geoJson` but has not yet been added to the map (still TODO)

[Test link](https://web-mapviewer.dev.bgdi.ch/base_drawing/index.html)


pakb:
integrated through https://github.com/geoadmin/web-mapviewer/pull/57 on `develop`

PR #47, opened 2021-03-24
Browserlist proposition

gberaudo:
It looks like current version of browserlist includes the not supported IE11.
Here is a proposition of update:
```bash
node_modules/.bin/browserslist 
and_chr 89
and_ff 86
and_qq 10.4
and_uc 12.12
android 89
baidu 7.12
chrome 89
chrome 88
chrome 87
edge 89
edge 88
firefox 86
firefox 85
firefox 78
ios_saf 14.0-14.5
ios_saf 13.4-13.7
kaios 2.5
op_mini all
op_mob 62
opera 73
opera 72
safari 14
safari 13.1
samsung 13.0
samsung 12.0
```

[Test link](https://web-mapviewer.dev.bgdi.ch/browserlist_update/index.html)

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -1,3 +1,5 @@
 > 1%
 last 2 versions
 not dead
+not IE 11
+Firefox ESR
--- a/public/index.html
+++ b/public/index.html
@@ -2,7 +2,6 @@
 <html lang="en">
   <head>
     <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width,initial-scale=1.0,user-scalable=no">
     <link rel="icon" href="<%= BASE_URL %>favicon.ico">
     <title><%= htmlWebpackPlugin.options.title %></title>


.browserslistrc
```diff
@@ -1,3 +1,5 @@
 > 1%
 last 2 versions
 not dead
+not IE 11
+Firefox ESR
```
pakb, line 5: I'm all for IE eradication, but I don't know if I want to specify this LTS version of FF specifically. If this version is used by more than 1% of the user base it will be added in the list, no?

IMHO, the least we customize, the better

davidoesch:
we officially do support current -1 Full Version of each browser
however ff ESR is the official swiss gov second browser (68.12.0 esr at the moment)
https://gs.statcounter.com/browser-version-market-share/all/switzerland/#monthly-202002-202102

pakb:
@davidoesch do you mean by that that we should specify it like so? Or can we remove it and assume that swiss gov will use a browser that as at least 1% of usages globally? (that's maybe optimistic, I know 😄)

davidoesch:
Yes.
Current and current -1 Version of each browser .
For gov: it has to work with our official edge version. And the ESR FF

pakb (approved):

PR #60, opened 2021-04-22
BGDIINF_SB-1702: Show only one layer details menu  at a time

ltshb:
Now the layer details menu is only shown one at a time like on the old viewer.

Also updated the CONTRIBUTING.md from my understanding of it but not sure if it is correct. For this to work correctly I think it is also missing the initial version tag `v0.0.0` on master. @pakb if you agree I'll create this tag on master ?

[Test link](https://web-mapviewer.dev.bgdi.ch/feature_bgdiinf_sb-1702_layer_visibility/index.html)


tests/e2e/specs/layerDetails.spec.js
```diff
@@ -0,0 +1,35 @@
...
+            .find('[class="menu-layer-list-item-details"]')
+            .as('wmsLayerDetail')
+            .should('be.visible')
+        // the other layer details should not be visible
+        cy.get('[data-cy="button-open-visible-layer-settings-test.wmts.layer"]')
+            .as('wmtsDetailBtn')
+            .parents('[class="menu-layer-list-item"]')
+            .find('[class="menu-layer-list-item-details"]')
+            .as('wmtsLayerDetail')
+            .should('not.be.visible')
```
pakb, line 25: I'd prefer that we don't use CSS class selector, but instead that we add a custom HTML attribute `data-cy="..."` to any element we need to grab in our test (see [Cypress's best practices on that](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements))

pakb (changes requested):
Great! Only one little thing in the test that I'd like to address.
Regarding your tagging of master, if that makes the GitHub action/plugin work, please do.

tests/e2e/specs/layerDetails.spec.js
ltshb: Good idea, thanks for the tip I'll change that.

pakb:
Could you change the color of the activated cog button to red (using the class `text-danger` on it or something similar) ?
I think it's written somewhere in the ticket to do so 😉

ltshb:
> Could you change the color of the activated cog button to red (using the class `text-danger` on it or something similar) ?
> I think it's written somewhere in the ticket to do so

Oops true I forget about that :+1:

pakb (approved):

PR #67, opened 2021-05-21
Refactor drawing

gberaudo:
- only store GeoJSON in the store when drawing is done

[Test link](https://web-mapviewer.dev.bgdi.ch/refactor_drawing_bgdiinf_sb-1832/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -14,7 +14,7 @@
                 :feature="selectedFeature"
                 @delete="deleteSelectedFeature"
                 @close="deactivateFeature"
-                @updateProperties="updateProperties"
+                @updateProperties="triggerKMLUpdate"
             />
         </div>
     </div>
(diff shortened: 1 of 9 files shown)


pakb (changes requested):
Please fix unit tests, after some of your refactor the function `lonLatToLv95` to `toLv95` doesn't work as expected in tests.
Maybe a default value for the arg `epsg` set to `proj4.WGS84` could help

gberaudo:
> Please fix unit tests, after refactoring the function from `lonLatToLv95` to `toLv95`, it doesn't work as expected in tests.
> Maybe a default value for the arg `epsg` set to `proj4.WGS84` could help

@pakb, the error is fixed. Please review.

src/modules/drawing/DrawingModule.vue
```diff
@@ -142,52 +126,51 @@ export default {
...
             }
         )
-        this.manager.on('change', (event) => {
-            this.setDrawingGeoJSON(event.geojson)
-            this.saveDrawing(event.kml)
+        // if we are testing with Cypress, we expose the store
+        if (IS_TESTING_WITH_CYPRESS) {
+            window['drawingMap'] = map
+            window['drawingManager'] = this.manager
+        }
```
pakb, line 133:
```suggestion
        // if we are testing with Cypress, we expose the drawing manager
        if (IS_TESTING_WITH_CYPRESS) {
            window.drawingMap = map
            window.drawingManager = this.manager
        }
```

pakb (approved):
👍

PR #68, opened 2021-05-25
BGDIINF_SB-1807 Add styling popup and capabilities for text

vladyslav-tk:
[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1807/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingStylePopup.vue
+++ b/src/modules/drawing/components/DrawingStylePopup.vue
@@ -9,11 +9,11 @@
         </div>
         <div class="card-body text-left">
             <form>
-                <div v-if="feature.getGeometry().getType() === 'Point'" class="form-group">
+                <div v-if="featureGeometry.getType() === 'Point'" class="form-group">
                     <label for="text">{{ $t('draw_popup_title_annotation') }}:</label>
                     <textarea id="text" v-model="text" class="form-control" rows="1"></textarea>
                 </div>
-                <div class="form-group">
+                <div v-if="isFeatureText" class="form-group">
                     <label for="description">{{ $t('modify_description') }}:</label>
                     <textarea
                         id="description"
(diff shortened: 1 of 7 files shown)


src/modules/drawing/components/DrawingStylePopup.vue
```diff
@@ -13,7 +13,7 @@
                     <label for="text">{{ $t('draw_popup_title_annotation') }}:</label>
                     <textarea id="text" v-model="text" class="form-control" rows="1"></textarea>
                 </div>
-                <div class="form-group">
+                <div v-if="feature.get('type') !== 'TEXT'" class="form-group">
```
pakb, line 16: as `feature.get('type')` is used more than once in the template, it would be more efficient to have a `computed` with it, so that it is calculated once (each time feature changes) and not 4 or 5 times in the template.

That's usually a good rule of thumb, do not place too much data access logic in the template, but prefer a `computed` properties for that.

src/modules/drawing/components/DrawingStylePopup.vue
```diff
@@ -27,6 +27,28 @@
...
+                        v-if="feature.get('type') === 'MARKER' || feature.get('type') === 'TEXT'"
+                        ref="textStylePopoverBtn"
+                        type="button"
+                        class="btn btn-default text-style"
+                    >
+                        <font-awesome-icon :icon="['fas', 'font']" />
+                    </button>
+                </div>
+                <div
+                    v-if="feature.get('type') === 'MARKER' || feature.get('type') === 'TEXT'"
```
pakb, line 40: it could even make more sense to have two computed properties called `isFeatureMarker` and `isFeatureText` as that's the only thing we use `feature.get('type')` for.

pakb (commented):
Overall very good code quality! I would just ask you to squash your 3 commits into one in order to clean up a little bit the git log.

I'm wondering if we should maybe have the popover at the right side of the marker (if the screen space authorize it) so that the user can see what he is changing without having to close the style popup. What do you think?

gberaudo (approved):
LGTM. @vladyslav-tk, feel free to merge, I can introduce the 2 computed properties while working on the marker symbols.

vladyslav-tk:
> I'm wondering if we should maybe have the popover at the right side of the marker (if the screen space authorize it) so that the user can see what he is changing without having to close the style popup. What do you think?

Currently, we don't have a live update of styles, I think changes will be visible when it is done

vladyslav-tk:
> LGTM. @vladyslav-tk, feel free to merge, I can introduce the 2 computed properties while working on the marker symbols.

I made it

PR #70, opened 2021-05-31
Create a polygon only when the first point is clicked

fredj:
[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1822/index.html)


pakb (approved):
👍 very good! I'm wondering if we shouldn't start splitting this big `DrawingManager.js` into helper functions / store interaction or something on those lines.
What do you think?

PR #75, opened 2021-06-03
Style lines

gberaudo:
[Test link](https://web-mapviewer.dev.bgdi.ch/lines_bgdiinf_sb-1806/index.html)

TODO:
-  [x] rebase on develop when #72  is merged;
-  [x] add tests.

### DIFF
--- a/src/modules/drawing/components/Color.vue
+++ b/src/modules/drawing/components/Color.vue
@@ -1,0 +1,80 @@
+<template>
+    <div class="setting-container">
+        <span>{{ $t('modify_color_label') }}:</span>
+        <div class="color-select-box">
+            <div
+                v-for="c in colors"
+                :key="c.name"
+                :class="{ selected: currentColor === c.fill }"
+                @click="() => onColorChange(c)"
+            >
+                <div :style="{ 'background-color': c.name }"></div>
+            </div>
+        </div>
+    </div>
+</template>
+
+<script>
+export default {
+    components: {},
+    props: {
+        colors: {
+            type: Array,
+            default() {
+                return []
+            },
+        },
+        currentColor: {
+            type: String,
+            default: '',
+        },
+    },
+    methods: {
+        onColorChange: function (c) {
+            this.$emit('onColorChange', c)
+        },
+    },
+}
+</script>
+
+<style lang="scss">
+.color-select-box {
+    position: relative;
+    overflow: hidden;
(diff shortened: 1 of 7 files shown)


gberaudo:
Depends on #72 .

src/modules/drawing/components/DrawingStylePopup.vue
```diff
@@ -157,14 +183,21 @@ export default {
                 return this.feature.get('type') === 'TEXT'
             },
         },
+        isFeatureLine: {
+            get() {
+                return this.feature.get('type') === 'LINE'
+            },
+        },
```
pakb, line 190: Here you can simplify the syntax to 
```
isFeatureLine: () => this.feature.get('type') === 'LINE',
```
if you want, same for the others above. Using the `get/set` syntax is good when your computed needs a setter, but if it's read only there's no need to use this more complex syntax.

src/modules/drawing/components/LineStylePopup.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+    <div class="line-style-popup">
+        <button type="button" class="close" @click="onClose">
+            <span aria-hidden="true">&times;</span>
+        </button>
+
+        <color :color="color" :colors="options.colors" @onColorChange="onColorChange"></color>
```
pakb, line 7: Could you use the uppercase name here? `<Color ...></Color>`
This way it's easier to know this is one of our custom component and not some custom tag coming from another library

src/modules/drawing/components/Color.vue
```diff
@@ -0,0 +1,80 @@
...
+export default {
+    components: {},
+    props: {
+        colors: {
+            type: Array,
+            default() {
+                return []
+            },
+        },
+        color: {
```
pakb, line 27: could be good to specify that this is the `currentColor`, it can be hard to know just by looking at the component why there is `colors` and `color`

pakb (commented):
Some little improvements there and there, overall pretty good!

src/modules/drawing/components/DrawingStylePopup.vue
gberaudo: It was not possible to use arrow functions in this case so I simplified the syntax with plain functions.

gberaudo:
@pakb, I implemented your suggestions.

pakb (approved):
👍

PR #76, opened 2021-06-11
Split drawing tests

gberaudo:
[Test link](https://web-mapviewer.dev.bgdi.ch/split_tests_bgdiinf_sb-1822/index.html)


pakb (approved):
What you added is good, now it would be good if you could review my part 😉

PR #78, opened 2021-06-15
Fix icons retrieval issues

gberaudo:
[Test link](https://web-mapviewer.dev.bgdi.ch/fix_icon_retrieval_issues_bgdiinf_sb-1779/index.html)

### DIFF
--- a/src/modules/drawing/lib/style.js
+++ b/src/modules/drawing/lib/style.js
@@ -72,6 +72,7 @@ export function featureStyle(feature) {
         // this might be expensive
         image = new Icon({
             src: icon,
+            crossOrigin: 'Anonymous',
             anchor: anchor,
             scale: markerScale,
         })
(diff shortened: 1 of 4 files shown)


gberaudo:
In addition, see https://github.com/geoadmin/service-icons/pull/33.

src/modules/drawing/lib/style.js
```diff
@@ -72,6 +72,7 @@ export function featureStyle(feature) {
         // this might be expensive
         image = new Icon({
             src: icon,
+            crossOrigin: 'Anonymous',
```
boecklic, line 75: Not exactly sure what this does, but if it sends `Anonymous` as Origin header to the backend, then it will fail.

gberaudo: @boecklic, the browser sometimes send the Origin headers and sometimes not. There are many cases.
This option asks the browser to always put the header, even if it is not "required" by the web standards. This fixes our issue because the server side has an hard requirement on this header (which is not the usual behaviour, hence the PR I created there).
No worry, the browser will not send the string "Anonymous", but the actual orgin of the website, which in this case will be `https://web-mapviewer.dev.bgdi.ch`.

pakb (approved):

PR #80, opened 2021-06-22
BGDIINF_SB-1804 Adapt Line/Polygon with Measurement style

vladyslav-tk:
[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1804/index.html)

### DIFF
--- a/src/modules/drawing/lib/DrawingManager.js
+++ b/src/modules/drawing/lib/DrawingManager.js
@@ -21,6 +21,7 @@ import { featureStyle } from './style'
 import Overlay from 'ol/Overlay'
 import i18n from '@/modules/i18n'
 import { Polygon, LineString } from 'ol/geom'
+import MeasureManager from '@/modules/drawing/lib/MeasureManager'
 
 const typesInTranslation = {
     MARKER: 'marker',
@@ -83,6 +84,8 @@ export default class DrawingManager extends Observable {
         })
         this.source = this.layer.getSource()
 
+        this.measureManager = new MeasureManager(this.map, this.layer)
+
         this.tools = {}
         this.sketchPoints = 0
         for (let [type, options] of Object.entries(tools)) {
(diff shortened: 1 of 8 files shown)


src/modules/drawing/lib/drawingUtils.js
```diff
@@ -15,14 +15,15 @@ export function formatPointCoordinates(coo) {
...
-export function formatMeters(value, { dim = 1, digits = 2 } = {}) {
+export function formatMeters(value, { dim = 1, digits = 2, applyFormat = true } = {}) {
     const factor = Math.pow(1000, dim)
     let unit = dim === 1 ? 'm' : 'm²'
     if (value >= factor) {
         unit = dim === 1 ? 'km' : 'km²'
         value /= factor
     }
-    return `${format(value, digits)} ${unit}`
+    value = applyFormat ? format(value, digits) : value.toFixed(digits)
```
pakb, line 25: What about calling this function with `digits=0` instead of adding a new boolean arg? It will have the same output I think

pakb (commented):
Some comments on the code.
Regarding UI, would it be possible to make the text displaying distance non selectable? First time I tried to create a measure line the text was selected when I double clicked to finish drawing.

src/modules/drawing/lib/DrawingManager.js
```diff
@@ -212,7 +218,13 @@ export default class DrawingManager extends Observable {
     }
 
     createGeoJSONAndKML() {
-        const geojson = this.geojsonFormat.writeFeaturesObject(this.source.getFeatures())
+        // todo Do we need overlays in geoJson? it makes loop during state update
+        const features = this.source.getFeatures().map((f) => {
```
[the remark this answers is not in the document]
vladyslav-tk: I am not sure, should we store overlays collection in geoJson or not?

src/modules/drawing/lib/drawingUtils.js
vladyslav-tk: No, it is not the same. `digits` defines the number of digits after the dot, `applyFormat=false` disables inserting of " ' " for thousands

pakb (approved):

PR #82, opened 2021-07-02
BGDIINF_SB-1802 Export drawing as KML

vladyslav-tk:
[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1802/index.html)


package.json
```diff
@@ -31,9 +31,11 @@
     "bootstrap": "^4.6.0",
     "core-js": "^3.9.1",
     "d3": "^6.7.0",
+    "file-saver": "^2.0.5",
     "hammerjs": "^2.0.8",
     "jquery": "^3.6.0",
     "mobile-device-detect": "^0.4.3",
+    "moment": "^2.29.1",
```
pakb, line 38: I see that we need this only to help format one date at a specific place. Please find another than to import `moment.js` to the whole project, momentjs is not yet a ES6 module ready project, and does not support tree shaking. Meaning the whole lib will be included in our deliverable JS, without removing what we don't use, resulting in very very big JS files for not much use.

Maybe a simple `new Date().toISOString()` could suffice for what you want to accomplish

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -6,13 +6,32 @@
...
                     </button>
-                    <DrawingToolboxButton
-                        v-for="drawingMode in drawingModes"
-                        :key="drawingMode"
-                        :drawing-mode="drawingMode"
-                        :is-active="currentDrawingMode === drawingMode"
-                        @setDrawingMode="bubbleSetDrawingEventToParent"
-                    />
+                    <div class="buttons-container">
+                        <div>
```
pakb, line 10: is this `<div>` really required?

pakb (commented):
I'd like to see some e2e tests covering that too 😉

vladyslav-tk:
@pakb Please review the last changes, I made updates according to comments and added tests. I can't check why CI fails but locally all tests passed.

pakb (commented):
I'll look into what is going wrong in the CI (I might need to rebase your branch after some fix on `develop`)
In the meantime, there's a little something you could change in the tests regarding (local) file download

tests/e2e/specs/exportDrawing.spec.js
```diff
@@ -0,0 +1,56 @@
...
+
+        recurse(
+            () => cy.task('findFiles', { folderName: downloadsFolder, extension: 'kml' }),
+            isNonEmptyString,
+            {
+                delay: 100,
+                timeout: 10000,
+            }
+        ).then((files) => {
+            const fileName = `${downloadsFolder}\\${files[files.length - 1]}`
```
pakb, line 43: we are mostly using UNIX based system, a `\` will not be a valid file separator

pakb (commented):
You could run `npm run test:headless` to see if things work on your side, or `npm run test:e2e` and choose specifically the one you have modified.
This way we don't have to do a ping-pong with the CI logs that you can't access

vladyslav-tk:
> You could run npm run test:headless to see if things work on your side, or npm run test:e2e and choose specifically the one you have modified.

I know, as I wrote, all tests passed locally

pakb (approved):

PR #87, opened 2021-07-15
BGDIINF_SB-1803 Delete drawing

vladyslav-tk:
[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb-1803/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingToolbox.vue
+++ b/src/modules/drawing/components/DrawingToolbox.vue
@@ -16,12 +16,17 @@
                         />
                         <br />
                         <div class="btn-group btn-group-sm draw-action-btns" role="group">
-                            <button type="button" class="btn btn-outline-secondary" disabled>
+                            <button
+                                type="button"
+                                class="btn btn-outline-secondary"
+                                :disabled="!drawingNotEmpty"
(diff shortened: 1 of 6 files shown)


src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -93,6 +104,9 @@ export default {
         }
     },
     methods: {
+        showClearConfirmation: function () {
+            this.$refs['clearConfirmation'].show = true
+        },
```
pakb, line 109: Wouldn't it be better (and more understandable) to use a `v-show` or `v-if` on the template itself instead of changing that here in a method?

pakb (commented):
Small question on this modal show/hide, otherwise looking good!

src/modules/drawing/components/DrawingToolbox.vue
vladyslav-tk: We should open the modal window on button click, so we will need a method to handle click events anyway. Also, in the case of use `v-show` or `v-if`, we will need an additional variable

pakb (approved):

PR #88, opened 2021-07-16
BGDIINF_SB-1633 : hiding scale line when zoom level is under 9

ltkum:
Issue : We want the scale line to not be present when it doesn't make sense. Webmercator deformations make it unusable when we are at high zoom levels, besides the fact that there is little need for such a functionality at such zoom levels.

Fix : added a v-if condition which checks the zoom level
        also added e2e tests which will check that the scaleline existence is behaving as expected.


ltkum:
From what I've seen, the error is in the profile tests, where it can't click on the profile delete button ( After checking out, this button doesn't exist) in the "test-empty-profile" test. Sounds like this is not related to this, issue, and should be ready to be merged.

pakb (approved):
Code looks good, I have to figure out what's happening in the e2e testing (which seems not related to what you have added...)

PR #97, opened 2021-09-15
BGDIINF_SB-1705: Leave menu open when topic is switched

IsabelleBzr:
[Test link](https://web-mapviewer.dev.bgdi.ch/feature_bgdiinf_sb-1705_leave_menu_open_when_topic_is_switched/index.html)

### DIFF
--- a/src/modules/menu/components/MenuSection.vue
+++ b/src/modules/menu/components/MenuSection.vue
@@ -28,7 +28,7 @@ export default {
         },
         showContent: {
             type: Boolean,
-            default: false,
+            default: true,
         },
     },
     data() {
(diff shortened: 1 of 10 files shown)


src/modules/menu/components/MenuSection.vue
```diff
@@ -28,7 +28,7 @@ export default {
         },
         showContent: {
             type: Boolean,
-            default: false,
+            default: true,
```
pakb, line 31: With what you have done below, I'm not sure this is still relevant.
I'd prefer to have its default behavior to be hidden

pakb (commented):
Very minor change left, otherwise looks good!

src/modules/menu/components/MenuSection.vue
pakb: About that, maybe we could add something in its `:show-content=""` part, that would open it only if it is not the default topic (the first of the bunch).

This way the topic tree would open only if you selected another topic (which is how map.geo.admin.ch does it)

pakb (approved):

PR #100, opened 2021-09-22
BGDIINF_SB-1866: Various small bugfix in drawing module

ltshb:
See commit message for details on the bugfixes.


package.json
```diff
@@ -49,6 +49,7 @@
     "vuex": "^3.6.2"
   },
   "devDependencies": {
+    "@babel/core": "^7.11.6",
```
pakb, line 52: I'm not sure it's really useful to describe this dependency in our `package.json` as it is already listed as a dep of `@vue/cli-plugin-babel` and `vue/cli` will make sure the version of `@babel/core` matches the one it was built for.
I'd prefer we let `@vue/cli-plugin-babel` set the correct version than to overwrite ourself.

package.json
```diff
@@ -64,6 +65,7 @@
     "babel-eslint": "^10.1.0",
     "caniuse-lite": "^1.0.30001255",
     "chai": "^4.3.4",
+    "cypress": "^7.7.0",
```
pakb, line 68: Same thing as with `@babel/core`, Cypress' version is already set as a byproduct of listing `@vue/cli-plugin-e2e-cypress` so we shouldn't force set it (as `@vue/cli-plugin-e2e-cypress` has been tested with a specific version, and not necessarily the latest)

package.json
```diff
@@ -75,6 +77,7 @@
     "git-branch": "^2.0.1",
     "googleapis": "^85.0.0",
     "mime-types": "^2.1.32",
+    "postcss": "^7.0.36",
```
pakb, line 80: Do we use postcss directly? I'm sure it is already added somewhere by `@vue/cli` or some other dependency (because even though you've set it here, no changes were made in `package-lock.json`)

If it's not something we use, I'd prefer avoiding listing it here (doesn't make sense IMHO)

pakb (commented):
Outside of some minor point to talk on `package.json` the rest looks good! thanks!

PR #104, opened 2021-09-27
Add support for topic URL param

pakb:


src/modules/store/modules/topics.store.js
```diff
@@ -47,6 +49,12 @@ const actions = {
     changeTopic: ({ commit }, topic) => {
         commit(CHANGE_TOPIC_MUTATION, topic)
     },
+    setTopicById: ({ commit, state }, topicId) => {
+        const topic = state.config.find((topic) => topic.id === topicId)
+        if (topic) {
```
boecklic, line 54: what happens if this evaluates to `false`? Are you printing errors on the console in such cases?

pakb: currently nothing happen for unkown topics, maybe we could log a little something in the console yeah...

PR #108, opened 2021-09-30
Visible layers attribution (copyrights)

pakb:
needs https://github.com/geoadmin/web-mapviewer/pull/105 to be merged first (require `bgLayer` URL param support in E2E tests)


src/api/layers/AggregateLayer.class.js
```diff
@@ -48,6 +50,8 @@ export default class AggregateLayer extends GeoAdminLayer {
         name,
         id,
         opacity,
+        attribution,
```
boecklic, line 53: maybe `attributionName` to make it more clear (i.e. attribution consists of attributionName and attributionUrl)

PR #112, opened 2021-11-08
Update pr-labeler.yml

ltshb:
Also trigger the PR Auto Labeler on push event
    
 Because the PR auto labeler workflow is required before merging, it needs to be runned
 also each time a push on the PR is done, otherwise the merge gets
 blocked until the workflow has been runned.

[Test link](https://web-mapviewer.dev.bgdi.ch/bug-workflow/index.html)

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -5,6 +5,7 @@ on:
     types:
       - opened
       - edited
+  push:
 
 jobs:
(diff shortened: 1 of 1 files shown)


ltshb:
I finally removed the `pr-labeler` from the required action before merging. The issue when having the action run on two different event, it will be run several time which is a bit overkill for such a simpler feature.

PR #120, opened 2021-11-30
BGDIINF_SB-2022 : Feature. Adding Disclaimer to drawing menu

ltkum:
Issue : On the old mapviewer, there was a disclaimer to tell people that their drawings would be stored and that they agree to the terms of use of the service when drawing. This was missing.
Fix: We add the disclaimer and ensure it is translated in the five languages.

[Test link](https://web-mapviewer.dev.bgdi.ch/bgdiinf_sb_2022/index.html)


pakb:
Have you updated the translations files through a `npm run update:translations` or manually?

ltkum:
I did that manually.

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -98,6 +98,13 @@
                             {{ $t('draw_button_delete_last_point') }}
                         </button>
                     </div>
+                    <div class="d-block">
+                        {{ $t('drawing_disclaimer_part1')
+                        }}<a :href="$t('drawing_disclaimer_url')">{{
```
pakb, line 103: can you add a `_target="blank"` to the `<a>` tag? That would open a new tab when we click on the link (not fun when it loads the link in the current tab...)

pakb: and I think you can put one statement per line (no need to stick them together in the code)

pakb:
I was thinking about this HTML piece in the translation file, and while going through the vue i18n doc I saw you can simply add à `v-html` attribute to the surrounding HTML tag, and the content will be interpreted as such.
Why don't we simplify this whole thing and do exactly that for this specific key? 😅

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -98,6 +98,14 @@
                             {{ $t('draw_button_delete_last_point') }}
                         </button>
                     </div>
+                    <!-- v-html could lead to some vulnerability if the html
+                    given is not thrustworthy. This makes the google doc a
+                    sensitive document (even more so than before) since an
+                    attacker which gained access to it could change the html
+                    here to cause damages to all users. Just leaving a comment
+                    here to remind us of that.-->
```
pakb, line 106: Technically the attacker would also need to have write access to this repo, and an AWS credentials that can write on our S3 buckets.
At this point, one would not just alter some obscure translation key if he has nefarious intent 😉

src/modules/i18n/locales/de.json
```diff
@@ -1 +1 @@
...
...edit_or_reset_style":"Diese Ebene verwendet einen Stil von Drittanbietern. Sie können diesen nicht ändern, aber Sie können den Standardstil verwenden, indem Sie auf diese Schaltfläche klicken.","basis":"Basiskarte light","labels":"Labels","roadtraffic":"Strassen","hydrology":"Hydrologie","landuse":"Grünflächen","settlement":"Siedlungsgebiet","territory":"Gelände","woodland":"Wald","edit_visibility":"Darstellung","current_value":"Aktueller Wert","mvt_ie11_alert":"Sie verwenden InternetExplorer ≤ 11:  Bitte nutzen Sie die neueste Version von Microsoft EDGE, Firefox oder Chrome, um optimale Ergebnisse zu erhalten","try_test_viewer":"Probieren Sie test.map.geo.admin.ch aus","obstacle_started_last_4_days":"Aktivierte Hindernisse letzte 4 Tage","obstacle_deleted_last_4_days":"Gelöschte Hindernisse letzte 4 Tage","draw_button_delete_last_point":"Letzten Punkt löschen","drawing_disclaimer_part1":"Ihre Zeichnung wird automatisch ein Jahr lang gespeichert. Durch die Nutzung dieses Dienstes stimmen Sie den ","drawing_disclaimer_url":"https://www.geo.admin.ch/de/about-swiss-geoportal/impressum.html#drawing","drawing_disclaimer_link_text":"Nutzungsbedingungen","drawing_disclaimer_ending":"zu."}
```
pakb, line 1: Is it needed to update/add this new keys as we are in fine not using them?

pakb (approved):

PR #126, opened 2021-12-16
BGDIINF_SB-1993: Added support for legacy KML adminid query parameter

ltshb:
The legacy URL for KML admin was as follow:

?adminid=ADMIN_ID

And now this needs to be transformed to

?layers=KML|https://public.geo.admin.ch/api/kml/files/KML_ID@adminid=ADMIN_ID

Also reorganized the code for a better readability.

And finally fixed a bug in the store sync.

### DIFF
--- a/src/api/files.api.js
+++ b/src/api/files.api.js
@@ -117,8 +117,8 @@ export const getKmlMetadataUrl = (id) => {
 /**
  * Publish KML on backend
  *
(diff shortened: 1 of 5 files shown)


pakb (approved):
Well done, nice way of solving the problem doing it at the end of the legacy param parsing. Was a bit messy doing it in the switch.

PR #127, opened 2021-12-17
BGDIINF SB-1854: add the position popup on right click

IsabelleBzr:
Adding the Position pop up when doing a right click on the map


src/modules/map/components/LocationPopup.vue
```diff
@@ -0,0 +1,228 @@
...
+import Overlay from 'ol/Overlay'
+import OverlayPositioning from 'ol/OverlayPositioning'
+import { printHumanReadableCoordinates, CoordinateSystems } from '@/utils/coordinateUtils'
+import ButtonWithIcon from '@/utils/ButtonWithIcon'
+import { registerWhat3WordsLocation } from '@/api/what3words.api'
+import { requestHeight } from '@/api/height.api'
+import { generateQrCode } from '@/api/qrcode.api'
+import { round } from '@/utils/numberUtils'
+
+export default {
```
pakb, line 89: Could you add a little comment (JSDoc) on what this component does (a quick description so that we don't have to read the whole component to know what it's doing)?

src/modules/map/components/LocationPopup.vue
```diff
@@ -0,0 +1,228 @@
...
+            if (this.clickCoordinates) {
+                requestHeight(this.clickCoordinates).then((height) => {
+                    this.height = height
+                })
+            }
+        },
+        generateQrCodeFromBackend: function () {
+            if (this.clickCoordinates) {
+                generateQrCode(window.location.href).then((qrCode) => {
+                    console.log('yay!', qrCode)
```
pakb, line 201: that shouldn't stay here 😉

src/modules/tooltip/TooltipModule.vue
```diff
@@ -1,5 +1,7 @@
 <template>
-    <TooltipBox />
+    <div>
+        <TooltipBox />
+    </div>
```
pakb, line 4: I don't think that's necessary anymore (might be a relic of us wanting to add this location popup at this location)

src/utils/coordinateUtils.js
```diff
@@ -318,7 +318,7 @@ export const CoordinateSystems = {
         epsg: 'EPSG:4326',
         label: 'MGRS',
         format: function (coordinate) {
-            return LLtoMGRS(coordinate, 5)
+            return LLtoMGRS(coordinate, 5).replace(/(.{5})/g, '$1 ')
```
pakb, line 321: maybe you could add a little comment on why we do this weird replace (the format we receive the value, and the wanted output)

tests/e2e/specs/layers.spec.js
```diff
@@ -123,6 +123,12 @@ describe('Test of layer handling', () => {
             cy.get('[data-cy="menu-active-layers"]').click()
             cy.get('[data-cy="menu-section-no-layers"]').should('be.hidden')
         })
+
+        it('Shows the LocationPopUp when doing a right click', () => {
+            cy.goToMapView()
+            // doing a right click
+            cy.get('[data-cy="map"]').rightclick()
+        })
```
pakb, line 131: did you forget to clear this test? it doesn't test much 😄

tests/e2e/specs/mouseposition.spec.js
```diff
@@ -20,32 +25,109 @@ function getMousePositionValue(x = 100, y = 100, coordStr = '671202, 6074296') {
...
-    it('goes back to LV95 display if selected again', () => {
-        // Change display projection without moving the mouse
-        getMousePositionAndSelect(CoordinateSystems.MGRS)
-        getMousePositionAndSelect(CoordinateSystems.LV95)
-        getMousePositionValue(100, 100, '2435419.18, 1297425.86')
+    context('LocationPopUp when rightclick on the map', function () {
+        it('Test the LocationPopUp is visible', () => {
+            cy.goToMapView()
+            // doing a right click
+            cy.get('[data-cy="map"]').rightclick(200, 100)
```
pakb, line 62: here you should then test that your popup is visibile on the screen by grabbing it with `cy.get('[data-cy="..."]').should('be.visible')` or something similar

pakb (changes requested):
Some little left over and some comments and we should be good to go 🚀

pakb (approved):

PR #130, opened 2021-12-30
BGDIINF_SB-2114: Update to Vue 3.x

ltshb:
[Test link](https://web-mapviewer.dev.bgdi.ch/feat-bgdiinf_sb-2114-update-vue-3/index.html)

PR #131, opened 2022-01-05
BGDIINF_SB-2114: Preparation to Vue 3 migration and various bug fixes

ltshb:
This PR has some preparation work for the Vue 3 migration and also fixing various bugs. See separate commits for more details.

Note the bug that have been fixed don't have yet e2e test, but I've created jira ticket for this task.

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -149,3 +149,9 @@ export const TILEGRID_RESOLUTIONS = [
     4000, 3750, 3500, 3250, 3000, 2750, 2500, 2250, 2000, 1750, 1500, 1250, 1000, 750, 650, 500,
     250, 100, 50, 20, 10, 5, 2.5, 2, 1.5, 1, 0.5, 0.25, 0.1,
 ]
+
+/**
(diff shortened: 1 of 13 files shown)


pakb (approved):
Nice!
What a load of bad code I have written this year... thanks for fixing some of it 😉

src/router/index.js
```diff
@@ -33,15 +32,6 @@ Vue.use(VueRouter)
...
         path: '/',
-        redirect: { name: 'LoadingView' },
-    },
-    {
-        path: '/startup',
-        name: 'LoadingView',
-        component: LoadingView,
-    },
-    {
-        path: '/map',
```
ltshb, line 44: Have you looked at the routing work that I've done in #130 ? This will conflict with this commit. Why removing the `LoadingView` ? It might be a good idea to discuss the goal of this LoadingView and some edge case like for instance if mf-chsdi3 is not available.

ltshb:
Because this work is already part of #130 we close this PR.

PR #133, opened 2022-01-13
BGDIINF_SB-2114: Fixed drawing layer name translation

ltshb:
The drawing layer name did not changed when the language setting was changed.
    
Because this name is for the moment hardcoded to `Drawing`, `Dessin`,  ... We made the KMLLayer name property dynamic with a getter returning the actual translation. The drawing-layer-management.plugin.js has been a bit cleaned up, but its logic didn't changed.


ltshb:
@pakb I resolve the e2e test, you can do the review.

pakb (approved):

PR #134, opened 2022-01-13
BGDIINF_SB-2137: Removed Turf.js

fkm:
We bundled the Turf.js library to make just one calculation.

While there's nothing wrong with the library it seemed overkill to include it for a single use. Especially as the calculation is rather trivial.

This hard-codes the necessary part directly where it is needed to generate as little overhead as possible.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2137-remove-turf/index.html)


pakb (approved):
Nice! Thanks

PR #138, opened 2022-01-18
BGDIINF_SB-2134: Added e2e test for layer toggling from topic menu

fkm:
During the update to Vue3 a bug was fixed that prevented unchecking a layer via the topic menu.

This adds an e2e test to check that the bug has been fixed and the layers can be toggled via the topic menu.

[Test link](https://web-mapviewer.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2134-e2e-toggle-layer-topicmenu/index.html)

### DIFF
--- a/tests/e2e/specs/layers.spec.js
+++ b/tests/e2e/specs/layers.spec.js
@@ -270,16 +270,17 @@ describe('Test of layer handling', () => {
                 })
             })
         })
-        it.only('allows toggling layers from the topic menu', () => {
+        it('allows toggling layers from the topic menu', () => {
             const testLayerId = 'test.wmts.layer'
+            const testLayerSelector = `[data-cy="topic-tree-item-${testLayerId}"]`
             cy.get('[data-cy="menu-topic-section"]').click()
             // Find the test layer and open the appropriate menu entries.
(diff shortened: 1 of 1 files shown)


tests/e2e/specs/layers.spec.js
```diff
@@ -270,6 +270,35 @@ describe('Test of layer handling', () => {
                 })
             })
         })
+        it.only('allows toggling layers from the topic menu', () => {
```
ltshb, line 273: Remove `only`.

ltshb (approved):
Very good, just don't forget to remove the `only` before merging.

tests/e2e/specs/layers.spec.js
fkm: Good idea 🙈
**Thank you!**

fkm:
I removed the `.only` modifier.

While I was at it I also did some minor refactoring.
> In addition this changes `menuItems.get()` to `menuItems.toArray()`. Both methods lead to the same result but the latter makes the code easier to understand.
>
>And finally this refactors the selector of the test layer into a variable (`testLayerSelector`) to make the code easier to read.

ltshb (approved):
👍🏼

PR #139, opened 2022-01-24
[Draft] Mobile tooltip

pakb:
Change the way features are handled after a click, so that the app can show them either in a floating popup (desktop) or in a bottom tray (mobile).

This also affects how the drawing is working, as we also want to have features from the drawing side in the mobile tray.

[Test link](https://web-mapviewer.dev.bgdi.ch/feat-jira-bgdiinf_sb-1857-mobile-tooltip/index.html)


pakb:
will be done through https://jira.swisstopo.ch/browse/BGDIINF_SB-2159

PR #140, opened 2022-01-24
BGDIINF_SB-2133: Added e2e test for language switch

fkm:
During the update to Vue3 a bug was fixed that reverted the configuration to its initial state and prevented the displayed language in the active layers and topic menu to change when the user switched the language in the settings.

The test changes the layer configuration after the initial load and checks if the configuration persists after changing the language.

This commit also introduces a new fake request so we can wait until the layers have been configured (@layers-configured).

[Test link](https://web-mapviewer.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2133-e2e-change-language/index.html)

### DIFF
--- a/tests/e2e/specs/layers.spec.js
+++ b/tests/e2e/specs/layers.spec.js
@@ -402,7 +402,7 @@ describe('Test of layer handling', () => {
                     })
                 })
                 context('Language settings in menu', () => {
-                    it.only('keeps the layer settings when changing language', () => {
+                    it('keeps the layer settings when changing language', () => {
                         const langBefore = 'en'
                         const langAfter = 'de'
                         const visibleLayerIds = [
@@ -431,7 +431,7 @@ describe('Test of layer handling', () => {
                         })
 
                         // Open the menu and change the language.
-                        cy.get('[data-cy="menu-button"]').click()
+                        clickOnMenuButtonIfMobile()
                         cy.get('[data-cy="menu-settings-section"]').click()
                         cy.get(`[data-cy="menu-lang-${langAfter}"`).click()
 
--- a/tests/e2e/specs/layers.spec.js
+++ b/tests/e2e/specs/layers.spec.js
@@ -25,6 +25,15 @@ describe('Test of layer handling', () => {
         } while (randomTimestampFromLayer === defaultTimestamp)
         return randomTimestampFromLayer
     }
+    /**
+     * This function is used as a parameter to `JSON.stringify` to remove all properties with the name `lang`.
+     *
+     * @param {String} key The current property name.
+     * @param {any} value The current value to stringify.
+     * @returns {String} The string representation of the object.
+     * @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#the_replacer_parameter
+     */
+    const stringifyWithoutLang = (key, value) => (key === 'lang' ? undefined : value)
 
     forEachTestViewport((viewport, isMobile, isTablet, dimensions) => {
         context(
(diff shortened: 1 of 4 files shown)


tests/e2e/specs/layers.spec.js
```diff
@@ -300,6 +310,70 @@ describe('Test of layer handling', () => {
...
+                cy.fixture('layers.fixture.json').then((fixtures) => {
+                    const layerId = 'test.timeenabled.wmts.layer'
+                    const layerMetadata = fixtures[layerId]
+                    const layerTimestamp = getRandomTimestampFromSeries(layerMetadata)
+                    store.dispatch('setTimedLayerCurrentTimestamp', {
+                        layerId,
+                        timestamp: layerTimestamp,
+                    })
+                })
+            })
```
pakb, line 336: In order to change the app at startup, I would prefer that we leverage the fact that the URL represents the app state too.
So instead of changing store values with dispatch (that will be hard to track if they are renamed later on), maybe use a `cy.goToMapView('lang', { layers: '...' })` such as [here](https://github.com/geoadmin/web-mapviewer/blob/26c89c3b09b863ba8e91054790d3def36d12b3b2/tests/e2e/specs/layers.spec.js#L47-L49)

You can find more information on how to describe all parts of a layer configuration in [this ADR](https://github.com/geoadmin/web-mapviewer/blob/develop/adr/2021_03_16_url_param_structure.md#decision)

tests/e2e/specs/layers.spec.js
```diff
@@ -300,6 +310,70 @@ describe('Test of layer handling', () => {
...
+                )
+            })
+
+            // Wait until the layers have been updated. We have to wait here
+            // to make sure the next time we wait we catch the language change.
+            cy.wait('@layers-configured')
+
+            // Change the language to a non-default value.
+            cy.get('[data-cy="menu-settings-section"]').click()
+            cy.get(`#menu-lang-selector button:contains("${langAfter.toUpperCase()}")`).click()
```
pakb, line 357: would it be possible to use a custom HTML attribute here (`data-cy-...`) instead of a normal CSS selector? That's a best practice with Cypress, so that we are not depending on the HTML/CSS styling couple, and clearly state in the HTML that this element will be used in further tests.

tests/e2e/specs/layers.spec.js
```diff
@@ -300,6 +310,70 @@ describe('Test of layer handling', () => {
...
+
+            // Wait until the layers have been updated. This happens after
+            // the @layers have been loaded and the @topics are also too fast.
+            cy.wait('@layers-configured')
```
pakb, line 361: IMO you could place this line directly in the helper function `goToMapView`, so that all tests uses it, but I'm not sure it will be needed anymore if you change the code as described in my comment above

tests/e2e/specs/layers.spec.js
fkm: Good idea! I changed the configuration to only use the goToMapView command.

To avoid running it twice (and also to separate layer settings from language settings) I created a new context ("Language settings in menu") where I put the test.

tests/e2e/specs/layers.spec.js
fkm: Yeah. That was messy of me. I added the data-attributes and updated the selector in the test.

tests/e2e/specs/layers.spec.js
fkm: I moved the first `wait` to the `goToMapView` command.

Unfortunately, I couldn't remove either of the two. It looked good at first but during some runs the first `@layers-configured` was triggered too late and the second `wait` caught the first event. Removing the second `wait` also passed during some runs but less often.

fkm:
Finished rebasing.

pakb (approved):

PR #142, opened 2022-01-28
BGDIINF_SB-2161: Fixed calls to non-existant error function

fkm:
Probably due to a change of the log function we have calls to `log.error` which doesn't exist.

This modifies the calls to work with the current implementation of the log function.

It also modifies the function's default handling to include/preserve the fist parameter passed to the function.

[Test link](https://web-mapviewer.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2161-log-function/index.html)

### DIFF
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -20,7 +20,7 @@ const log = function (level, ...message) {
                 console.debug(...message)
                 break
             default:
-                console.log(...message)
+                console.log(level, ...message)
         }
     }
     /* eslint-enable  no-console */
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -1,22 +1,32 @@
 import { DEBUG } from '../config'
 
 /**
- * Logs something if the env is not PROD, otherwise does nothing
+ * A logging level reference can be passed as first parameter to the log function. The levels
+ * correspond to the nativr console methods.
+ */
+export const LOGLEVEL = {
+    ERROR: 'ERROR',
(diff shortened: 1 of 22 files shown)


src/utils/logging.js
```diff
@@ -1,28 +1,48 @@
...
             default:
-                console.log(...message)
+                console.log(level, ...message)
         }
     }
     /* eslint-enable  no-console */
 }
+
+/* eslint-disable no-console */
+/** Same as console.error but doesn't log in production. */
```
pakb, line 40: this is not meant to be here no?

src/utils/logging.js
```diff
@@ -1,28 +1,48 @@
...
+log.error = (...message) => console.error(...message)
+/** Same as console.info but doesn't log in production. */
+log.info = (...message) => console.info(...message)
+/** Same as console.debug but doesn't log in production. */
+log.debug = (...message) => console.debug(...message)
```
pakb, line 45: Maybe we could call back to `log()` function here, and modify how we call `console.xxx` there. This way we only use console.log and other in the main `log()` function

fkm: Yeah. The shorthands were a bit ugly. I updated them to use our `log` function so they can respect the `DEBUG` constant.

src/utils/logging.js
fkm: Makes sense. I changed it.

pakb (approved):

PR #143, opened 2022-01-31
BGDIINF_SB-2154: Use icon anchors from backend

fkm:
Until now we used hard-coded anchors to position the icons on the map.

The backend service has been updated to provide an anchor for each icon and this change updates the code to make use of this information.

This PR also includes the fix for the missing anchors.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2154-icon-anchor-from-backend/index.html)

### DIFF
--- a/src/api/icon.api.js
+++ b/src/api/icon.api.js
@@ -13,10 +13,11 @@ export class IconSet {
 }
 
 export class Icon {
-    constructor(name, imageURL, imageTemplateURL) {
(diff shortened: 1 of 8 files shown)


pakb (approved):
I'm just checking before merging if, we some of your E2E test fixes (waiting on icons etc...) we could re-activate some tests that I disabled after mobile layout work / vue3 migration

PR #145, opened 2022-02-02
Rework HTML hierarchy

pakb:
Merges some modules (related to the main UI/menu) together in order to simplify the placement of HTML elements in the DOM. This way it is possible to achieve a fluid placement of the tools on the UI (no more absolute placement)

[Test link](https://web-mapviewer.dev.bgdi.ch/feat-jira-bgdiinf_sb-2160_rework_html_hierarchy/index.html)


src/App.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-    <div id="app">
+    <div id="main-component">
```
fkm, line 2: :sunglasses:

src/modules/menu/MenuModule.vue
```diff
@@ -119,70 +124,107 @@ export default {
...
-        .menu-tray-content {
-            transition: opacity 0.3s;
-        }
-        max-width: 22rem;
+    height: 100%;
+    // so that the user can click through this element (and we don't block interaction with the map)
+    pointer-events: none;
+    & > * {
+        // re-activate interaction with all children of the menu
+        pointer-events: all;
```
fkm, line 140: This is a bit hacky. But I don't have anything better on hand.
I'll create a PR when/if I have a better idea.

src/modules/store/plugins/drawing-overlay-menu-management.plugin.js
```diff
@@ -5,17 +5,10 @@ import { UIModes } from '@/modules/store/modules/ui.store'
...
         if (mutation.type === 'setShowDrawingOverlay') {
-            store.dispatch('toggleHeader')
-            store.dispatch('toggleZoomGeolocationButtons')
-            if (state.ui.showDrawingOverlay) {
-                // keep overlay's callback otherwise the menu will not know what to do later
-                // when it will be back on the screen and the user clicks on the overlay
-                store.dispatch('hideOverlayIgnoringCallbacks')
-            } else if (state.ui.mode === UIModes.MENU_OPENED_THROUGH_BUTTON) {
+            if (state.ui.mode === UIModes.MENU_OPENED_THROUGH_BUTTON) {
                 // store.dispatch('toggleMenuTray')
```
fkm, line 12: I assume the dispatch will be uncommented at some point. Otherwise, let's kill the plugin :smiley:

src/scss/webmapviewer-bootstrap-theme.scss
```diff
@@ -11,4 +11,5 @@ $primary: $red;
   @each $key, $value in $grid-breakpoints {
     #{unquote($key)}: $value;
   }
+  test: $red;
```
fkm, line 14: Is that an artifact from a test you did or is this meant to be here?

fkm (approved):
Mostly good stuff! :smiley: 
The rest can be 'fixed' in post.

src/modules/menu/MenuModule.vue
pakb: yeah, I wanted to find a way to let event go through a div that would take the entire screen... Maybe there's a better way of placing elements in this div so that it doesn't take so much space

src/modules/store/plugins/drawing-overlay-menu-management.plugin.js
pakb: Yep, good catch, this one can die

src/scss/webmapviewer-bootstrap-theme.scss
pakb: oops, yeah I can also remove the part above, as with Vue3 the webpack config has been changed and SCSS files aren't compiled to JS anymore (I couldn't find an easy way to reactivate that, so instead I added the threshold in `config.js`...)

src/modules/menu/MenuModule.vue
fkm: Yeah. I'm sure there's a way to get the sizing we want without the container. But I'm drawing a blank at the moment 🙈

PR #149, opened 2022-02-07
BGDIINF_SB-2110: Fixed drawing cursor/tooltip

fkm:
Currently the cursor is incorrect when hovering over features in the drawing view.

* Unselected marker has a drag cursor/tooltip.
* Tooltip considers visible shape while clicks don't.
* Polygon areas show drag cursor/tooltip.
* Marker sometimes shows wrong tooltip (new vertex)

This commit fixes all these problems and does some minor refactoring along the way.

Unfortunately, there is still a discrepancy betweeen the cursor/tooltip and the Modify/Select interactions. I tried playing around with their tolerance properties but couldn't make them behave identically. I created the ticket BGDIINF_SB-2176.


ltshb:
@fkm there is still a wrong cursor when a marker/text/... is selected and we hover it the cursor stay as normal pointer.

PR #150, opened 2022-02-07
BGDIINF_SB-2170: Changed layer copyright position

fkm:
The problem with the force-click in the test had already been addressed via a z-index. But that hides the copyright notice if the profile is present (minimized or not).

This change positions the copyright notice above the profile so that it is always in the bottom right-hand corner of the visible part of the map. That not only makes Cypress happy but also copyright holders.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2170-copyright-position/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -22,12 +22,14 @@
             @change="triggerKMLUpdate"
         />
         <div ref="draw-help" class="draw-help-popup"></div>
-        <ProfilePopup
-            :feature="selectedFeature"
-            :ui-mode="uiMode"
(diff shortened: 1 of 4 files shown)


fkm:
And some screenshots:

copyright-no-profile copyright-min-profile copyright-max-profile

pakb (approved):
As we've discussed, it would be good to later move those copyrights in a Footer component, so that we can get rid of this teleport.
Active layers are already described by the store, so there's no need of a link to the map component. But we can do that in a separate task.

Edit: task created for that : https://jira.swisstopo.ch/browse/BGDIINF_SB-2179

PR #152, opened 2022-02-08
Updating all libs to latest release

pakb:


tests/e2e/plugins/index.js
```diff
@@ -25,11 +25,13 @@ module.exports = (on, config) => {
             })
         },
         deleteFolder(folderName) {
+            // eslint-disable-next-line no-console
             console.log('deleting folder %s', folderName)
 
             return new Promise((resolve, reject) => {
                 fs.rmdir(folderName, { maxRetries: 10, recursive: true }, (err) => {
                     if (err) {
+                        // eslint-disable-next-line no-console
```
ltshb, line 34: would it not be better to have the `// es-lint-disable no-console` at the top of the file instead ?

tests/e2e/specs/drawing/marker-text.spec.js
```diff
@@ -10,7 +10,7 @@ const drawingStyleDescription = '[data-cy="drawing-style-feature-description"]'
 
 const drawingStyleMarkerButton = '[data-cy="drawing-style-marker-button"]'
 const drawingStyleMarkerPopup = '[data-cy="drawing-style-marker-popup"]'
-const drawingStyleMarkerShowAllIconsButton = '[data-cy="drawing-style-show-all-icons-button"]'
+// const drawingStyleMarkerShowAllIconsButton = '[data-cy="drawing-style-show-all-icons-button"]'
```
ltshb, line 13: Could this be removed ?

tests/e2e/plugins/index.js
pakb: Maybe for this one, but then I don't want any other console usage to be permitted anywhere else

pakb:
Regarding tests on the CI, I'm very close to kill two tests regarding drawings (kml.spec.js and marker-text.spec.js) which are failing in an erratic way on the CI, but go green whenever I run them locally.
I even tried to run them on a docker image (the same as the CI uses) and everything went smoothly 4 time in a row.

So I've reactivated (temporarily) the ability to bypass a CI green to merge, and whenever tests are failing on the CI I'm checking if it is those two that are failing.
If it's the case, I'm just ignoring it...

fkm (approved):

PR #153, opened 2022-02-08
BGDIINF_SB-2010: add a test case to cover case

pakb:
There was nothing to fix regarding what was described in this ticket, so here's a e2e test to be sure it won't come back later on.
If the app is loaded with a layer defined in the URL, but without a background, the bg layer of the topic is correctly applied (no void layer)


pakb:
Tests are failing because of the same reason stated here : https://github.com/geoadmin/web-mapviewer/pull/152#issuecomment-1034658383

fkm (approved):

PR #154, opened 2022-02-11
BGDIINF_SB-2179: Reorganized and updated footer components

fkm:
We currently have the problem that we collect more and more teleports.

This commit reorganizes all footer related components (except the infobox and profile) and updates them to make use of Vuex instead of values provided by parent components.

With the reorganisation also came a new HTML structure and the CSS to go with it.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2179-new-footer-structure/index.html)


fkm:
As this is a rather large change I made some screenshots to show the important aspects.

# Small, portrait
01-small-portrait-default 02-small-portrait-hover 03-small-portrait-wheel 04-small-portrait-fullscreen 05-small-portrait-profile-min 06-small-portrait-profile-min-wheel 07-small-portrait-profile-max

# Small, landscape
01-small-landscape-default 02-small-landscape-hover 03-small-landscape-wheel 04-small-landscape-fullscreen 05-small-landscape-profile-min 06-small-landscape-profile-min-wheel 07-small-landscape-profile-max

# Big, landscape
01-big-landscape-default
02-big-landscape-hover
03-big-landscape-wheel
04-big-landscape-fullscreen
05-big-landscape-profile-min
06-big-landscape-profile-min-wheel
07-big-landscape-profile-max

src/modules/map/store/map.store.js
```diff
@@ -60,6 +62,18 @@ export default {
          *   null, no pin will be shown.
          */
         pinnedLocation: null,
+        /**
+         * A reference to the OpenLayers map instance.
+         *
+         * @type {ol/Map}
+         */
+        instance: null,
```
pakb, line 70: I'm not sure this is a good idea. If Vuex wraps anything it receive in some kind of deep proxy, this will kill a lot of performance.
Maybe you could add the footer module in the slot of the map module, so that you can receive the map instance through inject (see how the DrawingModule does)

src/modules/menu/scss/topics-image.scss
```diff
@@ -1,3 +1,36 @@
...
+  (key: cadastre,      position: -220px -300px),
+  (key: meteoschweiz,  position: -220px -360px),
+  (key: swisstopo,     position: -330px 0),
+  (key: ivs,           position: -330px -60px),
+  (key: wildruhezonen, position: -330px -120px),
+  (key: inspire,       position: -330px -180px),
+  (key: geothermie,    position: -330px -240px),
+  (key: astra,         position: -330px -300px),
+  (key: bfs,           position: -330px -360px),
+);
```
pakb, line 32: 😎

pakb (commented):
A little detail, don't know if it easy to fix : 
would it be possible to set the z-index (or place in the DOM) of the profile box so that it covers the zoom/geolocation button and is not underneath?

src/modules/map/store/map.store.js
fkm: That should be doable. I'll check.

fkm:
> A little detail, don't know if it easy to fix : would it be possible to set the z-index (or place in the DOM) of the profile box so that it covers the zoom/geolocation button and is not underneath?

Good point! I'll see what I can do.

pakb (approved):

PR #155, opened 2022-02-14
Adding 3 retries on two tests

pakb:
Thoses tests are very often problematic on the CI, trying to see if a couple retry would fix our issue with it.


fkm (approved):
Interesting idea 😎 Let's go for it!

pakb:
I'm wondering if I shouldn't have that set globally, it's possible to have as many retries as you want while `cypress run` (which is what our CI is doing) but no retry when we `cypress open` (which is what we do when we have the GUI open)...

PR #157, opened 2022-02-15
BGDIINF_SB-2132: Changed text-add icon (drawing)

fkm:
We currently use a plus icon for the text-add button which doesn't
communicate the purpose of the button as the other buttons from the
drawing toolbox also add things.

This changes the icon from a "+" to a "T".

More information about the choice can be found in the Jira ticket.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2132-drawing-text-icon/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingToolboxButton.vue
+++ b/src/modules/drawing/components/DrawingToolboxButton.vue
@@ -2,19 +2,18 @@
     <ButtonWithIcon
         small
         outline-light
-        :button-title="drawingModeToLabel(drawingMode)"
+        :button-title="buttonLabel"
         :danger="isActive"
         :button-font-awesome-icon="buttonIcon"
         :icons-before-text="true"
         direction="column"
-        :style="buttonStyle"
+        :class="buttonClasses"
         :data-cy="`drawing-${drawingMode.toLowerCase()}`"
         @click="emitSetDrawingMode"
     />
 </template>
 
 <script>
-import { mapState } from 'vuex'
 import { drawingModes } from '@/modules/store/modules/drawing.store'
 import ButtonWithIcon from '@/utils/ButtonWithIcon.vue'
 import { UIModes } from '@/modules/store/modules/ui.store'
@@ -30,12 +29,13 @@ export default {
             type: Boolean,
             default: false,
         },
+        uiMode: {
+            type: String,
+            default: UIModes.MENU_ALWAYS_OPEN,
+        },
     },
     emits: ['setDrawingMode'],
     computed: {
-        ...mapState({
-            uiMode: (state) => state.ui.mode,
-        }),
         buttonIcon() {
             switch (this.drawingMode) {
                 case drawingModes.LINE:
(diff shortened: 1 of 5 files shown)


davidoesch:
nice find ! ok for me

src/modules/drawing/components/DrawingToolboxButton.vue
```diff
@@ -27,7 +33,10 @@ export default {
     },
     emits: ['setDrawingMode'],
     computed: {
-        buttonIcon: function () {
+        ...mapState({
+            uiMode: (state) => state.ui.mode,
+        }),
```
pakb, line 38: Could you move this Vuex binding to the parent component (the one that is already bound with it) and pass that as a prop?
We try to keep at a minimum the amount of component that are linked to the store, typically only binding the store on the top component of each modules.

src/modules/drawing/components/DrawingToolboxButton.vue
```diff
@@ -40,9 +49,24 @@ export default {
...
         },
+        buttonStyle() {
+            return {
+                // The buttons having different/dynamic widths looks unintentional.
+                // While inline styles are bad, there is no Bootstrap class we can
+                // apply to the component and this width is too use-case specific
+                // to be added as a class in the ButtonWithIcon component.
+                width: this.uiMode === 'MENU_ALWAYS_OPEN' ? '4.5em' : '',
+            }
+        },
```
pakb, line 60: I would prefer that we set a custom class dynamically, and then have this 4.5rem defined as a CSS property for this class (if we can avoid mixing CSS and JS, it's better)

Or specify something in ButtonWithIcon. As you've stated you felt it's too specific to do so, but I wouldn't mind an extra flag to cover this case if it's too complicated with a class/CSS approach (or it requires some !important)

src/modules/drawing/components/DrawingToolboxButton.vue
fkm: Ok 😃

src/modules/drawing/components/DrawingToolboxButton.vue
fkm: I'll see what I can do 👍

src/utils/ButtonWithIcon.vue
```diff
@@ -131,6 +123,9 @@ export default {
         background: none;
         border: none;
     }
+    &-w5 {
+        width: 5em;
+    }
```
pakb, line 128: Shouldn't that be in the DrawingToolboxButton component? As this style is scoped I think this will not be applied if the class is not applied in this component.

fkm: This works. But thank you for the question! Because I thought having the styles in the DrawingToolboxButton wouldn't work because of the scoping. I very much prefer to have the styles with the toolbox instead of the icon-button. I'll change it 😎

fkm: Done 😃

pakb (approved):

PR #159, opened 2022-02-18
BGDIINF_SB-2212: Made position content selectable

fkm:
Currently, the content of the position overlay is not selectable.

This seems to have been broken in OpenLayers 4.1 and since then we need to add an extra class to make the content selectable again.

I added the class and a comment and the content is now selectable.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2212-position-selectable-content/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -1,5 +1,4 @@
 <template>
-    <!-- Stop right click propagation to allow the user to interact with the popup. -->
     <div v-if="isRightClick" class="location-popup" data-cy="location-popup" @contextmenu.stop>
         <div class="card">
             <div class="card-header d-flex">
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -164,7 +164,9 @@ export default {
         this.overlay = new Overlay({
             offset: [0, 15],
             positioning: OverlayPositioning.TOP_CENTER,
-            className: 'location-popup-overlay',
+            // Selection of overlay content was broken in OL v4.1 so we need an extra class.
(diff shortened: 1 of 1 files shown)


ltshb (approved):
Very nice.
I noticed also two other issues with this tooltip, the first one might be also resolved with this PR ?

1. Tooltip is missing the small triangle at the top to display the pixel that has been clicked
2. Clicking on the right button on the tooltip don't open the context menu to be able to copy the selected text. Don't know if this can be easily done and the issue is also on the old viewer, but might be good to have a ticket to fix it if possible ?

fkm:
Found a solution for both. The pop-up now has an arrow on top that points to the clicked position on the map and a right click doesn't spawn a new location pop-up anymore but shows the context menu.

_I just realized that I forgot to run the tests again locally. Let's hope and see :smile:_

pakb:
I think there's something that should be edited, the location popup isn't placed anymore and is under the menu 
image

fkm:
Turns out that adding comments can break code :smile: 

The HTML comment about the `@contextmenu` I added broke `$el` (`this.overlay.setElement(this.$el)`.

pakb (approved):

ltshb (commented):
Great thanks. I found other issues with the content popup but nothing related to this, I'll open new jira ticket.

pakb:
> Turns out that adding comments can break code 😄
> 
> The HTML comment about the `@contextmenu` I added broke `$el` (`this.overlay.setElement(this.$el)`.

Maybe it would be better to use some `ref` next time, you might be able to keep comments with that

fkm:
> Maybe it would be better to use some `ref` next time, you might be able to keep comments with that

Good idea! :sunglasses:

PR #163, opened 2022-02-22
BGDIINF_SB-2226: Fixed position of share link

fkm:
The link to share the current position points to the center of the
map. This is because the link uses the current URL as-is instead
of the ClickInfo position.

This replaces lat/lon when generating the URL so that the marker
is at the expected position when opening the link.

[Test link](https://web-mapviewer.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2226-share-link-clickinfo/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -90,6 +90,7 @@ import { registerWhat3WordsLocation } from '@/api/what3words.api'
 import { requestHeight } from '@/api/height.api'
 import { generateQrCode } from '@/api/qrcode.api'
 import { round } from '@/utils/numberUtils'
+import stringifyQuery from '@/router/stringifyQuery'
 
 /** Right click pop up which shows the coordinates of the position under the cursor. */
 export default {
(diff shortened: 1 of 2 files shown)


src/modules/map/components/LocationPopup.vue
```diff
@@ -143,7 +144,14 @@ export default {
...
         shareLinkUrl: function () {
-            return `${window.location}&crosshair=marker`
+            let [lon, lat] = this.reprojectClickCoordinates('EPSG:4326')
+            let query = {
+                ...this.$route.query,
+                crosshair: 'marker',
+                lat,
+                lon,
+            }
+            return `${location.origin}/#/map?${stringifyQuery(query)}`
```
ltshb, line 154: Just a newbea javascript question, why not using `window.location.origin` here ?

ltshb (commented):
Very nice. 
While testing I've found another issue (which also exists in the old viewer), the qrcode generated doesn't contain the crosshair=marker in the url, I think it would be nice that the qrcode generates the same url as the share link.
I've also noticed that the qrcode is not reactive, it is generated once, if we then change the layer configuration the cqrcode is not re-generated with the new params.

src/modules/map/components/LocationPopup.vue
fkm: I'll see what I can do about the QR code 😃 

Concerning the use of `window`. There is no definitive answer. Generally speaking you never need to write `window` as every variable that is not defined will be set/looked up on the global scope (i.e. window).

So if you write `foo = 1` somewhere in your code (without having defined `foo` somewhere), you will set `window.foo`. The same goes for reading variables.

So. Whether or not you write `window` comes down to convention and/or preference. I generally avoid `window` unless the code is not clear otherwise. In essence, I'll write window for custom variables on the global scope but not for existing properties from the browser APIs (location, document, console, localStorage, ...). But I make exceptions in cases like `window.innerHeight`, etc.

The MDN pages of the window properties sometimes say what the convention is. But I'm not really going after that anymore. But it's a good start until you have your own preferences.

ltshb: Thanks for the explaination :-)

pakb (approved):

PR #164, opened 2022-02-23
Replacing in-house print code with lib

pakb:
As this will be much more reliable with future browser changes. The current in house code wasn't working anymore on the latest Chrome version anyway.

[Test link](https://web-mapviewer.dev.bgdi.ch/replace_print_by_lib/index.html)


fkm (approved):
👍

PR #165, opened 2022-02-23
BGDIINF_SB-2159: add a new popover OL component

pakb:
Use this new popover in the LocationPopup component
Use this new popover to show highlighted features on the map (with absolute positioning on the coordinates of the feature)
Add in this popover for features a button (in header) that enables the user to place the tooltip at the bottom of the app (with a grid system that gets wider as the screen gets bigger)
Switch to this footer tooltip when on mobile (and back to floating tooltip if going back to desktop)
Removing the usage of the swipe tray (needs more work if we want to use it)

[Test link](https://web-mapviewer.dev.bgdi.ch/feat-jira-bgdiinf_sb-2159_highlighted_feature_overhaul/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -1,17 +1,13 @@
 <template>
-    <div v-if="isRightClick" class="location-popup" data-cy="location-popup" @contextmenu.stop>
-        <div class="card">
-            <div class="card-header d-flex">
-                <span class="flex-grow-1 align-self-center">
-                    {{ $t('position') }}
-                </span>
-                <ButtonWithIcon
-                    data-cy="profile-popup-close-button"
-                    :button-font-awesome-icon="['fa', 'times']"
-                    @click="onClose"
-                />
-            </div>
-            <div class="card-body coordinates-list text-start">
+    <div v-if="isRightClick" class="location-popup" @contextmenu.stop>
+        <OpenLayersPopover
+            v-if="clickCoordinates"
+            data-cy="location-popup"
+            :title="$t('position')"
+            :coordinates="clickCoordinates"
+            @close="clearClick"
+        >
+            <div class="coordinates-list text-start">
                 <div>
                     <a :href="$t('contextpopup_lv95_url')" target="_blank">CH1903+ / LV95</a>
                 </div>
@@ -78,27 +74,28 @@
                     />
                 </div>
             </div>
-        </div>
+        </OpenLayersPopover>
     </div>
 </template>
 
 <script>
 import proj4 from 'proj4'
 import { mapState, mapActions } from 'vuex'
-import { ClickType } from '@/modules/map/store/map.store'
 import Overlay from 'ol/Overlay'
 import OverlayPositioning from 'ol/OverlayPositioning'
-import { printHumanReadableCoordinates, CoordinateSystems } from '@/utils/coordinateUtils'
-import ButtonWithIcon from '@/utils/ButtonWithIcon.vue'
+
 import { registerWhat3WordsLocation } from '@/api/what3words.api'
 import { requestHeight } from '@/api/height.api'
 import { generateQrCode } from '@/api/qrcode.api'
+import { ClickType } from '@/modules/map/store/map.store'
+import OpenLayersPopover from '@/modules/map/components/openlayers/OpenLayersPopover.vue'
+import { printHumanReadableCoordinates, CoordinateSystems } from '@/utils/coordinateUtils'
 import { round } from '@/utils/numberUtils'
 import stringifyQuery from '@/router/stringifyQuery'
 
 /** Right click pop up which shows the coordinates of the position under the cursor. */
 export default {
-    components: { ButtonWithIcon },
+    components: { OpenLayersPopover },
     inject: ['getMap'],
     data() {
         return {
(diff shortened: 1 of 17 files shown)


src/modules/map/components/LocationPopup.vue
```diff
@@ -1,17 +1,13 @@
...
-                    {{ $t('position') }}
-                </span>
-                <ButtonWithIcon
-                    data-cy="profile-popup-close-button"
-                    :button-font-awesome-icon="['fa', 'times']"
-                    @click="onClose"
-                />
-            </div>
-            <div class="card-body coordinates-list text-start">
+    <div v-if="isRightClick" class="location-popup" @contextmenu.stop>
```
fkm, line 2: Side note. I tried to use a `ref` so I can add the HTML comment. But that didn't work. No idea what I did wrong :smile:

src/modules/map/components/openlayers/OpenLayersMap.vue
```diff
@@ -33,11 +33,25 @@
         />
         <!-- Adding highlighted features -->
         <OpenLayersHighlightedFeature
-            v-for="(feature, index) in highlightedFeatures"
+            v-for="(feature, index) in selectedFeatures"
             :key="`${index}-${feature.id}`"
```
fkm, line 37: We don't need to change this. But I'm curious. Is the `feature.id` necessary?

src/modules/store/modules/ui.store.js
```diff
@@ -68,43 +68,72 @@ export default {
...
+        },
+        toggleLoadingBar({ commit, state }) {
+            commit('setShowLoadingBar', !state.showLoadingBar)
+        },
+        toggleDrawingOverlay({ commit, state }) {
+            commit('setShowDrawingOverlay', !state.showDrawingOverlay)
+        },
+        toggleFloatingTooltip({ commit, state }) {
+            commit('setFloatingTooltip', !state.floatingTooltip)
+        },
```
fkm, line 108: :sunglasses:

src/modules/store/modules/ui.store.js
```diff
@@ -68,43 +68,72 @@ export default {
...
+        setShowDrawingOverlay(state, flagValue) {
+            state.showDrawingOverlay = flagValue
+        },
+        setFloatingTooltip(state, flagValue) {
+            state.floatingTooltip = flagValue
+        },
+        setUiMode(state, mode) {
+            state.mode = mode
+        },
```
fkm, line 137: :sunglasses:

src/utils/EventEmitter.class.js
```diff
@@ -0,0 +1,40 @@
...
+        this._getEventListByName(eventName).add(fn)
+    }
+
+    once(eventName, fn) {
+        const self = this
+
+        const onceFn = function (...args) {
+            self.removeListener(eventName, onceFn)
+            fn.apply(self, args)
+        }
```
fkm, line 25: An arrow function would be nice here.

fkm (approved):
Looks good! Can be merged.

src/modules/map/components/openlayers/OpenLayersMap.vue
pakb: not required per say, but `v-for` wants to have unique keys for each element of the collection. So the easier way is to ask each feature to have an ID (or to build it through the index of the loop and some other string?!)

src/utils/EventEmitter.class.js
pakb: you mean so that I could use `this` inside instead of `self`?

src/modules/map/components/openlayers/OpenLayersMap.vue
fkm: I don't know enough yet about how Vue handles its components. What I wonder is if there is a situation where the index of an object changes. I guess that would be possible with splice or something similar.

src/utils/EventEmitter.class.js
fkm: Yeah. I guess the code is copy/pasted. And it's ok to keep close to the original.
Just something I stumbled over while reading it :smiley:

src/modules/map/components/openlayers/OpenLayersMap.vue
pakb: yep that could, the goal here was to have a unique identifier for each features, and as stated in Feature class JSDoc, it's possible to have to distinct features (unrelated, from different layers or sources) that have the same ID, and that would be valid.
So to make sure it doesn't happen, I've appened the index (of the loop) to the feature ID, I could also use the layer id (as now I've access to it in the feature metadata)

src/utils/EventEmitter.class.js
pakb: > Yeah. I guess the code is copy/pasted.

Nope

src/modules/map/components/LocationPopup.vue
pakb: that is perplexing...

fkm: Indeed :smile: Try it when your working on the file anyway. It's very well possible that I just did some newbie mistake.

src/utils/EventEmitter.class.js
fkm: :laughing:

src/modules/map/components/openlayers/OpenLayersMap.vue
fkm: :+1:

ltshb (commented):
When testing I noticed two issues:

1. The location popu (right click on the map) has now a scrolling bar and you have to scroll done to see the entire qrcode ! When testing the PR #166 it was not the case.
2. When putting a tooltip down on the screen, which is a very nice feature by the way, you also have sometimes unneeded scrollbar.

pakb:
ok thanks for the feedback! I'll open a ticket with this two issues

pakb:
Thanks for the feedback, I've created a ticket to address that (as this PR is already merged...) https://jira.swisstopo.ch/browse/BGDIINF_SB-2245

PR #167, opened 2022-02-25
BGDIINF_SB-2236: Hid mouse position on mobile

fkm:
As we don't have a mouse (hover) on mobile devices
we can hide the mouse position in the footer.

This makes that change and includes some improvements
to the footer's CSS to make it more flexible to cope
with the disappearing mouse position and map scale.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2236-hide-mouse-position/index.html)

### DIFF
--- a/src/modules/map/components/footer/MapFooter.vue
+++ b/src/modules/map/components/footer/MapFooter.vue
@@ -12,18 +12,15 @@
             <!-- Infobox, Profile, ... -->
         </div>
         <div class="map-footer-bottom">
-            <div class="map-footer-bottom-left">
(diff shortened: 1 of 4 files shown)


pakb (commented):
Why not hide also the projection selector? I think it only relates to this mouse position.
I would maybe also lower the threshold to `md` instead of `lg`, otherwise smaller desktop screen won't have it

fkm:
Good point! I'll remove it as well.

pakb (approved):

PR #168, opened 2022-02-25
BGDIINF_SB-2232: Removed fake tell-cypress requests

fkm:
Until now we had some fake requests that we intercepted
in order to be able to wait for certain events.

This removes those requests and replaces the waiting with
the Cypress plugin waitUntil. This also removes the warning
about this being a test site. The warning obscured a button
and we should test for what we deploy in production.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2232-dont-tell-cypress/index.html)


pakb (approved):
Nice!

PR #170, opened 2022-02-28
BGDIINF_SB-2237: Fixed missing update on URL change

fkm:
While the underlying link did update properly the QR code generation wasn't triggered.

This fixes that by watching the current route for changes and adds an e2e test to make sure this doesn't happen anymore.

[Test link](https://web-mapviewer.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2237-update-on-url-change/index.html)


pakb (approved):
Good catch!

PR #172, opened 2022-03-02
BGDIINF_SB-2245: Fixed height issues for context boxes

fkm:
The updated context box has some height issues where it
shows a scrollbar when at the bottom and only one row of
items. And the box is too long when floating because of
the QR code.

* Added fixed width (CSS) to pop-over.
* Added dynamic max-height to bottom drawer.
* Added scrollTo top on highlighted features change.
* Added toggle between QR-code and text to location.
* Added test for the QR-code / text-info toggle.
* Added tests for infobox module.
* Fixed missing context-menu propagation stop.
* Fixed styling (scoping) of highlighted features.
* Fixed conditions to avoid unnecessary requests.
* Fixed some bugs in GeoJSON layer component.
* Some refactoring along the way

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2245-fix-sizing-issues/index.html)


pakb (approved):
We will see if this new toggle between QR code and plain text is well received 😉 
Just some minor changes (maybe) but overall very good!

fkm:
> We will see if this new toggle between QR code and plain text is well received 

True :smile: Especially, as the toggle button is rather small.

src/modules/map/components/openlayers/OpenLayersMap.vue
```diff
@@ -253,11 +259,11 @@ export default {
                     const olLayer = this.map
                         .getLayers()
                         .getArray()
-                        .find((layer) => layer.get('id') === geoJsonLayer.id)
+                        .find((layer) => layer.get('id') === geoJsonLayer.getID())
```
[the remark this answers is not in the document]
fkm: No problem!

tests/e2e/specs/infobox.spec.js
```diff
@@ -0,0 +1,107 @@
...
+
+                        expect($element.height()).to.be.equal(maxHeight)
+                    })
+                })
+            }
+        )
+    })
+})
+
+var htmlPopup = `
```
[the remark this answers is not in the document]
fkm: The reasoning behind it was hoisting. But thinking about it now it isn't necessary as the reference is only used within a function. But putting it into a fixture file would be much better! I'll move it.

tests/e2e/specs/infobox.spec.js
```diff
@@ -24,7 +25,7 @@ describe('The infobox', () => {
                 beforeEach(() => {
                     const layer = 'test.wmts.layer'
                     cy.intercept('**/MapServer/identify**', features)
-                    cy.intercept('**/MapServer/**/htmlPopup**', htmlPopup)
+                    cy.intercept('**/MapServer/**/htmlPopup**', htmlPopupTable)
```
pakb, line 28: Oops I didn't see that before, to use fixtures with Cypress you don't need to import them
```
    cy.intercept('**/MapServer/**/htmlPopup**', {
        fixture: 'html-popup.fixture',
    })
```
but then you need to modify the fixture file so it is not a JS file but a plain text file (or more a HTML file in your case)

pakb: It would then be better to name it `html-popup.html` though...

fkm: True. I forgot about the intercept property. Thank you! I'll update it.

So. Not `html-popup.fixture.html`?

pakb: I don't know if it makes sense, it's already located in the `fixtures` folder anyway. See if the other files there have a `.fixture.` part, if not there's no need 😉

fkm: :+1:

tests/e2e/specs/infobox.spec.js
```diff
@@ -0,0 +1,80 @@
...
+        context(
+            `viewport: ${viewport}`,
+            {
+                viewportWidth: dimensions.width,
+                viewportHeight: dimensions.height,
+            },
+            () => {
+                beforeEach(() => {
+                    const layer = 'test.wmts.layer'
+                    cy.intercept('**/MapServer/identify**', features)
```
pakb, line 26: Sorry to come back again, but could you also change how this fixture is imported/used 😖

fkm: The thing here is that I need to import the features anyway because I need to only send one feature for the intercept on line 30.

pakb: Then you need to access the fixture in a `cy.fixture` method.
see https://docs.cypress.io/api/commands/fixture#Accessing-Fixture-Data
and an actual exemple in our code here : https://github.com/geoadmin/web-mapviewer/blob/ea5adc10dc74ac1ae694587d28784e3d90d199a3/tests/e2e/specs/layers.spec.js#L115-L130

fkm: Ah nice :sunglasses: Thank you!

pakb (approved):

PR #173, opened 2022-03-03
BGDIINF_SB-2105: Send drawings as KMZ (gzipped)

fkm:
To enable the migration of the service-kml it was requested that
we send the drawings gzipped (i.e. as KMZ).

This commit introduces pako (as pure JS library) to compress the
KML before sending it to the backend.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2105-send-as-kmz/index.html)


pakb (approved):
🚀

PR #175, opened 2022-03-03
BGDIINF_SB-1389: Updated search result styling

fkm:
This updates the search result styling to be more in line with the
implementation in mf-geoadmin3. This is not 100% identical but as
close as it gets with our current base styles.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-1389-search-results-styling/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchBar.vue
+++ b/src/modules/menu/components/search/SearchBar.vue
@@ -20,7 +20,7 @@
         <SearchResultList ref="results" @close="closeSearchResults" />
         <button
             v-show="searchQuery?.length > 0"
-            class="btn bg-transparent"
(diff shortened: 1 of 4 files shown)


pakb (approved):
I think we could move in the future the disclaimer so that it doesn't cover the header of the "location" part of the search result.
Or maybe add it in the flow of the header (or with a z-index that doesn't cover search results)

fkm:
Agreed. I'll try to come up with something that's enough in-your-face and still out-of-the-way :smiley:

fkm:
I created a ticket (BGDIINF_SB-2259).

PR #176, opened 2022-03-04
BGDIINF_SB-2228: Updated UI for link sharing

fkm:
The current implementation is somewhat misleading as it is presented as a link which must be copied via context menu.

To improve usability this commit takes a design used in mf-geoadmin3 where the link is (partially) shown and there is a button to copy the link to the clipboard

Screenshot_20220304_105713
.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2228-share-link-redesign/index.html)


fkm:
I updated the commit message. It's not reverting back. The link in an input is not used in the location pop-up but the share function in the menu.

ltshb (changes requested):
Unfortunately this break the cursor on the popup content, cursor don't change anymore to the `text` when hovering and content is not selectable anymore.
It might be also a good idea to have the copy icon also next to each coordinate ? This way any coordinates type can be easily copied. What do you think @davidoesch ?

davidoesch:
- [ ] copy icon for each value (and the values in selectable field) : excellent idea, +1
- [ ] Share link: as in mf-geoadmin3  ( user knows how to copy a link )

fkm:
Something like this?

Screenshot_20220307_100155

davidoesch:
maybe having the values in a box

fkm:
Next proposal :smiley: 

## Initial
location-popup-initial

## Hover
location-popup-hover

## Clicked
location-popup-clicked

pakb:
Good direction! I would suggest that a click anywhere on the coordinates (not necessarily on the copy button) should trigger a copy

davidoesch:
+1 for pascals comments

davidoesch:
Copy coordinates : you always want the complete part . It is even intended to ensure that you copy it completely(and don’t forget the first or last digit) so pakb proposal a clear go

copy links: Share my map. Left click on link is doable- or we use the same approach as in „share link“ as we have it today in „share „ in mr-geoadmin3

fkm:
If we make the whole thing clickable. How do we treat links? (what3works and "Share position")

I.e. what should the following do?
location-popup-link

ltshb:
@davidoesch if you think as the everyone user (not a geo user) I still think that the simplest and more intuitive is to have as follow:

1. add a copy button on hover, click the button to copy
2. cursor change to text on hover, select text that you want to copy, right click copy. This also allow to copy one part at a time to paste them in a tool that would require part separately (for example tools that have two field one for the latitude and one for the longitude, in this case is more practical to copy them separately !)
3. Nice to have: double click on a word (e.g. first coordinate x) => select the word
4. click `share position` or `what3words` go to link (new tab)
5. hover `share position` or `what3words` add copy button to copy link

fkm:
Using viewport size to determine interactions is something I generally recommend to avoid as there are people with large tablets and others (like me) who sometimes use non-maximized windows on their desktops. But pointer-events come with an information about what type of pointer triggered the event (mouse, touch, ...). I will have a look (tomorrow) if we get that information from the OpenLayers Overlay.

Apart from that. Do we need the copy buttons if I manage to fix the bug that prevents the user from selecting the text? Text selection is something that works on touch as well as with a mouse. And we wouldn't have to spam the popup with all those visible buttons on touch devices. I don't really see a way to make them pretty / less obnoxious while preserving usability 🙈

fkm:
How about this?

Small icons with light color for everyone and the border and darker icon for those with a mouse.

## Initial
location-popup-initial

## Hover
location-popup-hover

## Clicked
location-popup-clicked

fkm:
Unfortunately, `PointerEvent.pointerType` is not supported by Firefox.

ltshb:
What about always having the box and align the icon on the right inside the box ? might look a bit cleaner.

davidoesch:
"what do you think about different behavior depending on screen size ?"
a challenge for the user because ... it is different on every device he uses (see wheel to choose time for zeitreise on iOS, 'droid, mac, win ---)

So the best solution is the solution which is the same for every device. believe me---- it is as well simpler to maintain
the one in https://github.com/geoadmin/web-mapviewer/pull/176#issuecomment-1061500968 makes sense. but do make sure that selecting text works--- because this what user are used to in web: mark/select text and copy .

fkm:
Like this? :smiley: 

## Initial
location-popup-initial

## Hover
location-popup-hover

## Clicked
location-popup-clicked

pakb:
If there's a backdrop behind number, it should do something IMHO. So clicking anywhere on the backdrop color should copy the coordinates, or open the coordinates as an input in which the user can copy them himself (I prefer the simpler click = copy approach though).

ltshb:
@pakb  if you mean by `backdrop` the gray color on hover, I agree with you that it should do something. But I would say that it is not needed. For me I see 2 acceptable solutions:

1. As last example, having a box with copy icon inside right aligned (looks much cleaner as previous proposition with icons). Copy only on click of icon. Text selectable. Remove the backdrop on hover. Remove the box for the `Share Position` as we don't want to select this text and because the previous element `Elevation` doesn't have a copy icon it is not a big deal if the icon is not aligned with the others. 
Should work fine on both desktop and touch device (to be tested though) (My prefer solution 😃)
2. No icon, keeping the white box with backdrop on hover, click anywhere on the box = copy (prefered solution of @pakb 😄 ?).

If anyone agree with one of both of these solutions, then let's be democratic and vote 😄 
@fkm @davidoesch @boecklic what is you prefered solutions ?

davidoesch:
Number 2 it is. 
Icon are not understood . If we add an icon it is like in share : we would add a text like “copy coordinates”. The AB Testing in the past showed that. Important: the mark and select and copy as for every text in Internet has to work

fkm:
If we go with version 2, we still need to decide how we want to handle link(s).
Should we place the w3w link in an input too? Or drop the link and only keep the three words?

davidoesch:
- Option 2 @ltshb 

the success of map geo is based on teh fact, that we designed it for non-websavy users. even 2022. We, you and me , work daily in the web context --- you would be amazed about how many of our users ( from citizen, to gov to companies) are really struggling with 2022 tech -- there a re just to many icons out there which are not clear to a clerk, to organizer of local rally organizer etc 
The AB testing and feedback, they change, I  agree, but 

1) key point: " icons : your brain has to translate it into text then into action this is why providing the text directly is very helpful ". The "iconography of confusion" https://mapbrief.com/2013/02/19/an-iconography-of-confusion-why-map-portals-dont-work-part-iv/ is still valid today--- the less icons the better the UX. Another really oldie but goldie is "Don't Make Me Think!: A Common Sense Approach to Web Usability "https://www.amazon.de/Dont-Make-Me-Think-Usability/dp/0321344758 ( it is in my office , feel free to grab it...) which is all about this. See as well "my gmaps" the key functions have icons and text...
Even more: we are not targeting only  "the" web user  but moreover *every* person in Switzerland- a more divers community : people which have a hard time to deal with today's technology ( elderly, from different cultural bg etc) think of it as  "a very easy web viewer " The approach we follow is as in  https://www.bag.admin.ch/bag/fr/home/krankheiten/ausbrueche-epidemien-pandemien/aktuelle-ausbrueche-epidemien/novel-cov/barrierefreie-inhalte/leichte-sprache.html to represent a topic with  very simple low level communication means --- which might not  be win a design award but do let *everyone* do the job. I understand your hear as a dev: lets do sth very nice and visually nice--- but with icons only map portals: there are enough out there: we as gov agency fill the gap to offer a tool for all which are overwhelmed by the latest design.

2)  the *only* complaint we had  for the tooltip was:  "sometimes users do not copy all the Digits because they did not select  *all* the digits" , we had the same issue in "Share " (parts of the shortlink not copied)  this is why we have  today  there 3 measures in place to ensure the complete URL is taken into account:

- preselect the short url when openeing share (we can't do that in the tooltip, since we have multiple coords)
- the "copy link" button  (I agree, it is nasty big.. but it does the job very well... it is "dont' make me think"/WYSWG: if you click it, you have the link , so copy bas text button would to the job)
- Contextual Help coming up " Press Ctrl+C to copy" . This we can do with a mouseover on desktop

Hope this helps.... 

@fkm w3w link
-> we can omit the link: we already do today: the main purpose of providing the link is: to copy it ( and not to link to w3w)

@pakb QR Code
if we have the shortened QR code: it will be smaller no need to provide a link to a new tab. we can add it at the bottom for desktop (there is enough space on the screen) and not provide it in the mobile view. As today
the big advantage
- one click less to get to the QR code (which is priceless)
- one link less in the tooltip and a clear promotion of: hey you can scan it and get a mobile version of what you see

fkm:
Next proposal :smiley: 

As we don't have much space horizontally as well as vertically I didn't find a way to use text-buttons for the values that should be completely visible (everything except the link). My proposal uses `user-select: all` to automatically select the values on click. This enforces that the whole value is selected and makes copying the values via right click (context menu) easy.

*I pushed these changes so you will be able to try this out on your devices when the pipeline finishes.*

## Values selected on click
All values. Works for left click / touch as well as a context menu click (right click / long tap).
location-popup-userselect

## Tooltip on focus
When clicking on the input the address will be selected and a tooltip is shown.
location-popup-tooltip

## Copy via button
Text button will copy the address and change text for a second (as feedback).
location-popup-copied

ltshb:
I ve just find out that the location popup doesn’t work on mobile (iPhone Safari) not related to these changes though, it doesn’t work on the current develop deployment

pakb:
> I ve just find out that the location popup doesn’t work on mobile (iPhone Safari) not related to these changes though, it doesn’t work on the current develop deployment

yep : https://jira.swisstopo.ch/browse/BGDIINF_SB-2235 😉

pakb:
This `user-select: all` is pretty good looking!
I would even go a step further (I know, I'm repeating myself...) and straight copy the content as soon as the user has clicked it.
No need for an icon IMHO, but maybe some feedback (same as with the link copy) that the stuff selected has been copied.
What do you think @fkm ?

ltshb:
The current version deployed in the test link for coordinates is IMHO very confusing, you got the Text cursor but cannot select text like you are used to select with this cursor ! You cannot select only one part of the coordinate and you still need to press CTRL+C to copy. So no plus in comparison to the old viewer. Here below the old viewer functionality for coordinates:

- double click on coordinate => select only the part (latitude or longitude)
- three click on coordinate => select the whole coordinate

For the share position I'm ok with this solution.

@fkm I found an issue with the `copy link` button, it is not reactive and the language doesn't change when changing the website language.

fkm:
@pakb I like the idea!

@ltshb Do you still have the problem with the missing reactivity? I just tested it and it works for me. But it might be that the link wasn't updated when you tried. If you still have the problem I'll have to have another look.

fkm:
@ltshb I was able to reproduce the reactivity problem and fixed it. I also managed to find a proper fix for the selection problem.

@everyone I recommend that we go with the first proposal but with the text-button and toolip from the last proposals.

The popup/overlay is already too wide for mobile devices where some labels wrap to a new line. We could reduce the font-size but that doesn't help too much as some of the gains are 'used' to recover the wrapped labels.

And I'm also not sure if we find a solution to what should be copied. Is it the height in meters or feet, WGS84 in DD or DMS in lon/lat or lat/lon, etc.

## Default
location-default

## Focus
location-focus

## Clicked
location-clicked

davidoesch:
regarding  fontsize/width of pop up: try to use the same font and size /padding as in current mf-geoadmin3. it wokes -- no complaint so far.... and on desktop with shorted URL the QR fits as well below...

fkm:
Either GitHub had some problems Friday evening or I posted my reply somewhere else.

In any case here the latest proposal with the small font.

## Default
location-default

## Clicked
location-clicked

davidoesch:
I like
Try to reduce padding/space/fontsize So we have approximately the same space used by the new position box as in the old one. And the shortened QR will fit in the window. One click less for the user

davidoesch:
Try to reduce padding/space/fontsize So we have approximately the same space used by the new position box as in the old one. And the shortened QR will fit in the window. One click less for the user
image

fkm:
Et voilà :smiley: 

Even works on an iPhone SE. But I'll add a media query looking at the viewport height to hide the code when there's not space.

Screenshot_20220322_094201

davidoesch:
Size: 
- QR Code: if u use the shortened version: it will be way smaller
- "Position" bar: reduce height (too much space above and below)

Share position
add before the link /under elevation : "share position" so the user knows what he gets with the  link

fkm:
Good catch with the title size. Thank you! I set the font-size too high up in the tree.

I restored the original font-size and [created a new ticket](https://jira.swisstopo.ch/browse/BGDIINF_SB-2309) for that. The max-width of the component should probably also be less wide. But the component is shared with the highlighted features.

## Full
location-full

## No QR code when viewport not heigh enough
location-height

## No buttons when viewport not wide enough
location-width

fkm:
@pakb What would you say to the following translation keys?
* `copy_cta` "Copy" (I don't like to use just `copy` and it is a kind of call-to-action)
* `copy_done` "Copied!" (`copy_success` is already used for "Link copied!")

tests/e2e/specs/infobox.spec.js
```diff
@@ -53,7 +53,7 @@ describe('The infobox', () => {
                     cy.get('[data-cy="popover"]').should('be.visible')
                     cy.get('[data-cy="tooltip"]').should('not.exist')
 
-                    cy.get('[data-cy="toggle-floating-off"]').click()
+                    cy.get('[data-cy="toggle-floating-off"]').click({ force: true })
```
pakb, line 56: not sure if using `force` is a good idea, that means that your HTML element is certainly behind something while trying to reach it (so there is some design/CSS issue anyway), no?

fkm: The problem is that we would need to pan the map to see the controls. I added a media query so restrict the height on small viewports. This way we don't need to force the click.

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -0,0 +1,87 @@
...
+        resetButton() {
+            this.buttonText = 'copy_cta'
+        },
+        copyValue() {
+            // Don't use the Clipboard API just yet. While execCommand might be
+            // deprecated, browser support for the new API is not sufficient and
+            // would require to keep execCommand anyway as a fallback.
+            this.$refs.inputHidden.focus()
+            this.$refs.inputHidden.select()
+            document.execCommand('copy')
```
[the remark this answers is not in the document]
fkm: I just did a diff and the only things that are identical are `inheritAttrs` and `props`. I could rewrite the code in order to make `resetButton` and `copyValue` reusable. But IMO adding another layer for the two methods seems a bit overkill. I recommend to keep this separate. As soon as Firefox fixes its support of the Clipboard API the code will automatically become much simpler.

pakb (approved):

ltshb (approved):
I approved, from a usability point of view, it works fine. Still an issue on mobile (long press doesn't emulate the right click) but it is already present on DEV and probably not related to this PR.

I'm still not happy with the design of having the `copy` button for the coordinates, but if this is really wanted by @davidoesch then ok, I would have found the copy icon much better.
Especially on mobile the copy icon would be much much better IMHO.

davidoesch:
can anyone point me on to the link where it works on mobile?

ltshb:
> For me the current "Test link" works. https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2228-share-link-redesign/index.html

does it also work on your mobile ? On my iphone I can't open the position popup neither with this test link nor with the regular DEV deploy: https://web-mapviewer.dev.bgdi.ch

fkm:
Works for me on Android. But I remember that you mentioned this problem before. And I think I had something similar once with the dev-tools. I recommend we open a ticket for that and look into it separately.

Selection also works somewhat weirdly on mobile. But I think that's because the browser handles it differently than on desktop. E.g. when I long-press the altitude it also selects the unit (m). But that's out of our control.

ltshb:
I created a ticket about the mobile issue: https://jira.swisstopo.ch/browse/BGDIINF_SB-2323. And because it is not related to this PR it can be fixed separetely.

PR #178, opened 2022-03-10
BGDIINF_SB-2272: Inverted menu button on mobile

fkm:
This changes the color of the menu button on mobile when the menu is open to further indicate the toggled state.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2272-invert-menu-button/index.html)


pakb (approved):
I think we should switch `dark` to `#474949` (as in `mf-geoadmin3`), this dark is a bit too dark to my taste 😉

fkm:
Good idea! I'll make the change.

fkm:
I changed the background color and while I was at it I also replaced the `<strong>` tag inside the button with `font-weight:bold` on the button.

PR #179, opened 2022-03-14
BGDIINF_SB-226: Inverted add icon for topic tree

fkm:
The add and delete icons look quite similar which can be confusing.

This inverts the colors of the add (plus) icon to make them more distinguishable.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2265-inverted-plus-icon/index.html)


pakb:
I don't know if we should not stick to rouded icons for this part of the app (so that there is some consistency with layers settings) but use a non filled one for the topic tree. (https://fontawesome.com/icons/circle-plus?s=regular and the same with minus if they are free)

davidoesch:
Rounded  - like in the current viewer

fkm:
Ok. I'll add the borders. That will probably also fix the perceived size. Thank you for the feedback!

fkm:
A propos the FontAwesome Icon with the circle. It looks like only the "solid" version is free.
https://fontawesome.com/search?q=circle&m=free&s=regular

pakb (approved):

PR #180, opened 2022-03-17
Improvements (refactor, memory, tests)

fkm:
A collection of improvements.

* Fist some minor refactoring where I replaced `foo: function () {` with `foo() {`.
* [Then the important part with the fixes/improvements for the memory leaks.](https://github.com/geoadmin/web-mapviewer/commit/d87bab631231cf13456d5ec65b109855966cc896)
* And a fix for the geolocation test that failed because of `watchPosition`.
* And the removal of an `.only` that I (again...) forgot to remove.

[Test link](https://web-mapviewer.dev.bgdi.ch/improvements/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersMap.vue
+++ b/src/modules/map/components/openlayers/OpenLayersMap.vue
@@ -237,19 +237,46 @@ export default {
         this.map.setView(this.view)
 
         // Click management
-        let pointerDownStart = null
-        let lastClickTimeLength = 0
-        this.map.on('pointerdown', () => {
-            pointerDownStart = performance.now()
-        })
-        // TODO: trigger a click after pointer is down at (roughly) the same spot for longer than 1sec (no need to wait for the user to stop the click)
-        this.map.on('pointerup', () => {
-            lastClickTimeLength = performance.now() - pointerDownStart
-            pointerDownStart = null
-        })
-        // using 'singleclick' event instead of 'click', otherwise a double click (for zooming) on mobile
-        // will trigger two 'click' actions in a row
-        this.map.on('singleclick', (event) => {
+        this.map.on('pointerdown', this.onMapPointerDown)
+        // TODO: trigger a click after pointer is down at (roughly) the same spot
+        // for longer than 1sec (no need to wait for the user to stop the click)
+        this.map.on('pointerup', this.onMapPointerUp)
+        // using 'singleclick' event instead of 'click', otherwise a double click
+        // (for zooming) on mobile will trigger two 'click' actions in a row
+        this.map.on('singleclick', this.onMapSingleClick)
+        this.map.on('pointerdrag', this.onMapPointerDrag)
+        this.map.on('moveend', this.onMapMoveEnd)
+    },
+    unmounted() {
+        this.map.un('pointerdown', this.onMapPointerDown)
+        this.map.un('pointerup', this.onMapPointerUp)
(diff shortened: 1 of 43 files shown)


src/modules/map/components/openlayers/utils/addLayerToMap-mixins.js
```diff
@@ -26,8 +26,13 @@ const addLayerToMapMixin = {
     },
     unmounted() {
         if (this.layer && this.isPresentOnMap) {
+            this.layer.getSource()?.clear?.()
```
fkm, line 29: I'm a bit on the fence about this notation.
It's simple and short. But (at least for now) unfamiliar. Especially the `clear?.()` part.

Feel free to tell me to do it the classic way! :smiley:

src/modules/map/components/openlayers/OpenLayersMap.vue
```diff
@@ -237,19 +237,46 @@ export default {
...
+        this.map.un('pointerdown', this.onMapPointerDown)
+        this.map.un('pointerup', this.onMapPointerUp)
+        this.map.un('singleclick', this.onMapSingleClick)
+        this.map.un('pointerdrag', this.onMapPointerDrag)
+        this.map.un('moveend', this.onMapMoveEnd)
+        this.map.setTarget(null)
+        this.map.setView(null)
+
+        delete this.map
+        delete this.view
```
fkm, line 260: I'm not sure if this is necessary. I'd say that in theory those properties should go away automatically when the component is destroyed.

src/modules/map/components/openlayers/utils/addLayerToMap-mixins.js
pakb: I'm not sure you need the last `?.()` part, `this.layer.getSource()?.clear()` should be good enough (`clear()` will only be called if `getSource()` gives something)

pakb (approved):

src/modules/map/components/openlayers/utils/addLayerToMap-mixins.js
fkm: The problem is that not all sources have a clear method. So it is possible that `getSource()` returns something but `.clear` is undefined.

I might be thinking too far. But from what I see we have the following options:
* Optional chaining  
  Short but unfamiliar.
* Try/catch with an empty catch  
  Somewhat hacky.
* If-block  
  More lines but easy to read.

fkm:
I allowed myself to amend the relevant commit. Here are the changes:

A compromise between an if-block and optional chaining.
```js
// src/modules/map/components/openlayers/utils/addLayerToMap-mixins.js
    unmounted() {
        if (this.layer && this.isPresentOnMap) {
            if (this.layer.getSource()?.clear) {
                this.layer.getSource().clear()
            }
            this.layer.setSource(null)

            this.removeLayerFromMap(this.layer)
        }

        delete this.layer
    },
```

And I moved the whole teardown to `unmounted`.
```js
// src/modules/map/components/openlayers/OpenLayersMap.vue
    unmounted() {
        this.map.un('pointerdown', this.onMapPointerDown)
        this.map.un('pointerup', this.onMapPointerUp)
        this.map.un('singleclick', this.onMapSingleClick)
        this.map.un('pointerdrag', this.onMapPointerDrag)
        this.map.un('moveend', this.onMapMoveEnd)
        this.map.setTarget(null)
        this.map.setView(null)

        delete this.map
        delete this.view
    },
```

PR #185, opened 2022-03-22
BGDIINF_SB-2266: Added dynamic focus outlines

fkm:
Outlines, while useful, can be rather ugly. That's why we decided
we want to remove them where possible.

This change introduces the dynamic outlines from the Swiss Style
Guide (based on outline.js) that only show when using the keyboard.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2266-dynamic-outlines/index.html)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -60,10 +60,10 @@ export default {
 // to the output CSS as many time as this file is imported
 @import 'node_modules/bootstrap/scss/bootstrap';
 #main-component {
-    font-family: Avenir, Helvetica, Arial, sans-serif;
+    font-family: $frutiger;
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
-    color: #2c3e50;
+    color: $coal;
 }
 :focus {
     outline-style: none;
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,5 +1,10 @@
 <template>
-    <div id="main-component">
+    <div
+        id="main-component"
+        :class="{ outlines: showOutlines }"
+        @keydown="setOutlines(true)"
+        @pointerdown="setOutlines(false)"
+    >
         <router-view />
     </div>
 </template>
@@ -14,6 +19,11 @@ import { mapActions, mapState } from 'vuex'
  */
 export default {
     name: 'App',
+    data() {
+        return {
+            showOutlines: false,
+        }
+    },
     computed: {
         ...mapState({
             currentUiMode: (state) => state.ui.mode,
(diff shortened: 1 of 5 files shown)


src/App.vue
```diff
@@ -1,5 +1,10 @@
 <template>
-    <div id="main-component">
+    <div
+        id="main-component"
+        :class="{ outlines: showOutlines }"
+        @keydown="setOutlines(true)"
+        @pointerdown="setOutlines(false)"
```
pakb, line 6: Good idea!
I was wondering how we could get rid of this and still enable keyboard navigation.

src/scss/webmapviewer-bootstrap-theme.scss
```diff
@@ -2,6 +2,13 @@ $red: #dc3545 !default;
 $primary: $red;
```
pakb, line 2: could we straight use the newly defined `$venetian-red` here?

src/scss/variables-admin.scss
```diff
@@ -0,0 +1,49 @@
+/* ==========================================================
+ * admin-variables.scss
+ * Variables set for our project
+ *
+ * Author: Yann Gouffon, yann@antistatique.net
+ * Date:   2014-04-30 09:38:36
+ *
+ * Copyright 2014 Federal Chancellery of Switzerland
+ * Licensed under MIT
+ =========================================================== */
```
pakb, line 10: Maybe a little comment that this comes from some GitHub repo would be great (if this style variables were to be changed there for instance)

pakb (approved):

src/App.vue
fkm: Yeah. Simple but effective 😎 
It can be a bit weird when you click into an input and the outline appears when you start typing. But all in all it's great.

Here's the original implementation used by the style guide:
https://github.com/lindsayevans/outline.js

src/scss/variables-admin.scss
fkm: Good idea. I'll add one.

src/scss/webmapviewer-bootstrap-theme.scss
fkm: I updated the custom `$red` to `$venetian-red`. I kept the styles that were using `$red`.

pakb (approved):
One weird thing : the test banner / disclaimer is still using the old red color, but I don't understand where this color is coming from as you've replaced it entirely in the SCSS....

Maybe it's the `dange` color from bootstrap (unthemed)?

fkm:
Yes. It's `$danger`. I'm thinking of using the emergency widget for the updated version of the development warning.
https://swiss.github.io/styleguide/en/widgets.html#03-widgets-02-emergency

But I'm still not sure where to place the thing. Even a small line like we have now is the horror to place. It always gets in the way in some viewport-size and/or mode. I'll probably have to add some media queries for that message.

PR #186, opened 2022-03-22
Miscellaneous refactoring

fkm:
I'm not 100% sure about the `getMap` commit. But I think we can get rid of the if-clauses.

And the rest is more style than substance :smile:

What do you think?

[Test link](https://web-mapviewer.dev.bgdi.ch/refactor-20220302/index.html)

### DIFF
--- a/src/modules/i18n/index.js
+++ b/src/modules/i18n/index.js
@@ -6,27 +6,18 @@ import fr from './locales/fr.json'
 import it from './locales/it.json'
 import rm from './locales/rm.json'
 
-export const languages = {
-    en: en,
-    de: de,
-    fr: fr,
-    it: it,
-    rm: rm,
-}
+export const languages = { en, de, fr, it, rm }
 
-// detecting navigator's locale as the default language (if it is a language served by this app)
-let matchingLang = null
+// detecting navigator's locale as the default language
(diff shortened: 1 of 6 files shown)


src/modules/i18n/index.js
```diff
@@ -6,27 +6,23 @@ import fr from './locales/fr.json'
...
-        if (!matchingLang && lang in languages) {
-            matchingLang = lang
+    for (const language in navigator.languages) {
+        if (language in languages) {
+            matchedLanguage = language
+            // we keep the first match we found
+            break
         }
-    })
+    }
```
pakb, line 21: I would maybe simplify that to
```
let matchedLanguage = null
if (navigator.languages) {
    // we keep the first match we found
    matchedLanguage = navigator.languages.find((lang) => lang in languages)
}
```
What do you think of that?

pakb (approved):
I think we can go your way with `getMap()` simplification. I was keeping those `if()` blocks in case we would export our Vue/OL binding into an external lib (so that it would avoid bad user implementation error). But I don't think it will happen anytime soon 😉

src/modules/i18n/index.js
fkm: I like! :sunglasses:

PR #190, opened 2022-03-24
BGDIINF_SB-1388: Added preview for search results

fkm:
In line with mf-geoadmin3 this adds a preview on hover to the search
results. The layers use the same functionality as the topic list and
the locations use the pinnedLocation that was already present.

This also includes some refactoring of the layer preview code and
adds the e2e tests for both the topic list and search results.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-1388-search-layer-preview/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchResultListEntry.vue
+++ b/src/modules/menu/components/search/SearchResultListEntry.vue
@@ -9,8 +9,8 @@
         @keydown.home.prevent="goToFirst"
         @keydown.end.prevent="goToLast"
         @keypress.enter.prevent="selectItem"
-        @mouseenter="previewResult(true)"
-        @mouseleave="previewResult(false)"
+        @mouseenter="startResultPreview"
+        @mouseleave="stopResultPreview"
     >
         <!-- eslint-disable vue/no-v-html-->
         <div class="search-category-entry-main p-2" @click="selectItem" v-html="entry.title"></div>
@@ -44,7 +44,7 @@ export default {
             required: true,
         },
     },
-    emits: ['showLayerLegendPopup', 'preview'],
+    emits: ['showLayerLegendPopup', 'previewStart', 'previewStop'],
     computed: {
         resultType() {
             return this.entry.resultType.toLowerCase()
@@ -77,12 +77,15 @@ export default {
                 this.$refs.item.tabIndex = '-1'
             }
         },
-        previewResult(show) {
+        startResultPreview() {
             if (this.resultType === 'layer') {
-                this.$emit('preview', show ? this.entry.layerId : null)
+                this.$emit('previewStart', this.entry.layerId)
             } else {
-                this.$emit('preview', show ? this.entry.coordinates : null)
+                this.$emit('previewStart', this.entry.coordinates)
             }
+        },
+        stopResultPreview() {
+            this.$emit('previewStop')
         },
     },
 }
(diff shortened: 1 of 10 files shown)


src/modules/store/modules/layers.store.js
```diff
@@ -39,10 +39,10 @@ const getters = {
      * @param state
      * @returns {AbstractLayer[]} All layers that are currently visible on the map
      */
-    visibleLayers: (state) =>
-        [...state.activeLayers.filter((layer) => layer.visible), state.previewLayer].filter(
-            (layer) => layer !== null
-        ),
+    visibleLayers: (state) => {
+        const activeVisible = state.activeLayers.filter((layer) => layer.visible)
+        return state.previewLayer === null ? activeVisible : [...activeVisible, state.previewLayer]
```
pakb, line 44: instead of a ternary operator, you could use a `if` block that pushes the preview layer if defined
```
const activeVisible = state.activeLayers.filter((layer) => layer.visible)
if (state.previewLayer) {
   activeVisible.push(state.previewLayer)
}
return state.previewLayer
```
I find that more readable than a ternary (even though it achieves the same result)

src/modules/menu/components/search/SearchResultListEntry.vue
```diff
@@ -75,6 +77,13 @@ export default {
                 this.$refs.item.tabIndex = '-1'
             }
         },
+        previewResult(show) {
+            if (this.resultType === 'layer') {
+                this.$emit('preview', show ? this.entry.layerId : null)
+            } else {
+                this.$emit('preview', show ? this.entry.coordinates : null)
+            }
+        },
```
pakb, line 86: wouldn't be clearer (for us human) to fire a distinct event when the preview is done?
here only the payload tells us if it is ongoing or ended, I would maybe prefer to have a `preview-start` and `preview-stop` events if you don't mind

src/modules/store/modules/layers.store.js
```diff
@@ -221,19 +221,20 @@ const actions = {
...
     },
-    showLayerPreview({ commit, getters }, layerId) {
-        const layer = getters.getLayerConfigById(layerId)
-        if (layer) {
-            const cloned = layer.clone()
-            cloned.visible = true
-            commit('setPreviewLayer', cloned)
+    setPreviewLayer({ commit, getters }, layerId) {
+        if (!layerId) {
+            commit('setPreviewLayer', null)
```
pakb, line 226: Here if we implement a double event approach you could add a new action called `clearPreviewLayer` that would then call the mutation `setPreviewLayer` with `null` as param.

src/modules/store/modules/layers.store.js
fkm: Works for me :+1: I'll change it.

src/modules/menu/components/search/SearchResultListEntry.vue
fkm: Yeah. Makes sense! Like the `mouseenter` and `mouseleave` events.

src/modules/store/modules/layers.store.js
fkm: Will do :+1:

pakb (approved):

PR #192, opened 2022-03-25
BGDIINF_SB-2263: Added trigger to open/close menu

fkm:
To show more of the map we close the menu on smaller viewports.

It opens the menu (desktop mode) when window gets bigger than 768px
and closes the menu when the window gets smaller than 768px.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-jira-bgdiinf_sb-2263-uimodes-tablet/index.html)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -11,6 +11,7 @@
 
 <script>
 import { mapActions } from 'vuex'
+import debounce from '@/utils/debounce'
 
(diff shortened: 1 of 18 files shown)


pakb (approved):

fkm:
Like that? Not sure if that's enough.

PR #193, opened 2022-04-05
BGDIINF_SB-2323: Added condition to ignore faultily event

fkm:
Currently, when someone tries to open the location pop-up on iOS
Safari the pop-up directly closes again. This is because the browser
sends a click event in addition to the context-menu event.

As there is nothing we can do against this event, I added a check
to the "singleclick" event to see if the last event was a context-
menu and if it was on the same position.

[Test link](https://sys-map.dev.bgdi.ch/bugfix-jira-bgdiinf_sb-2323-ios-safari-longpress/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -1,6 +1,6 @@
 <template>
     <OpenLayersPopover
-        v-if="isRightClick && clickCoordinates"
+        v-if="displayLocationPopup"
         :title="$t('position')"
         :coordinates="clickCoordinates"
         class="location-popup"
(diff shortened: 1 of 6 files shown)


pakb:
But as we have hijiacked the single click with a fullscreen toggle, the location popup shouldn't appear at all. What succession of clicks or gesture to you do to make it appear?

ltshb:
I just tested it on my iphone and it still don't work on safari and chrome.

fkm:
> But as we have hijiacked the single click with a fullscreen toggle, long click for identify, and no right click support (because of pan) the location popup shouldn't appear at all. What succession of clicks or gesture to you do to make it appear?

Good point! The thing is that this always worked for me. All I do is a long-press (Android, Chrome + Firefox).

fkm:
Ok. The problem is that we only handle left-clicks. But the long-press on mobile is a context-menu or right-click.

Probably best to handle the location pop-up through the store plugin so that we can suppress that in our own long-press logic. I look into it.

fkm:
'New' implementation:
* Desktop
  * Left-click -> Identify
  * Right-click -> Location
* Mobile
  * Left-click (short) -> Fullscreen
  * Left-click (long) -> Identify
  
 @ltshb and @pakb: Can you test the new version on your iDevices?

ltshb:
Still not working for me

pakb:
On my iOs15.3.1 it works well (on Safari). Which device to you have @ltshb ?

ltshb:
Iphone 11 IOS 15.4.1, I tried on chrome, safari and firefox On firefox and chrome i tried also to clear my phone cache to make sure to have the latest version but still not working

fkm:
Turns out that the problem is that the "expected" behavior has changed.

As this PR fixes the current expected behavior I will merge it.
To restore the behavior from mf-geoadmin3 I will assign [the relevant ticket](https://jira.swisstopo.ch/browse/BGDIINF_SB-2235) to @pakb.

pakb (approved):

PR #194, opened 2022-04-07
Changed the event to prevent too many triggers

fkm:
Listening to the add events from the selected features triggered
the listener too often.

By listening to the select event of the interaction we get each
event once.

[Test link](https://web-mapviewer.dev.bgdi.ch/develop-refactor-drawing-cursor/index.html)


pakb (approved):
good catch!

PR #195, opened 2022-04-08
Fixed drawing tooltip and tolerances

fkm:
This fixes the drawing tooltip (especially for lines/polygons) and
updates the tolerances of the different interactions so that they
match visually. At least as good as it gets.

[Test link](https://web-mapviewer.dev.bgdi.ch/develop-refactor-drawing-cursor/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingSelectInteraction.vue
+++ b/src/modules/drawing/components/DrawingSelectInteraction.vue
@@ -90,7 +90,7 @@ export default {
             layers: [this.getDrawingLayer()],
             // As we've seen with the old viewer, some small features were hard
             // to select. We will try to add a bigger hit tolerance to mitigate that.
-            hitTolerance: 3,
+            hitTolerance: 6,
         })
     },
     mounted() {
--- a/src/modules/drawing/components/DrawingSelectInteraction.vue
+++ b/src/modules/drawing/components/DrawingSelectInteraction.vue
@@ -7,6 +7,7 @@ import { EditableFeature } from '@/api/features.api'
 import { extractOpenLayersFeatureCoordinates } from '@/modules/drawing/lib/drawingUtils'
 import { editingFeatureStyleFunction } from '@/modules/drawing/lib/style'
 import SelectInteraction from 'ol/interaction/Select'
+import { DRAWING_HIT_TOLERANCE } from '@/config'
 
 /**
  * Manages the selection of features on the drawing layer. Shares also which features are selected
@@ -90,7 +91,7 @@ export default {
             layers: [this.getDrawingLayer()],
             // As we've seen with the old viewer, some small features were hard
             // to select. We will try to add a bigger hit tolerance to mitigate that.
-            hitTolerance: 6,
+            hitTolerance: DRAWING_HIT_TOLERANCE,
         })
     },
     mounted() {
--- a/src/modules/drawing/components/DrawingTooltip.vue
+++ b/src/modules/drawing/components/DrawingTooltip.vue
@@ -7,6 +7,8 @@
 <script>
 import Overlay from 'ol/Overlay'
 import { pointWithinTolerance, getVertexCoordinates } from '@/modules/drawing/lib/drawingUtils'
+import { DRAWING_HIT_TOLERANCE } from '@/config'
+import { EditableFeatureTypes } from '@/api/features.api'
 
 const cssPointer = 'cursor-pointer'
 const cssGrab = 'cursor-grab'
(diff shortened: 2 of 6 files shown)


src/modules/drawing/components/DrawingSelectInteraction.vue
```diff
@@ -90,7 +90,7 @@ export default {
             layers: [this.getDrawingLayer()],
             // As we've seen with the old viewer, some small features were hard
             // to select. We will try to add a bigger hit tolerance to mitigate that.
-            hitTolerance: 3,
+            hitTolerance: 6,
```
pakb, line 93: as we are using this value in `DrawingTooltip.vue` too, it would maybe good to export it to `config.js` as a const

src/modules/drawing/components/DrawingTooltip.vue
```diff
@@ -34,113 +44,149 @@ export default {
...
-                // if so, we suppress the flag "hoveringSelectedFeature" as it is of no value
-                // to know that we are hovering the area (only borders have interactivity)
-                if (featureUnderCursor?.getGeometry() instanceof Polygon) {
-                    hoveringSelectedFeature = false
+
+                // Display a help tooltip when modifying
+                if (hoveringVertex || ['marker', 'annotation'].includes(featureType)) {
+                    translationKeys = `modify_existing_vertex_${featureType}`
+                } else {
+                    translationKeys = `modify_new_vertex_${featureType}`
```
pakb, line 131: Here I would prefer that we lower the case of `featureType` when we build our string, and then use `EditableFeatureType` enum in the `.includes` (it will be more resilient in case of refactoring)

src/modules/drawing/components/DrawingTooltip.vue
```diff
@@ -34,113 +44,149 @@ export default {
...
+                }
+            } else {
+                mapElement.classList.remove(cssPointer)
+                mapElement.classList.remove(cssGrab)
+
+                // Display a help tooltip when drawing
+                if (this.currentDrawingMode) {
+                    let type = this.currentDrawingMode.toLowerCase()
+
+                    if (this.currentlySketchedFeature && !['marker', 'annotation'].includes(type)) {
```
pakb, line 151: same as above, I would prefer we use the enum instead of hard coded values

src/modules/drawing/lib/drawingUtils.js
```diff
@@ -154,3 +154,43 @@ export function extractOpenLayersFeatureCoordinates(feature) {
...
+/**
+ * Checks if point is at target within tolerance.
+ *
+ * @param {Number[]} point The point in question.
+ * @param {Number[]} target The target to check against.
+ * @param {Number} [tolerance=0] Distance from target that still counts. Default is `0`
+ * @returns {Boolean} If the point is close enough to the target.
+ */
+export function pointWithinTolerance(point, target, tolerance = 0) {
+    return (
```
pakb, line 167: Maybe add some defensive measure to make sure `point` is defined 😉

pakb (commented):
in `DrawingTooltip.vue` it would be great if you could explain/comment each case covered in more details, what kind of tooltip are we adding on which ground. That would help a lot to fix things later.
Documentation is boring to make, but it will save us later 😉

src/modules/drawing/components/DrawingTooltip.vue
fkm: What do you mean by "lower the case"?

src/modules/drawing/lib/drawingUtils.js
fkm: As in throwing an error? And for target as well?

pakb: more something like
```
if (!point || !target) {
    return false
}
```

src/modules/drawing/components/DrawingSelectInteraction.vue
fkm: I'll amend the config and check if a tolerance of 6 works with the modifyInteraction.

src/modules/drawing/components/DrawingTooltip.vue
fkm: Do you mean `EditableFeatureTypes` from features.api.js?

src/modules/drawing/components/DrawingTooltip.vue
fkm: Ok

fkm: Ok

src/modules/drawing/components/DrawingTooltip.vue
fkm: I'm not sure if these values ever get serialized or have to be compared to serialized values. But if not we could convert them to symbols.

```js
export const EditableFeatureTypes = {
    MARKER: Symbol('MARKER'),
    ANNOTATION: Symbol('ANNOTATION'),
    LINEPOLYGON: Symbol('LINEPOLYGON'),
    MEASURE: Symbol('MEASURE'),
}
```

fkm: Hmmm... We'd have to add a mapping for the instances we want to use the value/label as with the translation keys. Probably best to keep it as-is :smiley:

pakb (approved):

PR #196, opened 2022-04-11
Consolidated store on top-level

fkm:
Having the store modules with the components that use them makes
a lot of sense on paper but in practice it makes finding a module
often more difficult.

This consolidates all store modules and plugins in the src directory
for a better overview.

[Test link](https://web-mapviewer.dev.bgdi.ch/feature-consolidate-store/index.html)


src/store/README.md
```diff
@@ -0,0 +1,28 @@
+# Store
+
+This is the [Vuex](https://vuejs.org) store used to manage the state of the entire application.
```
pakb, line 3: wrong URL 😄

pakb (approved):
Good idea to also move the store in `src/`'s root, much clearer this way.
I think we could also find another name than `src/modules/...` as many other libraries we are using have this name to describe part of what we can provide (that adds some confusion...)

Maybe you could also quickly go through all `.md` files for each `src/modules` that had a store module, there might be some docu here to remove/move as well.

fkm:
I've been thinking about how to name the modules for a while while working on my design proposal.

What I had with React/Redux was a separation into "containers" (the modules) and "components". But I'm not that big a fan as the resulting split structure can be bad. And just calling it "components" doesn't really work either as we also have other things in our modules (e.g. assets). Plus, there is a real distinction between modules and components that would get lost.

I'll get back to you if I have an idea.

And in the meantime I'll make a pass over the markdown files

pakb (approved):

PR #197, opened 2022-04-14
Split profile popup component

pakb:
So that it can be used both in the bottom tray and the floating popup (they are not in the same module, and one of them doesn't require the popup wrapper around)

@fkm I still have an issue with d3 plot width in the floating popup, and I don't find why this happens. If you could give it a look that would be great, I'm maybe missing some CSS properties that make it function properly in the bottom tray but not in the floating popup.
(or it is, as I fear, a time/sync issue with rendering and the moment we read the width of the element...)


fkm:
I fixed the tooltip styles. They were in the wrong component.

The sizing issue is because we don't have a min width. Although it is curious that the chart doesn't use 0 as width. 

Btw. is `drawing/components/ProfilePopup.vue` still in use?

fkm (approved):

PR #200, opened 2022-04-21
BGDIINF_SB-2359: Added workaround for double click

fkm:
Using the Cypress "dblclick" command doesn't work on mobile.

This rewrites the test to use two "click" commands. Until we find
a better solution this has to be enough.


tests/e2e/specs/drawing/kml.spec.js
```diff
@@ -15,7 +15,7 @@ describe('Drawing saving KML', () => {
                     cy.goToDrawing(isMobileViewport || isTabletViewport)
                 })
 
-                it('saves a KML on draw end', () => {
+                it.only('saves a KML on draw end', () => {
```
pakb, line 18: don't forget to remove your `.only` calls after fixing the test 😉

fkm: Dang! :laughing:

fkm: Done. I allowed myself to amend the last commit.

The only changes are the removal of the `.only` and I updated the clicks by taking the values from the `drawGeoms` custom task.

pakb (approved):

PR #203, opened 2022-04-25
Share menu section

pakb:
[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-1855-share_menu_section/index.html)


src/modules/menu/components/share/MenuShareShortLinkInput.vue
```diff
@@ -0,0 +1,57 @@
...
+    methods: {
+        clearIsCopiedInClipboard() {
+            this.copiedInClipboard = false
+            if (this.timeoutCopied) {
+                clearTimeout(this.timeoutCopied)
+                this.timeoutCopied = null
+            }
+        },
+        copyShortLinkInClipboard() {
+            navigator.clipboard.writeText(this.shortLink)
```
fkm, line 51: Did you test this in Firefox?
Mine (v99.0) doesn't have `navigator.clipboard`.

pakb: maybe I should use this signature instead : https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText
but it should be supported across all browser : https://caniuse.com/mdn-api_clipboard_writetext

pakb: well, in the exemple given on the mozilla webpage, it's the exact same signature... With your version of Firefox it should be supported, weird...

pakb:
We have to wait for https://github.com/nicolasbeauvais/vue-social-sharing/pull/312 to be accepted (or edited), merged and deployed in a newer version in order for this PR to work properly (otherwise QR code definition for `vue-social-sharing` is not done properly)

src/modules/menu/components/share/MenuShareShortLinkInput.vue
fkm: Sorry! My bad. I tend to use about:blank to test browser APIs and it seems that clipboard is not available there.

pakb: nope, you need to be either in HTTPS or on localhost for this API to be enabled 😉

src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -0,0 +1,176 @@
...
+    methods: {
+        /**
+         * Generates an iframe HTML code snippet pointing to the short link
+         *
+         * @param {EmbedSizes} size
+         * @returns {String} HTML iframe code snippet
+         */
+        iFrameLink(size) {
+            let width = this.customWidth,
+                height = this.customHeight
```
fkm, line 137:
```suggestion
            const width = this.customWidth
            const height = this.customHeight
```

src/modules/menu/components/share/MenuShareSection.vue
```diff
@@ -0,0 +1,46 @@
+<template>
+    <MenuSection
+        :title="$t('share')"
+        data-cy="menu-share-section"
+        secondary
+        @click.once="generateShortLinkIfMNeeded"
```
fkm, line 6: That will probably lead to a ticket down the road. I'm not really familiar with the Vue router. Could we save a kind of snapshot of the routing state and then use that to check if we need to regenerate the short-link?

src/modules/menu/components/share/MenuShareSection.vue
```diff
@@ -0,0 +1,46 @@
...
+    >
+        <FontAwesomeIcon v-if="!shortLink" icon="spinner" spin size="2x" class="p-2" />
+        <div v-if="shortLink" class="p-2">
```
fkm, line 9:
```suggestion
        <div v-else class="p-2">
```

src/modules/menu/components/share/MenuShareShortLinkInput.vue
```diff
@@ -0,0 +1,61 @@
...
+            this.clearIsCopiedInClipboard()
+        },
+    },
+    methods: {
+        clearIsCopiedInClipboard() {
+            this.copiedInClipboard = false
+            if (this.timeoutCopied) {
+                clearTimeout(this.timeoutCopied)
+                this.timeoutCopied = null
+            }
```
fkm, line 52:
```suggestion
            clearTimeout(this.timeoutCopied)
```

src/modules/menu/components/share/MenuShareShortLinkInput.vue
```diff
@@ -0,0 +1,61 @@
...
+        },
+        copyShortLinkInClipboard() {
+            navigator.clipboard.writeText(this.shortLink)
+            this.copiedInClipboard = true
+            this.timeoutCopied = setTimeout(this.clearIsCopiedInClipboard, 2500)
```
fkm, line 57: This needs a `clearTimeout` in an `unmounted/beforeUnmount`.

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -64,6 +64,11 @@ function storeMutationWatcher(store, mutation, router) {
                     log.info('Error while routing to', query, error)
                     routeChangeIsTriggeredByThisModule = false
                 })
+            // if the short linked version of the URL is already defined,
+            // we must refresh it as the URL has just changed
+            if (store.state.position.shortLink) {
```
fkm, line 69: Won't this regenerate the short-link with each map action (pan, zoom, etc) even if no short-link is visible?

fkm (changes requested):
The code-suggestions are just that. Implement as you see fit :smiley: 

But the short-link issues are something we need to look at.

src/router/storeSync/storeSync.routerPlugin.js
ltshb: We need to be careful not to generate too many useless shortlink !

pakb: yep, I think we will need to monitor in the state when the menu share section is opened (so when an update is needed if the URL changes)

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -65,8 +65,8 @@ function storeMutationWatcher(store, mutation, router) {
                     routeChangeIsTriggeredByThisModule = false
                 })
             // if the short linked version of the URL is already defined,
-            // we must refresh it as the URL has just changed
-            if (store.state.position.shortLink) {
+            // and the menu opened, we must refresh it as the URL has just changed
+            if (store.getters.showMenu && store.state.share.keepUpdatingShortLink) {
```
fkm, line 69: As all the information comes from the store you could move the check into the `generateShortLink` action.

src/store/plugins/stop-updating-shortlink-when-menu-closed.plugin.js
```diff
@@ -0,0 +1,23 @@
...
+ *
+ * @param {Vuex.Store} store
+ */
+const stopUpdatingShortLinkWhenMenuClosedPlugin = (store) => {
+    store.subscribe((mutation, state) => {
+        // when the user closes the menu, even though the share section is visible
+        // we shouldn't keep updating the short link (as the user has no intention to copy it right now, so we
+        // will just create unused short links if we do so)
+        // as soon as the menu is opened again (and the share section is still visible) we reactivate the update of the short link
+        if (mutation.type === 'setMenuDesktopOpen') {
```
fkm, line 12: Wouldn't it be easier to create an "action" for `setMenuDesktopOpen` / `setShowMenuTray` and dispatch `toggleKeepUpdatingShortLink` from there?

fkm (dismissed):
Much better :sunglasses: 

The only thing missing is the clearTimeout for the MenuShareShortLinkInput component.

src/store/plugins/stop-updating-shortlink-when-menu-closed.plugin.js
pakb: I'm in the process of replacing all the logic with one flag for the menu and the UI mode. I'm struggling with the tablet/desktop version to make it work properly...

src/router/storeSync/storeSync.routerPlugin.js
pakb: yes and no, `store.getters.showMenu` comes from `ui.store.js` and `store.state.share.keepUpdatingShortLink` from `share.store.js`. The philosophy we try to follow is that any cross store module interaction should be done in plugins 😉

ltshb (commented):
I've noticed that the shortlink is always created twice on each user action, do you know why ?
Otherwise IMHO it is ok to always update the shortlink on each user action when the share menu is open.

pakb:
> I've noticed that the shortlink is always created twice on each user action, do you know why ? Otherwise IMHO it is ok to always update the shortlink on each user action when the share menu is open.

We've re-worked how we generate shortlinks, it is now only generated if you open the share menu section. And as soon as the state of the app changes (the URL changes), the section will be closed, and the current shortlink deleted.

You then have to actively open the share menu section once again if you want a shortlink to be generated.
This way, we avoid any excessive shortlink generation, as the user opened the corresponding section, we can be (nearly) certain that this shortlink is required.

src/main.js
```diff
@@ -63,6 +64,12 @@ const app = createApp(App)
 app.use(router)
 app.use(i18n)
 app.use(store)
+app.use(VueSocialSharing, {
+    // defining service-qrcode as a network for the sharing library
+    networks: {
+        qrcode: `${API_SERVICES_BASE_URL}qrcode/generate?url=@url`,
+    },
```
ltshb, line 71: This seems not to work correctly, I tested the qrcode and it had an invalid short url. In my test the share menu had https://sys-s.dev.bgdi.ch/991e0ea1ec while the qrcode used https://sys-map.dev.bgdi.ch/api/qrcode/generate?url=https%3A%2F%2Fsys-s.dev.bgdi.ch%2F991e0ea1ecrl ! Which indeed did not exists in the  DB backend.

pakb: that's weird that it could slip by testing... We are specifically testing the URL that is opened by the link, and yeah there's this extra "rl" at the end that is coming from the configuration here : https://github.com/geoadmin/web-mapviewer/blob/86fa42572232b2b21219bfd2e4232a5a4ea205ad/src/modules/menu/components/share/MenuShareSocialNetworks.vue#L54-L58

I'm guessing the doc wasn't updated but now we should declare it `...?url=@u`, it removes this extra `rl` if I do so.
But then it should have been picked by https://github.com/geoadmin/web-mapviewer/blob/86fa42572232b2b21219bfd2e4232a5a4ea205ad/tests/e2e-cypress/integration/shareShortLink.cy.js#L57-L63
I'll look a bit into it, but that's confusing 😕

src/store/plugins/stop-updating-shortlink-when-menu-closed.plugin.js
```diff
@@ -0,0 +1,23 @@
...
+            // de-activating short link update as menu is closed
+            if (!state.ui.menuDesktopOpen && state.share.keepUpdatingShortLink) {
+                store.dispatch('toggleKeepUpdatingShortLink')
+            } else if (state.ui.menuDesktopOpen) {
+                // TODO document why this block is empty
```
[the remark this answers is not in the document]
pakb: or to remove it altogether 😄 it's not used anymore, it was a planned plugin to update the shortlink as soon as the state changed, but now we are not doing that anymore, I'll remove this file

PR #206, opened 2022-05-04
Change build tools to reflect Vue3 default tooling

pakb:
Vue3 now uses Vite instead of Webpack to process the build. This also will remove most (if not all) our problematic dependencies (security alerts on GitHub).

All folders and files have been moved/changed to reflect what a default/pristine vue3 project looks like when it is set up with `npm init vue@latest`

### Change e2e test viewport paradigm

In order to use Vite as our pre-processor, and such to use `import '@/...` notation, it was not possible to have exports in `cypress/plugins/index.js` anymore and so our viewport helper function here must disappear.
Viewport is now defaulted to mobile (mobile first approach) but can be defined otherwise in the Cypress config (or through ENV variable, see package.json run scripts)
Multi-viewport testing will be addressed by having multiple `npm run test:e2e:...` targets, each for one viewport. This will also enable us to parallelize down the line on the CI.

[Test link](https://sys-map.dev.bgdi.ch/feat-upgrade-vue-tooling/index.html)


cypress/integration/mouseposition.spec.js
```diff
@@ -113,7 +113,7 @@ describe('Test mouse position', () => {
                 cy.get('[data-cy="location-popup-height"]').contains(fakeheight.height)
             })
         })
-        context('Coordinates system tests', () => {
+        context('Coordinates system cypress', () => {
```
fkm, line 116: Looks like a find & replace change.

vite.config.js
```diff
@@ -0,0 +1,23 @@
+import { fileURLToPath, URL } from 'url'
+
+import { defineConfig } from 'vite'
+import vue from '@vitejs/plugin-vue'
+import vueJsx from '@vitejs/plugin-vue-jsx'
```
fkm, line 5: Do we use JSX?

src/api/topics.api.js
```diff
@@ -131,7 +131,7 @@ export const loadTopicTreeForTopic = (lang, topic) => {
 const loadTopicsFromBackend = (layersConfig) => {
     return new Promise((resolve, reject) => {
         if (!API_BASE_URL) {
-            // this could happen if we are testing the app in unit tests, we simply reject and do nothing
+            // this could happen if we are testing the app in unit cypress, we simply reject and do nothing
```
fkm, line 134: This also looks like find & replace.

src/api/layers/layers.api.js
```diff
@@ -172,7 +172,7 @@ export const getLayerLegend = (lang, layerId) => {
 export const loadLayersConfigFromBackend = (lang) => {
     return new Promise((resolve, reject) => {
         if (!API_BASE_URL) {
-            // this could happen if we are testing the app in unit tests, we simply reject and do nothing
+            // this could happen if we are testing the app in unit cypress, we simply reject and do nothing
```
fkm, line 175: This also looks like find & replace.

cypress/support/drawing.js
```diff
@@ -58,7 +59,8 @@ Cypress.Commands.add('goToDrawing', (menuIsClosed = false) => {
     addDefaultIconsFixtureAndIntercept()
     addSecondIconsFixtureAndIntercept()
     cy.goToMapView()
-    if (menuIsClosed) {
+    const viewportWidth = Cypress.config('viewportWidth')
+    if (viewportWidth && viewportWidth < BREAKPOINT_PHONE_WIDTH) {
```
fkm, line 63: Do we still need the `menuIsClosed` parameter (in the future)?

cypress/support/commands.js
```diff
@@ -127,7 +127,7 @@ Cypress.Commands.add(
             const targetTopic = state.topics.current?.layersToActivate.length
             const targetLayers =
                 'layers' in otherParams
-                    ? // Legacy layers come with an additional param. At least in our tests.
+                    ? // Legacy layers come with an additional param. At least in our cypress.
```
fkm, line 130: This also looks like find & replace.

fkm (changes requested):
There are a few details here and there but it looks ok.

vite.config.js
pakb: I don't think we do, that's just a poor copy paste on my part with default Vite/Vue3 config...

cypress/support/drawing.js
pakb: no I don't think so, I can clean that up

fkm (approved):

package.json
```diff
@@ -4,20 +4,18 @@
     "private": true,
     "scripts": {
         "start": "npm run serve",
```
fkm, line 6: While trying to debug the "preview" timeout problem I realized that `npm start` is broken.

fkm (changes requested):

pakb:
I had to disable one test file so that it passes, as I've forced the viewport to be mobile that was to be expected.
This test can be reactived when we work on multi-viewport testing (in parallel!)

pakb:
I might have some more work on the deploy script part though, it seems the branch name is not passed correctly in the build (with Vite) and so the app takes everything from the root of the bucket instead of the branch folder.

.env
```diff
@@ -0,0 +1,7 @@
+VITE_API_BASE_URL=https://mf-chsdi3.int.bgdi.ch/
+VITE_API_PUBLIC_URL=https://public.int.bgdi.ch/
+VITE_DATA_BASE_URL=https://data.geo.admin.ch/
+VITE_WMTS_BASE_URL=https://wmts{5-9}.int.bgdi.ch/
+VITE_WMS_BASE_URL=https://wms-bgdi-cdn-0.int.bgdi.ch/
+VITE_API_SERVICES_BASE_URL=https://sys-map.dev.bgdi.ch/api/
+VITE_API_SERVICE_KML_BASE_URL=https://sys-public.dev.bgdi.ch/
```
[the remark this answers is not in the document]
ltshb: I don't fully understand why `.env.development` has been deleted ? For which purpose `.env` is ? For local development or for DEV staging ?

buildspec.yml
```diff
@@ -4,6 +4,8 @@ env:
   shell: bash
   variables:
     AWS_DEFAULT_REGION: eu-central-1
+    AWS_REGION: eu-central-1
```
ltshb, line 7: Why those two variables ?

buildspec.yml
```diff
@@ -14,12 +16,13 @@ phases:
     runtime-versions:
       nodejs: 14
     commands:
-      # updating npm to version 8.x.x (it is not done automatically by settings env to nodejs v14)
-      # see package.json for version requirements
-      - npm install -g npm@8.4.1
+      # updating node to v16
+      - n 16
+      # As the cypress/download folder is not added to git, and (somehow) not created by Cypress at startup, we create it
+      - mkdir cypress/downloads/
```
ltshb, line 22: `mkdir -p` to create parents if `cypress` dir would not exist might be less brittle.

ltshb (commented):
Why using `/cypress/...` directory structure instead of `/tests/cypress/...` ? I found having all the tests in `/tests/...` easier to understand for new bee.

.env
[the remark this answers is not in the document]
pakb: both, I didn't really know if I should name it `.env.development` or `.env.local` and as Vite uses the `.env` by default if there is no stagging defined, I wanted to a have the default env set to local/dev.

Regarding stage mixing, I'll do some clean up, thanks for pointing it out

`VITE_API_PUBLIC_URL` is an remnant of what was used before we migrate to the new `service-kml` I'll remove it.

buildspec.yml
pakb: That was for testing with the local build bash script, it was complaining that no region was set.

fkm (approved):

PR #207, opened 2022-05-09
BGDIINF_SB-2361: Fixed infobox after refactoring

fkm:
This is the last clean-up task after the refactoring. It resolves
the problem of the infobox having to cover all the use-cases from
the current version in its new form.

I also included some minor bug fixes that I found during development.

What is missing from this commit is:
* Fix of the delete feature function.
* I disabled some tests.
* Style-only refactoring.

_And some refactoring:_
This mostly introduces some empty lines to give the
code some room to breathe and make it easier to scan.

I also moved the dragging setter/getter from the Feature
to the EditableFeature class and sorted the store modules
alphabetically.

And I fixed some English mistakes in the store ADR.


tests/e2e/specs/drawing/profile.spec.js
```diff
@@ -38,20 +38,21 @@ describe('Profile popup', () => {
...
                 ).should('have.text', '0.00m')
             }
             cy.get('[data-cy="profile-popup-area"]').should('not.have.attr', 'd')
-            cy.get('[data-cy="profile-popup-delete-button"]').click()
-            cy.get('[data-cy="profile-popup"]').should('not.exist')
-            cy.get('[data-cy="drawing-style-popup"]').should('not.exist')
+            // Delete is currently not implemented. Will be added in a later commit.
+            // cy.get('[data-cy="profile-popup-delete-button"]').click()
+            // cy.get('[data-cy="profile-popup-content"]').should('not.exist')
+            // cy.get('[data-cy="drawing-style-popup"]').should('not.exist')
```
pakb, line 55: Is it something you wanted to address in this PR?

tests/e2e/specs/drawing/profile.spec.js
```diff
@@ -75,57 +76,46 @@ describe('Profile popup', () => {
...
-            cy.get('[data-cy="profile-popup-close-button"]').click()
-            cy.get('[data-cy="profile-popup"]').should('not.exist')
-            cy.get('[data-cy="drawing-style-popup"]').should('not.exist')
+            cy.get('[data-cy="infobox-close"]').click()
+            cy.get('[data-cy="infobox"]').should('not.be.visible')
         })
         it('closes the profile popup, while minimized, when the user clicks on the X button', () => {
-            cy.get('[data-cy="profile-popup-minimize-button"]').click()
-            // TODO: find a way to place the profile popup's close button out of the layers copyrights (and then remove the { force: true })
-            // see : https://jira.swisstopo.ch/browse/BGDIINF_SB-2170
```
pakb, line 125: Is https://jira.swisstopo.ch/browse/BGDIINF_SB-2170 fixed by your changes?

src/modules/infobox/components/FeatureCombo.vue
```diff
@@ -0,0 +1,108 @@
...
+    methods: {
+        ...mapActions([
+            'changeFeatureTitle',
+            'changeFeatureDescription',
+            'changeFeatureColor',
+            'changeFeatureTextSize',
+            'changeFeatureTextColor',
+            'changeFeatureIcon',
+            'changeFeatureIconSize',
+        ]),
```
pakb, line 52: As many parts of the Vue code is similar between the 3 different flavour of Feature***.vue I would maybe ask you to try to either use a mixin, but that is not the recommended way to do in Vue3 anymore, so maybe have a look at https://vuejs.org/guide/reusability/composables.html#composables and see if you can make it happen here 😉

tests/e2e/specs/drawing/profile.spec.js
fkm: This was fixed long ago when we moved the profile to `#ap-footer-middle`. But it seems I forgot to update the test back then.

pakb (approved):

PR #210, opened 2022-05-16
BGDIINF_SB-2361: Added delete capability to drawing

fkm:
We currently have the problem that we cannot delete a drawing feature
from the infobox components are there is no link between the store and
the drawingLayer used by the drawing module.

This adds a list of feature IDs to the drawing store in order to track
the existing features so that we can remove them from the drawing when
they are deleted from the store.

[Test link](https://sys-map.dev.bgdi.ch/feature-jira-bgdiinf_sb-2361-delete-button/index.html)


pakb:
replaced by https://github.com/geoadmin/web-mapviewer/pull/221 (so that it is based on `develop` branch)

PR #218, opened 2022-05-31
BGDIINF_SB-2405: Added the profilechart axis labels and added translation in the tooltip

jedef:
[Test link](https://sys-map.dev.bgdi.ch/feat-2405-profilechart-axis-labels-and-translation/index.html)


src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -79,7 +79,7 @@ export default {
         return {
             showTooltip: false,
             options: {
-                margin: { left: 35, right: 15, bottom: 25, top: 15 },
```
pakb, line 82: Is that wanted? That's something you changed back with https://github.com/geoadmin/web-mapviewer/pull/215

src/modules/drawing/lib/ProfileChart.js
```diff
@@ -279,9 +279,12 @@ export default class ProfileChart {
...
     updateLabels() {
         this.group
-            .select('text.profile-label-x')
+            .select('text.ga-profile-label-x')
+            .attr('x', this.width / 2)
             .text(`${i18n.global.t(this.options.xLabel)} [${this.unitX}]`)
-        this.group.select('text.profile-label-y').text(`${i18n.global.t(this.options.yLabel)} [m]`)
+        this.group
+            .select('text.ga-profile-label-y')
+            .text(`${i18n.global.t(this.options.yLabel)} [m]`)
```
pakb, line 287: Could you replace the class `ga-profile-label` with its version without `ga-`? That was the prefix used in `mf-geoadmin3` and I'd like to remove it so that we can differentiate part that are straight copy paste and those that we have already processed

src/modules/infobox/components/FeatureProfile.vue
jedef: Yes. On #215 I needed bigger margins on the right and the top but wanted to keep overall size so I reduced left and bottom. But on this commit I see that I actually need this big margins also on the left and bottom to display the x and y labels.

src/modules/drawing/lib/ProfileChart.js
jedef: I fixed that

pakb:
replaced by https://github.com/geoadmin/web-mapviewer/pull/222

PR #227, opened 2022-06-10
Bugfix 2410 drawing should be editable when reopened

jedef:
The bug was due to a call to the measure manager in the drawing module, in which it was undefined. Tof fix it, the measure manager was moved to the Drawing Layer, so that it can also be called when adding a saved KML Layer. It was also added in the map KML Layer.

[Test link](https://sys-map.dev.bgdi.ch/bugfix-2410-drawing-should-be-editable-when-reopened/index.html)

### DIFF
--- a/src/utils/MeasureManager.js
+++ b/src/utils/MeasureManager.js
@@ -1,9 +1,3 @@
-/**
- * Handels the overlays that are shown on top of a each measure drawing
- *
- * It is used in the Drawing module as well as in the map KMLLayer, as the drawing is displayed with
- * a KML Layer when outside of the drawing mode.
- */
 import Overlay from 'ol/Overlay'
 import {
     canShowAzimuthCircle,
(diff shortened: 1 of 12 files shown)


src/utils/MeasureManager.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * Handels the overlays that are shown on top of a each measure drawing
+ *
+ * It is used in the Drawing module as well as in the map KMLLayer, as the drawing is displayed with
+ * a KML Layer when outside of the drawing mode.
+ */
```
pakb, line 6: you can move that a couple lines below (just before the component declaration)

pakb (commented):
Could you add a test that make sure this is always enabled in the future?

Also I think there's something fishy when editing a marker, I re-edited a KML with a marker and changed its color, and the icon disapeared (it unselected which icon was shown, when I selected another one from the UI it came back)

pakb (approved):
🚀

PR #228, opened 2022-06-20
App version is only shown when also the dev disclaimer is shown.

jedef:
Added a global constant so that the app version and dev disclaimer will always be shown based on the same if condition

[Test link](https://sys-map.dev.bgdi.ch/feat-2438-only-show-version-when-dev-disclaimer-is-shown/index.html)


pakb (approved):
That was quite difficult 😉

PR #230, opened 2022-06-23
Add a link to the service-alti api to download a CSV of the profile data

jedef:
A Link is shown to the user to download the CSV of the profile data

[Test link](https://sys-map.dev.bgdi.ch/feat-2432-add-export-to-csv-to-profile-popup/index.html)


pakb:
> * [ ]  the csv is missing a header info , renmae the 4 cols to Distance [m],Altitude[m],X,Y

We still need to switch on to `service-alti` on Frankfurt and see if this would fix this issue (Brice has done some bugfix on the latest version only deployed in Frankfurt). Otherwise, that means we will have to fix that in `service-alti` directly

src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -218,26 +229,40 @@ export default {
...
+        triggerDownload(blob, fileName) {
+            const link = document.createElement('a')
+            link.style.display = 'none'
+            link.href = URL.createObjectURL(blob)
+            link.download = fileName
+
+            document.body.appendChild(link)
+            link.click()
+            URL.revokeObjectURL(link.href)
+            link.parentNode.removeChild(link)
```
pakb, line 265: maybe a `window.open('...')` could also work here, and simplify this function

src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -192,6 +198,12 @@ export default {
             const id = this.feature.id.replace('drawing_feature_', '')
             this.deleteDrawingFeature(id)
         },
+        onCSVDownload() {
+            this.getProfile(profileCsv).then((data) => {
+                const dataBlob = new Blob([data], { type: 'text/csv', endings: 'native' })
+                this.triggerDownload(dataBlob, generateFilename('.csv'))
+            })
+        },
```
pakb, line 206: it might be good to also `.catch(...)` here and log into the console any error (or transform that into an `async` function with a `try...catch` block)

src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -218,27 +230,35 @@ export default {
...
             return this.profileChart.element
         },
         async updateProfileChart(size) {
-            const data = await this.getProfile(this.featureCoordinates)
+            const data = await this.getProfile()
             this.profileChart.update(data, size)
             return this.profileChart.element
         },
-        async getProfile(coordinates) {
+        async getProfile(apiFunction = profileJson, coordinates = this.featureCoordinates) {
```
pakb, line 245: it's declared async, but there's no `await`, the promise is returned directly.
As it's the same method as the one called by `onCSVDownload`, you could `await` here the API call, and `try...catch` and log any error

src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -218,27 +230,35 @@ export default {
...
             })
         },
+        triggerDownload(blob, fileName) {
+            const link = document.createElement('a')
+            link.href = URL.createObjectURL(blob)
+            link.download = fileName
+
+            link.click()
+            URL.revokeObjectURL(link.href)
+        },
```
pakb, line 261: could you add a little comment that we must use this `<a>` approach so that we can define the downloaded file name?
otherwise the next one to come here will want to switch to a `window.open(...)` 😉

tests/e2e-cypress/integration/drawing/profile.cy.js
```diff
@@ -34,7 +34,7 @@ const testInfo = ['0.00m', '0.10m', '0.10m', "1'342m", "1'342m", '4.50m', '4.51m
 describe('Profile popup', () => {
     context('check how app behave with backend output', () => {
         const goToDrawingWithMockProfile = (mockValue) => {
-            cy.mockupBackendResponse('rest/services/profile.json', mockValue, 'profile')
+            cy.mockupBackendResponse('rest/services/profile.json?*', mockValue, 'profile')
```
pakb, line 37: why not `cy.mockupBackendResponse('rest/services/profile.json**', mockValue, 'profile')`?
now if the app calls the backend without any URL argument (without the `?...` part of the URL) it might not be caught anymore, no?

pakb:
Please add some Cypress test that covers this new button and the output it gives 😉

src/api/profile.api.js
```diff
@@ -63,3 +78,23 @@ export const profile = (data) => {
             })
```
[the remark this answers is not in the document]
jedef: I am not sure if i understand you correctly here, but the .catch block is already present and is doing what you describe.

pakb (approved):

PR #232, opened 2022-06-30
Feature: BGDIINF_SB-2431: close share menu on map movement

ltkum:
Issue : the short link generation call was made every time there was a change in the url, causing a lot of unnecessary calls to the underlying service, beginning the first time a short link was generated and never stopping afterwards.
Fix : Whenever there is a change, if there is an existing short link, this will reset its state to null and close the share menu should it be open. Meaning the only calls made to the shortLink service should be for links people actually want to share.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2431-close-share-menu-on-map-movement/index.html)

### DIFF
--- a/src/modules/menu/components/menu/MenuSection.vue
+++ b/src/modules/menu/components/menu/MenuSection.vue
@@ -11,6 +11,7 @@
             data-cy="menu-section-header"
             data-toggle="collapse"
             @click="toggleShowBody"
+
         >
             <span class="menu-section-title">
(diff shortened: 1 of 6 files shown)


pakb:
Looks good.
I had an issue when I tried to open the embed part of the menu section, it closed the menu while saying in the console : `Uncaught TypeError: this.clearShortLink is not a function`
Could you check that the embed part of the share section still works with your work?

PR #233, opened 2022-07-06
Replace jQuery tooltip with Bootstrap tooltip

pakb:
Also fix dispatch to action that was previously renamed

This was not caught by our tests because we have skipped the location popup for now, as we only test viewport sizes for phones (and thus, not location popup on phones, as long as we do not re-implement click management as discussed https://github.com/geoadmin/web-mapviewer/pull/208)

[Test link](https://sys-map.dev.bgdi.ch/bug_fix_location_popup/index.html)


jedef (dismissed):

src/utils/DropdownButton.vue
```diff
@@ -0,0 +1,160 @@
...
+            data-bs-reference="parent"
+            :aria-expanded="expanded"
+            @click="toggleExpanded"
+        >
+            <span class="visually-hidden">Toggle Dropdown</span>
+        </button>
+        <ul class="dropdown-menu" :class="{ show: expanded }" :aria-labelledby="uniqueHtmlId">
+            <li v-for="item in items" :key="item.value">
+                <a
+                    class="dropdown-item"
```
ltshb, line 35: It seems that there is an issue with the `dropdown-item` css class. The cursor onhover on a dropdown item is not set to `pointer`

src/utils/DropdownButton.vue
```diff
@@ -0,0 +1,160 @@
...
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+    },
+    emits: ['click', 'select:item'],
+    data() {
+        return {
+            // generating a unique HTML ID for this dropdown
+            uniqueHtmlId: `dropdown-${randomIntBetween(0, 10000)}`,
```
ltshb, line 118: Not sure here that `randomIntBetween(0, 10000)` is safe, if the application grows and that you have many drop downs items you might have collisions ! Question does the ID needs to be unique within the dropdown button or within the whole application ?
Maybe something like vue-uid (https://dev.to/jshimkoski/ssr-friendly-unique-html-ids-in-vue-3-1og9) plugin might be a good idea ? Or using a more robust uuid like `uuid` or `nanoid` npm packages.

pakb: I was wondering if I should add `uuid` lib or not, and thought that we shouldn't have more than 3 or 4 dropdown at the same time on the screen, and so that a simple random number should be good enough instead of adding yet another lib to the stack already present in `package.json`

ltshb: Ok I see your point

ltshb (approved):
Looks good now. 👍🏼

src/utils/DropdownButton.vue
pakb: there might be a possibility to use `getUid()` from OpenLayers, but I also didn't want to add OL import in part of the app that doesn't really use mapping...

PR #236, opened 2022-07-21
Feat 2434 show compass on map rotate

jedef:
A compass is shown if the map is rotated. Clicking on the compass reputs the north at the top.
Although not strictly part of the original ticket, this PR also uniformizes all map buttons in the right toolbox. In particular, they are now all "\<button\>" elements and so can be cycled through by pressing TAB.

Open bug:
- On my Linux machine with a chrome based browser, the DragRotateInteraction of Openlayers does not seem to work (as others cannot replicate it, this is probably a bug with my distro / configuration).

Possible Improvements:
- The rotation of the map is not saved in the URL. So if someone shares a rotated map, the shared map will not look as expected. It is to be discussed if this is something that needs to be fixed or not.
- It could be useful to also add another shortcut (like ctrl+drag) to rotate the map, as it would probably be more intuitive than alt+shift+drag and as alt+shift is used to change keyboard layout on some OSes. However Ctrl+drag is already used to select multiple features in the old viewer, so this still needs clarifications.

[Test link](https://sys-map.dev.bgdi.ch/feat-2434-show-compass-on-map-rotate/index.html)


src/modules/map/components/footer/MapFooterBackgroundSelector.vue
```diff
@@ -106,7 +107,7 @@ $map-button-gap: 0.5rem;
     background-image: url('../../../menu/assets/backgrounds_mobile.png');
     background-repeat: no-repeat;
     border-style: solid;
-    border-color: $gray-800;
+    border-color: $map-button-border-color;
```
boecklic, line 110: :+1: to put this in variables

src/modules/menu/components/toolboxRight/GeolocButton.vue
```diff
@@ -1,13 +1,15 @@
 <template>
-    <div
+    <button
```
boecklic, line 2: :+1: for changing to `button` (dunno if there was a reason not to use button, but `button` seems right)

boecklic (commented):
Nice! Clean code, good comments, meaningful naming of variables and functions. I like! ;-) 

We need to check for the CTRL (therefore only comment and not approve): currently it's used to allow a selection of features by draging a rectangular selection box with CTRL. Dunno if this is still required. On the other hand though, the CTRL-drag works on my Chrome on Linux!

jedef:
> Nice! Clean code, good comments, meaningful naming of variables and functions. I like! ;-)
> 
> We need to check for the CTRL (therefore only comment and not approve): currently it's used to allow a selection of features by draging a rectangular selection box with CTRL. Dunno if this is still required. On the other hand though, the CTRL-drag works on my Chrome on Linux!

I have reverted the commit in question, so I think that now it can be merged. We can still discuss about the possibility to add another shortcut later on.

boecklic (approved):

PR #238, opened 2022-07-29
Feat 2406 preview position on map when moving mouse on profile graph

jedef:
An overlay (circle) that shows the corresponding position on the map when hovering over the profile graph was added.
In addition, an infinite looop bug was fixed that sometimes happend when moving the mouse over the profile graph. I also used this opportunity to make the calculations of the height more precise.

[Test link](https://sys-map.dev.bgdi.ch/feat-2406-preview-position-on-map-when-moving-mouse-on-profile-graph/index.html)

### DIFF
--- a/src/modules/drawing/lib/ProfileChart.js
+++ b/src/modules/drawing/lib/ProfileChart.js
@@ -36,6 +36,37 @@ export default class ProfileChart {
         return this.lineString.getCoordinateAt(ratio)
     }
 
+    /**
+     * Given a distance from origin, calculate the height.
+     *
+     * @param {Number} dist Distance (must be in the unit specified by this.unitX, i.e. depending on
+     *   the length of the path, either km or m)
+     * @returns The height in meter at the given distance
+     */
+    getHeightAtDist(dist) {
+        const maxDist = this.data[this.data.length - 1].domainDist || 0
+        dist = Math.min(Math.max(dist, 0), maxDist)
+        let lowerIndex = Math.trunc(((this.data.length - 1) * dist) / maxDist)
+        lowerIndex = Math.min(lowerIndex, this.data.length - 1)
+        while (dist < this.data[lowerIndex].domainDist) {
+            lowerIndex--
+        }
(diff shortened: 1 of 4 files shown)


src/modules/drawing/lib/ProfileChart.js
```diff
@@ -36,6 +36,37 @@ export default class ProfileChart {
...
+        const lowerDist = this.data[lowerIndex].domainDist
+        if (lowerIndex >= this.data.length - 1) {
+            return lowerAlti
+        }
+        let higherIndex = lowerIndex + 1
+        const higherAlti = this.data[higherIndex].alts[this.elevationModel] || 0
+        const higherDist = this.data[higherIndex].domainDist
+        const higherRatio = (dist - lowerDist) / (higherDist - lowerDist)
+        const lowerRatio = 1 - higherRatio
+        return lowerRatio * lowerAlti + higherRatio * higherAlti
```
pakb, line 67: I would be happy if you could explain a bit your thought process in this function.
Input outputs are clear, but not so what you do in the middle.

You could also add a couple unit tests for this, as it is well isolated 😉

src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -306,22 +310,34 @@ export default {
...
+                    'EPSG:2056',
+                    'EPSG:3857',
+                    this.profileChart.lineString.getCoordinateAt(x / plotWidth)
+                )
+                this.currentHoverPosOverlay.setPosition(coords)
             })
             glass.on('mouseover', () => {
                 this.showTooltip = true
+                this.currentHoverPosOverlay.getElement().style.backgroundColor =
+                    this.feature.fillColor.fill
```
pakb, line 335: As there's no unset of this property, wouldn't it be better to set that up when creating the instance, and not each time we hover?

jedef: The thing is that the color can change if the user draws a line and then changes the color of that line. So I think that it's necessary to do it like this if we want the color of the circle to always match the color of the line.

pakb (approved):
Nice!

PR #239, opened 2022-08-11
Feat 2434 show compass on map rotate second PR

jedef:
This is the second PR for the feat 2434 ticket.

- The map can now also be rotated by clicking on Ctrl (but the default alt+shift still works)
- The compass is now updated in real time when rotating the map. To be able to do this, the menu had to be put inside the openlayers map, so that it can register openlayers events, as using the store for real time updates of the map rotation would be way too slow.

[Test link](https://sys-map.dev.bgdi.ch/feat-2434-show-compass-on-map-rotate/index.html)

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,4 +1,6 @@
 <template>
+    <!-- Disallow zooming on touch, as the zoom gesture is already reserved for zooming the map -->
+    <meta name="viewport" content="width=device-width, user-scalable=no" />
     <div
         id="main-component"
         :class="{ outlines: showOutlines }"
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,6 +1,4 @@
 <template>
-    <!-- Disallow zooming on touch, as the zoom gesture is already reserved for zooming the map -->
-    <meta name="viewport" content="width=device-width, user-scalable=no" />
     <div
         id="main-component"
         :class="{ outlines: showOutlines }"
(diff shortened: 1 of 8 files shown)


src/modules/menu/components/toolboxRight/CompassButton.vue
```diff
@@ -23,19 +26,33 @@
...
+    inject: ['getMap'],
     emits: ['northen'],
-    computed: {
-        ...mapState({
-            rotation: (state) => state.position.rotation,
-        }),
+
+    data() {
+        return {
+            rotation_: 0,
```
pakb, line 36: Is the underscoring meaning it is a private prop?
If so I prefer a prefixed underscoring, so you don't stumble upon a private prop with autocompletion by accident, you have to actively write first a underscore to list them

src/App.vue
```diff
@@ -1,4 +1,6 @@
 <template>
+    <!-- Disallow zooming on touch, as the zoom gesture is already reserved for zooming the map -->
+    <meta name="viewport" content="width=device-width, user-scalable=no" />
```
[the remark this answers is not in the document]
jedef: Strange, I am not able to replicate this. This code should inhibit the zooming of the ui (you should still be able to zoom the map of course). I have an android, so maybe it does not work on IOS? (but I have searched the internet and it seems that it should also work on IOS)

src/modules/menu/components/toolboxRight/CompassButton.vue
jedef: I removed the underscore entirely, as we don't use this convention anywhere else in the code.

pakb (approved):

PR #240, opened 2022-08-16
Changes API base URL

pakb:
As we have finished our migration of most (if not all) our backends to Frankfurt, and thus to the new sys-... URL naming convention, old URLs starts to be decommissioned. We then have to change the last one that was still pointing to the old backend's DNS entry.

[Test link](https://sys-map.dev.bgdi.ch/backends_url_after_migration/index.html)


jedef (approved):
With this fix the map viewer indeed seems to load faster again.

PR #242, opened 2022-08-17
Add ch.swisstopo.leichte-basiskarte_world.vt as default BG layer

pakb:
Incorporate @geoblocks/ol-maplibre-layer to create the link between the OpenLayers instance and the Maplibre renderer
Config for VT layer is hard coded for now, it should be replaced later by a new entry in the layers config served by the backend (but that will have to wait till the very end as we serve the same config to the current productive viewer, and we don't want this VT layer to appear there...)


davidoesch:
Would make sense to start with default : the national map vector tiles version - preview

pakb:
You mean a different layer than `ch.swisstopo.leichte-basiskarte_world.vt`?

pakb:
Sadly we currently have some issues with our CI, so it won't build the test link yet... We need to fix something beforehand

davidoesch:
send you an email regarding " LK Vectortiles"

pakb:
Testlink is now functional!

davidoesch:
The default background layer for Switzerland is the landeskarte . So, currently this will be wmts pixelkarte-Farbe . Once Karto has a vector version of Landeskarte we will add this. Outside Switzerland: for the time being we can use light base map

the grey background will be light base map

pakb:
We wait for feedback from Kartografie to see if we can have a more detailed map as BG

davidoesch:
one option as backgroudn is as well : https://test.map.geo.admin.ch/?lang=en&topic=ech&layers=ch.swisstopo.pixelkarte-farbe&lon=8.77997&lat=46.85560&zoom=3.998876289191964

pakb:
There might be some OpenLayers configuartion to achieve what is shown there. If I had the layer as is in web-mapviewer we have this white zone around the extent. Maybe you can force the extent of a layer, and OL crops anything outside of this.

davidoesch:
what is interesting: there is no white border in the current https://test.map.geo.admin.ch/?lang=en&topic=ech&layers=ch.swisstopo.pixelkarte-farbe&lon=8.77997&lat=46.85560&zoom=3.998876289191964 ...why? both is openlayers AFAIK the curremt.test.map.geo.admin.ch with mapbox ....

pakb:
Both have the same approach, only show VectorTiles with Mapbox/MapLibre, the rest is managed by OpenLayers (WMTS, WMS, etc...)

davidoesch:
Nice!!

pakb:
it's not they are not loaded, it's there is no data, and we deliver them as JPEG, so no transparency...

pakb:
I think I've had a discussion with Marc where, in the past, they would ship tiles JPEG when wholly defined, and switch to PNG on the fly when the tiles were on the border of the data (so mixed data) so that we could achieve transparency...
@procrastinatio @ltclm @rebert Are we still able to deliver the pixelkarte-farbe map with mixed PNG/JPEG tiles? or was it just a POC and it never went further?

procrastinatio:
Very cool!
Maybe we should consider again the [WebP](https://developers.google.com/speed/webp) image format, whch combines the best of `png`and of `jpeg` . Looks like it's fairly widely supported (except IE6, well 11). See https://caniuse.com/?search=webp

davidoesch:
Nice find @procrastinatio  worth investigating --- an estimate is needed @ltclm . Nothing beats the beauty and performance of good old wmts pixelkarte

ltclm:
mapserver is currently not supporting the rendering of webp.
but mixed content type tiles are supported (png/jpeg). https://mapserver.org/sq/mapfile/outputformat.html

However, it has to be clarified whether the GIS FatClients can handle webp or mixed content types.

pakb:
In QGIS it should be alright : https://issues.qgis.org/issues/21083 through the fact that OSGeo supports it since 3 years ago, so I don't see why ESRI's stuff wouldn't follow

IE11 is out of our scope of supported browser, so we could do a switch to WebP for the new viewer. Your thoughts?

src/api/layers/VectorLayer.class.js
```diff
@@ -0,0 +1,27 @@
...
+     */
+    constructor(layerId, opacity, styleUrl) {
+        super(
+            layerId,
+            LayerTypes.VECTOR,
+            layerId,
+            opacity,
+            'swisstopo',
+            'https://www.swisstopo.admin.ch/en/home.html',
+            true
```
jedef, line 19: Are all these options of the constructor hardcoded because the backend does not provide them? If yes should the class name maybe be a bit more specific (like swisstopoBackgroundVectorLayer) or these options defined when the constructor is called?

davidoesch:
Ie11 is still in heavy use at some gov  offices 
Do we have some kibana stats?

Changing the most used wmts layer - need to be careful, in-depth analysis first.

WebP and all clients (including swisstopo app ) and mixed content : we need to be careful

src/modules/map/components/openlayers/OpenLayersVectorLayer.vue
```diff
@@ -0,0 +1,54 @@
...
+            },
+        })
+        // we load the style on the side in order to remove data over Switzerland (we are currently showing our standard National map as there is not enough details with our light vector map)
+        axios.get(this.styleUrl).then((response) => {
+            const vectorStyle = response.data
+            // filtering out any layer that uses swisstopo data (meaning all layers that are over Switzerland)
+            vectorStyle.layers = vectorStyle.layers.filter((layer) => layer.source !== 'swissmaptiles')
+            this.layer.maplibreMap.setStyle(vectorStyle)
+        })
+    },
```
ltshb, line 52: As the name of the component is very generic I would keep it generic and move the filter as parameter, e.g. `excludeSource` parameter and set it in the configuration to `swissmaptiles`

src/store/plugins/load-layersconfig-on-lang-change.js
```diff
@@ -28,25 +30,38 @@ function loadLayersConfig(lang) {
...
-                }
-            })
-        })
-        .catch((error) => log.error(error))
+const loadLayersAndTopicsConfigAndDispatchToStore = async (store) => {
+    try {
+        // adding vector tile backend through a hardcoded entry (for now)
+        // this should be removed as soon as the backend delivers a proper configuration
+        // for our vector tile background layer
+        const bgVectorLayer = new VectorLayer(VECTOR_TILES_STYLE_ID, 1.0, VECTOR_TILES_STYLE_URL)
```
ltshb, line 38: See comment above as this is here already hardcoded, I would put the attributionName and co here and not in the component.
From where comes from the vector tiles backend ?

ltshb (commented):
image
Changing the background is not very nice, see the layout above. Should we not have the vector tiles bg on the `Carte Couleurs` bg instead of a new background ?

src/modules/map/components/footer/MapFooterBackgroundSelector.vue
```diff
@@ -198,18 +198,12 @@ $menu-button-diameter: 3px;
...
-    }
-    .bg-index-2 {
-        $offset: calc(($map-button-diameter + $map-button-gap) * -1);
-        transform: translateY($offset);
+    $numberOfBackgrounds: 4;
+    @for $i from 0 through $numberOfBackgrounds {
+        .bg-index-#{$i} {
+            $offset: calc(($map-button-diameter + $map-button-gap) * ($i - $numberOfBackgrounds));
+            transform: translateY($offset);
+        }
```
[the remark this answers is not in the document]
pakb: I had to adjust a little bit the background selector with your latest work, it should be all good now.

src/api/layers/VectorLayer.class.js
pakb: That's a good point that was also raised by Brice, it should now be more generic, with hardcoded value only found in one place

src/modules/map/components/openlayers/OpenLayersMap.vue
```diff
@@ -173,8 +181,23 @@ export default {
...
+            return this.backgroundLayers.find(
+                (layer) => layer.getID() === 'ch.swisstopo.pixelkarte-farbe'
+            )
+        },
         // zIndex calculation conundrum...
         startingZIndexForVisibleLayers() {
+            // if vector tile background, we currently add our WMTS map on top so there are two background layers stacked
+            if (this.isBackgroundVectorTile && this.pixelKarteFarbeLayer) {
+                return 2
+            }
```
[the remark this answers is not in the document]
pakb: I've replaced a bit the variable names and the logic so that it should be more generic.

src/store/plugins/load-layersconfig-on-lang-change.js
pakb: So now all specific code is on this line (hardcoded values and such).
I don't really know how the backend is served, I'm just aware of this style ID and that if I put that in a Maplibre instance, cool maps are rendered 😉

pakb:
> Hi We will have the vector tiles instead of carte grey. Everything else is like map.geo.admin.ch-geoadmin3 And we will move the topic which are now pixelkarte grey to vectortiles

So no possibility to choose to have a gray background in the wheel?
Isn't that a bit cumbersome for people that want to draw something complex, and want to remove the "noise" around their drawing?
We do not serve the LBM layer without the national map (colorized!) so this means there would be no possibility to have a light swiss background anymore

pakb:
image
why? 🤷

davidoesch:
@pakb  you still load the hillshade wmts when in Switzerland

src/modules/map/components/openlayers/OpenLayersMap.vue
```diff
@@ -172,8 +180,30 @@ export default {
...
+            return null
+        },
         // zIndex calculation conundrum...
         startingZIndexForVisibleLayers() {
+            // checking if the BG layer is a vector layer that has anoter
+            if (this.isBackgroundVectorTile && this.backgroundLayerOnTopOfVectorBackground) {
+                return 2
+            }
             return this.currentBackgroundLayer ? 1 : 0
         },
```
ltshb, line 208: I think we could do this a bit more generic. If I understood correctly we already have in the config a flag telling if the layer is a background or not ? We could extend this flag to an int specifying the bg priority/zindex; e.g. 
- `background = 0` => not a background
- `background = 1` => upper background
- `background = 2` => lower background 
- ...

ltshb:
> image why? shrug

Could it be due to coordinates conversion between LV95 and WebMercator ?

src/modules/map/components/openlayers/OpenLayersMap.vue
pakb: I can't just set fixed value and not use the index, otherwise OL complains that it's not possible to insert a layer at an index greater than the current length of the array... Hence this "complicated" zIndex calculation

pakb: So I have to dynamically have a sequence of index, where all are used to insert a layer, starting from the small value to the biggest

pakb:
I've asked on OL, and it seems I have to define the extent once again at the layer level, having it on the TileGrid isn't sufficient to avoid a doubling of the map.
But yes, it is related to the reprojection method used

pakb:
> @pakb you still load the hillshade wmts when in Switzerland

I can't remove it, otherwise it also removes it from the rest of the world

pakb:
image
we've not become insularly yet 🎉

src/modules/map/components/openlayers/OpenLayersMap.vue
ltshb: 👍🏼

ltshb (approved):

PR #245, opened 2022-08-22
Bugfix 2539 profile issue with closed polygon

jedef:
Fixed an issue where the profile was not showing up after drawing a closed polygon. Also added some cypress tests to check for this bug in the future.

[Test link](https://sys-map.dev.bgdi.ch/bugfix-2539-profile-issue-with-closed-polygon/index.html)


pakb (approved):
👍

PR #246, opened 2022-08-23
Update translations

pakb:
and fix translation script that was unnecessarily migrated to Vite env variable handling (it's not a JS module, so it should stick to standard Node.js process env variable access)

[Test link](https://sys-map.dev.bgdi.ch/update_translations/index.html)


pakb:
I will rename the branch to match JIRA ticket

PR #249, opened 2022-08-26
Bugfix 2462 marker icon undefined after reload and color change

jedef:
This bug was caused by an incorrect recreation of the editable feature, as openlayers can only save strings in the KML and not whole objects, which was assumed. To simplify the code and serialization / deseralization, now the openlayers features directly save their respective editable feature, which is serialized before creating the KML.

Things that still need to be done, but probably in another ticket, as this ticket is already too big:
- When loading an old KML file (or any KML file that did not save the editable feature in the extra properties), the map viewer crashes, the drawing is not displayed correctly and you are not able to edit the drawing. This needs to be fixed e.g. by generating a default "editable featue" in this case.

Open Bugs: (where already existent before this PR)
- When drawing a line or a measure consisting of 2 points, the user must tripple click the second point to finish, which is unexpected. When drawing a line or a measure consisting of more than two points, the user must, as expected, only double click the last point to finish.
- When  drawing, the user can move already drawn markers and text fields by clicking anywhere on the marker / text. This is the expected behaviour. However, if the drawing is reloaded, the user is now only able to move the features that were drawn before the reload by clicking on their white dot.

Also, it seems to me that the Jira tickets [2109](https://jira.swisstopo.ch/browse/BGDIINF_SB-2109) and [2165](https://jira.swisstopo.ch/browse/BGDIINF_SB-2165) can be closed as "Could not reproduce", as it seems to me that the bug described in these two tickets was rather caused for the exact same reason as the bug of this ticket. I tried to reproduce them by not applying our style function when displaying the drawed kml outside of the drawing mode (as else our custom style function would directly overwrite the anchor value of the hotspot kml tag), but was not able to.

[Test link](https://sys-map.dev.bgdi.ch/bugfix-2462-marker-icon-undefined-after-reload-and-color-change/index.html)


src/modules/drawing/lib/export-utils.js
```diff
@@ -93,6 +90,7 @@ export function generateKmlString(features = [], styleFunction = null) {
             `<Document><name>${i18n.global.t('draw_layer_label')}</name>`
         )
     }
+    console.log(kmlString)
```
pakb, line 93: another to remove 😄

src/modules/drawing/lib/style.js
```diff
@@ -48,6 +46,7 @@ const sketchPoint = new Circle({
  * Note that the style differs when the feature is selected (or drawn for the first time) or when
  * displayed without interaction (see {@link featureStyleFunction} for this case)
  */
+//export const editingFeatureStyleFunction = featureStyleFunction
```
pakb, line 49: that could also be deleted no?

src/modules/drawing/lib/style.js
```diff
@@ -93,64 +92,47 @@ export const editingFeatureStyleFunction = (feature) => {
...
                           width: 3,
                       }),
             // filling a polygon with white if first time being drawn (otherwise fallback to user set color)
             fill: isDrawing
                 ? whiteSketchFill
                 : new Fill({
-                      color: fillColor,
+                      color: [...editableFeature.fillColor.rgb.slice(0, 3), 0.4],
                   }),
         }),
```
pakb, line 133: way cleaner this way, nice!

pakb (approved):
Very good rework of the drawing module!
A couple `console.log` to remove and we are good to go

PR #250, opened 2022-08-31
When in desktop mode, background selector is now as in mf-geoadmin3

jedef:
In desktop mode, background selector is now a square on the bottom left
instead of round in the bottom right.

[Test link](https://sys-map.dev.bgdi.ch/feat-2559-background-selector-desktop/index.html)


pakb:
I think it would be better if we also add the little button with the arrow, when the selector is opened :
image

This way users can decide not to change the background, and close the selector in a meaningful way.

Could you also switch the BG color to white? It pops out a bit too much with a dark color

jedef:
Personally I like the white border a bit less, because then the border of the buttons are harder to see, especially on a white background, but I still changed it. All in all, it should now be identical to mf-geoadmin3, except that the back button also has visual feedback, as mf-geoadmin3 lacked any visual feedback for the back button (expect the cursor change). But maybe the red feedback for the back button is a bit too strong? Or do you think its ok?

pakb (approved):
👍

PR #251, opened 2022-09-13
Use of LV95 as projection for internal layers

pakb:
In order to fix our "white border" issue, I've switched the default projection system we use to request tiles from our backend to LV95. I've also declared a proper TileGrid for WMTS layers, which should cut the white part that was present on the side of tiles (OpenLayers will not require tiles outside of the extent, plus seems to crop images to fit this extent). This way we should be ready to place the pixelkarte-farbe map on top of a vector tile layer, in order to achieve world wide coverage without compromising on details over Switzerland.

[Test link](https://sys-map.dev.bgdi.ch/use_lv95_for_internal_layers/index.html)

### DIFF
--- a/src/api/layers/AbstractLayer.class.js
+++ b/src/api/layers/AbstractLayer.class.js
@@ -23,9 +23,11 @@ export default class AbstractLayer {
 
     /**
      * @abstract
+     * @param {Number} projection The EPSG number of the projection system to use (for instance,
+     *   EPSG:2056 will require an input of 2056)
      * @returns {String} The URL to use to request tile/image/data for this layer
      */
-    getURL() {
+    getURL(projection = 3857) {
         throw new Error('You have to implement the method getURL!')
     }
 
--- a/src/api/layers/AbstractLayer.class.js
+++ b/src/api/layers/AbstractLayer.class.js
@@ -1,3 +1,5 @@
+import { CoordinateSystems } from '@/utils/coordinateUtils'
+
 /**
  * Base class for layers' config description, must be extended to a more specific flavor of Layer
  * (e.g. {@link WMTSLayer}, {@link WMSLayer}, {@link GeoJsonLayer}, {@link AggregateLayer} or {@link KMLLayer})
@@ -18,16 +20,17 @@ export default class AbstractLayer {
         this.opacity = opacity
         this.hasTooltip = hasTooltip
         this.visible = false
-        this.projection = 'EPSG:3857'
+        // default projection used, as we want to achieve worldwide coverage, is web mercator metric
+        this.projection = CoordinateSystems.WEBMERCATOR.epsg
     }
 
     /**
      * @abstract
-     * @param {Number} projection The EPSG number of the projection system to use (for instance,
+     * @param {Number} epsgNumber The EPSG number of the projection system to use (for instance,
      *   EPSG:2056 will require an input of 2056)
      * @returns {String} The URL to use to request tile/image/data for this layer
      */
-    getURL(projection = 3857) {
+    getURL(epsgNumber = CoordinateSystems.WEBMERCATOR.epsgNumber) {
         throw new Error('You have to implement the method getURL!')
     }
 
--- a/src/api/layers/WMTSLayer.class.js
+++ b/src/api/layers/WMTSLayer.class.js
@@ -53,9 +53,9 @@ export default class WMTSLayer extends GeoAdminLayer {
     }
 
     /** @returns {String} A XYZ type URL to request this WMTS layer's tiles */
-    getURL() {
+    getURL(projection = '3857') {
         return `${this.baseURL}1.0.0/${this.getID()}/default/${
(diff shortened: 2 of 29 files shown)


pakb:
@davidoesch you should like this one 😉

davidoesch:
I do!
cant wait to have the LBM behind it

some tiles are not loading 
image

pakb:
Yeah I've seen that... don't know what's happening but a small dezoom and it was like that :
image

As we are on dev, I'll check the console tomorrow to see if it's not simply a backend not delivering the tile

davidoesch:
Can't see anything in the console currently- but I'm sure it will work out in the end

jedef (changes requested):

src/api/layers/AbstractLayer.class.js
```diff
@@ -23,9 +23,11 @@ export default class AbstractLayer {
 
     /**
      * @abstract
+     * @param {Number} projection The EPSG number of the projection system to use (for instance,
+     *   EPSG:2056 will require an input of 2056)
      * @returns {String} The URL to use to request tile/image/data for this layer
      */
-    getURL() {
+    getURL(projection = 3857) {
```
ltshb, line 30: I would define all EPSG numbers as constants and use the constant name instead of those magic numbers, this will make the code much clearer. Always avoid magic numbers.
```javascript
EPSG_3857 = 3857
// or even better
EPSG_MERCATOR = 3857
WEB_MERCATOR = 3857
LV95 = 2056
...
```
And use the constant everywhere in the code instead of the hardcoded number

src/api/layers/WMTSLayer.class.js
```diff
@@ -53,9 +53,9 @@ export default class WMTSLayer extends GeoAdminLayer {
     }
 
     /** @returns {String} A XYZ type URL to request this WMTS layer's tiles */
-    getURL() {
+    getURL(projection = '3857') {
```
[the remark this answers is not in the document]
ltshb: This would be solved by my mention above

src/modules/map/components/openlayers/OpenLayersInternalLayer.vue
```diff
@@ -4,16 +4,18 @@
             v-if="layerConfig.type === LayerTypes.WMTS"
             :layer-id="layerConfig.getID()"
             :opacity="layerConfig.opacity"
-            :url="layerConfig.getURL()"
+            :url="layerConfig.getURL(2056)"
```
ltshb, line 7: See my first comment

src/modules/map/components/openlayers/OpenLayersInternalLayer.vue
```diff
@@ -4,16 +4,18 @@
...
             :z-index="zIndex"
+            :projection="'EPSG:2056'"
         />
         <OpenLayersWMSLayer
             v-if="layerConfig.type === LayerTypes.WMS"
             :layer-id="layerConfig.getID()"
             :opacity="layerConfig.opacity"
-            :url="layerConfig.getURL()"
+            :url="layerConfig.getURL(2056)"
```
ltshb, line 15: See my first comment

src/api/layers/AbstractLayer.class.js
pakb: So I've re-used what was declared in `coordinateUtils.js`, there was already an enum that was only missing the EPSG:3857 in it.
I think I've replaced all direct declaration of number or string regarding EPSG in the code with this enum

src/modules/map/components/openlayers/OpenLayersWMTSLayer.vue
```diff
@@ -56,6 +58,11 @@ export default {
             source: new XYZSource({
                 projection: this.projection,
                 url: this.url,
+                tileGrid: new TileGrid({
+                    resolutions: TILEGRID_RESOLUTIONS,
+                    extent: TILEGRID_EXTENT,
+                    origin: TILEGRID_ORIGIN,
+                })
```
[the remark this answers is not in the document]
pakb: I've only added the TileGrid declaration if the wanted projection is LV95, this should be more consistent this way, thanks!

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -55,10 +55,10 @@ export default {
...
                 opacity: this.opacity,
                 source: new TileWMS({
+                    projection: this.projection,
                     url: this.url,
                     gutter: this.gutter,
                     tileGrid: new TileGrid({
-                        projection: this.projection,
                         tileSize: WMS_TILE_SIZE,
                         origin: TILEGRID_ORIGIN,
                         resolutions: TILEGRID_RESOLUTIONS,
```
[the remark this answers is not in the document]
pakb: Here too, I've declared the tilegrid only if the wanted coordinate is LV95

ltshb (approved):
👍🏼

PR #254, opened 2022-09-14
Deactivated "embedding" part of share menu for MVP

jedef:
[Test link](https://sys-map.dev.bgdi.ch/feat-2598-deactivate-embed-in-share-menu/index.html)


pakb (commented):
Maybe we should comment it instead? So it will be easier to put it back.
And also, you might want to remove the component import and usage declaration in the JS.

jedef:
You're right. I have commented it in the html template so that we now where we removed it. And I have deleted the import and usage declaration.

pakb (approved):

PR #255, opened 2022-09-15
Feat 2440 add new topics bav and isos

jedef:
- Added the icon assets for topcis BAV and ISOS
- Number of columns adjust to the width of the screen in the topics popup.

[Test link](https://sys-map.dev.bgdi.ch/feat-2440-add-new-topics-bav-and-isos/index.html)


pakb (approved):
🚀

PR #257, opened 2022-09-19
Add contrast to the main BG selector button on desktop

pakb:
So that it catches more the eye, even when you are using a gray BG map (otherwise it blends too much with the map and it's difficult to see there's a button there)

[Test link](https://sys-map.dev.bgdi.ch/add_contrast_to_bg_text/index.html)


jedef (approved):
👍

PR #258, opened 2022-09-19
BGDIINF_SB-2009: share geolocation position when creating shortlink

pakb:
If the user has activated his geolocation, the generated share link (shortlink) will point to his current position with a marker. Geolocation URL param is also removed from the generated shortlink, so that the person opening the link is not placed where he currently stands, but where the person that has generated the link was.

Quick note on static image handling : I had to fix how images were imported in OpenLayersMarker.vue,. Seems like it went through without us noticing, but since Vite migration, static assets can't be imported through a require anymore, but through an import.

[Test link](https://sys-map.dev.bgdi.ch/feat_bgdiinf_sb-2009_sharelink_with_position/index.html)

### DIFF
--- a/src/store/modules/share.store.js
+++ b/src/store/modules/share.store.js
@@ -22,9 +22,14 @@ export default {
     },
     getters: {},
     actions: {
-        async generateShortLink({ commit }) {
+        async generateShortLink({ commit }, withBalloonMarker = false) {
             try {
-                const shortLink = await createShortLink(window.location.href)
+                const shortLink = await createShortLink(
+                    // we do not want the geolocation of the user clicking the link to kick in, so we force the flag out of the URL
+                    window.location.href.replace('&geolocation=true', '') +
+                        // if the geolocation was being tracked by the user generating the link, we place a balloon (dropped pin) marker at his position (center of the screen, so no need to change any x/y position)
+                        (withBalloonMarker ? '&crosshair=marker' : '')
+                )
                 if (shortLink) {
                     commit('setShortLink', shortLink)
                 }
--- a/src/modules/map/components/openlayers/OpenLayersMarker.vue
+++ b/src/modules/map/components/openlayers/OpenLayersMarker.vue
@@ -5,13 +5,18 @@
 </template>
 
 <script>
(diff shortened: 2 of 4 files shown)


src/store/modules/share.store.js
```diff
@@ -22,9 +22,14 @@ export default {
...
-        async generateShortLink({ commit }) {
+        async generateShortLink({ commit }, withBalloonMarker = false) {
             try {
-                const shortLink = await createShortLink(window.location.href)
+                const shortLink = await createShortLink(
+                    // we do not want the geolocation of the user clicking the link to kick in, so we force the flag out of the URL
+                    window.location.href.replace('&geolocation=true', '') +
+                        // if the geolocation was being tracked by the user generating the link, we place a balloon (dropped pin) marker at his position (center of the screen, so no need to change any x/y position)
+                        (withBalloonMarker ? '&crosshair=marker' : '')
+                )
```
ltshb, line 32: Because you have to check and remove the `geolocation=true` here, would it not make more sense to do add the marker based on this ?
```javascript
async generateShortLink({ commit }) {
    let href = window.location.href.replace("&geolocation=true", "&crosshair=marker")
    try {
        const shortLink = await createShortLink(window.location.href)
        ...
    }
    ...
}
```
This way you don't have to pass any parameter and in the code it is clear that the url contains a `?`. Because in you code if for any reason the `withBalloonMarker` is set to `true` but the `href` doesn't contain any query param (I know this is very unlikely), the url will be brocken.

pakb: the issue is, you could have activated your GPS, but not be tracking your position at the moment (you've panned somewhere else on the map, and want to share this portion of the map, not your current position).
That's why this balloon marker is placed only when `state.geolocation.active && state.geolocation.tracking`.

But we do not want the geolocation URL param to be passed along in any case

pakb: and yes, placing a `&...=...` is "risky", but at this point in time of the app's life cycle, there's no way lat/lon/zoom or other mandatory params haven't been placed, so it's not that big of a risk

ltshb: Ok 
```

ltshb (approved):

PR #260, opened 2022-09-26
BGDIINF_SB-2503 : matching old viewer style for textual buttons in drawing mode

ltkum:
[Test link](https://sys-map.dev.bgdi.ch/feature_bgdiinf-sb-2503-match-old-style-for-buttons/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingExporter.vue
+++ b/src/modules/drawing/components/DrawingExporter.vue
@@ -1,11 +1,12 @@
 <template>
     <DropdownButton
-        class="m-1"
+        class="m-1 drawing-button border-right"
         :title="$t('export_kml')"
         :current-value="$t('export_kml')"
         :items="exportOptions"
         :disabled="isDrawingEmpty"
         with-toggle-button
+        text-bold
         data-cy="drawing-toolbox-export-button"
         @select:item="onExportOptionSelected"
         @click="exportDrawing(false)"
--- a/src/modules/drawing/components/DrawingExporter.vue
+++ b/src/modules/drawing/components/DrawingExporter.vue
@@ -1,6 +1,6 @@
 <template>
     <DropdownButton
-        class="m-1 drawing-button border-right"
+        class="m-1"
         :title="$t('export_kml')"
         :current-value="$t('export_kml')"
         :items="exportOptions"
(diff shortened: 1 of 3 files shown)


src/modules/drawing/components/DrawingExporter.vue
```diff
@@ -1,11 +1,12 @@
 <template>
     <DropdownButton
-        class="m-1"
+        class="m-1 drawing-button border-right"
```
jedef, line 3: As these classes are defined in the parent module (and also scoped only to the parent module), I think you should also assign them in the parent class, as else it breaks the independence of the modules and is harder to understand. I.e. like this:
`<DrawingExporter :is-drawing-empty="isDrawingEmpty" class="drawing-button border-right" />`

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -206,6 +209,21 @@ $zindex-drawing-toolbox: -1;
...
 }
+.drawing-button {
+    touch-action: manipulation;
+    cursor: pointer;
+    background-image: none;
+    background-color: transparent;
+    border: 1px solid transparent;
+    border-radius: 4px;
+    width: 33%;
+}
```
jedef, line 220: Lots of the properties in the "drawing-button" class (like background-image and background-color) seem to be the default values. Maybe you could remove the attributes that are not necessary.

jedef (commented):
image

The dropdown menu of the drawing exporter is placed on the separator. Could you please fix that?

src/modules/drawing/components/DrawingExporter.vue
ltkum: This seems like a good idea. It is now done.

src/modules/drawing/components/DrawingToolbox.vue
ltkum: I removed what was necessary to remove. I left the background-color as it is not transparent by default and we need it to be transparent.

jedef (commented):
image

Sorry if I am nitpicking, but now it seems that the englisch version is not that beautiful, as the spacing seems uneven. As you seem to be in a flex container, you can probably achieve what you want by setting 'flex: 1' on the buttons.

Also with the button on the left, hovering over it changes the color of the text, while hovering over the two other buttons changes their background color, which is a bit inconsistent. So I would change it that hovering the left button also changes the background color and not the text color.

pakb:
The "export" button still has some border around it (and between the caret and button), is it possible to easily remove it?

pakb:
As discussed in our "Button talk" we will close this one, and switch the button to the default bootstrap `btn-light` instead of replicating mf-geoadmin3's code

PR #266, opened 2022-10-25
BGDIINF_SB-2611: add Swiss flag as favicon

pakb:
Use a SVG favicon in the process instead of the old .ico file. There's a fallback to a PNG image in case the browser doesn't support SVG favicons. Those favicons were generated with the SVG in src/assets/svg/swiss-flag.svg using the online tool https://realfavicongenerator.net/

I've also set the title of the page to the official english title in the process, so that it may not accidentally show as "Vue app" anymore...

[Test link](https://sys-map.dev.bgdi.ch/feat_bgdiinf_sb-2611_swiss_flag_as_favicon/index.html)


index.html
```diff
@@ -2,10 +2,11 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8" />
-        <link rel="icon" href="/favicon.ico" />
+        <link rel="icon" type="image/svg+xml" href="/favicon/favicon.svg">
+        <link rel="icon" type="image/png" href="/favicon/favicon.png">
         <!-- Disallow zooming on touch, as the zoom gesture is already reserved for zooming the map -->
         <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
-        <title>Vite App</title>
+        <title>Maps of Switzerland - Swiss Confederation - map.geo.admin.ch</title>
```
ltshb, line 9: Should not be this localized based on the selected language ?

pakb: It is later on in `App.vue`, I just wanted to avoid having "dev" placeholder further down the line (if for instance something goes wrong with JS and nothing is loaded)

index.html
```diff
@@ -2,10 +2,11 @@
 <html lang="en">
     <head>
         <meta charset="UTF-8" />
-        <link rel="icon" href="/favicon.ico" />
+        <link rel="icon" type="image/svg+xml" href="/favicon/favicon.svg">
+        <link rel="icon" type="image/png" href="/favicon/favicon.png">
```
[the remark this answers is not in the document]
pakb: I'll look into declaring a manifest or something like that, it could help the browser decide which one to load and not do this double/triple load at startup

jedef:
In my browser (chromium) it loads correctly the svg image (https://sys-map.dev.bgdi.ch/feat_bgdiinf_sb-2611_swiss_flag_as_favicon/icon.svg) but not the png image, as it tries to access https://sys-map.dev.bgdi.ch/icon-192.png where https://sys-map.dev.bgdi.ch/feat_bgdiinf_sb-2611_swiss_flag_as_favicon/icon-192.png would be the correct URL. As firefox does not load the png, the 404 error is not visible there.
I'm not sure if this behavior is expected or not. It surely will work after merging because then the beforementioned URL will be correct.

pakb:
oh right, maybe I should define the icons relatively in the webmanifest. As we are deploying branches in subfolder it plays a huge role not to access stuff from `/` but from `./`

jedef (approved):
This seems good to me

PR #267, opened 2022-10-25
Update libs

pakb:
Also fixes in the process of updating libs the d3-color vulnerability, with a npm audit fix.

Meaning that further npm install with new materials will also certainly require a npm audit fix as long as d3 is not updated with the latest version of d3-color

[Test link](https://sys-map.dev.bgdi.ch/update_node_requirement_and_update_libs/index.html)


ltshb (approved):
LGTM but CI is not passing.

PR #268, opened 2022-10-26
BGDIINF_SB-2508: Added feedback on selection of drawing icon color.

ltshb:
Also improve the Text color selection and made the drawing tool button the same color as the drawing option buttons.

Here some screen shot of initial implementation:
image
image

Current implementation screenshot:
image

NOTE: the user feedback in the icon selections will be done in a separate PR.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2508-drawing-marker/index.html)


pakb (commented):
I'm wondering if we should keep the black border, especially with the red color.
With the red button, when red is selected, the contrast between the button background and the color is not sufficient, it's hard to read what's going on.
What do you think?

ltshb:
@pakb if we keep the black border then I cannot use the icon. Let discuss it live.

src/modules/infobox/components/styling/DrawingStyleColorSelector.vue
```diff
@@ -1,33 +1,39 @@
...
-    >
-        <div
+    <div class="color-select-box bg-light border-light" data-cy="drawing-style-color-select-box">
+        <ButtonWithIcon
             v-for="color in colors"
             :key="color.name"
             :data-cy="`color-selector-${color.name}`"
+            :button-font-awesome-icon="['fas', 'circle']"
+            :style="{ color: color.name }"
+            :primary="currentColor === color"
```
jedef, line 9: You should probably implement a better check than "currentColor === color" (maybe implement an equals function that checks the values instead of the reference) as for now if you close the drawing, reload the page, and reopen the drawing, the selection is lost (no button is highlighted anymore).

ltshb: Very good catch

ltshb: I corrected it thanks @jedef

ltshb:
@pakb I used the white/black stroke for the color icons as you mentioned. I agreed it looks better so. I've also added a light grey background to the button group, in the following PR #274 would be the same without this light grey background (similar to other button groups that we have outside the drawing styling), but I prefer the solution from this PR.

pakb (approved):

PR #269, opened 2022-10-27
BGDIINF_SB-2509: Improve configuration menu (lang and ui mode)

ltshb:
Now we use the btn-light and btn-primary for the button for a better nice looking
and also for ui consistency accross other buttons.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2509-lang-menu-style/index.html)

### DIFF
--- a/src/modules/i18n/components/LangSwitchToolbar.vue
+++ b/src/modules/i18n/components/LangSwitchToolbar.vue
@@ -1,27 +1,29 @@
 <template>
-    <div class="lang-switch-toolbar p-1 btn-group">
-        <ButtonWithIcon
+    <div class="lang-switch-toolbar p-1">
+        <button
             v-for="lang in languages"
             :key="lang"
-            :lang="lang"
-            :danger="lang === currentLang"
-            :transparent="lang !== currentLang"
-            :button-title="lang.toUpperCase()"
-            :small="isDesktopMode"
+            class="btn"
+            :class="{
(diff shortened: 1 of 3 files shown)


ltshb:
@davidoesch I would suggest to change the translation for the ui mode from `Link zur Desktop Version` to `Desktop Version` so we are consistent with the mobile version button and it take less places, what do you think ?
@pakb how can we change this ?

pakb:
We've used here an existing translation, not the most relevant, but close enough. We could add a new translation to the translation file to fit our needs if need be.

But IMO we shouldn't do too much work on this part of the menu as it is meant to disappear in the future. We have to find a way to let the user change the lang on mobile, but for desktop we want to have the same look and feel as the one currently on map.geo.admin.ch
Regarding the layout switch button, it was more there for dev/debug purposes, and we should get rid of it before publishing the viewer as MVP.

ltshb:
@pakb Ok for the translation if we remove it in future.

davidoesch:
Desktop/ mobile switch we won't have on mobile

pakb (approved):
Lang change looks better!
Let's keep it that way translation wise for the layout switch, I'll create a ticket with the task to remove this layout switch button, so that we do not forget to do it before MVP

jedef (approved):
This also looks good to me, the language change buttons look indeed better. :)

PR #272, opened 2022-10-31
BGDIINF_SB-2502: add map saved text to drawing toolbox

jedef:
Add a text in the toolbox to inform the user if the KML is saved.

Remarks:
- The translation file is missing a translation in case of an error, so for now the message is in English only in this case.
- I myself would never have seen the message if the ticket did not explain where it is in the old viewer, as it is very well hidden and not at the expected location. (I would expect such messages to appear on the top right of the screen), so maybe we should consider putting it at another place?

[Test link](https://sys-map.dev.bgdi.ch/feat-2502-add-map-saved-text-to-drawing-toolbox/index.html)


davidoesch:
The position is ok: it's where the user expects the save button (export delete etc is already there)
Proximity of features ...

jedef:
> The position is ok: it's where the user expects the save button (export delete etc is already there) Proximity of features ...

Ok I see your point, so let's leave it there in this case.

src/modules/drawing/DrawingModule.vue
```diff
@@ -293,22 +296,44 @@ export default {
             clearTimeout(this.KMLUpdateTimeout)
             const kml = generateKmlString(this.drawingLayer.getSource().getFeatures())
             if (kml && kml.length) {
+                this.savingStatus = 'draw_file_saving'
                 try {
                     await this.saveDrawing(kml)
                 } catch (e) {
                     log.error('Could not save KML layer: ', e)
+                    // We should probably add a translation key for that
+                    this.savingStatus = 'Error while saving the drawing'
```
ltshb, line 305: Agree we MUST add a translation here.

src/modules/drawing/DrawingModule.vue
```diff
@@ -293,22 +296,44 @@ export default {
...
+         * Call this method when there are or when there will be unsaved changes. Change the saving
+         * status to "possibly unsaved changes"
+         */
+        willModify() {
+            if (
+                this.savingStatus === 'draw_file_saved' ||
+                this.savingStatus === 'draw_file_saving'
+            ) {
+                this.savingStatus = ''
+            }
```
ltshb, line 336: Would it not make more sense to have something like this
```javascript
clearStatus() {
  if (this.savingStatus !== '') {
    this.savingStatus = ''
  }
}
```
Because with you code above if the status is an error then it will not be erased.

jedef: The idea of this if statement was to make the error message a bit more persistant / to not discard the error message before the next saving attempt is made. So the error message can be replaced the the draw_file_saved or draw_file_saving status, but not by the empty status.

src/modules/drawing/DrawingModule.vue
jedef: Right. Do you now how to do that? Can I try to add it myself or is there some kind of translation team?

ltshb: I don't know at best ask @pakb

src/modules/drawing/DrawingModule.vue
ltshb: I see make sense. I'm not a big fan of such string compare which to make it worse are translation dependent. 
It might wise to either use a `savingStatus` as an enum and in the rendering translate the enum to the corresponding string internationalized. This way the `DrawingModule` is not responsible to choose what will be presented to the user, it is only responsible to notify the status as an enum, then `DrawingToolbox` would be responsible for the rendering choosing the correct way to render (using translation or any fancy stuff). `DrawingToolbox` is the component that will render the status, then it should have fully control on what it will display based on a status flag.

src/modules/drawing/DrawingModule.vue
pakb: we could use use this existing key for now :

| key | de | en | fr | it | rm |
| ----| --- | --- | -- | -- | -- |
| upload_failed | Fehler beim Hochladen! | Upload error! | Erreur de chargement! | Caricamento fallito! | Errur da chargiar |  

Or we could add one more, but that require some translation effort

pakb (approved):

PR #274, opened 2022-10-31
BGDIINF_SB-2508:  Added feedback on selection of drawing icon color - 3

ltshb:
Alternative to #268 without light grey background for the button group. This would be more consistent with other groups like the languages or the drawing tools.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2508-drawing-marker-3/index.html)


ltshb:
We keep the original design as in #268

PR #275, opened 2022-10-31
BGDIINF_SB-2508: Added feedback in drawing marker icon selection

ltshb:
This PR depends on #268.
[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2508-drawing-marker-icons/index.html)

NOTE: There is still a bug on iphone with the stroke of icons. When the icon selection is close, everything works as intended but when it is opened and we change the color, the icon stroke won't but correctly updated (changing from black to white depending on the color). To fix the stroke the use as to close the icon selection and reopen it. I could not find any solution to this bug and it seems that only iphone is affected. Tested on following devices
- Ubuntu Chrome and Firefox : OK
- Android tablet: OK
- **Iphone 11 IOS 16.1 Safari, Chrome, Firefox: NOK**
- Windows Edge: OK

### DIFF
--- a/src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
+++ b/src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
@@ -192,5 +192,9 @@ export default {
         width: 2rem;
         height: 2rem;
     }
+
+    button {
+        --bs-btn-padding-x: 0.25rem;
+    }
 }
 </style>
--- a/src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
+++ b/src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
@@ -57,7 +57,13 @@
                         :alt="icon.name"
                         :src="generateColorizedURL(icon)"
                         class="marker-icon-image"
-                        :class="getImageStrokeClass(feature.fillColor)"
+                        :class="
+                            getImageStrokeClass(
+                                currentIconSet.isColorable,
+                                feature.icon.name === icon.name,
+                                feature.fillColor.name
+                            )
+                        "
                         crossorigin="anonymous"
                     />
                 </button>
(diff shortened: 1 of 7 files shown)


davidoesch:
@ltshb : no issue on iphone7 and iOS 16.1

src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
```diff
@@ -131,6 +147,15 @@ export default {
...
         },
+        getImageStrokeStyle(isColorable, isSelected, color) {
+            if (isColorable) {
+                return {
+                    filter: `drop-shadow(1px 1px 0 ${color.border}) drop-shadow(-1px -1px 0 ${color.border})`,
+                }
+            } else if (isSelected) {
+                return { filter: 'drop-shadow(0px 0px 0 white)' }
+            }
+        },
```
pakb, line 158: I've tested on iOs, and also Safari desktop, and Safari is really playing with us here... The rule was CSS clearly set when I've played with colors, so on the CSS part everything was correct but the rendering was not. I then unchecked, and checked back the rule in the CSS inspector, and voilà, it was now correctly rendered.
So yeah, Safari issue, we can't do much about it...

src/utils/PopoverButton.vue
```diff
@@ -136,6 +136,10 @@ export default {
             // shows an arrow at the border of the popover that points to the button location
             arrow: true,
             trigger: 'click',
+            // The default tippy maxWidth is 350px, we need to add one more pixel here in order
+            // to avoid a useless horizontal scrollbar in the drawing icon selection menu on
+            // Windows Edge Browser.
+            maxWidth: 351,
```
pakb, line 142: in Bootstrap's variable, there's one that is called
```
$toast-max-width:                   350px !default;
```
We could maybe use it with a `calc($toast-max-width + 1);` so that we avoid having fixed value in our component's (S)CSS

src/modules/infobox/components/styling/DrawingStyleIconSelector.vue
pakb: I imagine you've tested both ways, having the filter defined in the SCSS part, or here?
If both are not doing a good job on Safari, and thus we can just ignore this ***** browser, I would prefer that we define a class here, and not directly CSS rules.

ltshb: I leave it like this as discussed, this allow to not touch the component if the `border.color` changes in the future.

src/utils/PopoverButton.vue
ltshb: I removed and resize the icons to fix the issue, see https://github.com/geoadmin/web-mapviewer/pull/275/commits/992467c8590c4122b569241235084cf3ba505b74

pakb (approved):

PR #277, opened 2022-11-01
BGDIINF_SB-2476: Debounce the search bar

ltshb:
Debounce the search bar input keys.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2476-search-bar/index.html)


pakb (approved):
I would even raise that to a bigger time window, maybe a quarter of a second or more.
It's still requesting for each character I'm writting if I write not too fast "Ibex" for instance.

ltshb:
@pakb I tried several timeout value up to 300ms but with a higher value the search bar is then less responsive, that's why I chose a small timeout to keep a very responsive search bar.

PR #279, opened 2022-11-02
Harmonize button spacer and remove duplicate delete button in drawing

ltshb:
Use a global variable to add space between adjacent buttons.

Also removed useless legacy css code in FeatureStyleEdit.

[Test link](https://sys-map.dev.bgdi.ch/feat-design-tune/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureProfile.vue
+++ b/src/modules/infobox/components/FeatureProfile.vue
@@ -25,7 +25,7 @@
             </div>
             <div ref="profileTooltipArrow" class="profile-tooltip-arrow"></div>
         </div>
-        <div v-if="profileInformation.length" class="d-flex p-2">
+        <div v-if="profileInformation.length" class="d-flex p-2 profile-popup-info-buttons">
             <div
                 class="flex-grow-1 profile-info-container d-flex border border-light ps-1 pe-4 py-1"
                 data-cy="profile-popup-info-container"
@@ -54,6 +54,7 @@
                 @click="onCSVDownload"
             />
             <ButtonWithIcon
+                v-if="showDeleteButton"
                 :button-font-awesome-icon="['far', 'trash-alt']"
                 data-cy="profile-popup-delete-button"
                 @click="onDelete"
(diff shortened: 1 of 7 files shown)


src/modules/infobox/components/FeatureProfile.vue
```diff
@@ -53,11 +53,6 @@
                 data-cy="profile-popup-csv-download-button"
                 @click="onCSVDownload"
             />
-            <ButtonWithIcon
-                :button-font-awesome-icon="['far', 'trash-alt']"
-                data-cy="profile-popup-delete-button"
-                @click="onDelete"
-            />
```
pakb, line 60: I think this should stay. How are we supposed to remove a measure line without this button?

pakb (commented):
Just some doubts about the delete button removal, otherwise nice looking!

src/modules/infobox/components/FeatureProfile.vue
ltshb: @pakb See screenshot below, this line remove the button crossed in red, but you still have the button in the description popup to delete the line. I don't think we need 2 buttons here.

pakb: that works for a line yes, and it makes sense to remove the duplicate in this case, but what about measure?
image

ltshb: Very good point ! I didn't know about this use case. So I'll check if I can easily remove only for the line otherwise I leave it.

davidoesch:
Hmm once you are in measure mode
image
 there is no pop up. the delete button was for : save the user the step for closing the reslt window and klick again on the line to delete it.... saves saves the user three klicks. 

the measurement function does not have a delete option... thsi is wy we added the delete button

even if it is a button more ...I would keep it in this case

ltshb:
@davidoesch @pakb I reworked the PR. Now in the profile the delete button is only shown for the measure tool and not for the line polygon. For the later the delete button is part of the description popup. This way for line polygon we only have 1 delete button instead of two. This way we save place on mobile device for the profile infos.
For a better nice looking I also added a small margin between buttons.

davidoesch (approved):

PR #280, opened 2022-11-02
BGDIINF_SB-2657: Added dynamic app versioning based on git tag

ltshb:
Unfortunately we cannot make use of the `version` field in `package.json` because it is static which means that we would need to modify it during a PR merge before tagging. They are already tools and github action for such use case, the issue is that they won't work with branch protection as we have.
[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2657-versioning/index.html)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,4 +1,5 @@
 import { fileURLToPath, URL } from 'url'
+import { gitDescribeSync } from 'git-describe'
 
 import { defineConfig } from 'vite'
 import vue from '@vitejs/plugin-vue'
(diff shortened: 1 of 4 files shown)


vite.config.ts
```diff
@@ -16,7 +17,7 @@ export default defineConfig(({ command, mode }) => {
         },
         define: {
             // Getting package.json version in order to expose it to the app
```
pakb, line 19: maybe also edit or remove this comment, it's now outdated 😉

pakb (approved):

PR #281, opened 2022-11-03
BGDIINF_SB-2657: Fixed versioning

ltshb:
The git hash was always appended to the version even if the version was from
a clean tag (not dirty and with extra commit).
By using semverString instead of raw, this solve the issue, but we need to
prepend the `v` prefix.

[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2657-versioning/index.html)

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -16,7 +16,7 @@ export default defineConfig(({ command, mode }) => {
             },
         },
         define: {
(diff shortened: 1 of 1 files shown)


pakb (approved):
Will that mean that we can have tags without smallhash on https://sys-map.dev.bgdi.ch/ as it is now?

PR #282, opened 2022-11-07
Improved look and feel of menu button

ltshb:
Removed the top corner of the open/close menu button and increase a bit spaces.

[Test link](https://sys-map.dev.bgdi.ch/feat-menu-button-look/index.html)

### DIFF
--- a/src/utils/ButtonWithIcon.vue
+++ b/src/utils/ButtonWithIcon.vue
@@ -27,7 +27,7 @@ export default {
         },
         buttonFontAwesomeIcon: {
             type: Array,
-            required: true,
+            default: new Array(),
         },
         iconSize: {
             type: String,
@@ -127,10 +127,12 @@ export default {
         },
         labelClasses() {
             let className
-            if (this.direction === 'column') {
-                className = this.iconsBeforeText ? 'mt-2' : 'mb-2'
-            } else {
-                className = this.iconsBeforeText ? 'ms-2' : 'me-2'
+            if (this.buttonFontAwesomeIcon.length > 0) {
+                if (this.direction === 'column') {
+                    className = this.iconsBeforeText ? 'mt-2' : 'mb-2'
+                } else {
+                    className = this.iconsBeforeText ? 'ms-2' : 'me-2'
+                }
             }
             return className
         },
--- a/src/utils/ButtonWithIcon.vue
+++ b/src/utils/ButtonWithIcon.vue
@@ -128,9 +128,9 @@ export default {
         labelClasses() {
             let className
             if (this.direction === 'column') {
-                className = this.iconsBeforeText ? 'mt-1' : 'mb-1'
+                className = this.iconsBeforeText ? 'mt-2' : 'mb-2'
             } else {
-                className = this.iconsBeforeText ? 'ms-1' : 'me-1'
+                className = this.iconsBeforeText ? 'ms-2' : 'me-2'
             }
             return className
         },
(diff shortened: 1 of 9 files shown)


jedef:
The new visual appearance of the menu button looks good to me.

It looks like the modifications in the "buttonwithmodal" component offcentered the text of the confirmation dialog.
image
The other buttons that are affected by the change still seem to look good, but I don't know if I have seen all buttons affected.

The menu in the drawing mode has also a "close menu" button that should match the style of the main menu button.
image

src/utils/ButtonWithIcon.vue
```diff
@@ -128,9 +128,9 @@ export default {
         labelClasses() {
             let className
             if (this.direction === 'column') {
-                className = this.iconsBeforeText ? 'mt-1' : 'mb-1'
+                className = this.iconsBeforeText ? 'mt-2' : 'mb-2'
             } else {
-                className = this.iconsBeforeText ? 'ms-1' : 'me-1'
+                className = this.iconsBeforeText ? 'ms-2' : 'me-2'
```
pakb, line 133: I have the impression that this pushes the text a bit off center, we should either compensate on the other side or go back to a smaller margin.
What was your goal by setting a bigger margin here?

ltshb: The goal is to have a margin between icon and label, not have them too close
Before:
image
After:
image

The issue also mentioned by @jedef is that the `ButtonWithIcon` class can be used without icon ! It is not very user friendly. I'll quickly fix this but we should consider some rework here to have the class either renamed and only used with icon. IMHO a button without icon doesn't require a component as we use bootstrap.

ltshb: By the way this of center was already present, just less bigger as it used a `me-1` instead of `me-2`.

jedef (commented):
It seems like you have forgotten to change this button:
image

I would also suggest to define a global css class (e.g. "default-button") that includes the needed bootstrap  classes "btn" and "btn-light" and use this class instead in the code. This would make the code less tightened to bootstrap, make it easier to know which button class to use when writing new code and make it easier to change the styles of all our buttons in one go if we want to.

ltshb:
@jedef the time series button is already done or is in progress by @ltkum.
Using custom classes that uses bootstrap class to decouple our code from bootstrap is good idea but require quite some effort as we would need to check the whole code, not only for buttons.

src/utils/ButtonWithIcon.vue
pakb: the goal having a single component to manage the button was specifically to encapsulate bootstrap usage in one place, so that if we wanted to switch to something else, or tweak the design of existing buttons, it would only need a change in one place.

ltshb:
@pakb I corrected as discussed using `ps-x` and `pe-x` for the padding.

pakb (approved):

jedef:
> @jedef the time series button is already done or is in progress by @ltkum. Using custom classes that uses bootstrap class to decouple our code from bootstrap is good idea but require quite some effort as we would need to check the whole code, not only for buttons.

Ok we can do it in another ticket then (or in multiple, among others one for the buttons, as we do not necessarily need to switch all at once). The rest looks good to me now.

jedef (approved):

ltshb:
@pakb after rebasing to dev I did some small improvement see latest commits. I also created a ticket https://jira.swisstopo.ch/browse/BGDIINF_SB-2680 to solve the button jungle after MVP ;-)

pakb (approved):

PR #283, opened 2022-11-07
Feat 2235 rework click on mobile device

jedef:
Reworked click on mobile device. Implementation is as stated in the ticked, but with one small modification. If one needs to right click or long click to display the location popup only depends on the input device (mouse or touch screen) and not on the the ui mode (desktop or phone ui). On the other hand, if fullscreen mode can be activated or not only depends on the ui mode and not on the input device.

@pakb Can you please test the if long click to display the location popup works on IOS? I have read that safari may not implement correctly the oncontextmenu event. I have already a fix ready, but only want to add it if it is really necessary.

[Test link](https://sys-map.dev.bgdi.ch/feat-2235-rework-click-on-mobile-device/index.html)

### DIFF
--- a/src/store/modules/map.store.js
+++ b/src/store/modules/map.store.js
@@ -112,6 +112,13 @@ export default {
                 commit('setDisplayedProjection', CoordinateSystems[projectionId])
             }
         },
+
+        displayLocationPopup({ commit }) {
+            commit('setDisplayLocationPopup', true)
+        },
+        hideLocationPopup({ commit }) {
+            commit('setDisplayLocationPopup', false)
+        },
     },
     mutations: {
         setHighlightedLayer: (state, layer) => (state.highlightedLayer = layer),
@@ -120,7 +127,6 @@ export default {
         mapStoppedBeingDragged: (state) => (state.isBeingDragged = false),
         setPinnedLocation: (state, coordinates) => (state.pinnedLocation = coordinates),
         setDisplayedProjection: (state, projection) => (state.displayedProjection = projection),
-        displayLocationPopup: (state) => (state.displayLocationPopup = true),
-        hideLocationPopup: (state) => (state.displayLocationPopup = false),
+        setDisplayLocationPopup: (state, display) => (state.displayLocationPopup = display),
     },
 }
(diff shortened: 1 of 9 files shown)


pakb:
sadly iOs is not triggering the `contextmenu`, as you feared... So you can add your fix if you have one

jedef:
> sadly iOs is not triggering the `contextmenu`, as you feared... So you can add your fix if you have one

Ok thanks, I have pushed the fix. Could you please tell me if this fixed the problem?

pakb:
Yes now it works!
I'll start the code review then

src/store/plugins/click-on-map-management.plugin.js
```diff
@@ -59,25 +55,45 @@ const clickOnMapManagementPlugin = (store) => {
...
+                        store.dispatch('toggleFullscreenMode')
+                    }
+                    store.dispatch('setSelectedFeatures', newSelectedFeatures)
+                })
+            }
+            if (isContextMenuClick) {
                 store.dispatch('clearAllSelectedFeatures')
+                store.commit('displayLocationPopup')
+            } else {
+                store.commit('hideLocationPopup')
```
pakb, line 93: I would prefer that we stick to `store.dispatch`ing instead of using directly mutations, it is then easier to see what's happnening when debugging as we only have to watch actions to see what the app is doing (otherwise we have to go back and forth between actions and mutations to grasp what was happening).

src/store/modules/map.store.js
```diff
@@ -129,5 +120,7 @@ export default {
         mapStoppedBeingDragged: (state) => (state.isBeingDragged = false),
         setPinnedLocation: (state, coordinates) => (state.pinnedLocation = coordinates),
         setDisplayedProjection: (state, projection) => (state.displayedProjection = projection),
+        displayLocationPopup: (state) => (state.displayLocationPopup = true),
+        hideLocationPopup: (state) => (state.displayLocationPopup = false),
```
pakb, line 124: If you implement this as I've suggested in the store plugin, it would maybe be better to have this two distinct mutations moved to actions, and have only one mutation called `setDisplayLocationPopup` here

pakb (commented):
A little change in how the store is used in the plugin, and we should be good to go, nice work!

Could you check if `cy.longClickOnMap()` is used elsewhere, and if not remove it from `commands.js`?

pakb (approved):

PR #284, opened 2022-11-07
BGDIINF_SB-2608: Fixed missmatch between profile and geodesic line

ltshb:
When drawing a line or a measurement, we use a geodesic line. However the profile
backend did a flat filling to have at least 200 points. This lead to a missmatch
of a few pixels when drawing a line across switzerland.

To solve the issue we use the geodesic points as coordinate to pass to the
profile backend, this way the filling is done by the frontend instead of the
backend.

[Test link](https://sys-map.dev.bgdi.ch/bug-bgdiinf_sb-2608-profile/index.html)

### DIFF
--- a/src/modules/drawing/lib/drawingUtils.js
+++ b/src/modules/drawing/lib/drawingUtils.js
@@ -14,8 +14,8 @@ export function toLv95(input, epsg) {
 }
 
 /**
- * Wraps the provided webmercator coordinates in the world extents (i.e. the coordinate range
- * that if equivalent to the wgs84 [-180, 180))
+ * Wraps the provided webmercator coordinates in the world extents (i.e. the coordinate range that
+ * if equivalent to the wgs84 [-180, 180))
  *
  * @param coords The coordinates (or array of coordinates) to wrap
  * @param inPlace If false, the original coordinates remain untouched and only a copy is modified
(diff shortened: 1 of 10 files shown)


src/modules/drawing/lib/drawingUtils.js
```diff
@@ -104,6 +104,14 @@ export function formatAngle(value, digits = 2) {
 
 export function extractOpenLayersFeatureCoordinates(feature) {
     let coordinates = feature.getGeometry().getCoordinates()
```
jedef, line 106: Surprisingly, this seems to work. Still, I would rather try to find another solution as this may have lots of undesired side effects. This seems to store all computed geodesic coordinates directly in the editable feature. This means that the kml will store lots useless coordinates. This may in turn have unintended effects on other code that tries to access these coordinates and expects the "correct" coordinates or on other code that accesses this helper function.
I would rather try to access the geodesic coordinates directly when doing the request to service alti (e.g. in featureprofile.vue).

jedef (commented):
To fix the problem in the fronted by passing directly the geodesic coordinates to service alti seems a very good idea to me. However, the proposed code change seems to be at the wrong place. See comment above.

src/modules/drawing/lib/drawingUtils.js
ltshb: Good point.

ltshb: I checked, effectively this solution adds the geodesic to the kml drawing which is not really what we want. I'll add a `geodesicCoordinates` property to the `EditableFeature` object instead.

pakb (approved):

PR #288, opened 2022-11-09
BGDIINF_SB-2671: Upload build artifacts to S3

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/feat-bgdiinf_sb-2671-artifact-bucket/index.html)


pakb (approved):
Nice!
I've seen that there is a second bucket named "something something build artefacts" in the CI account, maybe we should delete it.

ltshb:
Good point those other buckets are not used and I'll delete them

PR #292, opened 2022-11-15
Replace UNIX specific script in build pipeline

pakb:
so that the CI can still decipher which branch it is running on, and window users can still build the project

[Test link](https://sys-map.dev.bgdi.ch/feat-replace-unix-specific-script/index.html)

### DIFF
--- a/vite-plugins/generate-build-info.js
+++ b/vite-plugins/generate-build-info.js
@@ -22,14 +22,6 @@ function getGitBranch(gitHash) {
     return branches[0]
 }
 
-const currentHash = execSync('git rev-parse HEAD').toString().trim()
-
-// We take the version from GIT_BRANCH but if not set, then take it from git show-ref
-let gitBranch = process.env.GIT_BRANCH
-if (!gitBranch) {
-    gitBranch = getGitBranch(currentHash)
-}
-
 /**
  * Small Rollout / Vite plugin that writes a JSON containing build information, such as date or app
  * version.
@@ -43,6 +35,14 @@ export default function generateBuildInfo(version) {
             sequential: true,
             order: 'post',
             async handler() {
+                const currentHash = execSync('git rev-parse HEAD').toString().trim()
+
+                // We take the version from GIT_BRANCH but if not set, then take it from git show-ref
+                let gitBranch = process.env.GIT_BRANCH
+                if (!gitBranch) {
+                    gitBranch = getGitBranch(currentHash)
+                }
+
                 const now = new Date()
                 const localChanges = execSync('git status --porcelain').toString().trim()
                 this.emitFile({
--- a/vite-plugins/generate-build-info.js
+++ b/vite-plugins/generate-build-info.js
@@ -1,6 +1,36 @@
 import { execSync } from 'child_process'
 
 /**
+ * Javascriptifaction of
+ * https://github.com/geoadmin/web-mapviewer/blob/4826f4a05df9a5a12fc3eea2a902618800fc9425/buildspec.yml#L43-L44
+ * so that it also functions properly on a window10/11 machine in PowerShell (or Window terminal)
+ *
+ * Original cmd is git show-ref | grep "$(git rev-parse HEAD)" | awk '!/refs/tags// {print $2}' |
+ * head -1
+ *
+ * @param {String} gitHash The current git hash
+ * @returns {String} The git branch without "origin/" or "refs/" portion
+ */
+function getGitBranch(gitHash) {
+    const headRefs = execSync('git show-ref --heads').toString().trim().split('\n')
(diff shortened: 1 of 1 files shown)


vite-plugins/generate-build-info.js
```diff
@@ -1,5 +1,27 @@
...
+ * Javascriptifaction of
+ * https://github.com/geoadmin/web-mapviewer/blob/4826f4a05df9a5a12fc3eea2a902618800fc9425/buildspec.yml#L43-L44
+ * so that it also functions properly on a window10/11 machine in PowerShell (or Window terminal)
+ *
+ * Original cmd is git show-ref | grep "$(git rev-parse HEAD)" | awk '!/refs/tags// {print $2}' |
+ * head -1
+ *
+ * @return {String} The git branch without "origin/" or "refs/" portion
+ */
+function getGitBranchEvenWhenOnCI() {
```
ltshb, line 13: Would simply call it `getGitBranch`. In comment above I would add a comment that this function retrieve the current matching git branch even if we are on a detached HEAD.

vite-plugins/generate-build-info.js
```diff
@@ -1,5 +1,27 @@
...
+    const gitShowRef = execSync('git show-ref --heads').toString().trim().split('\n')
```
ltshb, line 14: `const headRefs = ...`

vite-plugins/generate-build-info.js
```diff
@@ -1,5 +1,27 @@
...
+    const gitRevParseHead = execSync('git rev-parse HEAD').toString().trim()
```
ltshb, line 15: `const commitSha = ...`

vite-plugins/generate-build-info.js
```diff
@@ -1,5 +1,27 @@
...
+    // we now do a grep-like operation, only keeping refs that have the hash found in gitRevParseHead
+    // we also have to remove all prefixes on the branch, i.e. /refs/tags or /origin/ etc...
+    const matchingRefsWithoutPrefix = gitShowRef
```
ltshb, line 18: `const branches = ...`

ltshb (approved):
Good I would just keep more explicit names, the name should say what it actually contains.

ltshb (changes requested):
mmh an issue is when we retry a build, the info.json is not correct
```json
{
  "version": "v0.1.0-beta.187-2-g37104fc",
  "build": {
    "date": "2022-11-15T13:32:37.060Z"
  },
  "git": {
    "hash": "37104fcc2c8be6337cc3689ee212cf491e1779ea",
    "dirty": false,
    "localChanges": ""
  }
}
```
See https://s3.console.aws.amazon.com/s3/object/build-artifacts-swisstopo?region=eu-central-1&prefix=dev/web-mapviewer/v0.1.0-beta.187-2-g37104fc/v0.1.0-beta.187-2-g37104fc/info.json and https://eu-central-1.console.aws.amazon.com/codesuite/codebuild/974517877189/projects/web-mapviewer/build/web-mapviewer%3Ad48a1d41-e674-4c2d-bbfd-be58fd1f26b9/?region=eu-central-1

ltshb:
An idea would be to pass the git branch as environment variable when built from he CI and keep the logic here only when no environment variable for the git branch is defined (when building locally). So the git branch logic on the CI would always be from the buildspec and we would not have two different logic, same as it is now for the version.

vite-plugins/generate-build-info.js
ltshb: actually this could be passed as parameter as it is needed also below in line 40, to avoid calling it twice.

pakb:
As discussed in our call, we will use the `GIT_BRANCH` env variable when defined (so on the CI) and resort to our node based code only when this variable is absent. The CI is too picky with this piece of code and fails for no apparent reasons sometimes...

vite-plugins/generate-build-info.js
```diff
@@ -1,5 +1,35 @@
...
+    return branches[0]
+}
+
+const currentHash = execSync('git rev-parse HEAD').toString().trim()
+
+// We take the version from GIT_BRANCH but if not set, then take it from git show-ref
+let gitBranch = process.env.GIT_BRANCH
+if (!gitBranch) {
+    gitBranch = getGitBranch(currentHash)
+}
```
ltshb, line 31: Would it not make more sense to keep this code inside the async handler function ?

ltshb (approved):
Looks all good

vite-plugins/generate-build-info.js
pakb: I think it would make more sense, as if something would fail in this piece of code it would happen in post-build, and not at startup, I'll move the code inside

PR #297, opened 2022-11-16
BGDIINF_SB-2610: enhance legacy kml import

jedef:
When importing a kml from mf-geoadmin3, the viewer is now able to reconstruct the editable feature based on the style infomration that is stored in the kml.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2610-enhance-legacy-kml-import/index.html)


jedef:
> @pakb @jedef I'm wondering how do you check if the KML is from the old viewer or from the new ? You don't make use of the `author` or `authorVersion` info from the backend ? Or did I missed something?

The check is made in "src/api/features.api.js" line 255 - 257 without the use of the metadata from the backend. It would surely be more robust to also check the metadata from the backend, but I think that this can be implemented when we also implement the check to make sure that old kmls are always read only, as I don't think that I have enough time this afternoon to do this.

pakb (approved):

pakb (changes requested):
there's some issue with unit tests, seems like it is using axios or some other libs (that shouldn't happen while unit testing)
Plus the parsing of the old KML doesn't parse the icon color correctly, I'll investigate while @jedef is in vacations

pakb (approved):

PR #304, opened 2022-11-22
Improved mouse position e2e tests

ltshb:
This test did sometimes not pass without a useful info of the failure reason.
Hopefully with this changes we can have a better understanding why this test
is sometimes failing.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-mouseposition-e2e/index.html)

### DIFF
--- a/tests/e2e-cypress/integration/mouseposition.cy.js
+++ b/tests/e2e-cypress/integration/mouseposition.cy.js
@@ -21,7 +21,9 @@ const defaultCenter = [47.5, 7.5]
  * @param {String} text A string containing an LV95 or LV03 coordinate.
  */
 function parseLV(text) {
-    const matches = text.match(/([-\d'.]+), ([-\d'.]+)$/)
+    const matches = text.match(/([-\d'.]+),\s*([-\d'.]+)$/)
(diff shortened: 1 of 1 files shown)


pakb (approved):
good idea, although I'm suspecting nothing will be present in text, and the issue is somewhat related to mouse position not being interpreted correctly (we use a `.click` to position the mouse)

ltshb:
@pakb could not be due to a timing problem between `click()` and `cy.get('[data-cy="mouse-position"]')` ? Should we not have some kind of wait to have the position popup visible ?
```javascript
function checkMousePositionNumberValue(expectedX, expectedY, parser) {
    cy.get('[data-cy="map"]').click()
    cy.get('[data-cy="mouse-position"]')
        .invoke('text')
        .then(parser)
        .then(checkXY(expectedX, expectedY))
}
```

pakb:
There should not be backend requests here, it should be seamless, so no alias to wait on... If we continue having issue with this test we could consider waiting for an arbitrary length of time (which is a bad practice with Cypress, but this test is way too prone to fail not to try fixing it)

PR #309, opened 2022-11-24
BGDIINF_SB-2662: Use shortlink for position popup url

ltshb:
Set fallback to non short url in case the backend service is not available.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2662-shortlink-position-popup/index.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "scripts": {
         "start": "npm run dev",
         "dev": "vite --port 8080",
+        "dev-host": "npm run dev -- --host --cors",
         "preview": "npm run build:prod && vite preview --port 8080 --outDir dist/production",
         "preview:dev": "npm run build:dev && vite preview --port 8080 --outDir dist/development",
         "preview:int": "npm run build:int && vite preview --port 8080 --outDir dist/integration",
--- a/package.json
+++ b/package.json
@@ -3,8 +3,7 @@
     "private": true,
     "scripts": {
         "start": "npm run dev",
-        "dev": "vite --port 8080",
-        "dev-host": "npm run dev -- --host --cors",
+        "dev": "vite --port 8080 --host --cors",
         "preview": "npm run build:prod && vite preview --port 8080 --outDir dist/production",
         "preview:dev": "npm run build:dev && vite preview --port 8080 --outDir dist/development",
         "preview:int": "npm run build:int && vite preview --port 8080 --outDir dist/integration",
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -61,12 +61,13 @@
             <div>
                 <a href="http://what3words.com/" target="_blank">what3words</a>
             </div>
-            <div>
+            <div v-if="what3Words">
                 <span v-show="what3Words" data-cy="location-popup-w3w">
                     {{ what3Words }}
                 </span>
                 <LocationPopupCopySlot :value="what3Words" />
             </div>
+            <div v-else>-</div>
             <div>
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -4,6 +4,7 @@
     "scripts": {
         "start": "npm run dev",
         "dev": "vite --port 8080",
+        "dev-host": "npm run dev -- --host --cors",
```
pakb, line 7: I think we could `--host --cors` on `npm run dev` directly, it is a target solely used for local development, and we should often check on our real phones how it looks like

src/modules/map/components/LocationPopup.vue
```diff
@@ -126,101 +135,140 @@ export default {
             currentLang: (state) => state.i18n.lang,
             displayLocationPopup: (state) => state.map.displayLocationPopup,
         }),
-        clickCoordinates() {
-            return this.clickInfo?.coordinate
+        coordinate() {
+            return this.clickInfo?.coordinate || null
```
pakb, line 139: I'm not sure if having an optional access here wouldn't return null anyway? (or `undefined` which is also considered `false` in an `if(...)` statement)

src/modules/map/components/LocationPopup.vue
```diff
@@ -126,101 +135,140 @@ export default {
...
+            createShortLink(url)
+                .then((shortUrl) => {
+                    this.shareLinkUrlShorten = shortUrl
+                    this.updateQrCode(shortUrl)
+                })
+                .catch((err) => {
+                    log.error(`Failed to shorten Share URL`)
+                    this.shareLinkUrlShorten = null
+                })
+        },
```
pakb, line 262: as you are calling two promise based function in a row (or nested), it could enhance readability to switch to an `async...await` call here, and wrap the whole thing in `try...catch` block

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -14,7 +14,7 @@
 export default {
     props: {
         value: {
-            type: String,
+            type: [String, null, undefined],
```
pakb, line 17: It would maybe be more elegant to say that `value` is not required, and it has a default value of `null`
```
{
   type: String,
   default: null,
}
```
And subsequently adapt the component (if need be) to manage a `null` value here

Or giving it a default value of `''`, so that if nothing (or `null` or `undefined`) is given, the prop has a valid default value

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -35,7 +35,7 @@ export default {
             this.buttonText = 'copy_cta'
         },
         async copyValue() {
-            await navigator.clipboard.writeText(this.value)
+            await navigator.clipboard.writeText(this.value || '')
```
pakb, line 38: if default value as described above is implemented, this `||` statement would be obsolete

pakb (approved):
Some small improvements, but overall very good 🚀

src/modules/map/components/LocationPopupCopySlot.vue
ltshb: Good idea 👍🏼

src/modules/map/components/LocationPopup.vue
ltshb: The optional access did returned `undefined` which caused a warning in the `LocationPopupCopySlot` but changing this component to an optional value solved this issue.

src/modules/map/components/LocationPopup.vue
ltshb: Done

PR #310, opened 2022-11-28
BGDIINF_SB-2662: Various bugfixes on position popup

ltshb:
Various bugfixes and improvements, see commits message for more details.

- Added border to position css triangle, needed on light background to add more contrast
- Fixed some error handling
- Do not display qrcode on mobile display
- Fixed long What3Words which pushed the copy button on a new line (see printscreen below)
  image
  Now the Whats3word is truncated with ellipsis, but you can still copy the whole address with the copy button. The position label space has been also reduced to let more space for the values. This reduced the risk of what3words truncate which now happens quite rarely.
  Here below is an example of a truncate what3words
  image

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2662-position-popup/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -78,7 +78,7 @@
             </div>
             <div v-else>-</div>
             <div class="location-popup-link lp-label">
-                {{ $t('link_bowl_crosshair') }}
+                {{ $t('share_link') }}
             </div>
             <div class="location-popup-link lp-data">
                 <LocationPopupCopyInput
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -325,7 +325,7 @@ export default {
         margin-bottom: 0.1rem;
     }
 }
-@media (min-height: 650px) {
+@media (min-height: 850px) {
     .location-popup-qrcode {
         display: block;
     }
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -270,6 +270,7 @@ export default {
 <style lang="scss" scoped>
 @import 'src/scss/webmapviewer-bootstrap-theme';
 .location-popup {
+    // Triangle border
     &::before {
         $arrow-height: 15px;
         position: absolute;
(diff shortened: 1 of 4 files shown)


src/modules/map/components/LocationPopup.vue
```diff
@@ -270,21 +270,47 @@ export default {
...
+    // Triangle border
     &::before {
         $arrow-height: 15px;
         position: absolute;
         top: -($arrow-height * 2);
         left: 50%;
         margin-left: -$arrow-height;
         border: $arrow-height solid transparent;
-        border-bottom-color: $light;
+        border-bottom-color: var(--bs-border-color-translucent);
```
pakb, line 281: as bootstrap is imported (in SCSS), you could (I think) directly use SCSS variable here.
So something like : `border-bottom-color: $border-color-translucent` (see https://getbootstrap.com/docs/5.2/utilities/borders/#css)

src/modules/map/components/LocationPopupCopyInput.vue
```diff
@@ -63,11 +64,14 @@ export default {
...
-            // Change button text and start the reset timer.
-            this.buttonText = 'copy_success'
-            this.resetTimeout = setTimeout(this.resetButton, this.resetDelay)
+            try {
+                await navigator.clipboard.writeText(this.value)
+                // Change button text and start the reset timer.
+                this.buttonText = 'copy_success'
+                this.resetTimeout = setTimeout(this.resetButton, this.resetDelay)
+            } catch (error) {
+                log.error(`Failed to copy to clipboard: ${error}`)
```
pakb, line 73: you could even pass the error as a secondary argument, in case this error is not a string but an object
```
log.error('Failed to copy to clipboard', error)
```
Otherwise it can sometimes flatten the object and output `Object: object` in the console, instead of flattening it into a string

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -35,11 +37,15 @@ export default {
...
+                await navigator.clipboard.writeText(this.value)
 
-            // Change button text and start the reset timer.
-            this.buttonText = 'copy_done'
-            this.resetTimeout = setTimeout(this.resetButton, this.resetDelay)
+                // Change button text and start the reset timer.
+                this.buttonText = 'copy_done'
+                this.resetTimeout = setTimeout(this.resetButton, this.resetDelay)
+            } catch (error) {
+                log.error(`Failed to copy to clipboard: ${error}`)
```
pakb, line 47: Same remark as above with error potentially being an object

pakb (approved):
Minor issues with logging, but not big of a deal

src/modules/map/components/LocationPopup.vue
ltshb: Done

src/modules/map/components/LocationPopupCopyInput.vue
ltshb: Done

PR #311, opened 2022-11-28
Use copy icon for coordinates in position popup

ltshb:
This allow more place for the data and avoid truncating the what3words.

Please give your feedback on the copy icon in the position popup, use the test link below and compare it to the current version in https://sys-map.dev.bgdi.ch

[Test link](https://sys-map.dev.bgdi.ch/preview/copy-icon/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopupCopySlot.vue
+++ b/src/modules/map/components/LocationPopupCopySlot.vue
@@ -43,8 +43,8 @@ export default {
         })
     },
     unmounted() {
-        this.copyTooltip?.destroy()
-        this.copiedTooltip?.destroy()
+        this.copyTooltip?.forEach((tooltip) => tooltip.destroy())
+        this.copiedTooltip?.forEach((tooltip) => tooltip.destroy())
     },
     methods: {
         async copyValue() {
--- a/src/modules/map/components/LocationPopupCopySlot.vue
+++ b/src/modules/map/components/LocationPopupCopySlot.vue
@@ -1,17 +1,19 @@
 <template>
-    <!-- eslint-disable vue/no-v-html-->
     <button
+        id="copyButton"
         ref="button"
         class="btn btn-light btn-sm"
         type="button"
         @click="copyValue"
-        v-html="$t(buttonText)"
-    />
-    <!-- eslint-enable vue/no-v-html-->
+    >
+        <FontAwesomeIcon class="icon" :icon="['far', 'copy']" />
+    </button>
 </template>
 
 <script>
 import log from '@/utils/logging'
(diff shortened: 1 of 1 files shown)


src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -1,48 +1,51 @@
...
+            content: this.$i18n.t('copy_done'),
+            arrow: true,
+            placement: 'right',
+            trigger: 'click',
+            onShow(instance) {
+                setTimeout(() => {
+                    instance.hide()
+                }, 1000)
+            },
+            allowHTML: true,
```
pakb, line 42: not sure if this flag is required, as it is only text that is being shown

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -1,48 +1,51 @@
...
     },
-    unmounted() {
-        clearTimeout(this.resetTimeout)
+    mounted() {
+        tippy('#copyButton', {
+            content: this.$i18n.t('copy_cta'),
+            arrow: true,
+            placement: 'right',
+            touch: 'hold',
+        })
```
pakb, line 31: I would maybe save the instances returned by both constructor, and then call https://atomiks.github.io/tippyjs/v6/methods/#destroy in the `unmount` lifecycle hook, so that we do not create unnecessary in-memory instances (could do a little bit of memory leakage there...)

pakb (commented):
I'm assuming this is based on https://github.com/geoadmin/web-mapviewer/pull/310 and only the last commit is to be reviewed? If so there it is 😉

src/modules/map/components/LocationPopupCopySlot.vue
ltshb: This is required because of non breaking space character in the `copied!` translation.

pakb (approved):

pakb:
I don't know if we have the right version of TippyJS for what I asked of you with `.destroy()`, whenever I close a location popup there is now an error in the console.

ltkum (approved):
Looks good to me, good work

boecklic (approved):
I like the copy icon. On desktop you have the popover that tells you the action, a mobile user is most likely used to that icon

hansmannj (approved):
+1 for the copy icon.
Probably a stupid question and probably a terrible design faux pas.
Might it make sense to shift the icon slightly to the left, directly following the coordinates?
With the distance in between it's a bit harder to see, which icon belongs to which coordinates.

But probably that looks ugly, and probably I'll don't think this is necessary once I got my new glasses :wink:

PR #312, opened 2022-11-28
BGDIINF_SB-2661: Use shortlink to share kml drawings

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2661-shortlink-kml/index.html)

### DIFF
--- a/src/modules/drawing/components/SharePopup.vue
+++ b/src/modules/drawing/components/SharePopup.vue
@@ -2,7 +2,7 @@
     <div class="ga-share">
         <div class="form-group">
             <label>{{ $t('draw_share_user_link') }}:</label>
-            <div class="input-group input-group-sm mb-3">
+            <div class="input-group input-group mb-3">
                 <input
                     type="text"
                     class="form-control"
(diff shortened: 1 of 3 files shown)


src/modules/drawing/components/SharePopup.vue
```diff
@@ -80,28 +80,63 @@ export default {
...
-                await navigator.clipboard.writeText(this.fileUrl)
+        async copyShareUrl() {
+            try {
+                await navigator.clipboard.writeText(this.shareUrl)
                 this.fileUrlCopied = true
                 this.fileTimeout = setTimeout(() => {
                     this.fileUrlCopied = false
                 }, 5000)
+            } catch (error) {
+                log.error(`Failed to copy: ${error}`)
```
pakb, line 110: same remark as the other PR : it might be better passing error as a secondary param instead of forcing it into a String here

src/modules/drawing/components/SharePopup.vue
```diff
@@ -80,28 +80,63 @@ export default {
...
+        },
+        async copyAdminShareUrl() {
+            try {
+                await navigator.clipboard.writeText(this.adminShareUrl)
+                this.adminUrlCopied = true
+                this.adminTimeout = setTimeout(() => {
+                    this.adminUrlCopied = false
+                }, 5000)
+            } catch (error) {
+                log.error(`Failed to copy: ${error}`)
```
pakb, line 121: 😉

src/modules/drawing/components/SharePopup.vue
```diff
@@ -112,5 +147,9 @@ export default {
 .ga-share {
     width: 100%;
     text-align: start;
+
+    .input-group {
+        width: 30rem;
+    }
```
pakb, line 153: as this is a Bootstrap class, I would maybe prefer that we add one custom class to the element you want to style this way, so that we do not overwrite Bootstrap's styling inadvertently (right now the style tag is scoped, so no big deal, but if someone were to remove the scoped flag, all Bootstrap `input-group` would be modified by this line)

pakb (approved):
Good, fallback to non shorten like, I like it!
Some minor adjustments and we are good to go

src/modules/drawing/components/SharePopup.vue
ltshb: 🙈 🙊 🙉 😆

src/modules/drawing/components/SharePopup.vue
ltshb: Good point.

PR #313, opened 2022-11-29
BGDIINF_SB-2683: Fixed GeoJSON layers

ltshb:
Some GeoJSON layers were not displayed correctly. Styling with a geometry type
of `point` where not displayed due to a wrong string comparison (`'point' === 'Point'`)

Also sometimes we could have a crash due to a non existing layer object which
was due to the fact that when hovering a search result a layer is created and
asynchronously the layer style was loaded, meanwhile when the hover is gone
the layer is destroyed which can happen before the layer style load from backend.

Updated some missing event in emit list to prevents vue warning.

- [x] Click on geojson feature don't open popup !

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2683-geojson/index.html)

### DIFF
--- a/src/api/features.api.js
+++ b/src/api/features.api.js
@@ -44,9 +44,9 @@ export class SelectableFeature extends EventEmitter {
         super()
         this._id = id
         // using the setter for coordinate (see below)
-        this._coordinates = coordinates
-        this._title = title
-        this._description = description
+        this.coordinates = coordinates
+        this.title = title
+        this.description = description
         this._isEditable = !!isEditable
         this._isDragged = false
     }
@@ -97,6 +97,10 @@ export class SelectableFeature extends EventEmitter {
                 this._coordinates = newCoordinates
             }
             this.emitChangeEvent('coordinates')
+        } else if (newCoordinates === null) {
+            this._coordinates = newCoordinates
+        } else {
+            log.error(`feature.api new coordinates is not an array`, newCoordinates)
         }
     }
     get lastCoordinate() {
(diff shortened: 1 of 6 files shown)


src/api/features.api.js
```diff
@@ -44,9 +44,9 @@ export class SelectableFeature extends EventEmitter {
         super()
         this._id = id
         // using the setter for coordinate (see below)
-        this._coordinates = coordinates
-        this._title = title
-        this._description = description
+        this.coordinates = coordinates
+        this.title = title
+        this.description = description
```
pakb, line 49: Why the change? Those properties are exposed as `instance.coordinates|title|description` through the fact that there is a getter set for them.
If you want to remove the underscored part, I don't mind, but you have to make sure any internal use below doesn't use this underscored name anymore

pakb (commented):
Not sure about this refactoring working as intended

src/api/features.api.js
ltshb: The changes is needed due to the setter, otherwise the setter is not called. For the coordinates the setter make sure to encapsulate single coordinate in an array, which was the bug because before when passing a single coordinate not in a array it was not correct and a geojson object selection was not marked as selected. I did not fully understand why we need to encapsulate single coordinate in an array though. Would be great to simplify the code by removing this setter and make sure to always path coordinate as follow: `[[x, y], ...]`. But that would require probably much more rework and effort.

src/api/features.api.js
```diff
@@ -97,6 +97,10 @@ export class SelectableFeature extends EventEmitter {
                 this._coordinates = newCoordinates
             }
             this.emitChangeEvent('coordinates')
+        } else if (newCoordinates === null) {
+            this._coordinates = newCoordinates
```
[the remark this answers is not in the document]
ltshb: I did not renamed it in constructor, I use the setter, `this.coordinates = coordinate` call the setter if any.

pakb (approved):

PR #314, opened 2022-11-29
BGDIINF_SB-2695: Fixed modal black drop

ltshb:
The modal view with the black drop did not took the whole screen but only the
size of the modal window. Also improved padding and margin for modal window

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2695-black-backdrop/index.html)

### DIFF
--- a/src/modules/drawing/components/SharePopup.vue
+++ b/src/modules/drawing/components/SharePopup.vue
@@ -144,12 +144,18 @@ export default {
 </script>
 
 <style lang="scss" scoped>
+@import 'src/scss/media-query.mixin';
 .ga-share {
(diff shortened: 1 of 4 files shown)


pakb (approved):
something is weird with test (maybe) unrelated with your change in drawing/sharing

ltshb:
@pakb The test were brocken due to a display bug on mobile device for the share modal view. I fixed it with the last commit. I check all the following views look nice on mobile and desktop version (only checked with chrome though)
- Share drawing
- Delete drawing
- Change topic
- Layer legend

PR #316, opened 2022-11-30
BGDIINF_SB-2708: Fixed map popover printing.

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2708-html-print/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersPopover.vue
+++ b/src/modules/map/components/openlayers/OpenLayersPopover.vue
@@ -19,7 +19,7 @@
                     @click="onClose"
                 />
             </div>
-            <div ref="mapPopoverContent" class="card-body">
+            <div id="mapPopoverContent" ref="mapPopoverContent" class="card-body">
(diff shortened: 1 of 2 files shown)


src/modules/map/components/openlayers/OpenLayersPopover.vue
```diff
@@ -79,7 +79,7 @@ export default {
             this.$emit('close')
         },
         printContent() {
-            promptUserToPrintHtmlContent(this.$refs.mapPopoverContent.outerHTML)
+            promptUserToPrintHtmlContent('mapPopoverContent')
```
pakb, line 82: I would maybe also change https://github.com/geoadmin/web-mapviewer/blob/develop/src/utils/print.js
at least update the JSDoc so that it is clear we can also give a HTML ID here (and not necessarily a HTMLElement as stated in the doc)

pakb (approved):
good catch

PR #317, opened 2022-12-01
Fixed elevation profile crash on windows resize

ltshb:
When resizing the windows or on mobile when switching between portrait
and landscape, the method updateElevationProfilePlot() crashed because it is
executed within a promise and due to timing the refs.profilePlot might be null.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-elevation-profile-crash/index.html)


pakb (approved):
I hope my refactoring hasn't brought up more bugs, at least it's a bit clearer when we need to fix something now 😜

PR #321, opened 2022-12-07
Fix_BGDIINF_SB-2682 no more doubling of wms layers on other side of globe

hansmannj:
Prevention of doubling of a WMS layer at the opposite site of the globe:

In order to prevent a WMS layer from appearing twice, once in CH and once close to New Zealand, pretty much the same fix as for WMTS was applied.
See: https://github.com/geoadmin/web-mapviewer/commit/c689f9a650c546c6e52a91fc2086d7cbbf48faa2

[Test link](https://sys-map.dev.bgdi.ch/preview/fix_bgdiinf_sb-2682_no_more_doubling_of_wms_layers_on_other_side_of_globe/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
+++ b/src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
@@ -8,9 +8,11 @@
 import { TILEGRID_EXTENT, TILEGRID_ORIGIN, TILEGRID_RESOLUTIONS, WMS_TILE_SIZE } from '@/config'
 import { CoordinateSystems } from '@/utils/coordinateUtils'
 import { Image as ImageLayer, Tile as TileLayer } from 'ol/layer'
+import { transformExtent } from 'ol/proj'
 import ImageWMS from 'ol/source/ImageWMS'
 import TileWMS from 'ol/source/TileWMS'
 import TileGrid from 'ol/tilegrid/TileGrid'
+import proj4 from 'proj4'
 import addLayerToMapMixin from './utils/addLayerToMap-mixins'
 
 /** Renders a WMS layer on the map */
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -51,35 +53,75 @@ export default {
...
+                    gutter: this.gutter,
+                    tileGrid: tileGridLV95,
+                })
+            } else {
+                source = new ImageWMS({
+                    projection: this.projection,
+                    url: this.url,
+                    tileGrid: tileGridLV95,
+                })
+            }
```
pakb, line 91: This looks really the same as line 57 to 68, would it be possible to use some setter for the TileGrid instead of recreating the same instances with this new option?

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -51,35 +53,63 @@ export default {
...
                 origin: TILEGRID_ORIGIN,
                 tileSize: WMS_TILE_SIZE,
             })
+
+            source.tileGrid = tileGridLV95
+
+            // tile grid and  reprojection to WebMercator is done in analogy to WMTS to prevent
+            // that the layer appears twice, once in CH and once near New Zealand.
+            // see: https://github.com/geoadmin/web-mapviewer/commit/c689f9a650c546c6e52a91fc2086d7cbbf48faa2
+            if (this.gutter !== -1) {
```
hansmannj, line 83: I am also not 100% sure if it is correct to do the reprojection only for the TileLayer and not in case of ImageLayer.
But for ImageLayer this did not seem to be possible and probably not necessary, if I correctly understood this.

pakb: I think you are right on this, I checked OpenLayer's API documentation and ImageWMS class doesn't have a function called `setTileGridForProjection`. So if you do not filter for the gutter options you might end up with an error in the console (and no layer shown on the map...)

pakb (approved):
🚀

PR #326, opened 2022-12-19
BGDIINF_SB-2729: Added multiple KML edit and simplify KML drawing logic

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2729-multiple-kml/index.html)


pakb (approved):
very nice! thanks a lot

src/modules/drawing/DrawingModule.vue
```diff
@@ -147,37 +150,34 @@ export default {
...
         async show(show) {
-            /**
-             * Makes it possible to abort the toggle on error by setting
-             * "this.abortedToggleOverlay=true" before calling "this.toggleDrawingOverlay()", as
-             * calling "this.toggleDrawingOverlay()" will retrigger this function.
-             */
-            if (this.abortedToggleOverlay) {
-                this.abortedToggleOverlay = false
-                return
-            }
```
jedef, line 167: Removing this protection while retrying the saving automatically when the prior saving failed (line 316) is very dangerous. Because now if you loose the internet connection and quit the drawing, not only will you loose your changes, but in the background the drawing module will still continuousely try to save the drawing. So when the internet connection comes again, as the drawing layer is empty when not in drawing mode, it will overwrite your kml drawing with a blank one.
So I would either improve the saving retry policy so that it stops to try to save the drawing when quitting the drawing mode, or readd the above protection, or both. Or/and clear the kml metadata when quitting drawing mode to prevent any accidental overwriting of the kml.

jedef (commented):
Hello
I could not finish the review today, as there are many changes. But I will still publish the comments that I have already added.

I have seen that you have not added any comments to explain that the savingStatus is for indicative uses only and is not always correct (see the comment in https://github.com/geoadmin/web-mapviewer/pull/325). Do you think that it is not necessary?

src/modules/drawing/DrawingModule.vue
ltshb: I removed this protection as it did not work correctly. It simply prevented to close the menu without retry of the failed saved and during my test (I changed the url of the backend to force a failed saving) the error message was not displayed. So that was not very user friendly as you could not do the action that you want without any indication why. I did not debug to see why the saving status of error was not displayed though. 
I think we should not try to make decision for the user, this will not be understood. Automatic retry on auto save is the best IMHO.
However you brought a very good point with the auto retry continuing when the drawing menu is closed ! I'll fix this, see #331

src/modules/drawing/DrawingModule.vue
```diff
@@ -316,19 +308,21 @@ export default {
...
                 try {
                     await this.saveDrawing(kml)
+                    this.savingStatus = SavingStatus.SAVED
                 } catch (e) {
                     log.error('Could not save KML layer: ', e)
                     this.savingStatus = SavingStatus.SAVE_ERROR
-                    throw e
-                }
-                if (this.savingStatus !== SavingStatus.UNSAVED_CHANGES) {
-                    this.savingStatus = SavingStatus.SAVED
```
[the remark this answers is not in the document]
ltshb: Good point I modify this in #331

PR #331, opened 2022-12-22
BGDIINF_SB-2729: Fixed save KML drawing error handling - from post review of #326

ltshb:
Now if the loading of the KML failed in the drawing mode, we remove the
kml layer as it is not selectable and we print an error message. The loading
of the kml to translate it to a drawing will be automatically retried. In case
of error we also display an error message at the same place as for saving error.

Those errors message are also now displayed in red.

Also renamed the savingStatus to drawingState which refer more to a drawing state
to make the code clearer as it is also used to check if any modification on
the drawing has been done.

Improved the error handling and user experience in case of network failure.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2729-save-drawing/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -152,7 +152,14 @@ export default {
             return null
         },
         isDrawingModified() {
-            return this.drawingState !== DrawingState.INITIAL
+            switch (this.drawingState) {
+                case DrawingState.INITIAL:
+                case DrawingState.LOADED:
+                case DrawingState.LOAD_ERROR:
+                    return false
+                default:
+                    return true
+            }
         },
     },
     watch: {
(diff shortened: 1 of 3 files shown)


src/modules/drawing/DrawingModule.vue
```diff
@@ -363,39 +382,36 @@ export default {
...
             this.drawingLayer.getSource().clear()
             this.onChange()
         },
-        async addKmlLayerToDrawing(layer) {
-            await this.addKmlDrawingLayer(layer)
-            // Remove the layer to not have an overlap with the drawing from
-            // the drawing manager
-            this.removeLayer(layer)
+        differAddKmlLayerToDrawing(layer, time = 2000) {
+            this.addKmlLayedTimeout = setTimeout(this.addKmlLayerToDrawing, time, layer)
```
jedef, line 393: Probably you wanted to spell it "addKmlLayerTimeout"

jedef (approved):
Apart from a small typo, its all good for me. Nice work.

PR #333, opened 2023-01-02
BGDIINF_SB-2738: Fixed location popup in drawing mode

ltshb:
If a location popup was displayed and then we were entering the drawing mode,
the location popup was not cleared and stayed in drawing mode.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2738-location-popup/index.html)


jedef (approved):
It's all good for me.
P.s. While reviewing I saw that on src/store/plugins/click-on-map-management.plugin.js Line 72 and 73 there are two comments that are not correct anymore, I probably forgot to delete them. Maybe you can do that as part of this pull request?

PR #334, opened 2023-01-03
Avoid Vue/Intlify Warnings and fixed missing translations

ltshb:
- Make sure to not call translation on empty strings.
- Use the translation for the icon sets name, due to this the max size of the
popover has been increased
- Avoid trying to translate icon size twice as well as trying to translate non
translable titles in DropdownButton like the KML|GPX dropdown items. For this
we let the user of DropdownButton the choice to translate or not the titles.

- [x] TODO add translation for `very_small_size`. @pakb I've seen that the old viewer don't have the very small size for the icon but just the small one, it is on purpose to have a new size in the new viewer ?

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-vue-warn/index.html)


src/modules/drawing/components/DrawingExporter.vue
```diff
@@ -31,9 +31,10 @@ export default {
         },
     },
     data() {
+        const options = [new DropdownItem('KML'), new DropdownItem('GPX')]
         return {
-            exportSelection: 'KML',
-            exportOptions: [new DropdownItem('KML'), new DropdownItem('GPX')],
```
jedef, line 36: "exportSelection" expects a String like 'KML' or 'GPX' and not a DropdownItem. I would initialize "exportSelection" with null (line 36).

Also I would not pass the "exportSelection" as "current-value" to the DropdownButton (see Line 5), as selecting a download option only triggers a download so it makes no sense for the button to be permanently highlighted after that.

src/utils/DropdownButton.vue
```diff
@@ -37,7 +37,7 @@
                     :data-cy="`dropdown-item-${item.title.toLowerCase()}`"
                     @click="selectItem(item)"
                 >
-                    {{ $t(item.title) }}
```
jedef, line 40: The documentation on line 55-57 should be fixed to reflect that change.

jedef (commented):
It seems that you have renamed "set-second.fixture.json" to "set-babs.fixture.json", but you have not replaced the occurrences of "second" in this file with "babs".

jedef:
Concerning "very-small-size": We added this size as we wanted all the sizes of the old viewer to match sizes from the new viewer. But the small icon size of the old viewer was too small to be very usefull, so we created the following mapping for the icon sizes:
```
Old Viewer     New Viewer
small          very small
medium         small
large          medium              
```

src/modules/drawing/components/DrawingExporter.vue
ltshb: Good catch about the value type of `exportSelection`.

However I would keep it as current value, because I think the dropdown selection is IMHO the selected action that will be performed by the button if clicked. For example you only click on the dropdown arrow to see which action will be performed, but without effectively running the action.

ltshb: @jedef see last commits, I reverted this.

src/utils/DropdownButton.vue
ltshb: Done

jedef:
> It seems that you have renamed "set-second.fixture.json" to "set-babs.fixture.json", but you have not replaced the occurrences of "second" in this file with "babs".
@ltshb Have you seen this comment? Is it ok to let it like this?

ltshb:
> > It seems that you have renamed "set-second.fixture.json" to "set-babs.fixture.json", but you have not replaced the occurrences of "second" in this file with "babs".
> > @ltshb Have you seen this comment? Is it ok to let it like this?

@jedef Oops while doing multitasking I forgot this, many thanks for the very thorough review. I fixed it in last commit.

jedef (approved):
Thanks, I think it's all good now!

pakb:
>  it is on purpose to have a new size in the new viewer ?

Yes, we thought it would be a good addition to have an extra options in between, and now that our backends facilitates size changes, it was easier to add than before (previously you had to create a new set of images for each size)

pakb (approved):

PR #335, opened 2023-01-04
Added JUNIT reporting

ltshb:
This helps finding quickly the issue on CI in case of failing tests.


jedef (approved):
I didn't totally understand how these vite plugins work (i.e. when and how the generate-build-info plugin is executed). But all other changes seem good to me.

PR #338, opened 2023-01-09
Menu shadow and corrected the red colors

ltshb:
Mainly we use 2 red colors, the primary (admin red) and the danger (for error
and dangerous actions).

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-look-and-feel/index.html)

### DIFF
--- a/src/modules/infobox/components/SwipableBottomSheet.vue
+++ b/src/modules/infobox/components/SwipableBottomSheet.vue
@@ -200,7 +200,7 @@ export default {
         width: 45px;
         height: 8px;
         border-radius: 14px;
-        background: $red;
+        background: $primary;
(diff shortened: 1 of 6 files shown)


pakb:
I don't know when we added that, but now there is a huge (and hugly) margin around the time selector tippy popup, we could fix that here too don't you think?
image

ltshb:
> I don't know when we added that, but now there is a huge (and hugly) margin around the time selector tippy popup, we could fix that here too don't you think? image

@pakb This has been fixed with the tippy PR #336 I'll rebase this PR on top of develop

pakb (approved):
yep, it disapeared with the rebase 👌

PR #339, opened 2023-01-10
BGDIINF_SB-2728: Fixed legacy drawing initial load

ltshb:
On initial load of legacy drawing, it was not visible. To make it visible we
had to toggle the visible flag.

This was a race condition between the icons backend and the kml backend. In
order to display legacy drawings we need to have the icons sets information
from the icons backend. In our case the kml was loaded and added to openlayer
before the icons backend answered.

Now we update the drawing features on changes from the icons backend.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2728-legacy-drawing/index.html)

### DIFF
--- a/src/api/features.api.js
+++ b/src/api/features.api.js
@@ -250,6 +250,8 @@ export class EditableFeature extends SelectableFeature {
      *
      * @param {Feature} olFeature An olFeature that was just deserialized with
      * @param {DrawingIconSet[]} availableIconSets {@link ol/format/KML}.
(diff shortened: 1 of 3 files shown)


pakb (approved):
🚀

jedef:
We should look for another way to fix this, as this partially breaks the geodesic geometries. As explained in a comment in "OpenLayersKMLLayer.vue", the features should be deserialized before they are added to the layer for the geodesic geometries to always show up correctly. Also, a feature should be deserialized only once, and with the modifications here, this is not anymore guaranteed.

PR #340, opened 2023-01-11
Update libs

pakb:
reverting git-branch to v1.0.0 as there is a vulnerability in one of the dependencies in v2

There are also some adaptation to migrate Cypress to v12 (no more `cy.server`, context is cleaned before each test now)

[Test link](https://sys-map.dev.bgdi.ch/preview/update-libs/index.html)


ltshb (approved):
👍🏼

PR #344, opened 2023-01-12
BGDIINF_SB-2733: Removed drawing adminId from url and open drawing mode automatically

ltshb:
Thee adminId is now removed from the URL and upon adminId detection, the drawing
menu is automatically opened.


src/api/layers/KMLLayer.class.js
```diff
@@ -53,11 +57,7 @@ export default class KMLLayer extends AbstractLayer {
 
     getID() {
         // format coming from https://github.com/geoadmin/web-mapviewer/blob/develop/adr/2021_03_16_url_param_structure.md
-        let id = `KML|${encodeURIComponent(this.kmlFileUrl)}|${this.name}`
-        if (this.adminId) {
-            id += `@adminId=${this.adminId}`
-        }
-        return id
+        return `KML|${this.kmlFileUrl}|${this.name}`
```
jedef, line 60: Just a short question for my comprehension. Is the admin ID not appearing anymore in the URL simply because you removed it from the ID here?

src/modules/drawing/DrawingModule.vue
```diff
@@ -418,6 +444,11 @@ export default {
                 }
             }
         },
+        async openDrawingOnAdminId(layer) {
+            if (layer?.adminId && !this.isDrawingOpen && !layer?.isLegacy()) {
+                await this.setShowDrawingOverlay(true)
+            }
+        },
```
jedef, line 451: There are a few times where this triggers incorrectly the drawing menu:
- When changing the language after that a drawing has been drawn. I think it is because internally the layers are removed and readded when changing the language.
- If you have multiple drawings and move them around in the active layers list, it will automatically open the drawing mode as soon as a new drawing with admin ID is at the top of the stack, which is unwanted.

Ideally the drawing mode should only be triggered someone pastes an URL with an adminid in the url bar.

src/api/layers/KMLLayer.class.js
```diff
@@ -70,10 +70,31 @@ export default class KMLLayer extends AbstractLayer {
...
+     * Clear the KML Admin ID
+     *
+     * Once the admin id has been cleared we cannot modify this KML anymore
+     */
+    clearAdminId() {
+        this.adminId = null
+        if (this.metadata) {
+            this.metadata.adminId = null
+        }
+    }
```
jedef, line 86: This function seems to be called nowhere.

src/api/layers/KMLLayer.class.js
```diff
@@ -1,28 +1,32 @@
...
-     *   (or equivalent in the current UI lang) will be defined
-     * @param {object | null} metadata Metadata of the KML drawing. This object contains all the
-     *   metadata returned by the backend.
-     * @param {boolean} isExternal Flag telling if this KML comes from our backend (false) or is
-     *   loaded from a different source (true)
+     * @param {string} kmlFileUrl The URL to access the KML data.
+     * @param {boolean} visible If the layer is visible on the map (or hidden).
+     * @param {number | null} [opacity=1.0] The opacity of this layer, between 0.0 (transparent) and
+     *   1.0 (opaque). When `null` is given, then it uses the default value which is `1.0`. Default
+     *   is `1.0`. Default is `1.0`
```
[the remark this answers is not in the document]
ltshb: Better 3 times that none :-P

src/api/layers/ExternalWMTSLayer.class.js
```diff
@@ -18,11 +18,11 @@ export default class ExternalWMTSLayer extends ExternalLayer {
     constructor(name, opacity, visible, getCapabilitiesUrl, externalLayerId, attributions) {
         super(
             // as we are encoding the name in getID(), we must decode it here when receiving it through the URL parser
-            decodeURIComponent(name),
+            name,
             LayerTypes.WMTS,
             externalLayerId,
             // same thing with the getCap URL, we encode it in getID, so it must be decoded
-            decodeURIComponent(getCapabilitiesUrl),
+            getCapabilitiesUrl,
```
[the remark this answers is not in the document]
ltshb: The decoding is totally removed as we let now vue do the encoding and decoding, we had a hack to prevent encoding of the layers query param and where handling the encoding/decoding in the layer classes but that causes troubles.

src/modules/drawing/DrawingModule.vue
```diff
@@ -164,69 +166,17 @@ export default {
...
-                    await this.addKmlLayerToDrawing(this.activeKmlLayer)
-                    // Remove the layer to not have an overlap with the drawing from
-                    // the drawing manager. We even remove this if the kml load failed as
-                    // this layer is not selectable.
-                    this.removeLayer(this.activeKmlLayer)
-                }
-                this.isDrawingEmpty = this.drawingLayer.getSource().getFeatures().length === 0
-                this.getMap().addLayer(this.drawingLayer)
+                this.isLoading = true
+                this.isDrawingOpen = true
```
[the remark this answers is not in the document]
ltshb: Because `show` trigger the closing of the drawing menu which is asynchrone due to the `saveDrawing()` and `addLayer()`. To avoid race condition at startup when we have a KML AdminID where we automatically open the drawing mode, I need to know that the mode is still open and not in a closing state.

src/api/layers/KMLLayer.class.js
ltshb: True this is a left over, a first implementation was to get rid of the admin id when leaving the menu.

src/store/modules/layers.store.js
```diff
@@ -181,9 +190,9 @@ const actions = {
                 // the KML layers, in this case we take the old active configuration as fallback.
                 // For KML layers the configuration doesn't requires translation as they only have
                 // the layer name to be translated and it is done dynamically using a getter name()
-                commit('addLayerWithConfig', layer.clone())
+                commit('addLayer', { layer: layer.clone() })
```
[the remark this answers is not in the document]
ltshb: done

src/api/layers/KMLLayer.class.js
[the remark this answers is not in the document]
ltshb: Done

src/modules/drawing/DrawingModule.vue
ltshb: Good findings I'll try to find a solutions for those use cases.

ltshb: @jedef this should be fixed.

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -138,20 +141,18 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
-            } else {
-                // if internal (or BOD) layer, we add it through its parsed config
-                log.debug(`  Add layer ${layer.id} to active layers`, layer)
-                promisesForAllDispatch.push(store.dispatch('addLayer', layer))
+            const layerObject = createLayerObject(parsedLayer)
+            if (layerObject.type == LayerTypes.KML && layerObject.adminId) {
+                promisesForAllDispatch.push(store.dispatch('setOpenOnAdminId', true))
             }
+            log.debug(`  Add layer ${parsedLayer.id} to active layers`, layerObject)
+            promisesForAllDispatch.push(store.dispatch('addLayer', layerObject))
```
jedef, line 155: Isn't there here a potential race condition? In the drawing module, you watch only for the active KML layer, and here you run "setOpenOnAdminId" and "addLyer" concurrently. So if the latter action finishes before the former, the drawing mode would not be opened I think.

src/store/modules/layers.store.js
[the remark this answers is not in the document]
jedef: I don't see the fix. See screenshot attached below. I have changed the language to french but the drawing layer name is still in german "Zeichnung". But this can also be fixed in another PR.
image

src/api/layers/KMLLayer.class.js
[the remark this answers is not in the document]
jedef: Not really important, but the same sentence is still there two times.

src/api/layers/KMLLayer.class.js
```diff
@@ -1,28 +1,32 @@
...
+     * @param {string | null} [fileId=null] The KML id (which is part of the kmlFileUrl). If null it
+     *   is parsed from kmlFileUrl. Default is `null`
+     * @param {string | null} [adminId=null] The admin id to allow editing. If null then the user is
+     *   not allowed to edit the file. Default is `null`
+     * @param {string | null} [name=nul] Name of this layer, if nothing is given a default name
+     *   "Drawing" (or equivalent in the current UI lang) will be defined. Default is `nul`
```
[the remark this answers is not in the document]
jedef: Here again, not really important but I don't see the fix.

src/modules/drawing/DrawingModule.vue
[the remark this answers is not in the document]
jedef: ok

jedef (commented):
seems good, but there's one place where I am not sure if there is a race condition.

src/store/modules/layers.store.js
```diff
@@ -65,6 +51,22 @@ const getters = {
...
+     *
+     * @returns {KMLLayer | null}
+     */
+    activeKmlLayer: (state) => {
+        const kmlLayers = state.activeLayers.filter((layer) => layer.type === LayerTypes.KML)
+        if (kmlLayers.length > 0) {
+            return kmlLayers[kmlLayers.length - 1]
+        }
+        return null
+    },
```
jedef, line 69: You should here probably use "visibleLayers.filter" instead of "activeLayers.filter", as for now you cannot unselect a (legacy) layer anymore to start a drawing.

src/store/modules/layers.store.js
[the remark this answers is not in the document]
ltshb: I changed the comment, I don't think this name should change by design, it is translated during the creation of the drawing using the current language at that time and then it is part of the URL. On the old viewer it is not re-translated if you don't have the adminId, only if you have adminId it is re-translated in case of language changes.
IMHO we should leave it for the moment as is, and we might in future improve it by allowing the user to define its own title and then it won't be anymore translated.

src/api/layers/KMLLayer.class.js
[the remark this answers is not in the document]
ltshb: Part of the sentence it automatically added by prettier... We should anyway in future have the formatting in an npm command for the moment is left to the IDE which is not the same for everyone and not configured the same everywhere...

src/router/storeSync/LayerParamConfig.class.js
ltshb: Good point. It works because for now both event occurs before the drawing module is created and that in created event of the module I call the `openDrawingOnAdminId` method. But for safety I'll watch also `setOpenOnAdminId` this would also cover the case where we would add the adminId manually in the url afterwards.

src/store/modules/layers.store.js
ltshb: I cannot use visible layers filter here, because the kml layer in drawing mode is on purpose left in the active layers with the invisible flag. This allow to keep the kml in the url and if the user press reload, then it keep its drawing (also it would lose its adminId but can easily continue working on his drawing with a new copy). To keep the kml in the url parameter without making it invisible but removing from layer stack when drawing, we would need another parameter. I'll check if I can achieve this easily without too much hacking.
Note if you have a legacy drawing, you can still remove it from the layers in order to do a new drawing. The legacy blocking is only done for the MVP and will be removed once we are live, so this is not an issue.
One drawback with the actual implementation is that when in drawing mode and press F5 the drawing is then in invisible mode instead of visible.

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -13,7 +13,7 @@ import log from '@/utils/logging'
  * layer and its state (visibility, opacity, etc...)
  *
  * @param {AbstractLayer} layer
- * @param {GeoAdminLayer[]} defaultLayerConfig
```
pakb, line 16: that might be the linter going nuts, or that we have different JSDocs parser you and I, but this is unwanted I assume 😉

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -43,51 +42,49 @@ export function transformLayerIntoUrlString(layer, defaultLayerConfig) {
...
         const splitLayerId = parsedLayer.id.split('|')
-        return new KMLLayer(
-            parsedLayer.opacity,
+        layer = new KMLLayer(
+            splitLayerId[1], // kml url
             parsedLayer.visible,
-            splitLayerId[1],
-            null,
+            parsedLayer.opacity,
+            null, // fileId, null := parsed from url
```
pakb, line 54: some python operator in the comments, now you've shown you true colours 🙈

src/store/modules/layers.store.js
```diff
@@ -163,27 +171,26 @@ const actions = {
         const activeLayerBeforeConfigChange = Array.from(state.activeLayers)
         commit('clearLayers')
         commit('setLayerConfig', config)
-        activeLayerBeforeConfigChange.forEach((layer) => {
+        for (const layer of activeLayerBeforeConfigChange) {
```
pakb, line 174: I'm not too big of a fan to use `for` loops when we have the `.forEach` function with arrays, but I don't mind.
It doesn't help the readability IMHO, did you find some improvement in performance with this notation?

tests/e2e-cypress/integration/drawing/kml.cy.js
```diff
@@ -24,6 +25,38 @@ describe('Drawing new KML', () => {
...
+        )
+        cy.reload(true)
+        cy.waitUntilState((state) => {
+            return state.layers.activeLayers.length > 0
+        })
+        cy.openDrawingMode()
+        cy.wait('@get-kml')
+        cy.readWindowValue('drawingLayer')
+            .then((layer) => layer.getSource().getFeatures())
+            .should('have.length', 1)
```
pakb, line 58: I'm not sure if this test thoroughly that you have a copy in your hands at this point.
Wouldn't it be a good thing to add another feature to the drawing, and check that the POST request is fired (instead of a PUT)?

pakb (approved):
discussing very minor changes
otherwise looks good!

src/router/storeSync/LayerParamConfig.class.js
ltshb: Nope based on jsdoc https://jsdoc.app/tags-param.html optional parameter are defined like this
```
/**
 * @param {string} [somebody] - Somebody's name.
 */
function sayHello(somebody) {
    if (!somebody) {
        somebody = 'John Doe';
    }
    alert('Hello ' + somebody);
}
```

src/store/modules/layers.store.js
ltshb: @pakb It's a leftover of a previous implementation where I updated the metadata hier and needed to break the loop in case of error, but I reverted that part, I can also revert to use the forEach. The only advantage of for over forEach is that you can break the loop in the first one if needed (although you can use `every()` for that purpose).

src/router/storeSync/LayerParamConfig.class.js
ltshb: @pakb nope it is not python `:=` doesn't exist neither in python. But its a mathematical notation for `defined to/as`  meaning that `null` is defined as parsed url.

src/api/layers/KMLLayer.class.js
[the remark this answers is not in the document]
ltshb: Done

src/router/storeSync/LayerParamConfig.class.js
ltshb: Done

src/store/modules/layers.store.js
ltshb: We also need to keep this like this to keep the same behavior as in old viewer. In older viewer, as long as you don't manually change the URL, you always have only one kml drawing in the layer list even if you deselect the kml layer (making is invisible) and you go in the drawing menu it will overwrite/copy the de-selected layer and make it visible again when leaving the drawing.
With my last commit I ensure to have only one drawing in the layer list.

src/modules/drawing/DrawingModule.vue
```diff
@@ -319,14 +329,32 @@ export default {
             clearTimeout(this.differSaveDrawingTimeout)
             const kml = generateKmlString(this.drawingLayer.getSource().getFeatures())
             try {
-                let metadata
-                if (!this.kmlMetadata || !this.kmlMetadata.adminId) {
+                if (!this.kmlLayer?.adminId) {
```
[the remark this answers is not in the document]
ltshb: indeed it was already a computed, I'll reuse the computed here.

src/store/modules/layers.store.js
ltshb: Done

src/store/modules/drawing.store.js
```diff
@@ -67,8 +71,23 @@ export default {
...
-        setActiveKmlLayer({ commit }, activeKmlLayer) {
-            commit('setActiveKmlLayer', activeKmlLayer)
+        setOpenOnAdminId({ commit }, value) {
+            commit('setOpenOnAdminId', value)
+        },
+        setKmlLayerAddToMap({ commit, rootGetters }, payload) {
+            if ('addToMap' in payload && 'layerId' in payload) {
+                const layer = rootGetters.getActiveLayerById(payload.layerId)?.clone()
+                if (layer) {
+                    layer.addToMap = payload.addToMap
```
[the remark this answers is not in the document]
ltshb: @pakb The store is not changed here, the layer is cloned and the clone is changed and pass to addLayer() where the store mutation happens in a mutation.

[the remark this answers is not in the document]
ltshb: Note we have the same code in the layers store actions, where the store value is cloned in order to make changes in the action which then trigger a mutation with the new value.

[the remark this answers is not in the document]
pakb: oh right, didn't catch the `.clone()` up there, so all is good

pakb (approved):

PR #346, opened 2023-01-12
BGDIINF_SB-2744: Switch VectorTile background layer usage

pakb:
Reverts default ECH BG layer to given value (from topics) and add light base map behind `pixelkarte-farbe`, instead of switching to light base map by default.
Light base map is now added as an option in the BG selection wheel, and if chosen will be rendered in full (no exclusion over Switzerland).
The exclusion of data over Switzerland with LightBaseMap still occurs if it is used behind `pixelkarte-farbe`

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2744_switch_vt_layer_usage/index.html)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -69,6 +69,7 @@
                 "mime-types": "^2.1.35",
                 "prettier": "^2.8.2",
                 "prettier-plugin-jsdoc": "^0.4.2",
+                "rimraf": "^3.0.2",
                 "sass": "^1.57.1",
                 "start-server-and-test": "^1.15.2",
                 "typescript": "^4.9.4",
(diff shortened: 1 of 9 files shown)


ltshb (approved):
From a code point of vue all goods.

However from a UI point of vue I think it doesn't make sense at all to add the world light base map to the pixel color map. I think from a user use case we would have the following:
1. User wants the whole world => use only base light map
2. User wants very detailed map of switzerland => use pixel color map (no need for the rest of the world)
3. User wants very detailed map of the whole word => not feasible as we don't have detailed map for the rest of the world

As @pakb already mentioned having two types of map displayed simultaneously looks very ugly and that users will notice it. @davidoesch what is the reason to add the world base light map to the pixel color map ? Which use case are you trying to cover ?

davidoesch:
The reason we have the pixel map as default on vectormap:
- user has outside Switzerland map coverage
- world coverage on all layers

- once there is a vector Landeskarte: it can be added to replace ok


Please proceed.
The setup has been evaluated in the Testphase: and user feedback was: Switzerland PK , outside world 

So: go on and implement it

PR #348, opened 2023-01-17
BGDIINF_SB-2750: Encode `|` in url

ltshb:
See https://jira.swisstopo.ch/browse/BGDIINF_SB-2750

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2750-url-encoding-2/index.html)

### DIFF
--- a/src/modules/drawing/components/SharePopup.vue
+++ b/src/modules/drawing/components/SharePopup.vue
@@ -47,6 +47,7 @@
 
 <script>
 import { createShortLink } from '@/api/shortlink.api'
+import { stringifyQuery } from '@/utils/url'
 import log from '@/utils/logging'
 
 export default {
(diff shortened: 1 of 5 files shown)


davidoesch (commented):
since this is backward compatible for the existing permalinks... LGTM

jedef:
@ltshb It seems that most or all of these changes are already present in https://github.com/geoadmin/web-mapviewer/pull/344 so I have already reviewed them. If there are changes that are not present in the beforementioned pull request, please rebase it in such a way that I only see the new things.

pakb (approved):
Don't know if it is not too cumbersome to achieve, but if you could grab the part of prior commits from https://github.com/geoadmin/web-mapviewer/pull/344 that removes `src/router/stringifyQuery.js` and merge it here so that we have proper refactor action, that would be neat (but that is not mandatory if too complicated)

pakb (approved):

PR #350, opened 2023-01-18
BGDIINF_SB-2751: Moved the language toolbar in desktop mode

ltshb:
Now in desktop mode the language toolbar is on top right corner using button
link.

Clean up some old css and made the header more reactive using bootstrap instead
of custom css rules.

Also improved search bar look and feel.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2751-language-menu/index.html)

### DIFF
--- a/src/modules/i18n/components/LangButton.vue
+++ b/src/modules/i18n/components/LangButton.vue
@@ -1,0 +1,93 @@
+<template>
+    <button
+        class="language-btn btn"
+        :class="buttonClass"
+        :title="lang.toUpperCase()"
+        :data-cy="`menu-lang-${lang}`"
+        @click="changeLang"
+    >
+        {{ lang.toUpperCase() }}
+    </button>
+</template>
+
+<script>
+import log from '@/utils/logging'
+import { mapActions, mapGetters, mapState } from 'vuex'
+
+export default {
+    props: {
+        lang: {
+            type: String,
+            required: true,
+        },
+    },
+    data() {
+        return {
+            selected: false,
+        }
+    },
+    computed: {
+        ...mapGetters(['isDesktopMode']),
+        ...mapState({
+            currentLang: (state) => state.i18n.lang,
+        }),
+        buttonClass() {
+            let classes = ''
+            if (this.isDesktopMode) {
+                classes += 'm-0 px-1 btn-xs btn-link custom-text-decoration'
+                if (this.selected) {
+                    classes += ' text-primary'
+                } else {
+                    classes += ' text-black'
+                }
+            } else {
+                classes += ' mobile-view btn-sm mx-1'
+                if (this.selected) {
+                    classes += ' btn-primary'
+                } else {
+                    classes += ' btn-light'
+                }
+            }
+            return classes
+        },
+    },
+    watch: {
+        currentLang(newLang) {
+            this.updateSelection(newLang)
+        },
+    },
+    mounted() {
+        this.updateSelection(this.currentLang)
+    },
+    methods: {
+        ...mapActions(['setLang']),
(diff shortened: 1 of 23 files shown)


davidoesch:
- [ ] order  is DE FR IT EN RM 
- [ ] CSS _ remove underline for links

davidoesch (commented):
order is DE FR IT EN RM
 CSS _ remove underline for links

jedef:
I haven't looked at the code for now but is seems strange the the length of the search bar changes depending on the language selected.

davidoesch:
LGTM

davidoesch (commented):
LGTM
I

ltshb:
@jedef good finding and it is only on desktop mode, mobile still works, I'll look into it.

pakb:
for the middle layout (tablets, iPad and such) I would maybe also move the lang selector in the menu. That looks like very tiny element to be able to touch them accurately with some big clumsy fingers ✋ 
image

davidoesch:
Just use the same settings as in mfgoadmin3
Mind the height settings --they seem to be different currently on dev

src/modules/i18n/components/LangButton.vue
```diff
@@ -0,0 +1,93 @@
...
+            } else {
+                classes += ' mobile-view btn-sm mx-1'
+                if (this.selected) {
+                    classes += ' btn-primary'
+                } else {
+                    classes += ' btn-light'
+                }
+            }
+            return classes
+        },
```
pakb, line 52: you can make an array of class if you want, or you could directly state classes in the template with 
```
<button
        class="language-btn btn"
        :class="{
            "m-0 px-1 btn-xs btn-link custom-text-decoration text-black": isDesktopMode,
            "text-primary": isDesktopMode && selected,
            "mobile-view btn-sm mx-1 btn-light": !isDesktopMode,
            "btn-primary": !isDesktopMode && selected
        }"
</button
```

If possible I tend to prefer this way of expressing classes, so that their definition stays in the HTML and is not mixed with the JS.

I'm not 100% sure about how `btn-primary` or `text-primary` will be interpreted with how I've written the class props above, maybe you will need to declare `btn-light` and `text-black` on a separate entry (they might not be overridden by the primary classes...)

src/modules/i18n/components/LangSwitchToolbar.vue
```diff
@@ -1,58 +1,40 @@
...
-            @click="changeLang(lang)"
-        >
-            {{ lang.toUpperCase() }}
-        </button>
+    <div
+        :class="{
+            'lang-switch-menu': !isDesktopMode,
+            'p-1': !isDesktopMode,
+            'lang-switch-header': isDesktopMode,
+            'me-2': isDesktopMode,
```
pakb, line 7: you can combine many classes into one statement
```
:class="{
            'lang-switch-menu p-1': !isDesktopMode,
            'lang-switch-header me-2': isDesktopMode,
```

pakb (approved):
Some minor improvements in Vue template/component
Looking good! 👍

src/modules/i18n/components/LangButton.vue
ltshb: 👍🏼

ltshb: Good idea I also did not like the way I did it but did not thought of this way above 😉

src/modules/i18n/components/LangSwitchToolbar.vue
ltshb: 👍🏼

pakb (approved):

PR #352, opened 2023-01-19
Avoid crash in openlayers in case of load kml failure

ltshb:
The crash could occure if the load kml failed or if the load kml results
arrive once the kml layer has been removed from the map (drawing mode open).

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-kml-crash/index.html)


jedef (approved):
👍

PR #353, opened 2023-01-19
Add icons API intercept in cypress

ltshb:
The icons API is called also outside of the drawing module, so make the intercept
global on navigating the page. This avoid calling the backend.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-cypress-icons/index.html)


jedef (approved):
Seems good to me.

pakb (approved):

PR #354, opened 2023-01-23
BGDIINF_SB-2618: Set the background selector button to swiss-image and fixed selected bug

ltshb:
Always uses the swiss-image background as selector button, this gives the best
contrast.

Also fixed the selected background selection feedback when void is selected,
in this case the void background button was not highlighted in red.


pakb (approved):
👍

pakb:
Background wheel should have the same image inside wether the background has changed or not, phone or desktop.
So this fix is the way to go

PR #357, opened 2023-01-25
BGDIINF_SB-2182: Fixed E2E tests issue with `click({force: true})`

ltshb:
Sometimes for an unknown reason we have to set `{force: true}` on click() in
cypress otherwise cypress return an errors saying that the component is not
visible also it is visible.

Now we make sure before the click({force: true}) that the component is visible.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2182-e2e/index.html)


pakb (approved):
Nice! Putting test back into the loop in the process 👍

PR #361, opened 2023-01-26
BGDIINF_SB-2674 : change Give Feedback button color to red

pakb:
So that it catches more the attention compared to the give information link.

Also refactoring a bit the way we define links for the header, as we do many time the same look and feel there's now a dedicated component for such links.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2674_feedback_button_in_red/index.html)


src/modules/menu/components/header/HeaderLink.vue
```diff
@@ -0,0 +1,44 @@
+<template>
+    <button
+        class="btn btn-xs btn-link m-0 px-1 custom-text-decoration"
+        :class="{
+            'text-black': !selected,
+            'text-primary': selected,
+        }"
```
ltshb, line 7: I would leave the logic of selection outside of this component, this is specific to language. It is also not intuitive to set `selected=true` for the feedback button just to have it in the primary color. So this class should be by default `black` and with a prop could be changed in `primary`, `secondary`, `danger`, ...
We need to keep in mind that we might have in future other LinkButton with other colors.

src/modules/menu/components/menu/FeedbackButton.vue
```diff
@@ -1,57 +1,34 @@
...
-            'text-black': isDesktopMode,
-            // Mobile/tablet mode classes
-            'mobile-view btn-sm mx-1': !isDesktopMode,
-            'btn-light': !isDesktopMode,
-        }"
-        :title="$t('app_feedback_button_title')"
-        @click="sendFeedback"
-    >
-    {{ $t('test_map_give_feedback') }}
+    <HeaderLink v-if="isDesktopMode" :selected="true" @click="sendFeedback">
```
ltshb, line 2: See comment above

src/modules/i18n/components/LangButton.vue
```diff
@@ -1,15 +1,13 @@
 <template>
+    <HeaderLink v-if="isDesktopMode" :selected="currentLang === lang" @click="changeLang">
```
ltshb, line 2: See comment below

src/modules/menu/components/menu/FeedbackMoreInfo.vue
```diff
@@ -0,0 +1,33 @@
+<template>
```
ltshb, line 1: Name of the component should `MoreInfo` it has nothing to do with the feedback.

ltshb (approved):
It looks good and works, but I have some comment about naming and logic location.

PR #362, opened 2023-01-30
BGDIINF_SB-2747: Add KML disclaimer icon and external disclaimer warning

ltshb:
Added the KML external disclaimer.

Also now on click of an external layer (kml or external layer) a modal window
with a data disclaimer is shown.

This PR replace #360, which had the data disclaimer warning in the tooltip instead of in a modal. The disadvantage is that tooltip are more designed and useful for short text. It is difficult on mobile to have tooltip that don't bother the user but stay long enough for reading.

[Test link with external layer mf-geoadmin3 syntax](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2747-kml-user-icon-3/index.html?lang=en&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=WMS%7C%7CAargauer%20Wanderweg%7C%7Chttps:%2F%2Fwms.geo.ag.ch%2Fpublic%2Fows%3FSERVICE%3DWMS%26%7C%7Cch_ag_geo_are_divwanderweg%7C%7C1.3.0)
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2747-kml-user-icon-3/index.html)


davidoesch (commented):
The modal window is really very heavy , It is not a warning as modal window, it is more like: you did something really bad... Therefore I would go for #360

ltshb:
@davidoesch what about this 
1. image
 ?
The tooltip is IMHO not very usable on mobile. I agreed that the all red modal is not ideal as well.
Other options:
2. 
image
3. 
image

4. 
image
5. 
image
6. 
image

My preference goes for 1.

davidoesch:
Number one: proceed as discussed today

pakb (approved):
Looking good!
Way better this way, it still looks like you have to watch out for something, but not like you just had a critical error with the app.

PR #363, opened 2023-01-30
BGDIINF_SB-2764 : add feedback form

pakb:
that enables the user to rate his experience with a 5 star scale, and write his message directly in the UI (instead of letting him write it in his email editor).

Also fixing a bug on tablet layout where the links were not rendered properly in the menu/settings section, as the UI was still considered being in desktop mode, thus the feedback and more info buttons were still rendered as links.

TODO :

- [x] Switch the `mailTo` link system to using the (not yet developped) `service-feedback` endpoint to register the ratings/feedbacks

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2764_feedback_form/index.html)


src/modules/menu/components/menu/feedback/FeedbackForm.vue
```diff
@@ -0,0 +1,75 @@
...
+    data() {
+        return {
+            rating: null,
+            maxRating: 5,
+            feedback: null,
+        }
+    },
+    computed: {
+        feedbackCanBeSent() {
+            return this.feedback && this.feedback.length > 0
```
ltshb, line 49: Should we not allow to also simply sent a rating without feedback ?

src/modules/menu/components/menu/feedback/FeedbackForm.vue
```diff
@@ -0,0 +1,75 @@
...
+            )
+            return `mailto:webgis@swisstopo.ch?subject=${title}`
+        },
+    },
+    methods: {
+        ratingChange(newRating) {
+            this.rating = newRating
+        },
+        sendFeedback() {
+            console.log('sending feedback with', {
```
ltshb, line 63: Use logging instead of `console`

src/modules/menu/components/menu/feedback/FeedbackForm.vue
```diff
@@ -0,0 +1,75 @@
...
+</template>
+<script>
+import FeedbackRating from '@/modules/menu/components/menu/feedback/FeedbackRating.vue'
+
+export default {
+    components: { FeedbackRating },
+    emits: ['cancel'],
+    data() {
+        return {
+            rating: null,
```
ltshb, line 42: Use `0` as default.

src/modules/menu/components/menu/feedback/FeedbackButton.vue
```diff
@@ -1,29 +1,37 @@
 <template>
-    <HeaderLink v-if="isDesktopMode" :primary="true" @click="sendFeedback">
+    <HeaderLink v-if="showAsLink" :primary="true" @click="showFeedbackForm = true">
         <strong>{{ $t('test_map_give_feedback') }}</strong>
     </HeaderLink>
-    <button v-else class="btn btn-primary btn-sm mx-1" @click="sendFeedback">
+    <button v-else class="btn btn-primary btn-sm mx-1" @click="showFeedbackForm = true">
         {{ $t('test_map_give_feedback') }}
     </button>
+    <ModalWithBackdrop v-if="showFeedbackForm" @close="showFeedbackForm = false">
```
ltshb, line 8: I would add a title to the modal window, e.g `$t('test_map_give_feedback')`

ltshb (commented):
Very Nice.

I would put the rating before the feedback and let the lazy user give a simple rating, but thats my personal opinion.

ltshb:
@pakb When using the `service-feedback` I proposed to use the following params:

- ua = current user agent
- version = version of the current application
- subject = [web-mapviewer] [rating: x/5] User Feedback

On DEV and INT emails can be seen on the `geoservices-tech@swisstopo.ch`

ltshb:
@pakb I would reduce the max size of the modal window
image
It is quite big.

ltshb:
@pakb the rating preview that you did was quite nice but don't work anymore or did you removed it ?

pakb:
I tried to simplify it as much as possible as it was on other website. Plus I didn't like that it was pushing the whole stars because they are centered, so the star you clicked wasn't under your cursor anymore.

Size : yes I could give a max size for very large screen maybe. I didn't see much issue because I've tested only on small "devices" 😉

src/modules/menu/components/menu/feedback/FeedbackButton.vue
```diff
@@ -0,0 +1,150 @@
...
+        v-else
+        class="btn btn-primary btn-sm mx-1"
+        data-cy="feedback-button"
+        @click="showFeedbackForm = true"
+    >
+        {{ $t('test_map_give_feedback') }}
+    </button>
+    <ModalWithBackdrop
+        v-if="showFeedbackForm"
+        :title="$t('feedback_rating_title')"
```
ltshb, line 20: I personally would use a short title for the window and at best the same as the feedback button. The `How would you rate your experience` would be better IMHO inside the window

src/api/feedback.api.js
```diff
@@ -0,0 +1,45 @@
...
+        })
+        const success = response?.data?.success
+        if (success) {
+            log.info('Feedback sent successfully')
+        } else {
+            log.error('Something went wrong while processing this feedback')
+        }
+        return success
+    } catch (err) {
+        log.error('Error while sending feedback', err)
```
ltshb, line 43: return false to be more explicit

src/api/feedback.api.js
```diff
@@ -0,0 +1,45 @@
...
+        const response = await axios.post(`${API_SERVICES_BASE_URL}feedback`, data, {
+            headers: {
+                'Content-Type': 'multipart/form-data',
+            },
+        })
+        const success = response?.data?.success
+        if (success) {
+            log.info('Feedback sent successfully')
+        } else {
+            log.error('Something went wrong while processing this feedback')
```
ltshb, line 39: Might be a good idea to print the response in case of error.

src/modules/menu/components/menu/feedback/FeedbackRating.vue
```diff
@@ -0,0 +1,59 @@
...
+</script>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+
+.ratings {
+    color: $gray-500;
+    .heart.checked {
+        color: $danger;
+    }
```
ltshb, line 54: For what is this ?

src/modules/menu/components/menu/feedback/FeedbackButton.vue
```diff
@@ -0,0 +1,150 @@
...
+                        icon="check"
+                        data-cy="feedback-success-icon"
+                    />
+                    <span v-else data-cy="feedback-send-text">{{ $t('send') }}</span>
+                </button>
+            </div>
+            <div v-if="request.completed || request.failed" class="text-end mt-3">
+                <span>
+                    <small v-if="request.failed" class="text-danger" data-cy="feedback-failed-text">
+                        {{ $t('upload_failed') }}
```
ltshb, line 68: I would create a new translation for this:

`send_failed`

| Lang | Translation | 
| en | Sending failed |
| fr | Envoi échoué |

ltshb (commented):
Very nice just some comments

src/api/feedback.api.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {String} kmlFileUrl
+ * @returns {Promise<Boolean>} True if successful, false otherwise
+ */
+export default async function sendFeedback(rating, maxRating, text, kmlFileUrl) {
+    try {
+        const shortLink = await createShortLink(window.location.href)
+        let kml = null
+        if (kmlFileUrl) {
+            kml = await getKmlFromUrl(kmlFileUrl)
+        }
```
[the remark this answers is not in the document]
pakb: I can maybe fallback to the long URL, but for KML the service will not accept it if I do not send a proper XML/KML.
I think we should just skip those parameters if we fail to generate them

src/modules/menu/components/menu/feedback/FeedbackRating.vue
pakb: left over 🙈

src/api/feedback.api.js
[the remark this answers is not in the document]
ltshb: For shortlink Fallbeil to long url ans for Kml fallback to no kml 😉

ltshb:
@pakb The deselect of rating (two click on a star) don't work on iphone, the star stay selected but the send button is then disable. I propose to totally remove the deselect feature as it is not needed.

ltshb (approved):
Just a minor bug on Iphone.

To be honest I would not shorten the url, this url is not used by the user, only by us, and I don't see the advantage of having it shorten. In case of problem report it might even be a disadvantage, as we might need to know the exact url that the user used (we would need to do a curl --head on the short url to get it).

pakb:
That's a good point, although the goal initially was to let the user know that we are going to send this "short permalink" to the service desk, in order to be fully transparent. I've not yet added this line saying "This permalink will be send alongside".

I can send the full URL, that would work too yes...

PR #368, opened 2023-02-06
try to us the same font size as in the current UI

procrastinatio:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2756_search_font_size/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchBar.vue
+++ b/src/modules/menu/components/search/SearchBar.vue
@@ -6,7 +6,7 @@
         <input
             ref="search"
             type="text"
-            class="form-control"
+            class="form-control search-bar-input"
             :class="{ 'rounded-end': searchQuery?.length == 0 }"
             :placeholder="$t('search_placeholder')"
             aria-label="Search"
@@ -91,3 +91,11 @@ export default {
     },
 }
 </script>
+
+<style lang="scss" scoped>
+
+.search-bar-input {
+    font-size: 0.825rem;;
+}
+
+</style>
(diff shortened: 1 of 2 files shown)


src/modules/menu/components/search/SearchBar.vue
```diff
@@ -91,3 +91,12 @@ export default {
     },
 }
 </script>
+
+<style lang="scss" scoped>
+
+.form-control {
+    font-size: 0.825rem;
+    font-style: italic;
+}
```
pakb, line 100: I don't think the current viewer has italic font, does it?
Also, it might be better to use a custom class here instead of overwriting things on a bootstrap class, it's all fine because you scoped the style, but as soon as someone will remove this `scoped` part (because he might want to do something globally) then your rule will be applied to all inputs styled with bootstrap.

pakb:
Are you trying to modify only the search bar font? if so please keep everything in the SearchBar component file.
If you wanted to modify the global font size, you should use the Bootstrap variable described here https://getbootstrap.com/docs/5.3/content/typography/#global-settings
in the file `webmapviewer-bootstrap-theme.scss`

src/scss/variables-admin.scss
```diff
@@ -43,6 +43,7 @@ $cadillac: #884488;
 //** Fonts variables
 $frutiger: 'Frutiger Neue', Helvetica, Arial, sans-serif;
 $admin-icons: 'Admin Icons';
+$admin-default-font-size: 0.825rem;
```
pakb, line 46: as said in my last comment, I think this should go away.
https://github.com/geoadmin/web-mapviewer/pull/368#issuecomment-1437934824

pakb (approved):

PR #369, opened 2023-02-06
Updating translations for search bar 

procrastinatio:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2755_search_text/index.html)

### DIFF
--- a/src/modules/menu/components/header/HeaderWithSearch.vue
+++ b/src/modules/menu/components/header/HeaderWithSearch.vue
@@ -17,7 +17,7 @@
                 class="search-bar-section d-flex-column flex-grow-1"
                 :class="{ 'align-self-center': !hasDevSiteWarning }"
             >
-                <span class="float-start search-title">{{ $t('search_title') }}</span>
+
                 <SearchBar />
             </div>
             <div


pakb (commented):
As it doubles down on the text present on desktop, we could get rid of this text oustide of the search bar to gain some screen space. What's your take on that?
image

procrastinatio:
OK, for me.

pakb:
Might be a bit hard to achieve, but would be great if the search bar was centered vertically. And the TESTSITE warning placed underneath, after having centered the searchbar.
You might need to rework the HTML hierarchy to achieve that, and use some of the helper classes found here : https://getbootstrap.com/docs/5.3/utilities/flex/#align-self

procrastinatio:
Looks like it centered now. No idea why, though.

src/modules/menu/components/search/SearchBar.vue
```diff
@@ -91,3 +91,17 @@ export default {
...
+.search-bar {
+  position: absolute;
+  top: 50%;
+  transform: translateY(-50%);
+}
+
+
+
+
+</style>
```
pakb, line 107: I'm not sure this is necessary, as the search bar is already centered by its parent (by HeaderWithSearchBar)

src/modules/menu/components/header/HeaderWithSearch.vue
```diff
@@ -105,6 +105,7 @@ $animation-time: 0.5s;
     // On desktop we limit hte maximum size of the search bar just
     // to have a better look and feel.
     max-width: 800px;
+    position: relative;
```
pakb, line 108: this too might be unnecessarry

pakb (commented):
As I've used some bootstrap Flexbox magic to center all of this, it should be easier for you yes 😉

pakb (approved):

PR #370, opened 2023-02-07
BGDIINF_SB-2769 : add email input in feedback form

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2769-email_field_feedback/index.html)


jedef (commented):
Nice! Just some thoughts:
- After having sent the feedback, the fields stay editable. Maybe this could confuse the user as it will simply close the window if he tries to edit the message and resend it again. So maybe you could deactivate the editing of the text fields when the message has been sent / is being sent.
-  There is no feedback on what is wrong when the send button is grayed out. Maybe the email text box could turn red if it looses focus and the formatting is wrong.

ltshb (commented):
Good, same comment as @jedef

src/modules/i18n/locales/de.json
```diff
@@ -1 +1 @@
...
...isse zu erhalten","try_test_viewer":"Probieren Sie test.map.geo.admin.ch aus","obstacle_started_last_4_days":"Aktivierte Hindernisse letzte 4 Tage","obstacle_deleted_last_4_days":"Gelöschte Hindernisse letzte 4 Tage","draw_button_delete_last_point":"Letzten Punkt löschen","elevation_href":"https://www.swisstopo.admin.ch/de/wissen-fakten/geodaesie-vermessung/bezugsrahmen/lokal/ln02.html","copy_cta":"Kopieren","copy_done":"Kopiert!","test_map_give_feedback":"Feedback geben","legacy_drawing_warning":"Es ist nicht möglich, eine von map.geo.admin.ch erstellte Zeichnung auf der Website test.map.geo.admin.ch zu bearbeiten. Wenn Sie eine neue Zeichnung erstellen möchten, müssen Sie entweder die aktuelle Zeichnung entfernen oder die Auswahl aufheben.","draw_file_load_error":"Laden von KML fehlgeschlagen","draw_file_save_error":"Speichern von KML fehlgeschlagen","very_small_size":"Sehr klein","feedback_rating_text":"Teilen Sie uns mit, wie Sie sich mit unserem Kartenviewer fühlen","feedback_rating_title":"Wie würden Sie Ihre Erfahrung bewerten ? (erforderlich)","send_failed":"Senden fehlgeschlagen","feedback_email":"Ihre E-Mail Adresse (optional)","feedback_invalid_email":"ungültige E-Mail"}
```
ltshb, line 1: Not related to this PR but for review it would be nice to have these json file formatted.

ltshb (commented):
The rating should also be disabled once the feedback has been sent.

Another small issue not related to this PR, is that the rating cannot be given via keyboard, I don't know if this is easily feasible but if it is it should be done, maybe open a ticket for this.

src/modules/i18n/locales/de.json
pakb: that's a good point, I've created a ticket so that we can tackle that later on
https://jira.swisstopo.ch/browse/BGDIINF_SB-2786

jedef (approved):
Can you deactivate the inputs also when request.pending is true as noted in a comment above?
But appart form this small fix, it seems all good to me now 👍

src/modules/menu/components/menu/feedback/FeedbackButton.vue
```diff
@@ -19,21 +19,39 @@
...
         fluid
-        @close="showFeedbackForm = false"
+        @close="closeAndCleanForm"
     >
         <div class="p-2" data-cy="feedback-form">
             <span>{{ $t('feedback_rating_title') }}</span>
             <FeedbackRating
                 class="my-4 text-center"
                 :max-rating="maxRating"
+                :disabled="request.completed"
```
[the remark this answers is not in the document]
pakb: Good point, in the end I also cleared the form when a successful request had been sent, so that only the "Your message was sent successfully..." text is present.
I've therefore only disabled inputs on `request.pending`

PR #373, opened 2023-02-15
BGDIINF_SB-2680 : remove ButtonWithIcon component

pakb:
and replace all usage of this component with standard Bootstrap button. Also simplifying PopoverButton component, so that the specialized version used by feature editing is stored there, and the simpler version uses plain HTML/JS for the time selector.

Reworking the drawing toolbar so that it is more Bootrap based, and less custom CSS based. Also replacing many instances of custom CSS code in component, that could be tackled by Bootstrap, with Bootstrap classes.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2680_remove_button_component/index.html)


jedef:
I haven't looked at he code yet, but just some comments while testing the site:
On tablet size, the "close drawing toolbox" button is missing while there is a "close drawing mode" button that should only appear in phone mode.
Here is a screenshot:
image
Here is how it should look like:
image


Also previously the saving status had a reserved place in the drawing toolbox. Now the drawing toolbox becomes smaller or bigger as the saving status appears or disappears.

ltshb (commented):
On mobile the drawing close button is not well placed:
image
When clicking on it we can see that the cross is not centered on the button.

Personally I don't really like big buttons, I would limit their size based on their content:
image
image

Before the drawing tool selection button and delete last point where not so wide, I like it better, but this might be just a preference taste 😉 

I haven't looked at the code yet.

pakb:
I've added back the reserved space for the "Saving..." / "Map saved." message, it doesn't wobble anymore when showing it the first time this way.
I also fixed some breakpoint in Bootstrap CSS class that I mixed up, the layout was switching to phone mode when we were on tablet size (what you just described @jedef )

Regarding button sice @ltshb I wanted to go back to something more standard, Bootstrap wise, and wanted to impose a no custom CSS rule. I also wanted to cover as many space as possible on mobile, as I found the previous button lackluster in space and difficult to touch.
We have to imagine that this UI might fall into the hands of people who are just brand new to the web (and technology in general), we can't make a specialized app for geo/tech savvy only. We could discuss that in a tech talk if you want, but I'm a firm advocate of having big button on mobile

ltshb:
@pakb I see your point, all good for mobile. However now with this PR the buttons on Desktop are quite much bigger, for Desktop I would reduce the size of the drawing button (icon and text), this would also reduce the size of the menu and let more space for the map. With this PR the size of drawing button are slightly bigger on mobile, which is totally fine there, but are also bigger on desktop, and more there as on mobile.

jedef:
There is still a mismatch between phone and tablet mode. This is probably due to using directly the bootstrap viewport size classes instead of "@respond-above(phone)" or using isPhoneMode.

image

pakb:
I've reverted back some layout choice from the drawing toolbox to using `isPhoneMode` or `isDesktopMode` from the store, instead of relying on Bootstrap breakpoints

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -3,78 +3,104 @@
...
+                                @click="openShare"
+                            >
+                                {{ $t('share') }}
+                            </button>
+                        </div>
+                    </div>
+                    <div v-if="isDrawingLineOrMeasure" class="row mt-2">
+                        <div class="col d-grid">
+                            <button
+                                v-if="isDrawingLineOrMeasure"
```
ltshb, line 76: Is this `v-if` needed if already in line 73

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -60,13 +76,26 @@ export default {
...
+        if (this.hasMultipleTimestamps) {
+            this.popover = tippy(this.$refs.timeSelectorButton, {
+                theme: 'popover-button light-border',
+                content: this.$refs.timeSelectorModal,
+                allowHTML: true,
+                placement: 'right',
+                interactive: true,
+                arrow: true,
+                trigger: 'click',
+            })
```
ltshb, line 89: Did I missed where you delete the tippy instance ?

ltshb (commented):
Overall very good job 👍🏼 

Just a minor issue with tippy.

Also not related to this PR but just an idea, the time button is quite big (was already before) it might be nicer to have a little bit smaller ? 
image

src/modules/drawing/components/DrawingToolbox.vue
pakb: no it is not required anymore... good catch

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
pakb: 🙈 nope

ltshb (approved):
👍🏼

PR #374, opened 2023-02-15
Fix Swiss flag size on smaller devices

pakb:
the flag was shrinking on smaller devices, when the menu is also smaller, not covering the full extent of the space it was given.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_fix_swissflag_size_on_mobile/index.html)


ltshb (commented):
I personally found the flag then too big on mobile. On old viewer it was also shrink. 
For my taste I would leave it the same size, but it could be aligned with the search bar, aka centered vertically.

pakb:
replaced by https://github.com/geoadmin/web-mapviewer/pull/375

PR #376, opened 2023-02-17
BGDIINF_SB-2783 : add back light base map copyrights

pakb:
whenever we show the lightbasemap layer behind pixelkarte-farbe, with the new extent detection we were not going through the standard layer pipeline, so we also need to add an exception in the attribution management for this case

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2783_lightbasemap_copyrights/index.html)


jedef (approved):
Ok, it seems good to me.

PR #377, opened 2023-02-20
feat_BGDIINF_SB-2788: removing pixelkarte grau background layer

procrastinatio:
It would be nice to have the original icons to be able to quicly reconstruct the `backgrounds.jpeg` and `backgrounds_mobile.jpeg` file (without quality loss). Left untouched for now.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2788_background_layer/index.html)

### DIFF
--- a/src/scss/variables-admin.scss
+++ b/src/scss/variables-admin.scss
@@ -14,34 +14,34 @@
 // COLORS
 //=============================================================================
 
-$venetian-red:                               #dc0018;
-$red:                                        #f7001d;
-$mocassin:                                   #fffab2;
-$cerulean:                                   #006699;
-$pattens-blue:                               #d8e8ef;
-$malibu:                                     #66afe9;
-$solitude:                                   #e7edef;
-$clear-day:                                  #f2f7f9;
+$venetian-red: #dc0018;
+$red: #f7001d;
+$mocassin: #fffab2;
+$cerulean: #006699;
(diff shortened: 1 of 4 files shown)


src/scss/variables-admin.scss
```diff
@@ -14,36 +14,38 @@
...
 //** Fonts variables
-$frutiger:    'Frutiger Neue', Helvetica, Arial, sans-serif;
+$frutiger: 'Frutiger Neue', Helvetica, Arial, sans-serif;
 $admin-icons: 'Admin Icons';
 
 // MISC
 //=============================================================================
 $focus-outline: 3px solid $malibu;
+// Number of background layers (in addition to 'blank')
+$numberOfBackgrounds: 3;
```
pakb, line 51: As no other file is using this, I would keep it in `MapFooterBackgroundSelector.vue`
but as you've correctly fixed it, maybe only declare it only once (at the top of the SCSS section)

pakb (commented):
Can you check if it's easy to also enforce light base map when we switch topics? (and the target topic has `pixelkarte-grau` as BG)
should be here : `src/store/plugins/topic-change-management.plugin.js`

src/scss/variables-admin.scss
procrastinatio: OK

PR #379, opened 2023-02-20
BGDIINF_SB-2793: leave menu open when searching

jedef:
- In desktop mode, do not hide the menu anymore when showing search results, but still do that in phone mode.
- To make both the menu and the search results visible at the same time, above the size lg, the search bar starts after the menu tray so that the search results do not overlap with the menu.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2793-open-menu-back-after-search-results-hidden/index.html)

### DIFF
--- a/src/modules/menu/MenuModule.vue
+++ b/src/modules/menu/MenuModule.vue
@@ -40,7 +40,7 @@
                     class="menu-tray"
                     :class="{
                         'desktop-mode': isDesktopMode,
-                        'desktop-menu-closed': isDesktopMode && !showMenu,
+                        'desktop-menu-closed': isDesktopMode && !isMenuShown,
                     }"
                     data-cy="menu-tray"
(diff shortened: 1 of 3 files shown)


src/modules/menu/MenuModule.vue
```diff
@@ -40,7 +40,7 @@
                     class="menu-tray"
                     :class="{
                         'desktop-mode': isDesktopMode,
-                        'desktop-menu-closed': isDesktopMode && !showMenu,
+                        'desktop-menu-closed': isDesktopMode && !isMenuShown,
```
pakb, line 43: Don't know if `isMenuShown` will ever be false on desktop (yet, as no fullscreen support on desktop implemented for now)

pakb (approved):

PR #381, opened 2023-02-22
Updated CI build status badge

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/norn-ltshb/index.html)


pakb (approved):
do IDs for projects change often? I remember updating them not so long ago...

ltshb:
> do IDs for projects change often? I remember updating them not so long ago...

@pakb good question, I think they should not changes unless we destroy/re-create the codebuild project, which we usually don't do unless we have issues with the github webhook.

PR #382, opened 2023-02-23
BGDIINF_SB-2789: search results full width on phone

jedef:
- Search results use full width on phone
- Search results take more screen real estate
- Do not blink when typing search terms

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2789-search-results-full-width-on-phone/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchResultList.vue
+++ b/src/modules/menu/components/search/SearchResultList.vue
@@ -1,6 +1,6 @@
 <template>
     <div
-        class="search-results-container"
+        class="search-results-container m-0"
         :class="{ 'search-results-dev-site-warning': hasDevSiteWarning && isPhoneMode }"
     >
         <div
@@ -72,7 +72,6 @@ export default {
     height: min(calc(100vh - $header-height), max(20rem, 45vh));
     left: 0;
     width: 100%;
-    margin: 0 !important; //overwrites a bootstrap property from parent module
     pointer-events: none;
 }
 .search-results-dev-site-warning {
--- a/src/modules/menu/components/search/SearchResultList.vue
+++ b/src/modules/menu/components/search/SearchResultList.vue
@@ -1,22 +1,20 @@
 <template>
     <div
-        id="search-results-container"
+        class="search-results-container"
         :class="{ 'search-results-dev-site-warning': hasDevSiteWarning && isPhoneMode }"
     >
         <div
             class="search-results bg-light"
-            :class="
-                isPhoneMode
-                    ? ['border-top', 'border-bottom', 'rounded-bottom']
-                    : ['border', 'rounded']
-            "
+            :class="{
+                'border-top border-bottom rounded-bottom': isPhoneMode,
+                'border rounded': !isPhoneMode,
+            }"
             data-cy="search-results"
             @keydown.esc.prevent="$emit('close')"
         >
             <div class="search-results-inner">
                 <SearchResultCategory
                     :title="$t('locations_results_header')"
-                    :half-size="results.locationResults.length > 0"
                     :entries="results.locationResults"
                     data-cy="search-results-locations"
                     @preview-start="setPinnedLocation"
@@ -24,7 +22,6 @@
                 />
                 <SearchResultCategory
                     :title="$t('layers_results_header')"
-                    :half-size="results.layerResults.length > 0"
                     :entries="results.layerResults"
                     data-cy="search-results-layers"
                     @preview-start="setPreviewLayer"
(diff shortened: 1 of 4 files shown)


src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -1,31 +1,42 @@
...
+                    ? ['border-top', 'border-bottom', 'rounded-bottom']
+                    : ['border', 'rounded']
+            "
+            data-cy="search-results"
+            @keydown.esc.prevent="$emit('close')"
+        >
+            <div class="search-results-inner">
+                <SearchResultCategory
+                    :title="$t('locations_results_header')"
+                    :half-size="results.locationResults.length > 0"
```
pakb, line 19: we could get rid of this props binding as you've removed it from the SearchResultCategory component

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -1,31 +1,42 @@
...
+                    :entries="results.locationResults"
+                    data-cy="search-results-locations"
+                    @preview-start="setPinnedLocation"
+                    @preview-stop="clearPinnedLocation"
+                />
+                <SearchResultCategory
+                    :title="$t('layers_results_header')"
+                    :half-size="results.layerResults.length > 0"
```
pakb, line 27: same as above

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -1,31 +1,42 @@
...
-            @preview-start="setPreviewLayer"
-            @preview-stop="clearPreviewLayer"
-        />
+        <div
+            class="search-results bg-light"
+            :class="
+                isPhoneMode
+                    ? ['border-top', 'border-bottom', 'rounded-bottom']
+                    : ['border', 'rounded']
+            "
```
pakb, line 12: I would prefer not using a ternary here, but more something like
```
:class="{
    'border-top border-bottom rounded-bottom': isPhoneMode,
    'border rounded': !isPhoneMode,
}"
```

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -54,18 +66,49 @@ export default {
...
+    left: 0;
     width: 100%;
-    max-height: calc(75vh - 3rem);
+    margin: 0; //overwrites a bootstrap property from parent module
+    pointer-events: none;
+}
+.search-results-dev-site-warning {
+    // Must be important as it needs to overrite the top defined in #search-results-container
+    top: $header-height + $dev-disclaimer-height !important;
+    height: min(calc(100vh - $header-height - $dev-disclaimer-height), max(20rem, 45vh)) !important;
```
pakb, line 84: don't know if the `!important` is required, if so it's maybe because we used an ID to style the element. Maybe if you remove this ID definition and switch to a class based styling for both approach you might be able to get rid of those `!important` (which can be quite dangerous to use...)

tests/e2e-cypress/integration/search/search-results.cy.js
```diff
@@ -301,7 +304,7 @@ describe('Test the search bar result handling', () => {
         cy.get(searchbarSelector).paste('test')
         cy.wait(`@search-locations`)
         cy.get('[data-cy="search-result-entry-location"]').should('be.visible')
-        cy.get('[data-cy="map"]').click()
+        cy.get('[data-cy="map"]').click(viewportWidth * 0.5, viewportHeight * 0.75)
```
pakb, line 307: was there an issue while clicking at the center of the map?

src/modules/menu/components/search/SearchResultList.vue
jedef: "SearchBar.vue" defines the bootstrap class input-group that defines some strange margines that I want to reset, thats why I used an ID (as an ID overwrites styles defined in a class). But this does not solve the problem as you mentioned, it only moves it.
So now I changed it into a class, removed the two important rules that you mentioned, but added one important rule for the margin instead. This is maybe a slight improvement.

tests/e2e-cypress/integration/search/search-results.cy.js
jedef: Yes as the search results go a bit further down than half the screen, so clicking on the center would actually not click on the map.

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -54,18 +63,48 @@ export default {
...
-    left: 0%;
+.search-results-container {
+    position: fixed;
+    top: $header-height;
+    // 45vh (so that previews are visible), but on small screen min 20rem (ca. 3 lines per category)
+    height: min(calc(100vh - $header-height), max(20rem, 45vh));
+    left: 0;
     width: 100%;
-    max-height: calc(75vh - 3rem);
+    margin: 0 !important; //overwrites a bootstrap property from parent module
```
pakb, line 75: then keep it bootsrapy and use `m-0` class 😉

pakb (approved):
very minor detail so we can get rid of this `!important` rule

PR #383, opened 2023-02-24
BGDIINF_SB-2792 : fix layer ordering

pakb:
There was a discrepancy between mf-geoadmin3 layer ordering (in the menu and URL) and web-mapviewer way of handling them. In the URL, layers should be ordered from bottom to top, and on the side menu from top to bottom (layer "on top" of the layer stack is the top layer). Topics were an issue as they list their layers to activate in the opposite order (top -> bottom) so I've inverted those layer before importing them.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2792-layer_ordering/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -72,7 +72,7 @@
             <ButtonWithIcon
                 :button-font-awesome-icon="['fas', 'arrow-up']"
                 :disabled="isFirstLayer"
-                :data-cy="`button-lower-order-layer-${id}`"
+                :data-cy="`button-raise-order-layer-${id}`"
                 :large="!compact"
                 transparent
                 @click="onOrderChange(1)"
@@ -80,7 +80,7 @@
             <ButtonWithIcon
                 :button-font-awesome-icon="['fas', 'arrow-down']"
                 :disabled="isLastLayer"
-                :data-cy="`button-raise-order-layer-${id}`"
+                :data-cy="`button-lower-order-layer-${id}`"
                 :large="!compact"
                 transparent
                 @click="onOrderChange(-1)"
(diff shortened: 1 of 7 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -72,15 +72,15 @@
             <ButtonWithIcon
                 :button-font-awesome-icon="['fas', 'arrow-up']"
                 :disabled="isFirstLayer"
-                :data-cy="`button-raise-order-layer-${id}`"
+                :data-cy="`button-lower-order-layer-${id}`"
```
jedef, line 75: I would keep the data-cy naming here as it was and instead change the naming where it is used. Because even if technically you move the layer further back in an array, this will still result in raising the layer and not lowering it.

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -2,6 +2,7 @@
     <PopoverButton
         v-if="hasMultipleTimestamps"
         ref="popover"
+        popover-position="left"
```
jedef, line 5: Why do you want the popover to open on the left?

jedef (approved):
Just a simple naming fix of a "data-cy" and its all good!

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
pakb: that's to fix an issue with timestamp testing in Cypress, if set to `auto` (left untouched) it would go on the top position, and with the dev warning, Cypress considers that it is covered by the disclaimer and fails the test of visibility. That was an "easy" fix.
We could move it to the right though...

PR #388, opened 2023-03-03
BGDIINF_SB_2673 show absence of data in the profile

pakb:
Brand new approach to the profile chart generation, we now only request `service-alti` backend when we deal with coordinates in LV95's extent, and show the user on the profile chart when there is no data.
Also adding the possibility to zoom into the profile with a pinch gesture (mobile) or through a wheel or shift+drag gesture (on desktop)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2673_profile_show_absence_of_data/index.html)


davidoesch:
Very nice.
I like the zoom approach and the no data approach.
Might be interesting to add for the missing data this source https://open-elevation.com/ but post post mvp

davidoesch (approved):

ltshb:
On mobile by default the tooltip is display below with the description and if we want to see the profile we need to scroll down which is not very user friendly. Especially if the description is empty we only see an empty description and have to scroll down to see the profile. Some possible mitigation
- Increase the height of the profile tooltip footer on mobile
- Put the description below the profile
- Reduce the description box size if it is empty

Before it was not optimal but we could see at least part of the profile on an Iphone SE, now we don't see anything and it is then not trivial to see that there is a profile below
image

pakb:
I agree with your remark but say this is out of scope of this PR. Here I try to replace the chart generation framework, which is already a big chunk of code to review, I would do what you propose in a second PR/ticket.
A possible solution would be to use the header of the infobox (the light gray part) to add two "tabs" there when we are on mobile and dealing with a feature that can be edited and has a profile chart at the same time, so that the user can show only the part that is of interest, and clearly sees that there is two "sides" to this feature.

davidoesch:
Use opentopography freemium API to cover missing data https://opentopography.org/developers

pakb:
> Use opentopography freemium API to cover missing data https://opentopography.org/developers

I prefer to wait for some sort of official solution from our backend, I don't know if users will find this particular solution precise enough when mixed with our data, and that adds a layer of complexity to show the user that this data isn't ours.

I think we should think of a more thorough solution and make our backend service world-wide and mercator compatible.

src/api/profile/ElevationProfile.class.js
```diff
@@ -1,67 +1,32 @@
...
-    /** @returns {Boolean} True if the profile has at least 2 points */
     get hasData() {
-        return this.length >= 2
+        return !!this.segments.find((segment) => segment.hasData)
     }
 
     /** @returns {Number} */
     get maxDist() {
-        return (this.hasData && this.points[this.length - 1].dist) || 0
+        return this.segments.slice(-1)[0].points.slice(-1)[0].dist
```
ltshb, line 29: What happen here if we have no segments ?

src/api/profile/ElevationProfileSegment.class.js
```diff
@@ -0,0 +1,11 @@
+export default class ElevationProfileSegment {
+    /** @param {ElevationProfilePoint[]} points */
+    constructor(points) {
+        /** @type {ElevationProfilePoint[]} */
+        this.points = [...points]
+    }
+
+    get hasData() {
+        return this.points.length >= 2 && this.points[0].elevation !== null
```
ltshb, line 9: Should you not use the `ElevationProfilePoint.hasData()` method here with the `any()`  method ?

davidoesch:
We do already have 2 datasets in the elevation, Swissalti and dhm25 . 

https://open-elevation.com/ would be my choice

src/api/profile/ElevationProfile.class.js
pakb: then it will throw an error 🙈

src/api/profile/ElevationProfileSegment.class.js
pakb: that would be better yes

src/api/profile/profile.api.js
```diff
@@ -0,0 +1,118 @@
...
+            )
+        )
+    })
+    return new ElevationProfileSegment(points)
+}
+
+/**
+ * @param {CoordinatesChunk} chunk
+ * @param {[Number, Number] | null} startingPoint
+ * @param {Number} startingDist
```
ltshb, line 28: Missing return value in docu

src/api/profile/profile.api.js
```diff
@@ -0,0 +1,118 @@
...
+            // or so of the LV95 bounds, resulting in an empty profile being sent by the backend even though our
+            // coordinates were inbound (hence the dataForChunk.data.length > 2 check)
+            if (dataForChunk && dataForChunk.data && dataForChunk.data.length > 2) {
+                return parseProfileFromBackendResponse(dataForChunk.data, startingDist)
+            } else {
+                log.error('Incorrect/empty response while getting profile', dataForChunk)
+            }
+        } catch (err) {
+            log.error('Error while trying to fetch profile data', err)
+        }
```
ltshb, line 57: Would be better to explicitely return a value in case of error, e.g. `null` or an empty `ElevationProfileSegment`

ltshb: Oh I just see that it return an empty segment below. I don't really like method that have a return value in the middle, it would be easier to read and understand to have something like
```javascript
let segment = null
if (chunk.isWithinLV95Bounds) {
  ...
  segment = parseProfileFromBackendResponse()
}

if (!segment) {
   // empty segment in case of error or non lv95 bound
   segment = ...
}

return segment
```

src/api/profile/profile.api.js
```diff
@@ -0,0 +1,118 @@
...
+        return chunk
+    })
+}
+
+/**
+ * Gets profile from https://api3.geo.admin.ch/services/sdiservices.html#profile
+ *
+ * @param {[Number, Number][]} coordinates Coordinates in EPSG:3857 (WebMercator) from which we want
+ *   the profile
+ * @returns {ElevationProfile}
```
ltshb, line 94: or null in case of error

src/api/profile/profile.api.js
```diff
@@ -0,0 +1,118 @@
...
+    if (!coordinates || coordinates.length === 0) {
+        const errorMessage = `Coordinates not provided`
+        log.error(errorMessage)
+        throw errorMessage
+    }
+    const segments = []
+    const coordinateChunks = splitIfOutOfLV95Bounds(coordinates)
+    if (!coordinateChunks) {
+        log.error('No chunks found, no profile data could be fetched')
+        return null
```
ltshb, line 106: or should not be an empty segment here ?

ltshb: If I understand it correctly if the request is totally out of switzerland this method return null, and if it is in switzerland but the backend is failing it return an empty segment ? Seems not very consistent.

ltshb (commented):
still on-going review

src/api/profile/profile.api.js
pakb: And I personally prefer a fail fast (return fast) approach so that we do not have too many indentation, rendering the code harder to grasp 😉 
That's totally a case of taste...

src/api/profile/profile.api.js
pakb: as the function I've written to separate in chunks can return null, whenever the input is malformed or invalid, I wanted to cover this case so that it wouldn't raise an unwanted exception (or console error that would break all the script) later on.

It should NEVER happen, but as the function can return null, well... let's cover it in case. That was why I've written this if block

src/modules/infobox/components/FeatureElevationProfilePlot.vue
```diff
@@ -105,238 +122,304 @@ export default {
...
+                        text: `${this.$t('distance_label')} [${this.unitUsedOnDistanceAxis}]`,
+                        font: {
+                            weight: 'bold',
+                        },
+                        // removing the padding so that we gain a bit of vertical space
+                        padding: 0,
+                    },
+                    ticks: {
+                        // processing distance number to be more human-readable
+                        callback: (val) => round(val * this.factorToUseForDisplayedDistances, 1),
```
ltshb, line 194: Not sure that returning a callback function in computed property is a good idea.

src/modules/infobox/components/FeatureElevationProfilePlot.vue
```diff
@@ -105,238 +122,304 @@ export default {
...
+                    min: this.yAxisMinimumValue,
+                    max: this.yAxisMaxValue,
+                },
+            }
+        },
+        /** Defines how the mouse-over tooltip will behave */
+        chartJsTooltipConfiguration() {
+            return {
+                enabled: false,
+                external: ({ chart, tooltip }) => {
```
ltshb, line 214: I would avoid returning complex method/callback in computed properties which are cached and reactive.

src/api/profile/ElevationProfileSegment.class.js
ltshb: 👍🏼

src/modules/infobox/components/FeatureElevationProfilePlot.vue
```diff
@@ -105,238 +122,304 @@ export default {
...
+                Math.max(Math.ceil(this.tenPercentOfElevationDelta), 5)
+            )
+        },
+        /** Definition of the data ChartJS will show, with some styling configuration too */
+        chartJsData() {
+            return {
+                datasets: [
+                    {
+                        label: `${this.$t('elevation')}`,
+                        data: this.elevationProfile.points,
```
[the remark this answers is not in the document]
pakb: it's an array of instances of `ElevationProfilePoint`, hence the parsing config below to help ChartJS sort what it should get out of those points

src/modules/infobox/components/FeatureElevationProfilePlot.vue
pakb: I don't know either. I don't think we can externalise that (that's a requirement from ChartJS to give a function to process this styling) as we are using some other computed/data props here (`this.factorToUseForDisplayedDistance`)

src/modules/infobox/components/FeatureElevationProfilePlot.vue
pakb: then maybe move that to some methods, and have watchers to update the ChartJS instances in case of update... I'll look into it, that's true it's not optimal to have such big objects in the reactive loop

jedef:
The old tooltip was always visible, even at the border:
image

This is not the case anymore of the new tooltip:
image

jedef:
The data model source was previously a link that was clickable:
image

This is not anymore the case:
image

Is this a wanted change?

jedef (commented):
There are lots of changes, so I could not review all of it today, so I publish what I already have. Overall I like very much the new style of the graph,  that the datapoints are visible and that you can zoom in the graph.
I think that by default the backend always returns 200 points for each request. (but at least every km as the geodesic gemoetry we send to the backend has at least a point every km) I was asking myself if maybe we should adapt the resolution depeding on the length?

jedef:
The problem with the lv95 boundings it that they seem pretty approximative. There is no guarantee that data is available even if within the bonds.
image

ltshb (approved):

PR #389, opened 2023-03-07
BGDIINF_SB-2743 : aggregate TippyJS usage for third party disclaimer

pakb:
There was sometimes issue with TippyJS tooltip not showing up in the footer attributions. As it was kind of the same usage of TippyJS as the one found in the active layer menu, I aggregated the use into one single component, which encapsulate all the TippyJS logic. This should fix the discrepancy between the two use cases.

[Test link with multiple external layers](https://sys-s.dev.bgdi.ch/9de74d1e62)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2743_third_party_tooltip_issues/index.html)


ltshb (approved):
👍🏼

PR #391, opened 2023-03-14
BGDIINF_SB-2820 : legacy parameters double encoded

ltkum:
Issue : Upon receiving legacy parameters, the application would encode them on reception (which is an expected behavior), then create a new query with those adapted parameters, which would encode them again (For example, a space would become `%20`, then `%2520`).

Fix : We decode the URI parameters before storing them in the new query Parameters, ensuring they won't be encoded twice.

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2820-swisssearch-legacy-compatibility/index.html)

### DIFF
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -50,6 +50,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
     const legacyCoordinates = []
     Object.keys(legacyParams).forEach((param) => {
         let value
+
         let key = param
         switch (param) {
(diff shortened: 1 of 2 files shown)


ltshb:
Legacy url with `+` don't work: https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2820-swisssearch-legacy-compatibility/index.html?swisssearch=hello+world

ltshb (approved):
👍🏼 

After testing on the old viewer, the `+` sign was not decoded in a space there so it is ok to have it in the search while translating legacy urls.

PR #392, opened 2023-03-14
BGDIINF_SB-2785: Infobox does not hide menu anymore in phone mode

jedef:
It makes sense for the menu to be above the infobox in phone mode, as in phone mode the menu is fullscreen and so opens on top of everything else. In desktop mode on the contrary, the menu is open most of the time so a user should never be forced to close the menu just to see a popup. Thats why we leave the popup on top of the menu in desktop mode.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2785-scroll-not-possible-in-catalog-when-tooltip-open-mobile/index.html)

### DIFF
--- a/src/modules/map/components/footer/MapFooter.vue
+++ b/src/modules/map/components/footer/MapFooter.vue
@@ -115,8 +115,11 @@ $flex-gap: 1em;
 
     &-middle {
         position: relative;
-        z-index: $zindex-footer-infobox;
+        z-index: $zindex-footer;
         background-color: $white;
+        @include respond-above(phone) {
+            z-index: $zindex-desktop-footer-infobox;
+        }
     }
 
     &-bottom {
--- a/src/scss/variables.scss
+++ b/src/scss/variables.scss
@@ -9,7 +9,9 @@ $zindex-map: 1;
 $zindex-footer: $zindex-map + 1; // The menu should be displayed above the footer in phone mode
 $zindex-warning: 1;
 $zindex-menu: $zindex-footer + 1;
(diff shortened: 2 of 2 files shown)


ltshb (commented):
Change the title to match the bug that we fix, e.g. similar to the ticket title. For release note it is not clear why the menu is above infobox in phone mode.

Also when testing on desktop if the tooltip is in the map overlay it is behind the menu, but if it is set below it is on top of menu also blocking the menu interaction ! You cannot close the menu anymore ! I think this should be also fixed for desktop.

jedef:
> Change the title to match the bug that we fix, e.g. similar to the ticket title. For release note it is not clear why the menu is above infobox in phone mode.
> 
> Also when testing on desktop if the tooltip is in the map overlay it is behind the menu, but if it is set below it is on top of menu also blocking the menu interaction ! You cannot close the menu anymore ! I think this should be also fixed for desktop.

I added a more accurate description at the top to explain the changes and tried to change title to something a bit more understandable. However, the original ticket title is not an accurate description, as technically scrolling was still possible. It's just that a part of the menu was hidden by the popup.

ltshb (approved):
ok as discussed in refinement meeting

pakb (approved):
As discussed together in refinement, this is a good enough solution and we will move on with this for the time being.

A future improvement would be to take the footer/infobox height into account when setting the height of the menu, but that requires extensive work (either render the infobox as pure CSS, and then adapt the placement with some CSS grid magic, or add the infobox height as Javascript computation in the menu height management)
We will wait for further negative feedback on this issue before spending any time on this complex topic

PR #394, opened 2023-03-15
Enforce presence of IDs in GeoJSON data

pakb:
And also reload data whenever lang has changed

The issue with ID was detected while migrating DIEMO layer, as one feature had no ID and was breaking the whole loading of features in OpenLayers.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_fix_diemo_migration/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersGeoJSONLayer.vue
+++ b/src/modules/map/components/openlayers/OpenLayersGeoJSONLayer.vue
@@ -10,8 +10,8 @@ import log from '@/utils/logging'
 import axios from 'axios'
 import GeoJSON from 'ol/format/GeoJSON'
 import { Vector as VectorLayer } from 'ol/layer'
+import { get as getProjection } from 'ol/proj'
 import { Vector as VectorSource } from 'ol/source'
-import { reproject } from 'reproject'
 import addLayerToMapMixin from './utils/addLayerToMap-mixins'
 import OlStyleForPropertyValue from './utils/styleFromLiterals'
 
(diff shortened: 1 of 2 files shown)


boecklic (commented):
Hmm, I get a bunch of errors if I open the testlink, search for 'elektro', and select the diemo layer. The layer eventually loads, I'm not sure if the errors in the console are related to this or not...

src/modules/map/components/openlayers/OpenLayersGeoJSONLayer.vue
```diff
@@ -46,66 +46,58 @@ export default {
...
+        async loadStyle() {
+            try {
+                const style = new OlStyleForPropertyValue((await axios.get(this.styleUrl)).data)
+
+                this.layer.setStyle(function (feature, res) {
+                    return style.getFeatureStyle(feature, res)
+                })
+            } catch (err) {
+                log.error('Unable to load style for GeoJSON layer', this.layerId, err)
+            }
```
ltshb, line 74: I would not put the try catch block around everything, but only around the axios call.
During my test the loadStyle() failed as follow:
```
Unable to load style for GeoJSON layer ch.bfe.ladestellen-elektromobilitaet TypeError: Cannot read properties of undefined (reading 'setStyle')
    at Proxy.loadStyle (index-373d0628.js:1410:17195)
```
The Axios call was successful which means that the data could not be understood by openlayer, I don't know if letting crash would provide more infos here but we clearly don't don't know why it happens.

src/modules/map/components/openlayers/OpenLayersGeoJSONLayer.vue
```diff
@@ -46,66 +46,58 @@ export default {
...
-            .catch((error) => {
-                log.error(
-                    'Error while fetching GeoJSON data/style for layer ' + this.layerId,
-                    error
-                )
-            })
+            } catch (err) {
+                log.error('Unable to load data for GeoJSON layer', this.layerId, err)
+            }
+        },
```
ltshb, line 100: Similar as above, might be a better idea to treat axios call and data parsing error differently ?

PR #396, opened 2023-03-20
BGDIINF_SB-2829: Text in tooltips is now selectable

jedef:
Text in the tooltips of features was previously not selectable. This
commit fixes this. The classname map-popover-overlay did not exist.
Removing this classname activates instead the default class name
ol-selectable and ol-overlay-container which makes the text in the
overlay selectable. (The overlay is a child of the openlayersmap which
deactivates selection)

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-bgdiinf_sb-2829-cant-copy-from-tooltip/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -8,7 +8,7 @@
         data-cy="location-popup"
         @close="onClose"
     >
-        <div class="location-popup-coordinates ol-selectable">
+        <div class="location-popup-coordinates">
             <div class="lp-label">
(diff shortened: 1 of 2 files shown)


jedef:
> Looks good but I don't understand why it did not work before, was it because of the `map-popover-overlay` class ? Where is this class defined ? Looking at the code diff I would have expect that the text selection of the location popup is removed ! So maybe better explain the code changes in commit.

I changed the commit message.

ltshb (approved):
Great thanks

PR #397, opened 2023-03-27
BGDIINF_SB-2748: Add selection feedback in topics.

jedef:
[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-bgdiinf_sb-2748-add-selection-feedback-in-topics/index.html)

### DIFF
--- a/src/utils/ModalWithBackdrop.vue
+++ b/src/utils/ModalWithBackdrop.vue
@@ -4,11 +4,12 @@
         sure that it is always on top of the reset. -->
         <div>
             <BlackBackdrop class="modal-view" @click="onClose(false)" />
-            <div class="modal-popup" :class="{ 'modal-popup-fluid': fluid }">
+            <div class="modal-popup">
                 <div
                     class="card"
                     :class="{
                         'border-primary': headerPrimary,
+                        'modal-popup-fluid': fluid,
                     }"
                 >
                     <div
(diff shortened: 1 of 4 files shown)


ltshb (commented):
As discussed it doesn't work now

ltshb:
buttons looks squashed, the padding on top/bottom could be increased or reduce the padding/margin on left/right to have kind of square button that don't look squashed.

ltshb:
I see your point about having the title on line. Looking at the old viewer it seems that the width is fixed but large enough to have all title on one line. In your current version depending on the screen width you might have very wide button which looks very ugly and a lots of white spaces.
image

ltshb:
Another idea would be to use the outline buttons to have less big fat red buttons.

jedef:
image

Some words don't break correctly. However this was already the case before. I think that this could be fixed by changing the translation (e.g. in the example above to "Grundstück\&shy;information"), so that the browser knows how to correctly split the word into two lines.) However, the translations are also used by the old viewer, so I'm not sure if I can do that safely.

src/utils/ModalWithBackdrop.vue
```diff
@@ -126,16 +126,27 @@ export default {
...
     @include respond-above(phone) {
         // But for desktop we let the size be dynamic with max to 90% of the view
         width: unset;
         max-width: 90vw;
+        max-height: 90svh;
     }
     .card {
+        // Same as above, 90vh will be used as fallback is 100dvh or 90svh is not recognized.
         max-height: 90vh;
+        max-height: 100dvh;
```
ltshb, line 145: You have twice `max-height` here

src/utils/ModalWithBackdrop.vue
```diff
@@ -126,16 +126,27 @@ export default {
         min-width: 75vw;
     }
     z-index: $zindex-modal;
+    // For phone we set the width and height fixed to 100% of the view.
+    // dvh takes into account the user interface in mobile browsers (with vh part of the modal is
+    // not visible if ui is shown). Is recognized by browsers from 2022 or newer. If the browser
+    // is older, 90vh will normally be used, which is a bit less clean but good enough.
```
ltshb, line 132: Although this is a good idea for the topic use case where the window take anyway the whole screen, this doesn't fit well with all use cases. For the `Delete` and `Share` use case of the drawing the modal is smaller and it doesn't make sense to have it as a banner in the middle. 
image
So I would either left it as before or only make it fullscreen when it make sense. When making it fullscreen the rounded corner needs to be removed though.

src/utils/ModalWithBackdrop.vue
jedef: Please read the comment just above that explains this

ltshb: Ok I did not see this, might be a good idea to put a `fallback` comment next to the line.

ltshb (approved):
👍🏼

PR #400, opened 2023-04-12
BGDIINF_SB-2849 : CMS post-migration chores

pakb:
Update copyright label and URL after CMS migration. Some other links in the translations have also been updated

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2849_cms_migration/index.html)


src/modules/i18n/locales/rm.json
```diff
@@ -81,7 +81,8 @@
 	"copy_success": "Copià il link!",
 	"copy_url": "Copiar il link",
 	"copyright_data": "© Datas:",
-	"copyright_label": "Copyright & decleraziun da protecziun da datas",
+	"copyright_label": "Nutzungsbedingungen",
```
jedef, line 84: Line 437 I see that this has already been trasnslated in Rm as "cundiziuns d'utilisaziun", so you could take the same translation here.

ltshb: Good point

src/modules/i18n/locales/rm.json
```diff
@@ -81,7 +81,8 @@
...
+	"copyright_url": "https://www.geo.admin.ch/de/allgemeine-nutzungsbedingungen-bgdi/",
```
ltshb, line 85: I see that the page for the romanche in CMS is missing, does the CMS webmaster knows about this and will add it ?

pakb: I don't know what's the stand with romanche, but for now if you try to access any page with this language it says that there is no translation yet. (https://www.geo.admin.ch/rm/)

src/modules/i18n/locales/rm.json
pakb: Yes we could "vampirize" this translation, good point

src/modules/i18n/locales/rm.json
ltshb: I know but I would ask if this is know and if it is planned to be fixed, if it is then I would use the correct path here even if it is a not found page, otherwise it will be forgotten.

PR #403, opened 2023-04-25
BGDIINF_SB-2859: Removing legend button

ltkum:
Issue : Some layers have no legend, and upon clicking the legend button, the legend popup shows up and looks like it's perpetually loading content.

Fix : We added a `hasLegend` getter in the abstractLayer class, which is then called to decide wheter we show the legend button or not.

Currently, we consider KML and external layers has having no legend. A refinement might be needed further down in the line to ensure layers with legends show the legend icon, but this is sufficient for now.

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2859-remomve-legend-icon-when-no-legend/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -97,6 +97,7 @@
                 <FontAwesomeIcon icon="arrow-down" />
             </button>
             <button
+                v-if="showLegendIcon"
                 class="btn d-flex align-items-center"
                 :class="{ 'btn-lg': !compact }"
                 :data-cy="`button-show-legend-layer-${id}`"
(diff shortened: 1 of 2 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -178,6 +179,12 @@ export default {
             }
             return 'square'
         },
+        showLegendIcon() {
+            if (this.layer !== null) {
+                return this.layer.hasLegend
+            }
+            return false
```
pakb, line 186: just for curiosity, you can "one line" this in JS with the optional chaining :
`return this.layer?.hasLegend`

with this notation, `hasLegend` will only be accessed if `this.layer` is valid (not null or undefined), otherwise it returns `null` (and as `null` is somewhat equivalent to `false` in JS, that works pretty well)

pakb (approved):

PR #405, opened 2023-04-27
BGDIINF_SB-2609: consistent rounding

ltkum:
Issue : When displaying longitude and latitude, when the last figure of a number was 0, we would display numbers with a different significant number after the decimal.

Fix : We now use the "toFixed" native js function to have a fixed amount of significant numbers after the decimal.

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2609-round-position/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -141,23 +141,19 @@ export default {
             return this.clickInfo?.coordinate
         },
         coordinateLV95() {
-            return printHumanReadableCoordinates(
-                reproject(CoordinateSystems.LV95.epsg, this.coordinate),
-                CoordinateSystems.LV95
-            )
+            const lv95_coordinates = reproject(CoordinateSystems.LV95.epsg, this.coordinate)
+            return `${round(lv95_coordinates[0], 2, true)}, ${round(lv95_coordinates[1], 2, true)}`
         },
         coordinateLV03() {
-            return printHumanReadableCoordinates(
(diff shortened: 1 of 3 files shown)


src/utils/numberUtils.js
```diff
@@ -4,15 +4,20 @@
...
+export function round(value, decimals = 0, enforcedigit=false) {
     if (!isNumber(value)) {
         return undefined
     }
     if (decimals === 0) {
         return Math.round(value)
     }
+    if (enforcedigit) {
+    return value.toFixed(decimals)
+    }
```
pakb, line 20: linting 😉 (missing tab here)
you should be able to auto-lint with Prettier, if you use VSCode or JetBrains products, you can add Prettier as a save action command (or there's typically a shortcut to launch it)

src/modules/map/components/LocationPopup.vue
```diff
@@ -157,7 +157,7 @@ export default {
         },
         coordinateWGS84Plain() {
             const wgsMetric = this.coordinateWGS84Metric
-            return `${round(wgsMetric[1], 5)}, ${round(wgsMetric[0], 5)}`
+            return `${round(wgsMetric[1], 5, true)}, ${round(wgsMetric[0], 5, true)}`
```
pakb, line 160: I would apply this forceDigit to all outputs of this LocationPopup component for consistency (so all other coordinate systems with floating numbers)

pakb (approved):

PR #407, opened 2023-05-25
BGDIINF_SB-2501 : add support for embed URL param

pakb:
when set to true, the UI is simplified to a minimum (zoom in zoom out and a link to open the app in a new tab)

I had to rework a bit the URL parsing tooling for this, as otherwise the embed=false or embed= were staying in the URL. But working on this opened another whole can of worms with app startup and I had to change the logic in some places in the storeSync plugin to fix that (layers and bgLayers were disappearing at startup somehow)

### DIFF
--- a/src/modules/i18n/locales/de.json
+++ b/src/modules/i18n/locales/de.json
@@ -614,5 +614,6 @@
 	"send_failed": "Senden fehlgeschlagen",
 	"feedback_email": "Ihre E-Mail Adresse (optional)",
 	"feedback_invalid_email": "ungültige E-Mail",
-	"profile_no_data": "kein Datum"
+	"profile_no_data": "kein Datum",
+	"view_on_mapgeoadminch_webmapviewer": "In {url} ansehen"
 }
--- a/src/modules/i18n/locales/en.json
+++ b/src/modules/i18n/locales/en.json
@@ -614,5 +614,6 @@
 	"send_failed": "Sending failed",
 	"feedback_email": "Your Email (optional)",
 	"feedback_invalid_email": "Invalid email",
-	"profile_no_data": "No data"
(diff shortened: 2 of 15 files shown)


ltshb:
@pakb switching infobox of a feature to the header don't work, the infobox disappear and the user need to refresh the windows to see it again.

ltshb:
After some thoughts if we need to touch the URL parameter logics, I think it would be wise to try to simplify it. It seems a bit complex to me especially with the `keepValueInUrlWhenEmpty` flag which is used in several place in code and not always straightforward. I'm thinking of a more generic way to handle parameters, something as follow:

1. Add a `defaultValue` member to the `AbstractParamConfig`
2. Change the `keepValueInUrlWhenEmpty` to `keepInUrlWhenDefault`
3. Always add the param to the store, either from query if found or using the `defaultValue`. In this logic no need to check for `keepInUrlWhenDefault`
4. When writing the URL, only add non default param unless `keepInUrlWhenDefault` is true
5. Make sure that boolean param without value are treated as `true` same as boolean param with value `1` 

This should allow to have the `embed` flag in url without value to keep url short. It would also allow to have `embed=true` or `embed=1`. We could also have a flag which is per default `true` if needed in future. Maybe we can have a quick call to discuss this idea ?

ltshb (approved):
Looks great thanks

pakb:
> @pakb I guess the old embed (old viewer) translation to the new viewer will be done via redirect ? embedeed.html -> #?embed, am I correct. Has it already be done ? If not is there already a ticket ?

that's a good point, as it was a different HTML endpoint in mf-geoadmin3 I think we might need to tackle that on the proxy side of things (what Bernie was working on recently), I'll create a ticket so that we do not forget to deal with it.

PR #408, opened 2023-05-30
BGDIINF_SB-1856 : share - iframe code snippet generator

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-1856_iframe-generator/index.html)

### DIFF
--- a/src/modules/menu/components/share/MenuShareEmbed.vue
+++ b/src/modules/menu/components/share/MenuShareEmbed.vue
@@ -6,7 +6,7 @@
             @click="toggleEmbedSharing"
         >
             <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
-            <span class="ms-1">{{ $t('share_more') }}</span>
+            <span class="ms-2">{{ $t('share_more') }}</span>
         </button>
         <CollapseTransition :duration="200">
             <div v-show="showEmbedSharing" class="p-2 card border-light bg-light">
@@ -46,13 +46,14 @@
                 >
                     <select
                         v-model="currentPreviewSize"
-                        class="form-select w-auto"
+                        class="embed-preview-modal-size-selector form-select w-auto"
                         data-cy="menu-share-embed-iframe-size-selector"
                     >
                         <option
                             v-for="size in EmbedSizes"
                             :key="size.i18nKey"
                             :value="size"
+                            class="embed-preview-modal-size-selector-option"
                             :data-cy="`menu-share-embed-iframe-size-${size.i18nKey.toLowerCase()}`"
                         >
                             {{ $t(size.i18nKey) }}
@@ -206,5 +207,8 @@ export default {
     @include respond-above(phone) {
         display: block;
     }
+}
+.embed-preview-modal-size-selector {
+    cursor: pointer;
 }
 </style>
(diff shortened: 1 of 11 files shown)


src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -1,25 +1,34 @@
 <template>
     <div class="menu-share-embed">
-        <button class="btn btn-light embedded-button" @click="toggleEmbedSharing">
-            <FontAwesomeIcon icon="plus" />
+        <button
+            class="btn btn-light embedded-button"
+            data-cy="menu-share-embed-button"
+            @click="toggleEmbedSharing"
+        >
```
ltshb, line 7: Maybe use a small button `btn-sm` here. The `+ Embed` is bigger than the menu title `Share`

ltshb: Instead of a `+` `-` icons I would use the arrow right and down icon like in the menu and drop down buttons.

src/store/modules/share.store.js
```diff
@@ -22,20 +27,30 @@ export default {
     getters: {},
     actions: {
         async generateShortLink({ commit }, withBalloonMarker = false) {
```
ltshb, line 29: Now the name should be plural as it generates two links

src/store/modules/share.store.js
```diff
@@ -22,20 +27,30 @@ export default {
...
-                        // if the geolocation was being tracked by the user generating the link, we place a balloon (dropped pin) marker at his position (center of the screen, so no need to change any x/y position)
-                        (withBalloonMarker ? '&crosshair=marker' : '')
-                )
+                const shortLink = await createShortLink(urlWithoutGeolocation)
                 if (shortLink) {
                     commit('setShortLink', shortLink)
                 }
             } catch (err) {
                 log.error('Error while creating short link for', window.location.href, err)
                 commit('setShortLink', null)
```
ltshb, line 41: Fallback to non shorten url would be good in case of errors

src/store/modules/share.store.js
```diff
@@ -22,20 +27,30 @@ export default {
...
+            try {
+                const embeddedShortLink = await createShortLink(
+                    urlWithoutGeolocation + '&embed=true'
+                )
+                if (embeddedShortLink) {
+                    commit('setEmbeddedShortLink', embeddedShortLink)
+                }
+            } catch (err) {
+                log.error('Error while creating embedded short link for', window.location.href, err)
+                commit('setEmbeddedShortLink', null)
```
ltshb, line 52: fallback to non shorten url would be good in case of error

src/store/modules/ui.store.js
```diff
@@ -56,7 +56,7 @@ export default {
          *
          * @type Boolean
          */
-        embeddedMode: false,
+        embeddedMode: 'true' === new URLSearchParams(window.location.href).get('embed'),
```
ltshb, line 59: Is this really necessary ?

ltshb:
In the shortlink input if you click inside it automatically select the whole input, but on the embed input if you click outside and then back inside nothing is selected. Would be good to have the same behavior for both inputs.

src/store/modules/ui.store.js
pakb: I was trying to start the app straight into embed mode, without having the UI flicker between the two states. It's not yet as I wanted it to be, so I'll remove that and create a ticket to tackle this detail

ltshb:
The dropdown button don't uses `pointer` cursor 
image

ltshb:
Personnally I would add a bit more space/padding between the icon and the text here 
image

ltshb (approved):
Just the two minor comment above

PR #412, opened 2023-06-15
BGDIINF_SB-2956 : add 3D toggle button in the toolbox

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2956_add-3d-button/index.html)


ltshb (approved):
What about the legacy URL ? Will this be done separately do you have a ticket for that ?

pakb:
good point, I've created a ticket for that yes : https://jira.swisstopo.ch/browse/BGDIINF_SB-2970
but I think it would make more sense to wait for other param to be implemented before doing this work (all the camera positioning / angle params)

PR #414, opened 2023-06-16
BGDIINF_SB-2429 : add a full-width checkbox to the embed code generator

pakb:
so that the generated code can be a fluid iframe (meaning it takes 100% of the width of its container)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2429_full_width_embed_snippet/index.html)

### DIFF
--- a/scripts/generate-i18n-files.js
+++ b/scripts/generate-i18n-files.js
@@ -55,7 +55,7 @@ sheets.spreadsheets.values.get(
                 if (index > 0) {
                     fs.writeFileSync(
                         `src/modules/i18n/locales/${lang}.json`,
-                        JSON.stringify(translations[lang], null, '    ')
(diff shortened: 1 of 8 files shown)


ltshb (commented):
image
image

Not a big fan of the `X` between the input box, it looks like a button but it is not. Might be better to have it as a label instead ?
Something like this 
image
Also the `100%` is centered in its input group while the other are left align, would be good to be consistent, also making the input group a bit smaller as they will probably never be used with a number bigger than 9'999. Might be also better to have them right align instead of left as they are numbers.

pakb:
then let's address that in the next ticket (https://jira.swisstopo.ch/browse/BGDIINF_SB-2430) if you don't mind, as it was written specifically for this.
I'll add your right text align suggestion in it.

ltshb (approved):

PR #416, opened 2023-06-27
 BGDIINF_SB-2018: catalogNodes parameter support

ltkum:
Issue : When specifying a catalogNodes in the URL, it didn't open the
corresponding catalog Nodes.

Fix : Created a type of URL parameter which is not bi-directional
(modifying the URL will modify the store, modifying the store won't
touch the URL), and added watchers and actions to track and make use of
it


pakb (commented):
you might want to clean your git log 😉 
a good help, if you're as me a bit too visual to do everything through git command, is to use some GUI like GitKraken, helps a lot to keep things clean

pakb (commented):
Can you resolve (or remove) all remaining TODOs ?

src/router/storeSync/storeSync.config.js
```diff
@@ -119,6 +120,22 @@ const storeSyncConfig = [
...
     ),
+    new SimpleUrlParamConfig(
+        '3d',
+        'setShowIn3d',
+        'setShowIn3d',
+        (store) => store.state.ui.showIn3d,
+        false,
+        Boolean,
+        false
+    ),
```
pakb, line 131: this is now a duplicate, you might have added that in your latest rebase
please remove it

tests/e2e-cypress/integration/topics.cy.js
```diff
@@ -270,4 +285,31 @@ describe('Topics', () => {
             })
         })
     })
+    context('Using the CatalogNodes URL parameter', () => {
+	it('Open the catalog and only the relevant topic and node elements when the URL contains the catalogNode parameter', () => {
+	    cy.goToMapView('en', {
+	    catalogNodes: '2'})
```
pakb, line 291: the format is a bit weird here, are you sure you've run prettier on this file?

src/router/storeSync/storeSync.config.js
ltkum: thanks for finding this.

tests/e2e-cypress/integration/topics.cy.js
ltkum: the prettier command we made is not formatting the tests. I'll add this to another PR if that's okay with you.

pakb: then fix your indentation (but you could also run prettier from your IDE on demand), the portion of code you're adding is not very well indented or has weird line breaks

for VS Code, there's a Prettier plugin, then you can use CTRL+SHIFT+P and boom, voilà, it's formatted

pakb (approved):

PR #417, opened 2023-06-27
BGDIINF_SB-2958 : Groundwork for future Cesium components

pakb:
Also moved many OL dependent components directly into the OpenLayers realm/folder, and teleporting them at the right spot.
This way we don't rely too much on the slot of the MapModule to render most of our UI. This was making the menu and footer disappear whenever we were switching between 2D and 3D (as the MapModule was moving the slot between OL and Cesium), very ugly switching...

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-2958_groundwork_cesium_components/index.html)


ltshb (approved):
LGTME

PR #418, opened 2023-06-30
BGDIINF_SB-2982: limit camera height

fredj:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2982-limit-cesium-camera/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumMap.vue
+++ b/src/modules/map/components/cesium/CesiumMap.vue
@@ -30,8 +30,15 @@ import {
 } from 'cesium'
 import { UIModes } from '@/store/modules/ui.store'
 import { mapGetters, mapState } from 'vuex'
-import { TERRAIN_URL } from './constants'
-import { IS_TESTING_WITH_CYPRESS, WMTS_BASE_URL } from '@/config'
+import {
+    TERRAIN_URL,
+    CAMERA_MIN_ZOOM_DISTANCE,
+    CAMERA_MAX_ZOOM_DISTANCE,
+    CAMERA_MIN_PITCH,
+    CAMERA_MAX_PITCH,
+} from './constants'
+import { limitCameraCenter, limitCameraPitchRoll } from './utils/cameraUtils'
+import { IS_TESTING_WITH_CYPRESS, WMTS_BASE_URL, TILEGRID_EXTENT_EPSG_4326 } from '@/config'
 import CesiumInternalLayer from './CesiumInternalLayer.vue'
 import GeoAdminWMTSLayer from '@/api/layers/GeoAdminWMTSLayer.class'
 import LayerTimeConfig from '@/api/layers/LayerTimeConfig.class'
@@ -134,7 +141,6 @@ export default {
         }
 
         const scene = this.viewer.scene
-        scene.screenSpaceCameraController.enableCollisionDetection = false
         scene.useDepthPicking = true
         scene.pickTranslucentDepth = true
         scene.backgroundColor = Color.TRANSPARENT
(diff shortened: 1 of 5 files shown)


tests/e2e-cypress/integration/3d/navigation.cy.js
```diff
@@ -0,0 +1,38 @@
...
+        it('minimum distance from the terrain', () => {
+            cy.waitUntilCesiumTilesLoaded().then((viewer) => {
+                // Move close to the ground and try to zoom closer with the mouse wheel
+                viewer.camera.flyTo({
+                    destination: Cartesian3.fromDegrees(7.451498, 46.92805, 500),
+                    duration: 0.0,
+                })
+                cy.get('[data-cy="cesium"] .cesium-viewer').trigger('wheel', { deltaY: -5000 })
+                cy.waitUntilCesiumTilesLoaded().then(() => {
+                    expect(viewer.scene.camera.positionCartographic.height).gt(25)
```
pakb, line 20: you could even import the `CAMERA_MIN_ZOOM_DISTANCE` constant here, meaning that if we change it in the future the test would dynamically adapt.

tests/e2e-cypress/integration/3d/navigation.cy.js
```diff
@@ -0,0 +1,38 @@
...
+        it('maximum distance from the terrain', () => {
+            cy.waitUntilCesiumTilesLoaded().then((viewer) => {
+                // Move far from the ground and try to zoom higher with the mouse wheel
+                viewer.camera.flyTo({
+                    destination: Cartesian3.fromDegrees(7.451498, 46.92805, 4999500),
+                    duration: 0.0,
+                })
+                cy.get('[data-cy="cesium"] .cesium-viewer').trigger('wheel', { deltaY: 5000 })
+                cy.waitUntilCesiumTilesLoaded().then(() => {
+                    expect(viewer.scene.camera.positionCartographic.height).lt(5000000)
```
pakb, line 33: same remark as above, might be a good idea to import the constant and use it here

tests/e2e-cypress/support/commands.js
```diff
@@ -339,3 +339,22 @@ Cypress.Commands.add('addProfileJsonFixture', (mockupData) => {
...
+        })
+    })
+}
+
+Cypress.Commands.add('waitUntilCesiumTilesLoaded', () => {
+    cy.readWindowValue('cesiumViewer').then((viewer) => {
+        cy.wrap(null).then({ timeout: 50000 }, async () => {
+            await cesiumTilesLoaded(viewer)
+            return viewer
+        })
```
pakb, line 358: what is achieved by wrapping a `null`?
if your goal was for Cypress to wait for an arbitrary duration, you can use `cy.wait(...)` which is synchronous

instead of using a Promise based wait, you could also ressort to a request intercept. For instance if you remove what you changed in `src/modules/map/components/cesium/CesiumMap.vue` at line 71, you could intercept requests to terrain tiles and serve a mockup (or no tiles) in the intercept. This would look something like :
```
cy.intercept('**.terrain**, { /* Mock up response, or path to a fixture */}).as('loadTerrainTile')
```
and then use something like
```
cy.goToMapView('en', { showIn3d: true })
cy.wait('@loadTerrainTile')
```
I don't know if this would simplify your workflow, it's usually a bad practice in Cypress to wait for arbitrary length of time.

fredj: I've found this `cy.wrap(null)` on a cypress blog somewhere (I will found it again and add a comment).
As for the `timeout` it's not a fixed amount to wait but the maximum duration to wait before failing, the default  10000ms if not enough

tests/e2e-cypress/integration/3d/navigation.cy.js
fredj: done

tests/e2e-cypress/integration/3d/navigation.cy.js
fredj: done

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -103,6 +115,10 @@ export default {
         )
         baseLayer.show = true
 
+        const rectangle = Rectangle.fromDegrees(5.1402988, 45.3981222, 11.4774363, 48.230617)
```
fredj, line 118: Todo, use `WGS84_EXTENT` once #423 is merged

pakb: little heads up : it is now renamed `TILEGRID_EXTENT_EPSG_4326`

tests/e2e-cypress/support/commands.js
```diff
@@ -344,3 +344,22 @@ Cypress.Commands.add('addProfileJsonFixture', (mockupData) => {
...
```
pakb, line 363: might be good idea to add this comment you were suggesting, about this blog post describing this `cy.wrap(null)`

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -131,7 +138,7 @@ export default {
             skyBox: false,
             imageryProvider: firstImageryProvider,
             useBrowserRecommendedResolution: true,
-            terrainProvider: await CesiumTerrainProvider.fromUrl(TERRAIN_URL),
+            terrainProvider: IS_TESTING_WITH_CYPRESS ? undefined : await CesiumTerrainProvider.fromUrl(TERRAIN_URL),
```
pakb, line 141: so with test, to make them faster, no terrain at all?

PR #419, opened 2023-07-03
Add Formatting Command

ltkum:
Issue : There was no formatting command for npm

Fix : We added a formatting command for npm and ran it.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-add-formatting/index.html)


src/modules/menu/scss/topics-image.scss
```diff
@@ -1,49 +1,145 @@
...
+        position: -330px -300px,
+    ),
+    (
+        key: bfs,
+        position: -330px -360px,
+    ),
+    (
+        key: bav,
+        position: -330px -420px,
+    )
```
pakb, line 129: this block I kind of dislike post format, less readability...
Could you wrap it around some "no format" comments or something that deactivate format for these lines?

pakb (commented):
might be a good idea to also spot the place where the translations files are created (should be in `scripts/generate-i18n-files.js`) and change the line creation to add directly 4 spaces instead of a tab char

pakb:
have you re-run the `npm run format-lint` command after having rebased on the latest work made on develop? there might be some small changes there too

src/modules/menu/scss/topics-image.scss
```diff
@@ -34,16 +35,16 @@ $topics: (
 );
```
pakb, line 35: you might want to reactivate prettier from here

pakb (approved):
on last detail and we are good to go, did you run this formatter once more after rebasing on develop?

src/modules/menu/scss/topics-image.scss
ltkum: prettier-ignore-start and prettier-ignore-end are only supported on markdown. I could make it work only for this block but I'd have to wrap it into a block with brackets, which I'm not sure how it would interact with the code (unless I'm wrong, which is perfectly possible ^^).

and yes, I did format after rebasing.

PR #420, opened 2023-07-03
BGDIINF_SB-2973 : add camera URL param

pakb:
As this clashes with the existing lon,lat and zoom URL param, I've condensed the management of all position param in a single param config class. This enables us to more efficiently add or remove relevant URL params depending on the state of the 3D or 2D rendering.

The behavior of this new config class has been extensively covered with unit tests to ensure it functions properly (and future changes will not break this essentiel group of URL params)


src/router/storeSync/PositionUrlParamConfig.class.js
```diff
@@ -0,0 +1,229 @@
+import AbstractParamConfig from '@/router/storeSync/abstractParamConfig.class'
+
+function dispatchCameraPositionFromUrlIntoStore(store, urlParamValue) {
+    const allValues = urlParamValue.split(',')
+    const [x, y, z, yaw, pitch, roll] = allValues.map((value) =>
+        value === '' ? 0 : parseFloat(value)
+    )
```
ltshb, line 7: I would put the parsing of the URL parameter in a separate function and reuse it in `parseCameraFromQuery`. Currently here you use `parseFloat` and in the other `Number`, does it return the same ? I would also handle/catch error due to invalid URL parameter to avoid a crash, basically the url parameter could be manually entered and invalid.

src/router/storeSync/PositionUrlParamConfig.class.js
```diff
@@ -0,0 +1,229 @@
...
+            dispatchCameraPositionFromUrlIntoStore,
+            generateCameraUrlParamFromStoreValues,
+            false,
+            String
+        )
+    }
+
+    /**
+     * Even though it is not used below in populate functions, this is required so that the
+     * routerStoreSyncPlugin can work properly
```
ltshb, line 98: I don't understand the comment, if this is not used why not simply return null ?

src/router/storeSync/PositionUrlParamConfig.class.js
```diff
@@ -0,0 +1,229 @@
...
+ * Definition of a set of URL params to store the position of the map in the URL. These params will
+ * change depending on the mode of rendering of the map (2D or 3D)
+ *
+ * In 2D it will generate lon/lat/z params.
+ *
+ * In 3D it will generate a unique camera URL param that will be a concat of all relevant camera
+ * values (x,y,z,yaw,pitch,roll)
+ *
+ * This param config class takes for granted that the store's flag showIn3d has been processed
+ * before it is then called to do its job. Meaning that if the flag has changed in the URL, but the
```
ltshb, line 81: How can this be guarantied that the showIn3d gets processed first ?

ltshb (commented):
While testing if I manually change the camera position in the url and refresh the values are lost ! Should not the camera position reflect the current position in 3d, or this will be done in a later PR ? 
Why not keeping the other lat, lon and z param in url in 3d mode and just ignore them ? It could make the code simpler and has the advantage that the user initial position in 2d would not be lost when switching to 3d, then refresh the browser and exiting the 3d view.

pakb:
> While testing if I manually change the camera position in the url and refresh the values are lost ! Should not the camera position reflect the current position in 3d, or this will be done in a later PR ? 

The Cesium component doesn't update the camera position yet, it will be done during the work on the 3D navigation Epic

> Why not keeping the other lat, lon and z param in url in 3d mode and just ignore them ? It could make the code simpler and has the advantage that the user initial position in 2d would not be lost when switching to 3d, then refresh the browser and exiting the 3d view.

That could be done, the only issue is that the old viewer already uses `lat` and `lon` in the case it is showing 3D. Maybe we can deal with that in the legacy URL param parser, and keep our `lat` and `lon` params (which do not represent the same thing as what `mf-geoadmin3` uses them for...)

src/router/storeSync/PositionUrlParamConfig.class.js
pakb: At first I removed it as I saw it wasn't used below, then I had issue at app startup where params were not parsed.
This was due to the fact that the router plugin uses `readQueryValue` to determine if a dispatch must occur (instead of using `valuesAreDifferentBetweenQueryAndStore(query, store)`, maybe a switch to this would be more logical?!)

src/router/storeSync/PositionUrlParamConfig.class.js
pakb: by adding AFTER the param parser for the 3D flag, see `src/router/storeSync/storeSync.config.js` line 31 and further

ltshb (approved):
👍🏼

PR #423, opened 2023-07-05
Add WMTS layer component for 3D

vladyslav-tk:


src/config.js
```diff
@@ -218,6 +218,13 @@ export const TILEGRID_EXTENT = [2420000, 1030000, 2900000, 1350000]
  */
 export const LV95_EXTENT = [572215.44, 5684416.96, 1277662.37, 6145307.4]
 
+/**
+ * TILEGRID_EXTENT (defined above) reprojected in EPSG:4326 through epsg.io website.
+ *
+ * @type {Number[]}
+ */
+export const WGS84_EXTENT = [5.1402988, 45.3981222, 11.4774363, 48.230617]
+
```
pakb, line 227: maybe we should rename this two so that it makes a little bit more sense that they are related.
what about :
- `TILEGRID_EXTENT_EPSG_3857`
- `TILEGRID_EXTENT_EPSG_4326`

So that it is clear that they derive from `TILEGRID_EXTENT`

src/modules/map/components/cesium/CesiumInternalLayer.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <div>
+        <CesiumWMTSLayer
+            v-if="layerConfig.type === LayerTypes.WMTS && !layerConfig.isExternal"
+            :wmts-layer-config="layerConfig"
+            :preview-year="previewYear"
+            :projection="WEBMERCATOR"
```
pakb, line 7: don't know if it would make sense to directly use the same logic as the one for OL, meaning :
```
:projection="layerConfig.isExternal ? WEBMERCATOR : LV95"
```
Some of our WMTS layers do not behave totally correctly when requested as `EPSG:3857` or `EPSG:4326`

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -11,25 +28,54 @@ import {
...
+    components: { CesiumInternalLayer },
     provide() {
         return {
             // sharing cesium viewer object with children components
             getViewer: () => this.viewer,
         }
     },
+    data() {
+        return {
+            viewerCreated: false,
```
pakb, line 49: is this really needed? As you create the Cesium instance in `beforeCreate()` lifecycle hook, it should always be set whenever the component is mounted, meaning the `v-if` above in the template should always be `true`

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,125 @@
...
+import { CoordinateSystem, WEBMERCATOR } from '@/utils/coordinateSystems'
+import { ImageryLayer, Rectangle, WebMapServiceImageryProvider } from 'cesium'
+import { WGS84_EXTENT } from '@/config'
+import { mapState } from 'vuex'
+import { YEAR_TO_DESCRIBE_ALL_OR_CURRENT_DATA } from '@/api/layers/LayerTimeConfigEntry.class'
+import GeoAdminWMSLayer from '@/api/layers/GeoAdminWMSLayer.class'
+import addImageryLayerMixins from '@/modules/map/components/cesium/utils/addImageryLayer-mixins'
+
+// todo should we adapt LOD to resolution?
+const MAXIMUM_LEVEL_OF_DETAILS = 18
```
pakb, line 17: What is the concept of LOD in Cesium, so that I understand what we could achieve by implementing what you suggest?

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,125 @@
...
+
+export default {
+    mixins: [addImageryLayerMixins],
+    props: {
+        wmsLayerConfig: {
+            type: [GeoAdminWMSLayer],
```
pakb, line 23: note sure it is correct that the type is wrapped in `[...]`?

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,125 @@
...
+                // if a time entry is defined, and is different from 'all'
+                // (no need to pass 'all' to our WMS, that's the default timestamp used under the hood)
+                if (
+                    this.wmsLayerConfig.timeConfig.currentYear !==
+                    YEAR_TO_DESCRIBE_ALL_OR_CURRENT_DATA
+                ) {
+                    return this.wmsLayerConfig.timeConfig.currentTimestamp
+                }
+            }
+            return ''
```
pakb, line 78: As this is essentially a copy past of `src/modules/map/components/openlayers/OpenLayersWMSLayer.vue` from line `65...86`, I would prefer that we move this logic in a small simple `wmsLayerUtils.js` in the `src/utils/` folder, or something in this direction.
If we want to change in the future the logic behind previewYear, it would not be good to have duplication of code such as this one.

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,125 @@
...
+            return this.wmsLayerConfig.opacity || 1.0
+        },
+        wmsVersion() {
+            return this.wmsLayerConfig.wmsVersion || '1.3.0'
+        },
+        format() {
+            return this.wmsLayerConfig.format || 'png'
+        },
+        url() {
+            return this.wmsLayerConfig.getURL(this.previewYear, this.projection.epsgNumber)
```
pakb, line 56: I think you can remove these two params, they were not removed when we worked on the preview year concept in the OL component 🙈

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,125 @@
...
+         * https://openlayers.org/en/latest/apidoc/module-ol_source_TileWMS-TileWMS.html and
+         * https://openlayers.org/en/latest/apidoc/module-ol_source_ImageWMS-ImageWMS.html have this
+         * option.
+         *
+         * If we let the URL have all the param beforehand (sending all URL param through the url
+         * option), most of our wanted params will be doubled, resulting in longer and more
+         * difficult to read URLs
+         *
+         * @returns Object
+         */
```
pakb, line 91: you might want to update or remove the docs here, as it's talking about OL stuff

src/modules/map/components/cesium/utils/addImageryLayer-mixins.js
```diff
@@ -0,0 +1,63 @@
+const addImageryLayerMixins = {
```
pakb, line 1: a little (big) docs here is wanted, to explain what is expected from a component using this mixin

You seem to be wanting a method instead of a property, is that something you ended up because just requiring `this.layer` was not working as intended? Otherwise I would prefer we stick to creating a `this.layer = new ...` in `beforeCreate()` hooks in the components using the mixin, and the mixin is responsible for mounting and unmounting it from Cesium (the component creating it is responsible for its deletion)

src/modules/map/components/cesium/CesiumWMTSLayer.vue
```diff
@@ -0,0 +1,79 @@
...
+            if (
+                this.previewYear &&
+                this.wmtsLayerConfig.timeConfig &&
+                this.wmtsLayerConfig.timeConfig.years.includes(this.previewYear)
+            ) {
+                return this.wmtsLayerConfig.timeConfig.getTimeEntryForYear(this.previewYear)
+                    .timestamp
+            }
+            return null
+        },
```
pakb, line 54: Maybe we should move this logic too, as it is the same as the one found in `src/modules/map/components/openlayers/OpenLayersWMTSLayer.vue`

I don't know if this is about the same as what is done in WMS, but if it is the case it could be a good opportunity to condense these two into one single function

src/config.js
vladyslav-tk: done

src/modules/map/components/cesium/CesiumInternalLayer.vue
vladyslav-tk: maybe we can investigate and try to use `WebMapServiceImageryProvider` instead of `UrlTemplateImageryProvider` if LV95 required for WMTS

src/modules/map/components/cesium/CesiumMap.vue
vladyslav-tk: yes, it is needed. We can't create a Cesium viewer in `beforeCreate` because Cesium requires a container for the viewer so Cesium instants are created when the component mounted

src/modules/map/components/cesium/CesiumWMSLayer.vue
vladyslav-tk: from Cesium docs:
> The maximum level-of-detail supported by the imagery provider, or undefined if there is no limit. If not specified, there is no limit.

In old geo.admin (https://github.com/geoadmin/mf-geoadmin3/blob/c21a537ea0f449279a7b715599d1ba64f601f718/src/components/map/LayersService.js#L424) maxResolution takes from layer config and then corresponding value takes from an array of LODs or 18 if not exists.

Should we implement something similar?

src/modules/map/components/cesium/CesiumWMSLayer.vue
vladyslav-tk: moved

src/modules/map/components/cesium/CesiumWMSLayer.vue
vladyslav-tk: yes, updated, thanks

src/modules/map/components/cesium/utils/addImageryLayer-mixins.js
vladyslav-tk: I added doc

I used the method because there are two places where the layer creates, when adding and when the URL changes (e.g. year changed). Also, WMTS and WMS are using different providers so we can't have it in mixin

src/utils/wmtsLayerUtils.js
```diff
@@ -0,0 +1,17 @@
...
+ */
+export function getTimestampForPreviewLayer(previewYear, wmtsLayerConfig) {
+    if (
+        previewYear &&
+        wmtsLayerConfig.timeConfig &&
+        wmtsLayerConfig.timeConfig.years.includes(previewYear)
+    ) {
+        return wmtsLayerConfig.timeConfig.getTimeEntryForYear(previewYear).timestamp
+    }
+    return null
```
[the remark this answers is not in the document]
vladyslav-tk: I merged this functions and moved it to `layerUtils`

src/utils/wmsLayerUtils.js
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function getWMSTimestampFromConfig(wmsLayerConfig) {
+    if (wmsLayerConfig.timeConfig) {
+        // if there is a preview year set, we search for the matching timestamp
+        if (this.previewYear) {
+            const matchingTimeEntry = wmsLayerConfig.getTimeEntryForYear(this.previewYear)
+            if (matchingTimeEntry) {
+                return matchingTimeEntry.timestamp
+            }
+        }
```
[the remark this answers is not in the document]
vladyslav-tk: I don't know how I didn't notice it :see_no_evil:

vladyslav-tk:
>You do not state that you require a createImagery(...) method, and its input/outputs
You are watching url in the mixin, but it is never stated that this is required by the mixin

I added this to the doc

> I don't know if this would align nicely with future GeoJSON and KML implementation, maybe you could require that the component using the mixin provides a Cesium.ImageryProvider() as this.provider and you could react to changes to this provider directly. Don't know if it would be better in regards to future GeoJSON and KML implementation?!

We can't use this mixin for GeoJSON and KML because it will be not imagery layers and zIndex, opacity, and URL will be handled differently. Maybe I will create one more mixin, more general

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -0,0 +1,104 @@
...
+                    subdomains: '0123',
+                    layers: this.wmsLayerConfig.geoAdminID,
+                    maximumLevel: MAXIMUM_LEVEL_OF_DETAILS,
+                    rectangle: Rectangle.fromDegrees(...TILEGRID_EXTENT_EPSG_4326),
+                }),
+                {
+                    show: this.wmsLayerConfig.visible,
+                }
+            )
+        },
```
pakb, line 99: `opacity` is not set here, meaning that if I change the opacity of the layer before making it visible, it is ignored

src/modules/map/components/cesium/CesiumWMTSLayer.vue
```diff
@@ -0,0 +1,72 @@
...
+                new UrlTemplateImageryProvider({
+                    rectangle: Rectangle.fromDegrees(...TILEGRID_EXTENT_EPSG_4326),
+                    maximumLevel: MAXIMUM_LEVEL_OF_DETAILS,
+                    url: url,
+                }),
+                {
+                    show: this.wmtsLayerConfig.visible,
+                }
+            )
+        },
```
pakb, line 67: `opacity` is not set here, meaning that if I change the opacity of the layer before making it visible, it is ignored

pakb (commented):
looking good! small issue with opacity to fix, and some unnecessary WMTS check in the utils function and we should be good to go!

src/modules/map/components/cesium/CesiumInternalLayer.vue
pakb: let's keep mercator there then, it's something we should investigate if we want to have LV95 projection for 3D in the future (I doubt it's a good idea anyway, for 3D at least)

src/modules/map/components/cesium/CesiumWMTSLayer.vue
vladyslav-tk: opacity will be applied correctly even if the layer is not visible but I added the `alpha` property anyway to be sure that it is always applied correctly

pakb (approved):

PR #424, opened 2023-07-06
BGDIINF_SB-2981: Add cesium-compass widget

fredj:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2981-add-cesium-compass/index.html)


pakb (commented):
looking good so far, I'll think about this "tool box" we discussed last time
it works also not too bad being at the bottom of the view

we should maybe hide it as soon as we are in mobile (there's a flag somewhere in `ui.store.js` that set this), and maybe even before (for tablets too) as I've not had the best experience with this compass on my iPhone.
The panning is correctly detected but then it never ends (`touchend` or `pointerup` is never caught)

fredj:
@pakb Done: the widget is now only displayed on desktop

PR #431, opened 2023-07-14
Feature : Adding test formatting command

ltkum:
Issue : There were no formatting command for tests

Fix : We created the command, included it into the format-lint command, then ran it.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-add-format-tests/index.html)


package.json
```diff
@@ -19,7 +19,8 @@
         "test:ci": "npm run test:unit && npm run test:e2e:ci",
         "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore",
 	"format": "prettier --write src",
-	"format-lint": "npm run format && npm run lint",
+	"format-test": "prettier --write tests",
+	"format-lint": "npm run format && npm run format-test && npm run lint",
```
pakb, line 23: I think you can add the `tests` folder directly into `format` rule, no need for a specific `format-test` rule

pakb: and you should maybe run prettier once on the package.json file directly, you are using tabs instead of spaces 😉

pakb (approved):
Thanks!
I've also added `lint` and `test:unit` as pre-commit hooks

PR #434, opened 2023-08-08
BGDIINF_SB-2927 : keep pin on search

ltkum:
Issue : When selecting a search option, the pin would then disappear, only remaining on preview.

Fix : The pin clear used to be done when the search selection was no longer selected. It's now called upon selecting a new search item, right before asking to pin the new location

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2927-add-marker-after-address-selection/index.html)


pakb (approved):
Sorry I overtook a bit your work, as I had some extensive plans of things to fix and a bit a refactoring in my head when I wrote this ticket. But I didn't write down everything in it 🤷

PR #441, opened 2023-08-21
BGDIINF_SB-3061 : keep where the camera is looking at when going back 2D

pakb:
and not where the camera is positioned, otherwise the view (top-down at the position of the camera) doesn't correspond at all with what the user was seeing in 3D

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3061_keep_camera_target_when_going_2d/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumMap.vue
+++ b/src/modules/map/components/cesium/CesiumMap.vue
@@ -78,10 +78,13 @@ import log from '@/utils/logging'
 import '@geoblocks/cesium-compass'
 import * as cesium from 'cesium'
 import {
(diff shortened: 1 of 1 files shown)


hansmannj (commented):
Nice!
Probably an edge case:
In 3D, when tilting the camera to a very low angle, it is possible that the camera position gets below the terrain and the user sees the "sky".
Tilting back to a more "normal" steeper angle, the camera position does not correspond with the initial position but can be completely off (i.e. even outside of CH).

ltshb (approved):
👍🏼

PR #442, opened 2023-09-08
BGDIINF_SB-3105 : layer's z-index rewrite

pakb:
In the process of calculating the z-index for group of layers, I've found some functionalities that were lacking precision or weren't well test covered or documented.

This includes changes in the layer module :
- Storing the background layer directly as a config object (instead of ID)
- Enabling the setting of the background layer, and visibility of layer, through the layer ID or the config object
- Reworking of addLayer to make sure it creates a clone of any given config object (the one in the args, or the one found in the config) so that the config stays pristine

I've also removed an unnecessary wait in the infobox Cypress test, that was sometime failing for unknown reasons on my local machine (might be also the case on the CI like on https://github.com/geoadmin/web-mapviewer/pull/441)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3105_layer_zindex_rewrite/index.html)

### DIFF
--- a/src/api/layers/ExternalGroupOfLayers.class.js
+++ b/src/api/layers/ExternalGroupOfLayers.class.js
@@ -1,0 +1,23 @@
+import ExternalLayer from '@/api/layers/ExternalLayer.class'
+import LayerTypes from '@/api/layers/LayerTypes.enum'
+
+/**
+ * Description of a group of layers, that could be added altogether or separately, that stems from a getCapabilities XML parsing. (see https://www.mediamaps.ch/ogc/schemas-xsdoc/sld/1.2/capabilities_1_3_0_xsd.html#Layer)
+ *
+ * If the group of layer is added to the map, all layers being part of it should be added under this group's name "banner"
+ */
+export default class ExternalGroupOfLayers extends ExternalLayer {
+    /**
+     * @param {String} name Name of this layer to be shown to the user
+     * @param {String} hostname getCapabilities URL host name, so that it can be used in the ID generation
+     * @param {ExternalLayer[]} layers Description of the layers being part of this group (they will all be displayed at the same time, in contrast to an aggregate layer)
+     */
+    constructor(name, hostname, ...layers) {
+        super(name, LayerTypes.GROUP, `${hostname}:${name.replaceAll(' ', '_')}`, null, 1, true)
+        this.layers = [...layers]
+    }
+
+    getID() {
+        return this.externalLayerId
+    }
+}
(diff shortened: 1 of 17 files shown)


src/store/modules/layers.store.js
```diff
@@ -135,27 +122,81 @@ const getters = {
         const layer = state.activeLayers.find((layer) => layer.getID() === layerId)
         return layer?.isExternal || (layer?.type === LayerTypes.KML && !layer?.adminId)
     },
+    /**
+     * Returns the z-index of a visible layer, taking into account the background layer(s) and/or
+     * if one of them is a group of layer (and thus all sub-layers are counted in the index)
+     *
+     * If the layer is not part of the visible layers (or is null or invalid), this will return -1 as a result
+     *
+     * @param {AbstractLayer} layerIdOrObject
```
ltshb, line 131: Should be `@param {AbstractLayer|string} layerIdOrObject` here.

src/store/modules/layers.store.js
```diff
@@ -135,27 +122,81 @@ const getters = {
...
+     * @returns {Number}
+     */
+    zIndexForVisibleLayer: (state, getters) => (layerIdOrObject) => {
+        let lookupId
+        if (layerIdOrObject instanceof AbstractLayer) {
+            lookupId = layerIdOrObject.getID()
+        } else {
+            lookupId = layerIdOrObject
+        }
```
ltshb, line 140: Might be wise to check for string type here and in case it is not a string to log an error, for debuging might be useful

src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -0,0 +1,23 @@
...
+     */
+    constructor(name, hostname, ...layers) {
+        super(name, LayerTypes.GROUP, `${hostname}:${name.replaceAll(' ', '_')}`, null, 1, true)
+        this.layers = [...layers]
+    }
+
+    getID() {
+        return this.externalLayerId
+    }
+}
```
ltshb, line 23: For which use case is this ExternalGroupOfLayers ? From a OOP stand point it is wrong to inherit from `ExternalLayer` here.

pakb: It's for the external layer UI tool, as some Kanton stores layers as group (we do not, AFAIK)
But you are correct that this could easily be renamed `GroupOfLayers` and extend from `AbstractLayer` (also using this kind of type for the internal layers). I might do that it's clearer this way.

ltshb: As discussed ok for now

ltshb (approved):

PR #443, opened 2023-09-11
BGDIINF_SB-2890 : LV95 as primary projection

pakb:
In order to facilitate the go-live, and as we are not yet ready with all services regarding mercator support, we will implement the necessary steps so that LV95 can be used as the projection for the mapping framework (and all services).

This PR is the representation of a work branch, and should not be merged as long as there are still work to do on its sub-tasks (as long as it is a draft...)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2890-lv95-pyramid/index.html)


ltshb:
This PR has been superseed by #470

PR #444, opened 2023-09-11
BGDIINF_SB-2890 : helper functions now takes projection as input

pakb:
preparation work to have the projection set in the store


src/utils/coordinateUtils.js
```diff
@@ -402,23 +402,24 @@ export const printHumanReadableCoordinates = (coordinates, coordinateSystem = LV
...
         return null
     }
     // checking that all coordinates are well-formed
     if (coordinates.find((coordinate) => coordinate.length !== 2)) {
         return null
     }
+    let bounds = LV95_BOUNDS
+    if (projection.epsg === WEBMERCATOR.epsg) {
+        bounds = LV95_IN_MERCATOR_BOUNDS
+    }
```
ltshb, line 419: Here I would use a switch case that handle both lv95 and webmercator projection and default to return null with eventually an error message to make it clear that the function only supports either lv95 or web mercator

ltshb: Maybe would be worse to also implement other projection like WGS84 anmd lv03

src/utils/coordinateUtils.js
```diff
@@ -431,26 +432,27 @@ export const splitIfOutOfLV95Bounds = (coordinates) => {
...
  * @param {Boolean} isFirstChunk
  * @returns {[CoordinatesChunk]}
  */
-function splitIfOutOfLV95BoundsRecurse(coordinates, previousChunks = [], isFirstChunk = true) {
+function splitIfOutOfLV95BoundsRecurse(
+    coordinates,
+    bounds,
+    previousChunks = [],
+    isFirstChunk = true
+) {
```
ltshb, line 445: This function seems to use `TILEGRID_EXTENT_EPSG_3857` at line 475 in a `clip` function, not sure if here we also needs some adaptation ?

ltshb: It should actually use the tile grid extent of the correct projection for the CH bound

pakb: I think we should maybe starts directly by having each projection system declares its extent, that would help readability and use cases. Instead of having `switch...case` clauses everywhere, I'll try going this way

src/api/profile/profile.api.js
```diff
@@ -79,9 +83,14 @@ async function getProfileDataForChunk(chunk, startingPoint, startingDist = 0) {
...
  */
-function chunksToLV95(chunks) {
+function chunksToLV95(chunks, fromProjection = WEBMERCATOR) {
+    if (fromProjection.epsg !== WEBMERCATOR.epsg) {
+        log.error('Unsupported projection system for profile API', fromProjection)
+        return []
+    }
     return chunks.map((chunk) => {
         chunk.coordinates = toLv95(chunk.coordinates, WEBMERCATOR.epsg)
         return chunk
```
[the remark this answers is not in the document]
pakb: I've reworked this part, it should now be able to take any type of coordinate in, processed them into LV95 to request the backend, and then reproject the result of the backend back to the wanted projection (if not LV95 already)

src/utils/coordinateUtils.js
pakb: Split is now made on the base of the CoordinateSystemBounds, and is therefor "projection agnostic"

src/utils/coordinates/CoordinateSystem.class.js
```diff
@@ -0,0 +1,63 @@
...
+     * projection's page
+     */
+    constructor(id, epsg, epsgNumber, label, proj4transformationMatrix, bounds = null) {
+        this.id = id
+        this.epsg = epsg
+        this.epsgNumber = epsgNumber
+        this.label = label
+        this.proj4transformationMatrix = proj4transformationMatrix
+        this.bounds = bounds
+    }
```
ltshb, line 26: Nice one 👍🏼

ltshb (approved):
Nice work

ltshb (approved):
👍🏼

PR #449, opened 2023-09-12
BGDIINF_SB-3071 : external layer providers list

vladyslav-tk:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3071_extarnal_layer_providers_list/index.html)


pakb (commented):
Could you run a format on this file or something similar (`npm run format` )? I think we should have 4 spaces per tab (instead of 2 here, see `.prettierrc`)

vladyslav-tk:
> Could you run a format on this file or something similar (`npm run format` )? I think we should have 4 spaces per tab (instead of 2 here, see `.prettierrc`)

done

ltshb (commented):
It seems to me that this PR is superseded by #450

PR #452, opened 2023-10-12
BGDIINF_SB-2430: Improved sharing embed previzualization

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2430-share-menu/index.html)

### DIFF
--- a/src/modules/menu/components/share/MenuShareEmbed.vue
+++ b/src/modules/menu/components/share/MenuShareEmbed.vue
@@ -39,14 +39,11 @@
             fluid
             @close="togglePreviewModal"
         >
-            <div class="embed-preview-modal">
-                <div
-                    class="d-flex embed-preview-modal-tools mb-1"
-                    :class="{ 'custom-size': isPreviewSizeCustom }"
-                >
+            <div class="embed-preview-modal" :style="embedPreviewModalWidth">
+                <div class="d-flex flex-row mb-2" :class="{ 'custom-size': isPreviewSizeCustom }">
                     <select
                         v-model="currentPreviewSize"
-                        class="embed-preview-modal-size-selector form-select w-auto"
+                        class="embed-preview-modal-size-selector form-select"
                         data-cy="menu-share-embed-iframe-size-selector"
                     >
                         <option
@@ -60,7 +57,7 @@
                         </option>
                     </select>
                     <MenuShareShortLinkInput
-                        class="flex-grow-1 ms-1"
+                        class="ms-1 flex-grow-1"
                         :with-text="false"
                         :small="false"
                         :short-link="iFrameLink"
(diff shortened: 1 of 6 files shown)


src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -197,6 +190,10 @@ export default {
         }
     },
     computed: {
+        embedPreviewModalWidth() {
+            // User the iframe with as maximal width for the embed preview modal window
```
pakb, line 194: there might be a couple misspelling in this comment, wouldn't it be more something like 
`Uses the iframe's width as maximal width for the entire modal window`?

src/modules/menu/components/share/MenuShareInputCopyButton.vue
```diff
@@ -1,52 +1,51 @@
 <template>
-    <div v-if="shortLink">
-        <label v-if="withText">{{ $t('share_link') }}: </label>
+    <div v-if="inputText">
+        <label v-if="labelText">{{ labelText }}: </label>
```
pakb, line 3: shouldn't you be running `labelText` through `$t(...)` as it was before? Only the i18n key selection was made dynamic, but I don't see that `$t(...)` is called somewhere anymore

src/modules/menu/components/share/MenuShareSection.vue
```diff
@@ -11,7 +11,13 @@
         <FontAwesomeIcon v-if="!shortLink" icon="spinner" spin size="2x" class="p-2" />
         <div v-if="shortLink" class="p-2">
             <MenuShareSocialNetworks :short-link="shortLink" class="pt-1" />
-            <MenuShareShortLinkInput :short-link="shortLink" class="px-2 py-3" />
+            <MenuShareInputCopyButton
+                :input-text="shortLink"
+                :label-text="$t('share_link')"
```
pakb, line 16: would be better, as stated above, to use the same logic as with `copyText` and `copiedText`, and pass only the i18n key for this prop too, letting the component manage the translation itself when it is needed

src/modules/menu/components/share/MenuShareInputCopyButton.vue
ltshb: The translation is done outside the component but you are right might be better to do it here for consistency

src/modules/menu/components/share/MenuShareSection.vue
ltshb: Done

src/modules/menu/components/share/MenuShareInputCopyButton.vue
ltshb: Done

src/modules/menu/components/share/MenuShareEmbed.vue
ltshb: Done

pakb (approved):

PR #455, opened 2023-10-13
BGDIINF_SB-2854 feature : Show area and length of drawings

ltkum:
BGDIINF_SB-2854

Issue : Are and length of polygons drawn were not shown

Fix : We now check if the polygon are closed. If they are, we calculate both area and perimeter. If not, we only calculate the perimeter. We then show them in the popup.

Still to do :
Font Awesome Icon for filled area is not loading properly, I need to check this

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-2854-calculat_area_closed_polygons/index.html)

### DIFF
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -30,7 +30,10 @@
                 <font-awesome-icon :icon="['far', 'square']" /> {{ length }}
             </div>
             <div v-if="isFeaturePolygon">
-                <font-awesome-icon :icon="['far', 'square']" style="color: #888a85" />
+                <font-awesome-icon
+                    :icon="['far', 'square']"
+                    style="background: #888a85; color: #888a85"
+                />
                 {{ area }}<sup>2</sup>
             </div>
         </div>
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -187,6 +187,12 @@ export default {
             return getLength(this.geometry)
         },
         isClosed() {
+            /*
+                The length parameter must be greater than 3, because the polygon has one point
+                twice : the first and last point are both existing in the same exact space.
+                A point would be length 2, a line would be length 3. We do not consider the
+                case where there are more than 3 points, but all in a single line.
+            */
             return (
                 this.feature.coordinates.length > 3 &&
                 this.feature.coordinates[0][0] ===
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -26,8 +26,14 @@
                 class="form-control"
                 rows="2"
             ></textarea>
+            <div v-if="isFeatureLine">
+                <font-awesome-icon :icon="['far', 'square']" /> {{ length }}
+            </div>
+            <div v-if="isFeaturePolygon">
+                <font-awesome-icon :icon="['fas', 'square']" style="color: #888a85" /> {{ area
+                }}<sup>2</sup>
+            </div>
         </div>
-
         <div class="d-flex">
             <SelectedFeatureProfile :feature="feature" />
 
@@ -101,6 +107,8 @@ import SelectedFeatureProfile from '@/modules/infobox/components/styling/Selecte
 import { allStylingColors, allStylingSizes } from '@/utils/featureStyleUtils'
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import { mapActions } from 'vuex'
+import { getArea, getLength } from 'ol/sphere.js'
+import { Polygon } from 'ol/geom'
 
 /**
  * Display a popup on the map when a drawing is selected.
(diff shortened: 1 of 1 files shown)


src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+                We don't have rings in this case, so we need to create an ol geometry
+                */
+
+                const geom = [this.feature.coordinates]
+
+                return new Polygon(geom)
+            },
+        },
+        isClosed: {
+            get() {
```
pakb, line 184: no need to declare a getter function if it's only used to get the information/the data, a get/set approach is when you want to link the result of the computed to a HTML Element, and have a handler/a callback when the user modify this value

Here you can simply write
```
isClosed() {
    return ...
}
```

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+                return (
+                    this.feature.coordinates[0][0] ===
+                        this.feature.coordinates[this.feature.coordinates.length - 1][0] &&
+                    this.feature.coordinates[0][1] ===
+                        this.feature.coordinates[this.feature.coordinates.length - 1][1]
```
pakb, line 189: you might want to check that `this.feature.coordinates.length > 1` in order to say that it is closed, otherwise a single point will be considered as closed too

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+                )
+            },
+        },
+        length: {
+            get() {
```
pakb, line 194: same remark as with `isClosed`, no need to declare a getter if the user will not modify this value

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+                if (length > 100) {
+                    output = Math.round((length / 1000) * 100) / 100 + ' ' + 'km'
+                } else {
+                    output = Math.round(length * 100) / 100 + ' ' + 'm'
+                }
+                return output
+            },
+        },
+        area: {
+            get() {
```
pakb, line 208: same as `isClosed`

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+                const length = getLength(this.geometry)
+
+                let output
+
+                if (length > 100) {
+                    output = Math.round((length / 1000) * 100) / 100 + ' ' + 'km'
+                } else {
+                    output = Math.round(length * 100) / 100 + ' ' + 'm'
+                }
+                return output
```
pakb, line 204: a bit of simplification could go a long way
```
// way above 
import { round } from '@/utils/numberUtils'

// then here
const length = getLength(this.geometry)
if (length > 100) {
    return `${round(length / 1000.0)} km`
}
return `${round(length)} km`
```
Should do about the same as what you have written

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +167,55 @@ export default {
...
+        area: {
+            get() {
+                const area = getArea(this.geometry)
+                let output
+                if (area > 10000) {
+                    output = Math.round((area / 1000000) * 100) / 100 + ' ' + 'km'
+                } else {
+                    output = Math.round(area * 100) / 100 + ' ' + 'm'
+                }
+                return output
```
pakb, line 216: could be simplified as above with `length`

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -171,6 +228,10 @@ export default {
         isFeatureMeasure() {
             return this.feature.featureType === EditableFeatureTypes.MEASURE
         },
+        isFeaturePolygon() {
+            console.log(this.isClosed)
```
pakb, line 232: please remove any console log 😉

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -26,8 +26,14 @@
                 class="form-control"
                 rows="2"
             ></textarea>
+            <div v-if="isFeatureLine">
+                <font-awesome-icon :icon="['far', 'square']" /> {{ length }}
+            </div>
+            <div v-if="isFeaturePolygon">
+                <font-awesome-icon :icon="['fas', 'square']" style="color: #888a85" /> {{ area
+                }}<sup>2</sup>
```
pakb, line 34: try to break the line after the `>` from FontAwesomeIcon, it's correct linter wise how it's done here, but it is not really easy to read

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -159,6 +168,41 @@ export default {
...
+            }
+            return `${round(length)} m`
+        },
+        area() {
+            const area = getArea(this.geometry)
+            if (area > 10000) {
+                return `${round(area / 1000000)} km`
+            }
+            return `${round(area)} m`
+        },
```
pakb, line 205: As you are already writing the `2` of `km2` and `m2` directly in the template, it might be better to move everything unit related there. Here it is a bit confusing to have a `area` props return a `m` or `km` number, it is the wrong unit.
So maybe it would be good to have `m` and `m2` in the template, and you only write the `k` whenever you reduce the size of the number by the corresponding factor. Or do something like

```
length() {
    return getLength(this.geometry)
}
lengthRounded() {
    if (this.length > 100) {
        return round(length / 1000.0)
    }
    return round(length)
}
// same for area
```
and in the template have something like
```
<div v-if="isFeatureLine">
    <font-awesome-icon :icon="['far', 'square']" /> {{ lengthRounded }}{{ length > 100 ? 'km' : 'm' }}
</div>
```

And don't forget that, in JS, if you divide by an integer number, the result of the division will be an integer too. If you want a float number result, you have to divide by a floating number (so you have to write the `.0` to the divisor)

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -167,54 +168,40 @@ export default {
...
-                return (
-                    this.feature.coordinates[0][0] ===
-                        this.feature.coordinates[this.feature.coordinates.length - 1][0] &&
-                    this.feature.coordinates[0][1] ===
-                        this.feature.coordinates[this.feature.coordinates.length - 1][1]
-                )
-            },
+        isClosed() {
+            return (
+                this.feature.coordinates.length > 3 &&
```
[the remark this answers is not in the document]
ltkum: a triangle has "4 points" because the last one is accounted for (We only remove the extra point in geodesic coordinates, not standard coordinates) . If I make this with greater than 2, it will give me the area of lines (which is 0, not a big deal, but still not an information we need)

I can put a small comment to make it clear in the code if you'd like

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -26,8 +26,18 @@
...
             ></textarea>
+            <div v-if="isFeatureLine">
+                <font-awesome-icon :icon="['far', 'square']" />
+                {{ lengthRounded }} {{ length > 100 ? 'km' : 'm' }}
+            </div>
+            <div v-if="isFeaturePolygon">
+                <font-awesome-icon
+                    :icon="['far', 'square']"
+                    style="background: #888a85; color: #888a85"
+                />
```
pakb, line 37: I would prefer that you use some kind of Bootstrap helper class here, `text-dark` or `bg-dark`, instead of hardcoding CSS values in the HTML.
If Bootstrap doesn't have the color you want, I would then prefer that you add a CSS class to this element, and style it under in the `<style>` portion of the Component

pakb (approved):
Small CSS/HTML mixing issue, and we are good to go

PR #459, opened 2023-10-17
BGDIINF_SB-2890: KML now stores only the geometries in standard KML element in WGS84

ltshb:
The editableFeature that is stored in the KML files doesn't have anymore the
coordinates to avoid projection issues. KML files have always coordinate in
WGS84 as specified in its official spec.

The editableFeature then has its coordinates in the corresponding application
projection. It gets its coordinates from the OL Feature geometries.

We use the OL KML parser properly in order to do the coordinates translation from
WGS84 into the application projection.

Due to these changes it appears that now the geometry parsed by the KML
parser have the 'xyz' layout while they had previously the 'xy' layout.


src/modules/drawing/lib/drawingUtils.js
```diff
@@ -27,12 +29,36 @@ export function wrapWebmercatorCoords(coords, inPlace = false) {
...
+        if (Array.isArray(coords[0])) {
+            coords.forEach((coords) => wrapXCoordinates(coords, projection, true))
+            return coords
+        } else {
+            return wrapX(coords, getProjection(projection.epsg))
+        }
+    } else {
+        return Array.isArray(coords[0])
+            ? coords.map((coords) => wrapXCoordinates(coords, projection, false))
+            : wrapX(coords.slice(), getProjection(projection.epsg))
```
pakb, line 61: this looks awfully the same as what is done above, with just a call to `slice()` before
Might be a good idea to structure this function so that we do things only once, such as (pseudo code)
```
let wrappedCoords = coords
if (!inPlace) {
    wrappedCoords = wrappedCoords.slice()
}
if (is array) {
    return wrappedCoords.map((coord) => wrapXCoordinates(coord, projection, inPlace))
}
return wrapX(wrappedCoords, getProjection(projection.epsg))
```

src/modules/map/components/cesium/CesiumKMLLayer.vue
```diff
@@ -38,26 +36,21 @@ export default {
...
                 if (features) {
                     this.olLayer.setSource(new VectorSource({ wrapX: true }))
-                    features.forEach((olFeature) => {
-                        EditableFeature.deserialize(olFeature, this.availableIconSets)
-                    })
                     // remove all old features first
                     this.olLayer.getSource().clear()
                     // add the deserialized features
                     this.olLayer.getSource().addFeatures(features)
-                    return WEBMERCATOR.epsg
```
pakb, line 60: is it really necessary that we return a String containing the EPSG? That sounds a bit weird

src/utils/__tests__/legacyKmlUtils.spec.js
```diff
@@ -79,7 +78,9 @@ function performStandardChecks(
     expectedFeatureType,
     expectedTitle = '',
     expectedDescription = '',
-    expectedCoordinateCount = 2
+    // TODO with the PR #X the ol geometries parsed from KML are with the layout 'xyz' instead
+    // of 'xy' therefore the expectedCoordinateCount has been increase to 3.
```
pakb, line 82: is that meant to be amended or removed? It is missing a proper PR number (or if not available, we could default back to a JIRA ticket ID)

tests/e2e-cypress/fixtures/html-popup.fixture.html
```diff
@@ -1,6 +1,9 @@
 <div class="chbabskulturgueter htmlpopup-container">
     <div class="htmlpopup-header">
-        <span>I'm a test window, I'm a test window, I'm a test window, I'm a test window, I'm a test window</span>
+        <span
+            >I'm a test window, I'm a test window, I'm a test window, I'm a test window, I'm a test
+            window</span
+        >
```
pakb, line 6: small detail, linter-formatter sometimes do that, but I would prefer that we do not wrap the tag itself, but just after the tag
```
<span>
...
</span>
```
It is then much more readable

src/api/features.api.js
```diff
@@ -107,6 +110,15 @@ export class SelectableFeature extends EventEmitter {
...
     }
 
+    /**
+     * Set the coordinates from the ol Feature
+     *
+     * @param {ol/Feature} olFeature Ol Feature to get the coordinate from
+     */
+    setCoordinates(olFeature) {
+        this.coordinates = extractOlFeatureCoordinates(olFeature)
+    }
```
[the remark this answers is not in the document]
ltshb: Good point, done

src/modules/drawing/lib/drawingUtils.js
```diff
@@ -27,12 +29,36 @@ export function wrapWebmercatorCoords(coords, inPlace = false) {
...
+ * @param {Array} coords The coordinates (or array of coordinates) to wrap
+ * @param {CoordinateSystem} projection Projection of the coordinates
+ * @param {boolean} inPlace If false, the original coordinates remain untouched and only a copy is modified
+ * @returns If "inPlace", then the same reference as "coords", else a reference to the modified copy
+ */
+export function wrapXCoordinates(coords, projection, inPlace = false) {
+    if (inPlace) {
+        if (Array.isArray(coords[0])) {
+            coords.forEach((coords) => wrapXCoordinates(coords, projection, true))
+            return coords
```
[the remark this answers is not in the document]
ltshb: I actually copied this function from `wrapWebmercatorCoords` and did only changed the minimum required, but you are right as I'm touching it I'll improve it at the same time.

src/modules/drawing/lib/drawingUtils.js
ltshb: This makes the code much easier to understand, thanks. I've also added unit test for this code for checking the inPlace and not inPlace.

src/modules/map/components/cesium/CesiumKMLLayer.vue
ltshb: Good question I did not check

src/utils/__tests__/legacyKmlUtils.spec.js
ltshb: I'll remove it, I think the changes of the coordinates from 'xy' to 'xyz' comes from the fact that we were overwriting the feature geometry coordinate from the one in the EditableFeature, while now we do the oposite.

tests/e2e-cypress/fixtures/html-popup.fixture.html
ltshb: Prettier has a very good reason for doing this, see https://prettier.io/blog/2018/11/07/1.15.0#whitespace-sensitive-formatting.
In this test fixture case I don't know if this is relevant though ?

src/modules/map/components/cesium/CesiumKMLLayer.vue
ltshb: It is indeed used and needed.

pakb (approved):

PR #463, opened 2023-10-18
BGDIINF_SB-2890 : add center as new URL param (replaces lat lon)

pakb:


src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -58,9 +57,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
             case 'zoom':
                 value = translateSwisstopoPyramidZoomToMercatorZoom(legacyParams[param])
                 if (!value) {
-                    // if the value is not defined in the old zoom system, we use the 'default' zoom level
-                    // of 8 (which will roughly show the whole territory of Switzerland)
-                    value = 8
+                    value = DEFAULT_PROJECTION.defaultZoom
```
ltshb, line 60: Should we not here translate the zoom to the correct projection based on the store `projection` value ? Also at line 58 we use mercator but line 60 might be lv95, depending on the default projection.

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -109,12 +106,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
         // if a legacy coordinate (x,y or N,E) was used, we need to guess the SRS used (either LV95 or LV03)
         // and covert it back to EPSG:4326 (Mercator)
         if (legacyCoordinates.length === 2 && legacyCoordinates[0] && legacyCoordinates[1]) {
-            const center = reprojectUnknownSrsCoordsToWGS84(
-                legacyCoordinates[0],
-                legacyCoordinates[1]
-            )
-            newQuery['lon'] = round(center[0], 6)
-            newQuery['lat'] = round(center[1], 6)
+            newQuery['center'] = legacyCoordinates.join(',')
```
ltshb, line 109: The comment is not up to date here

src/router/legacyPermalinkManagement.routerPlugin.js
pakb: you are totally right, and I'm working on that for a different PR right now as it's a bit extensive how much code must change to accommodate that...

ltshb (approved):

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -46,16 +48,21 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
...
+                // if we are using WebMercator as the map projection we need te re-evaluate LV95 zoom,
+                // as it was a zoom level tailor-made for this projection (and not made to cover the whole globe)
+                if (store.state.position.projection.epsg === WEBMERCATOR.epsg) {
+                    value = translateSwisstopoPyramidZoomToMercatorZoom(legacyParams[param])
+                } else {
+                    value = legacyParams[param]
+                }
                 if (!value) {
                     value = DEFAULT_PROJECTION.defaultZoom
                 }
```
ltshb, line 68: I guess this `if` could be removed due to previous `else`

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -103,9 +117,25 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
...
+            if (legacyCoordinatesAreExpressedInWGS84) {
+                legacyCoordinates = proj4(
+                    WGS84.epsg,
+                    store.state.position.projection.epsg,
+                    legacyCoordinates
+                )
+            }
+            // if the current projection is not LV95, we also need to re-project x/y or N/E
+            // (the legacy viewer was always writing coordinates in LV95 in the URL)
+            if (store.state.position.projection.epsg !== LV95.epsg) {
```
ltshb, line 132: Should not be here `else if` ?

PR #466, opened 2023-10-18
BGDIINF_SB-2890 : rework how Height and Search API handle projections

pakb:
they now "talk" only in LV95 with the backend, and re-project as needed the output to the required mapping projection

### DIFF
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -1,8 +1,9 @@
-import { API_SERVICE_SEARCH_BASE_URL } from '@/config'
-import { WEBMERCATOR } from '@/utils/coordinates/coordinateSystems'
+import { API_SERVICE_SEARCH_BASE_URL, DEFAULT_PROJECTION } from '@/config'
+import { LV95, WEBMERCATOR } from '@/utils/coordinates/coordinateSystems'
 import log from '@/utils/logging'
 import { translateSwisstopoPyramidZoomToMercatorZoom } from '@/utils/zoomLevelUtils'
 import axios from 'axios'
+import proj4 from 'proj4'
 
 // API file that covers the backend endpoint http://api3.geo.admin.ch/services/sdiservices.html#search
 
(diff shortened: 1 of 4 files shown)


src/api/search.api.js
```diff
@@ -180,23 +182,32 @@ async function search(queryString = '', lang = '') {
...
+            let coordinate = []
+            let zoom = location.attrs.zoomlevel
             if (location.attrs.x && location.attrs.y) {
                 coordinate.push(location.attrs.x)
                 coordinate.push(location.attrs.y)
             }
+            if (projection.epsg !== LV95.epsg) {
+                // re-projecting result coordinate and zoom to wanted projection
+                coordinate = proj4(LV95.epsg, projection.epsg, coordinate)
+                zoom = translateSwisstopoPyramidZoomToMercatorZoom(zoom)
```
ltshb, line 195: This should not be done only for mercator ?

src/modules/map/components/LocationPopup.vue
```diff
@@ -231,14 +230,14 @@ export default {
...
                 log.error(`Failed to get position height`)
                 this.height = null
             }
         },
         updateShareLink(coordinate, routeQuery) {
-            let [lon, lat] = reproject(this.projection.epsg, WGS84.epsg, coordinate)
+            let [lon, lat] = proj4(this.projection.epsg, WGS84.epsg, coordinate)
             let query = {
                 ...routeQuery,
                 crosshair: 'marker',
```
ltshb, line 243: This with your #463 PR won't be anymore valid, maybe rebase this on top of the other and use center instead of lon lat.

src/api/search.api.js
```diff
@@ -123,9 +124,10 @@ let cancelToken = null
 /**
  * @param {String} queryString The query string that describe what is wanted from the search
  * @param {String} lang The lang ISO code in which the search must be conducted
+ * @param {CoordinateSystem} projection the projection in which the search results must be returned
  * @returns {Promise<CombinedSearchResults>}
  */
-async function search(queryString = '', lang = '') {
+async function search(queryString = '', lang = '', projection = DEFAULT_PROJECTION) {
```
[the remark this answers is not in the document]
pakb: I will rename it `outputProjection` to make things clearer

src/api/search.api.js
pakb: Any output projection that is not LV95 should be reprojected, I'm thinking on a way to link the zoom calculation to the CoordinateSystem, so that we can make it flexible too

ltshb (approved):

PR #471, opened 2023-10-19
BGDIINF_SB-2890: better handling of zoom conversion

pakb:
between Swiss and non-Swiss projections, fixing the Zoom to extent functionality in the process
Most, if not all, zoom related utils function and constants have been moved to the zoomLevelUtils.js file

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2890_fix_zoom_to_extent/index.html)


src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -57,7 +57,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
             case 'zoom':
                 // if we are using WebMercator as the map projection we need te re-evaluate LV95 zoom,
                 // as it was a zoom level tailor-made for this projection (and not made to cover the whole globe)
-                if (store.state.position.projection.epsg === WEBMERCATOR.epsg) {
+                if (!store.state.position.projection.isSwissProjection) {
                     value = translateSwisstopoPyramidZoomToMercatorZoom(legacyParams[param])
                 } else {
                     value = legacyParams[param]
```
ltshb, line 63: Just looking at this diff, it looks wrong ! 
For future it may be wise to have a switch case here and only translate in projection that we support and in default case do a log error.

```javascript
switch(store.state.position.projection.eps) {
  case WEBMERCATOR.epsg:
    value = translateSwisstopoPyramidZoomToMercatorZoom(legacyParams[param])
    break
  case LV95.epsg:
  case LV03.epsg: // not sure this would be correct though ?
    value = legacyParams[param]
    break 
  default:
    log.error(`Unsupported projection`)
    throw Error()
}
```

src/utils/coordinates/CoordinateSystem.class.js
```diff
@@ -22,13 +22,16 @@ export default class CoordinateSystem {
      *   re-projections.
      * @param {[Number, Number]} defaultCenter Coordinate to be used as the default center, will use
      *   the center of the bounds if none are given (and bounds are defined)
+     * @param {Boolean} isSwissProjection If this projection is made specifically to cover
+     *   Switzerland (if it uses our national LV95 or LV03 bounds, zoom level, etc...)
```
ltshb, line 26: I wounder if it would not make the code easier to understand if the `isSwissProjection` would be set in the constructor based on the `epsgNumber` when searching for it in the code it would be directly clear that it is only for lv03 and lv95 even though the name say it.

src/store/modules/search.store.js
```diff
@@ -53,12 +53,12 @@ const actions = {
             const coordinate = coordinateFromString(query)
             if (coordinate) {
                 dispatch('setCenter', coordinate)
-                dispatch('setZoom', ZOOM_LEVEL_1_25000_MAP)
+                dispatch('setZoom', MERCATOR_ZOOM_LEVEL_1_25000_MAP)
```
ltshb, line 56: Why search is setting the zoom to web mercator ? Should it no set it depending on the current projecion ?

src/store/modules/search.store.js
```diff
@@ -53,12 +53,12 @@ const actions = {
...
                 dispatch('setPinnedLocation', coordinate)
             } else if (isWhat3WordsString(query)) {
                 retrieveWhat3WordsLocation(query).then((what3wordLocation) => {
                     dispatch('setCenter', what3wordLocation)
-                    dispatch('setZoom', ZOOM_LEVEL_1_25000_MAP)
+                    dispatch('setZoom', MERCATOR_ZOOM_LEVEL_1_25000_MAP)
```
ltshb, line 61: Dito

src/store/plugins/reproject-everything-on-projection-change.plugin.js
```diff
@@ -77,7 +76,7 @@ const reprojectEverythingOnProjectionChangePlugin = (store) => {
                     store.dispatch('setSelectedFeatures', reprojectedSelectedFeatures)
                 })
 
-                if (newProjection.epsg === LV95.epsg) {
+                if (newProjection.isSwissProjection) {
                     store.dispatch(
                         'setZoom',
                         translateMercatorZoomToSwisstopoPyramidZoom(state.position.zoom)
```
ltshb, line 82: Here it may be wise to have a new function from the zoom utilities called `translateZoom(fromProjection, toProjection, zoom)` and that this method will call the correct `translate...` method or throw and error if we don't support the translation.
So for the zoom here we would have a very short code 

```javascript
store.dispatch('setZoom', translateZoom(oldProjection.epsg, newProjection.epsg, state.position.zoom))
```

src/utils/coordinates/CoordinateSystem.class.js
ltshb: See final comment about refactoring, this could be then part of `SwissCoordinateSystem` or `Lv95CoordinateSystem`

ltshb (commented):
Ok after a call with @pakb here what we decided to make the code easier to understand and to enhanced with other coordinate systems:

1. Add a mandatory abstract method in `CoordinateSystem` class that would translate the zoom level. The method could be then named for example: `getZoomFromResolution(resolution, coordinate)`. I personally would provide the whole coordinate even if for the moment only the latitude is needed for mercator.
2. Use class inheritance for the coordinate system specialization `class Lv95 extends CoordinateSystem`. This specialization would implement the `getZoomFromResolution`
3. Add also a `getResolutionFromZoom` in `CoordinateSystem` this would remove the getters `resolution` in position.store.js

It may be a good idea to first merge this PR with just some minor changes before this big refactoring, and do the big refactoring in separate PR.

ltshb:
@pakb  Passing to 3d view crash the application

ltshb (commented):
You can (either by scrolling, clicking on zoom or entering in the url) go up to zoom 13, however you got only tiles up to zoom 9.4 above you have either empty tiles or failures from backend.

pakb:
There was a missing resolution, I forogt one in the process of moving them from config.js to the dedicated projection file somehow 🙈

src/utils/coordinates/WebMercatorCoordinateSystem.class.js
```diff
@@ -0,0 +1,82 @@
...
+    constructor() {
+        super(
+            3857,
+            'WebMercator',
+            '+proj=merc +a=6378137 +b=6378137 +lat_ts=0.0 +lon_0=0.0 +x_0=0.0 +y_0=0 +k=1.0 +units=m +nadgrids=@null +wktext  +no_defs',
+            new CoordinateSystemBounds(
+                -20037508.34,
+                20037508.34,
+                -20048966.1,
+                20048966.1,
```
ltshb, line 19: Comment about the origin of those numbers would be nice

src/utils/coordinates/WGS84CoordinateSystem.class.js
```diff
@@ -0,0 +1,77 @@
...
+            4326,
+            'WGS 84 (lat/lon)',
+            '+proj=longlat +datum=WGS84 +no_defs +type=crs',
+            new CoordinateSystemBounds(
+                -180.0,
+                180.0,
+                -90.0,
+                90.0,
+                // center of LV95's extent transformed with epsg.io website
+                [8.239436, 46.832259]
```
ltshb, line 19: Might be better to use `proj4` to dynamically re-project the LV95 extent here ?

ltshb: Or even better make this configurable in the config.js, this way if someone want to use our soft and want to center on france for example he only needs to change its config.js file and not the implementation.

src/utils/coordinates/WebMercatorCoordinateSystem.class.js
```diff
@@ -0,0 +1,82 @@
...
+                // center of LV95's extent transformed with epsg.io website
+                [917209.87, 5914737.43]
```
ltshb, line 21: Same comment as for WGS84

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -253,7 +247,9 @@ export default {
 
         this.viewerCreated = true
 
-        this.viewer.scene.postRender.addEventListener(limitCameraCenter(TILEGRID_EXTENT_EPSG_4326))
+        this.viewer.scene.postRender.addEventListener(
+            limitCameraCenter(LV95.getBoundsAs(WGS84).flatten)
+        )
```
ltshb, line 252: Similar comment as down below, would it make sense to have config value for the default bound/center ? To avoid hardcoding it here, so our default bounds and center is CH but for other it might something else. So could we have something like this

```javascript
// CesiumMap.vue
limitCameraCenter(DEFAULT_PROJECTION.getBoundsAs(WGS84).flatten)
```

Maybe then also add a comment in the config.js that the default projection is also used as default bound and center.

src/modules/map/components/cesium/CesiumWMSLayer.vue
```diff
@@ -89,7 +90,7 @@ export default {
                     subdomains: '0123',
                     layers: this.wmsLayerConfig.geoAdminID,
                     maximumLevel: MAXIMUM_LEVEL_OF_DETAILS,
-                    rectangle: Rectangle.fromDegrees(...TILEGRID_EXTENT_EPSG_4326),
+                    rectangle: Rectangle.fromDegrees(...LV95.getBoundsAs(WGS84).flatten),
```
ltshb, line 93: Same as above using a default projection configuration

src/modules/map/components/cesium/CesiumWMTSLayer.vue
```diff
@@ -55,7 +56,7 @@ export default {
         createImagery(url) {
             return new ImageryLayer(
                 new UrlTemplateImageryProvider({
-                    rectangle: Rectangle.fromDegrees(...TILEGRID_EXTENT_EPSG_4326),
+                    rectangle: Rectangle.fromDegrees(...LV95.getBoundsAs(WGS84).flatten),
```
ltshb, line 59: Dito

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -141,31 +133,13 @@ export default {
...
-                        ),
-                    })
-                )
+            if (this.projection instanceof CustomCoordinateSystem) {
+                source.tileGrid = new TileGrid({
+                    resolutions: this.projection.getResolutions(),
+                    extent: this.projection.bounds.flatten,
+                    origin: this.projection.getTileOrigin(),
+                    tileSize: WMS_TILE_SIZE,
+                })
```
ltshb, line 142: Very nice code simplification 👍🏼

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -57,8 +59,11 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
             case 'zoom':
                 // if we are using WebMercator as the map projection we need te re-evaluate LV95 zoom,
                 // as it was a zoom level tailor-made for this projection (and not made to cover the whole globe)
-                if (store.state.position.projection.epsg === WEBMERCATOR.epsg) {
-                    value = translateSwisstopoPyramidZoomToMercatorZoom(legacyParams[param])
+                if (!(projection instanceof SwissCoordinateSystem)) {
+                    value = LV95.transformCustomZoomLevelToStandard(legacyParams[param])
+                    if (projection instanceof CustomCoordinateSystem) {
+                        value = projection.transformStandardZoomLevelToCustom(value)
+                    }
```
ltshb, line 66: Nice one but the comment needs to be updated. The code is almost self explaining and I think that just a comment that `legacyParams['zoom']` is always in LV95 would be sufficient.

ltshb (approved):
Great works, looks much cleaner.

I have only a few minors comments/suggestions.

src/utils/coordinates/WGS84CoordinateSystem.class.js
pakb: I think this needs some kind of startup rework, as proj4 is not yet configured when we instantiate these coordinate systems.
Plus it would not be super robust, as if I start the app with `.../?sr=3857` without specifying a center, it would not reproject the default center automatically (so it needs some more thoughts)
But I like this idea!

src/modules/map/components/cesium/CesiumMap.vue
pakb: maybe we should do that only when the `DEFAULT_PROJECTION` is a `CustomCoordinateSystem`, as it would normally be the case that such a system will shrink its size to fit some part of the World.

PR #474, opened 2023-10-19
Fixed time slider toggling and button visibility

ltshb:
When only one time config layer was visible and you entered the time preview
without then changing any preview year, if you changed the year in the active
layer config the timeslider was not toggled (made invisible). This was only
the case with one layer visible at a time and when we did not changed the
preview years.

The bug was that in that case the previewYear in the store was not changed
and the menu module could not now of this change and react on it.

The time slider button was visible as soon as at least one layer was visible
even if this layer did not had any time configuration.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-timeslider/index.html)


src/store/modules/layers.store.js
```diff
@@ -110,6 +110,14 @@ const getters = {
     getActiveLayerById: (state) => (geoAdminLayerId) =>
         state.activeLayers.find((layer) => layer.getID() === geoAdminLayerId),
 
+    /**
+     * Get layers with time config
+     *
+     * @returns {GeoAdminLayer[]} List of layers with time config
+     */
+    visibleLayersWithTimeConfig: (state, getters) =>
+        getters.visibleLayers.filter((layer) => layer.timeConfig.timeEntries.length),
```
pakb, line 119: I'm not 100% all layers will have a `timeConfig`, maybe a little `?` in the mix could help
```
getters.visibleLayers.filter((layer) => layer?.timeConfig.timeEntries.length)
```

ltshb: Good point, I changed it to be on the safe side

pakb (approved):

PR #470, opened 2023-10-19
Default projection to lv95 

ltshb:
Relates to ticket BGDIINF_SB-2890

TODO:

- [x] drawing e2e tests failing
  - drawing/geodesicDrawing.cy.js ltshb
    - #483
    - #485 
    - #488
  - drawing/share.cy.js @ltshb 
    - #511 
- [x] geolocation e2e tests failing (ltshb)
    - #506 
- [x] infobox.cy.js tests failing (@pakb)
   - #513 
- [x] 3d tests failing (@pakb)
   - #512 
- [x] search e2e tsts failing (@pakb)
  - #498 
- [x] legacyParamImport.cy.js tests failing (@ltshb)
   - #509 
- [x] mouseposition.cy.js tests failing
   - #510
- [x] Add geodesie in lv95 e2e tests (currently only web mercator is tested) 
  - Will be done later with BGDIINF_SB-3160
- [x] 3d -> 2d goes to nirvana
  - #482
- [x] drawing line and measure profile mouse tracking
  - #477 
- [x] Zoom to extent doesn't work anymore (zoom is assumed to be mercator only) :
  - #471
- [x] Deactivate buttons in drawing mode 
  - #469 
  - #480
- [x] Offset with WMTS tiles when app is in WebMercator
  - A first analyze shows that on `develop` branch we don't have the issue but their we use the WMTS service with LV95 and do a reproject on the viewer side to Mercator, while now on `develop-lv95` we use directly the WMTS service with Mercator ! So it might be an issue with our WMTS backend. See ticket BGDIINF_SB-3164
- [x] Remove lightbasemap bg under pixelkarte bg in webmercator
  - #505 
- [x] In lv95 do not allow to move the view outside of CH otherwise it is hard to go back to CH
  - will be done later on with BGDIINF_SB-3161
- [x] geolocation
  - #495
- [x] Check that `zoomToExtent` is working properly after #476 is merged on `develop` and we've rebased this branch
  - #494 
  - `zoomToExtent` is still not working no changes between `develop` and `develop-lv95`. We need to fix it later on within BGDIINF_SB-3153
  -

TODO after merge on develop
  - https://jira.swisstopo.ch/browse/BGDIINF_SB-3166

[Test link](https://sys-map.dev.bgdi.ch/preview/develop-lv95/index.html)

PR #476, opened 2023-10-19
BGDIINF_SB-3084 : GUI for import results

vladyslav-tk:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3084-import-result-list/index.html)

### DIFF
--- a/src/api/layers/ExternalGroupOfLayers.class.js
+++ b/src/api/layers/ExternalGroupOfLayers.class.js
@@ -16,9 +16,19 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
      *   generation
      * @param {ExternalLayer[]} layers Description of the layers being part of this group (they will
      *   all be displayed at the same time, in contrast to an aggregate layer)
+     * @param {String} abstract Abstract of this layer to be shown to the user
      */
-    constructor(name, hostname, layers) {
-        super(name, LayerTypes.GROUP, `${hostname}:${name.replaceAll(' ', '_')}`, null, 1, true)
+    constructor(name, hostname, layers, abstract = '') {
+        super(
+            name,
+            LayerTypes.GROUP,
+            `${hostname}:${name.replaceAll(' ', '_')}`,
+            null,
+            1,
+            true,
+            [],
+            abstract
+        )
         this.layers = [...layers]
     }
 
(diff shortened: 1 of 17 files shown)


vladyslav-tk:
> 1. When opening the import you have a the place reserved for the result list, might be good to have it responsive and only have the place once you got the results.

then how dropdown should behave? Should we show the list at the top?

Screenshot 2023-10-24 at 13 55 23



> 2. Changing the language doesn't update the result list, would be good to have it updated (aka reloading the external url).

I will have a look

vladyslav-tk:
> 2. Changing the language doesn't update the result list, would be good to have it updated (aka reloading the external url).

done, list reloads on language change

ltshb:
@vladyslav-tk very good point about the dropdown. On the old viewer as the import toolbox is an overlay, the dropdown overflow the box what we cannot do here as it is in the footer. However if you look at this print screen the dropdown is not taking the whole space
image
We could reduce the white space to match the drowpdown list and increase it when the result is displayed, but this could be a bit strange/odd behaviour on the infox box on the footer as it will move the input and connect button higher up. On the old viewer as floating tooltip it looks more natural.
If you can do this behavior easily without spending much time, maybe it is worth a try to see how it behave from a user perspective, otherwise we can leave it as it is now. Thanks

ltshb (approved):
Very nice, Thanks a lot.

Just one last small issue with the `zoomToExtent`, for the `https://wms.geo.admin.ch` it is then not zoomed to the center of switzerland (comparing to the oldviewer where we can see the whole switzerland in the middle of the view). But as this is probably related to a projection issue and that the whole projection stuff is being reworked in #470 I would not spend time now to try to solve it but I would open a ticket to do it later on, @pakb what do you think.

pakb:
> what do you think.

Let's open a little line of bug/task on our PR so that we do not forget to address this on our projection work

ltshb:
@pakb @vladyslav-tk I created a ticket for the zoom to extent bug that we do it later on, see BGDIINF_SB-3153

vladyslav-tk:
>  We could reduce the white space to match the drowpdown list and increase it when the result is displayed, but this could be a bit strange/odd behaviour on the infox box on the footer as it will move the input and connect button higher up. On the old viewer as floating tooltip it looks more natural. If you can do this behavior easily without spending much time, maybe it is worth a try to see how it behave from a user perspective, otherwise we can leave it as it is now. Thanks

done

PR #477, opened 2023-10-19
BGDIINF_SB-2890: Fixed elevation profile tracking in lv95

ltshb:
Fixed the profile tracking when hovering the profile, the point is tracked on the map. This has been brocken when passing to lv95.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2890-profile-tracking/index.html)

### DIFF
--- a/src/api/features.api.js
+++ b/src/api/features.api.js
@@ -1,7 +1,6 @@
 import { DrawingIcon, DrawingIconSet } from '@/api/icon.api'
 import { API_BASE_URL } from '@/config'
 import { featureStyleFunction } from '@/modules/drawing/lib/style'
-import { WEBMERCATOR } from '@/utils/coordinates/coordinateSystems'
 import EventEmitter from '@/utils/EventEmitter.class'
 import {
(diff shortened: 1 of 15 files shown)


src/api/features.api.js
```diff
@@ -572,7 +572,7 @@ export const identify = (
  *   be expressed
  * @returns {Promise<LayerFeature>}
  */
-const getFeature = (layer, featureID, lang = 'en', projection = WEBMERCATOR) => {
+const getFeature = (layer, featureID, lang = 'en', projection) => {
```
pakb, line 575: It would be better to swap `lang` and `projection` if we remove the default value, it is unadvised to have non-default parameters declared after one that have default value

pakb (approved):
Minor details, and the comment on what we've talked about in our last call, and we are good to go! Thanks!

src/api/features.api.js
ltshb: you are right I'll change that

PR #484, opened 2023-10-24
BGDIINF_SB-3107: store the compare ratio

ltkum:
added a compare ratio in the mapviewer store, to keep track of where the compare slider should be
by default the value is -1, and it stops being visible when it's not between 0 and 1
added an url parameter to set it to a specific value, which will also update the url when we move the slider
added the legacy parameter support for swipe_ratio, converting it to compare_ratio

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-3107-store-compare-ratio/index.html)


src/modules/map/MapModule.vue
```diff
@@ -44,7 +44,7 @@ const isPhoneMode = computed(() => uiMode.value === UIModes.PHONE)
             <teleport to="#toolbox-compass-button">
                 <OpenLayersCompassButton />
             </teleport>
-            <CompareSlider />
+            <CompareSlider v-if="isCompareSliderActive" />
```
pakb, line 47: this `isCompareSliderActive` isn't defined above, maybe you might want to add the store binding too

src/store/modules/ui.store.js
```diff
@@ -103,6 +104,15 @@ export default {
          * @type Boolean
          */
         importFile: false,
+
+        importOverlay: false,
```
pakb, line 108: that might be some rebase left-over, but I think it is meant to be deleted

src/store/modules/ui.store.js
```diff
@@ -11,7 +11,8 @@ export const UIModes = {
     DESKTOP: 'DESKTOP', // formerly called "MENU_ALWAYS_OPEN", also used for tablets
     PHONE: 'PHONE', //  formerly called "MENU_OPENED_THROUGH_BUTTON"
 }
-
+// TODO : faire un commentaire
```
pakb, line 14: please add the wanted comment here

src/modules/map/components/CompareSlider.vue
pakb, line 1: Could you transform this into a Composition API component?

src/modules/map/MapModule.vue
```diff
@@ -20,7 +20,13 @@ const store = useStore()
 const is3DActive = computed(() => store.state.cesium.active)
 const uiMode = computed(() => store.state.ui.mode)
 const displayLocationPopup = computed(() => store.state.map.displayLocationPopup)
-
+const isCompareSliderActive = computed(() => {
+    if (store.state.ui.compareRatio > 0 && store.state.ui.compareRatio < 1) {
+        true
+    } else {
+        false
+    }
```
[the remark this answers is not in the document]
ltkum: I was dumb, I'll get a coffee and correct this

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,35 +1,71 @@
...
+function grabSlider() {
+    window.addEventListener('mousemove', listenToMouseMove, { passive: true })
+    window.addEventListener('touchmove', listenToMouseMove, { passive: true })
+    window.addEventListener('mouseup', releaseSlider, { passive: true })
+    window.addEventListener('touchend', releaseSlider, { passive: true })
+}
+
+function listenToMouseMove(event) {
+    const currentPosition = event.type === 'touchmove' ? event.touches[0].pageX : event.pageX
+    compareRatio.value = currentPosition / clientWidth.value
```
pakb, line 40: would be better to `round` to 2 decimal here (from `numberUtils.js`), the value in the URL can be very long otherwise

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,35 +1,71 @@
...
+watch(storeCompareRatio, (newValue) => {
+    compareRatio.value = newValue
+})
+
+onMounted(() => {
+    compareRatio.value = storeCompareRatio.value
+})
+function generateCSSPosition() {
+    return { left: compareSliderPosition.value }
+}
```
pakb, line 30: `compareSliderPosition` is only used to create this CSS style object, maybe we could create this object directly in the computed `compareSliderPosition` and remove this function.

PR #487, opened 2023-10-24
Updated timeout in Cypress config

ltshb:
Reduced the timeout in order to speed tests when they are failings

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-ci-timeout-retry/index.html)


pakb (approved):
I'm not sure we might not have some issue with bigger tests, such as the geodesic tests.
But if we do not encounter flakiness, that would improve build time when some tests are failing

PR #489, opened 2023-10-25
BGDIINF_SB-2890 : fix search tests

pakb:
search now uses lat/lon from the backend, as there was some issue while using x/y with sr=2056. Might be a bug on the backend, but those coordinates were out of bound of LV95 when searching for addresses... Lat/Lon were correct, so I used them instead. Also adapting the fixture file to have only LV95 values (as that is what we are requesting from the backend, even though x/y are somewhat wrong on the real backend)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2890_fix_search_test/index.html)


src/api/search.api.js
```diff
@@ -132,7 +132,7 @@ let cancelToken = null
  *   returned
  * @returns {Promise<CombinedSearchResults>}
  */
-async function search(queryString = '', lang = '', outputProjection) {
+async function search(queryString = '', lang = '', outputProjection = DEFAULT_PROJECTION) {
```
ltshb, line 135: I would not used any default projection in the api functions to avoid surprise when changing between projection.

ltshb (changes requested):

src/api/search.api.js
pakb: then I have to change the function signature, as it's the last param

ltshb (approved):

PR #490, opened 2023-10-25
BGDIINF_SB-2890 : fix 3d/layers e2e test

pakb:
I had to change how KML and GeoJSON layers were handled by Cesium

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2890_fix_geojson_in_3d/index.html)


src/modules/map/components/cesium/CesiumInternalLayer.vue
```diff
@@ -65,12 +68,14 @@ export default {
             type: Number,
             default: null,
         },
+        projection: {
+            type: CoordinateSystem,
+            default: DEFAULT_PROJECTION,
```
ltshb, line 73: I would try to avoid using the default projection whenever possible and make it required to avoid any confusion

src/modules/map/components/cesium/CesiumKMLLayer.vue
```diff
@@ -32,32 +34,36 @@ export default {
             type: Number,
             default: -1,
         },
+        projection: {
+            type: CoordinateSystem,
+            default: DEFAULT_PROJECTION,
```
ltshb, line 39: Same as above I would not use the default projection

src/store/plugins/2d-to-3d-management.plugin.js
```diff
@@ -0,0 +1,20 @@
...
+ */
+export default function from2Dto3Dplugin(store) {
+    store.subscribe((mutation, state) => {
+        if (DEFAULT_PROJECTION.epsg !== WEBMERCATOR.epsg && mutation.type === 'setShowIn3d') {
+            if (state.ui.showIn3d) {
+                store.dispatch('setProjection', WEBMERCATOR)
+            } else {
+                store.dispatch('setProjection', DEFAULT_PROJECTION)
+            }
+        }
```
ltshb, line 18: Here you set the projection on the store to WEBMERCATOR but in the Cesium Component the projection is hardcoded ! Doesn't seem logic. would it not make more sense to have the Cesium hardcoded to WEBMERCATOR if it need it and let the projection in the store to the on the user choosed ? I mean if in future we support more than 2 projections you will have an issue here.

ltshb (approved):
All good

PR #491, opened 2023-10-26
BGDIINF_SB-3154 : make crosshair URL param dynamic

### LINKED ISSUE #451: Add parameter for crosshair placement
Hello,
in the context of the BFS application "AREApp" which is a custom application that is used to generate the statistics about land usage and land coverage, there are some windows which include map.geo.admin.ch as auxiliary information.

The interpreters who use the software check all the points of the surveillance network (one each 100m), and when they click on "next" in the AREApp application, all the maps are re-centered on the new point.

Planning the migration to the new web-viewer when it will replace the current geoadmin3, I did a couple of test for the recentering of the map. Changing the lat and lon parameter does not reload the application, which is great for improved perfomance. But we need also to place a crosshair with `crosshair=cross` so that the users can locate the surveillance point on the map, and this forces us to reload the whole application anyway.

It would be great to have a similar parameter set, like `crosslat=46.0&crosslong=35.0` available in the url to recenter the crosshair location.

So, a few questions now:
- How much effort would be needed to implement this in your opinion?
- Do you think that this makes sense?
- Can you point out a couple of lines in the current code where we should check if we get a bit of budget to implement it?

pakb:
So that it may be changed while the app has been loaded, and moved too (without requiring a complete reload)

Will need to wait for #470 to be merged, and then adapt the crosshair center on a projection change
This fixes #451

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3154_dynamic_crosshair/index.html)

### DIFF
--- a/src/router/storeSync/CrossHairParamConfig.class.js
+++ b/src/router/storeSync/CrossHairParamConfig.class.js
@@ -1,0 +1,53 @@
+import AbstractParamConfig from '@/router/storeSync/abstractParamConfig.class'
+import { round } from '@/utils/numberUtils'
+
+function dispatchCrossHairFromUrlIntoStore(store, urlParamValue) {
+    const promisesForAllDispatch = []
+
+    const parts = urlParamValue.split(',')
+    if (parts.length === 1) {
+        promisesForAllDispatch.push(store.dispatch('setCrossHair', { crossHair: urlParamValue }))
+    } else if (parts.length === 3) {
+        const crossHair = parts[0]
+        const crossHairPosition = [parseFloat(parts[1]), parseFloat(parts[2])]
+        promisesForAllDispatch.push(
+            store.dispatch('setCrossHair', { crossHair, crossHairPosition })
+        )
+    }
+    return Promise.all(promisesForAllDispatch)
+}
+
+function generateCrossHairUrlParamFromStoreValues(store) {
+    if (store.state.position.crossHair) {
+        let crossHairParamValue = store.state.position.crossHair
+        const { center, crossHairPosition } = store.state.position
+        if (
+            crossHairPosition &&
+            (center[0] !== crossHairPosition[0] || center[1] !== crossHairPosition[1])
+        ) {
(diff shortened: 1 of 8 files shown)


src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -0,0 +1,53 @@
...
+    const parts = urlParamValue.split(',')
+    if (parts.length === 1) {
+        promisesForAllDispatch.push(store.dispatch('setCrossHair', { crossHair: urlParamValue }))
+    } else if (parts.length === 3) {
+        const crossHair = parts[0]
+        const crossHairPosition = [parseFloat(parts[1]), parseFloat(parts[2])]
+        promisesForAllDispatch.push(
+            store.dispatch('setCrossHair', { crossHair, crossHairPosition })
+        )
+    }
```
ltshb, line 16: I'm not sure this is a good idea to add the crosshair position into the crosshair parameter. At least my understanding of #451 this would not be needed as all they would need is to set the `center` parameter. All what we need here is to reset the crosshair position when the center changes. I tested it with this PR and it works.
Having the position in the crosshair make the application a bit more difficult to understand and could reveal some issue in the future as the behavior of the crosshair is then different when it has the position or not.

ltshb (commented):
From the code point of view, looks good to me. However about the behavior I've a comment, maybe we can have a chat today about this.
Finally once we have talked about the behavior some e2e test for this would be good. At least one simple test that checks that the crosshair position changed when the center changes.

src/router/storeSync/CrossHairParamConfig.class.js
ltshb: After a talk with @pakb saving the position with the crosshair is actually a good idea. It give the possibility to move the map around keeping the crosshair at its initial position. Note that on the old viewer the crosshair did not stayed permanently in url so when reloading it was lost, this pr as the advantage to keep it.

ltshb (approved):
I would suggest to add some e2e tests to cover the crosshair behaviour

PR #492, opened 2023-10-26
BGDIINF_SB-3071: Added CLI tool to check the external providers

ltshb:
This allow to check the validity of the provided external provider list.
    
For this I needed to reorganized a bit the utils in order to be able to import simple functions inside a node js script.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3071-import-url/index.html)


scripts/check-external-layers-providers.js
```diff
@@ -0,0 +1,229 @@
...
+            status: response.status,
+            headers: response.headers.toJSON(),
+        })
+    } else if (
+        !['*', 'https://map.geo.admin.ch'].includes(
+            response.headers.get('access-control-allow-origin').toString()?.trim()
+        )
+    ) {
+        console.error(
+            `Provider ${provider} does not geoadmin in its CORS: status=${
```
pakb, line 107: missing `have` here 😉

src/utils/external-provider.js
```diff
@@ -0,0 +1,110 @@
...
+export function isValidUrl(urlToCheck) {
+    let url
+
+    try {
+        url = new URL(urlToCheck)
+    } catch (_) {
+        return false
+    }
+
+    return url.protocol === 'http:' || url.protocol === 'https:'
```
pakb, line 69: shouldn't we enforce `https://` usage by filtering out non HTTPS URLs? (or we follow the redirect to see if they force HTTPS on their end?)

pakb (approved):

src/utils/external-provider.js
ltshb: I would let redirect to work if they are any.

PR #498, opened 2023-10-30
BGDIINF_SB-2890 : fix coordinate search

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2890-coordinate-search/index.html)

### DIFF
--- a/src/utils/coordinates/coordinateExtractors.js
+++ b/src/utils/coordinates/coordinateExtractors.js
@@ -1,20 +1,19 @@
-import { WEBMERCATOR, WGS84 } from '@/utils/coordinates/coordinateSystems'
+import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { reprojectUnknownSrsCoordsToWGS84 } from '@/utils/coordinates/coordinateUtils'
 import { toPoint as mgrsToWGS84 } from '@/utils/militaryGridProjection'
-import { round } from '@/utils/numberUtils'
 import proj4 from 'proj4'
 
 // 47.5 7.5
 const REGEX_WEB_MERCATOR = /^\s*([\d]{1,3}[.\d]+)\s*[ ,/]+\s*([\d]{1,3}[.\d]+)\s*$/i
 // 47°31.8' 7°31.8'
 const REGEX_MERCATOR_WITH_DEGREES =
-    /^\s*([\d]{1,3})[° ]+([\d]+[.,]?[\d]*)[']?\s*[,/]?\s*([\d]{1,3})[° ]+([\d.,]+)[']?\s*$/i
+    /^\s*([\d]{1,3})[° ]+([\d]+[.,]?[\d]*)['′]?\s*[,/]?\s*([\d]{1,3})[° ]+([\d.,]+)['′]?\s*$/i
 // 47°38'48'' 7°38'48'' or 47°38'48" 7°38'48"
 const REGEX_MERCATOR_WITH_DEGREES_MINUTES =
-    /^\s*([\d]{1,3})[° ]+([\d]{1,2})[' ]+([\d.]+)['"]{0,2}\s*[,/]?\s*([\d]{1,3})[° ]+([\d.]+)[' ]+([\d.]+)['"]{0,2}\s*$/i
+    /^\s*([\d]{1,3})[° ]+([\d]{1,2})[' ]+([\d.]+)['′"″]{0,2}\s*[,/]?\s*([\d]{1,3})[° ]+([\d.]+)['′ ]+([\d.]+)['′"″]{0,2}\s*$/i
 // 47°38'48''N 7°38'48''E or 47°38'48"N 7°38'48"E
 const REGEX_MERCATOR_WITH_DEGREES_MINUTES_AND_CARDINAL_POINT =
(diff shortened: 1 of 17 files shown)


src/utils/coordinates/coordinateExtractors.js
```diff
@@ -40,20 +39,20 @@ const webmercatorExtractor = (regexMatches) => {
     }
     let lon, lat
     if (regexMatches.length === 5) {
-        // 4 matches + global match, i.e. : (47)°(5.123)', (8)°(4.154)' (we inverse lat/lon to lon/lat in the process)
-        lon = Number(regexMatches[3]) + Number(regexMatches[4]) / 60.0
-        lat = Number(regexMatches[1]) + Number(regexMatches[2]) / 60.0
+        // 4 matches + global match, i.e. : (47)°(5.123)', (8)°(4.154)'
+        lon = Number(regexMatches[1]) + Number(regexMatches[2]) / 60.0
+        lat = Number(regexMatches[3]) + Number(regexMatches[4]) / 60.0
```
ltshb, line 44: For now it is ok but those extractors are not a good way as they are very dependent on the regex. I propose to rework this in a later PR to have the regex together with the extractor, this would make the code much easier to understand/read and less error prone

src/utils/coordinates/coordinateExtractors.js
```diff
@@ -40,20 +39,20 @@ const webmercatorExtractor = (regexMatches) => {
...
     }
     if (regexMatches.length === 7) {
-        // 6 matches + global match, i.e. : (47)°(5)'(41.61)", (8)°(4)'(6.32)" (we inverse lat/lon to lon/lat in the process)
+        // 6 matches + global match, i.e. : (47)°(5)'(41.61)", (8)°(4)'(6.32)"
```
ltshb, line 47: Why removing the inverse of lat/lon here ? Would be good to document it in the commit message 😉

src/utils/coordinates/coordinateUtils.js
```diff
@@ -11,13 +14,13 @@ import { LV03, LV95, WEBMERCATOR, WGS84 } from './coordinateSystems'
...
  * @param x {Number} X part of the coordinate (Easting/Longitude)
  * @param y {Number} Y part of the coordinate (Northing/Latitude)
- * @returns {Number[]} Re-projected coordinate as [lat, lon], or undefined if the input was not
+ * @returns {Number[]} Re-projected coordinate as [lon, lat], or undefined if the input was not
  *   convertible to WebMercator
  */
 export const reprojectUnknownSrsCoordsToWGS84 = (x, y) => {
-    if (LV95.getBoundsAs(WEBMERCATOR)?.isInBounds(x, y)) {
+    // guessing if this is already WGS84, or Mercator or a Swiss projection (if so, we need to reproject it to WGS84)
+    if (LV95BoundsAsWebMercator.isInBounds(x, y)) {
```
ltshb, line 22: In this function why checking backward ? I think it would make our life easier if we only support normal axis order and not try to guess wrong axis order.

tests/e2e-cypress/integration/search/coordinates-search.cy.js
```diff
@@ -15,35 +21,49 @@ describe('Testing coordinates typing in search bar', { testIsolation: false }, (
...
     })
-    // using the same values as coordinateUtils.spec.js (in Unit cypress)
-    const coordEpsg3857 = [773900, 5976445]
-    const WGS84 = [47.2101583, 6.952062]
-    const LV95 = [2563138.69, 1228917.22]
-    const LV03 = [563138.65, 228917.28]
+    const expectedCenter = DEFAULT_PROJECTION.bounds.center.map((value) => value - 1000)
+    const expectedCenterLV95 = proj4(DEFAULT_PROJECTION.epsg, LV03.epsg, expectedCenter).map(
+        LV03.roundCoordinateValue
+    )
```
ltshb, line 32: Should it not be 
```javascript
const expectedCenterLV95 = proj4(DEFAULT_PROJECTION.epsg, LV95.epsg, expectedCenter).map(
        LV95.roundCoordinateValue
    )
```

src/utils/coordinates/coordinateUtils.js
pakb: That's because of the issue with `CRS:84` and `WGS:84`, some maps function with a lat/lon pairing, and others with lon/lat, and you can't be sure it will be either one of them you receive.

For LV95 and LV03 backward checking, it is to mimic what was done in `mf-geoadmin3` and help automatically correct users' mistake with X/Y

src/utils/coordinates/coordinateExtractors.js
pakb: As we now not know the target projection (or default projection) I thought it would make sense to always return coordinate as `[x, y]` (meaning `[lon, lat]`) instead of doing some specifics for WGS84 and return `[lat, lon]` because that's how some of our frameworks want that. I prefer that we invert this coordinates when they need to be (for OpenLayers for instance) but store them as X, Y, much more logical

src/utils/coordinates/coordinateExtractors.js
pakb: I think this part of the app should be reworked too, yes.
It might have been one of the first thing I did while working on the project, and it pains my eyes to see some of my code from 4 years ago 🙈

pakb: I've created a ticket to do that, after we've merged our work branch on `develop` https://jira.swisstopo.ch/browse/BGDIINF_SB-3165

ltshb (approved):

PR #499, opened 2023-10-30
BGDIINF_SB-2890: Fixed projection error with external WMS layer

ltshb:
Some external WMS layer may use CRS:84 as CRS instead of ESPG:4326. This would
make the re-project failed.

Now we support the CRS:84 alias.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-2890-projection-alias/index.html)


ltshb:
NOTE: We cannot use CRS:84 as alias for ESPG:4326 as the axis is inverted, currently how our coordinate system is constructed we cannot support CRS:84, but looking at the special case for https://wms.geo.ag.ch/public/?service=wms&version=1.3.0&request=GetCapabilities it always put the bounding box in two projection system; LV95 and CRS:84. We should change the code to loop over the bounding box in order to take the first matching.

ltshb:
Fixed by #500

PR #503, opened 2023-10-31
BGDIINF_SB-3071: Moved the external-layer-providers.json in its module

ltshb:
Move it into its module as well as its utilities, which are highly related to
the import tool.

Also avoid a typescript failure in vscode by adding explicitely json file in
the tsconfig.json include pattern.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3071-url-list/index.html)


src/modules/infobox/components/ImportContent.vue
```diff
@@ -323,3 +323,4 @@ export default {
     right: 50px;
 }
 </style>
+../utils/external-layer-providers.js
```
pakb, line 326: I'm not sure this is wanted? If yes : what does it achieve?

ltshb: Oops you are right this is not intended, I'll remove that.

ltshb: done

pakb (approved):

PR #504, opened 2023-10-31
BGDIINF_SB-3071: Removed non working external layer provider

ltshb:
Those provider that have been removed have been removed using the script.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3071-url-list-clean-up/index.html)


src/modules/infobox/utils/external-layer-providers.json
```diff
@@ -295,41 +274,40 @@
     "https://wms.geo.gr.ch/wildschutzgebiete",
     "https://wms.geo.sh.ch/wms?SERVICE=WMS",
     "https://wms.geo.ti.ch/service",
-    "https://wms.geoportal.ch:8080/geoserver/AVAI/wms",
+    "https://wms.pcn.minambiente.it/ogc?map=/ms_ogc/WMS_v1.3/Vettoriali/Rete_ferroviaria.map",
```
pakb, line 277: is the script ordering alphabetically but takes upper cased letter as "bigger" as lower cased?
might be better to lower everything before ordering, don't you think?

pakb (approved):

src/modules/infobox/utils/external-layer-providers.json
ltshb: The script is ordering alphabeticaly, here is the sort compare method used; https://github.com/geoadmin/web-mapviewer/blob/37bb4cceee89bbb1aace276c863a27dda0a3e47c/scripts/check-external-layers-providers.js#L50
But that's a good idea to make the sorting non case sensitive, I'll change that

ltshb: Missed that was for the yaml files sorting, for the url sorting I use the regular `sort()` method.

ltshb: Done

scripts/check-external-layers-providers.js
```diff
@@ -54,9 +54,18 @@ function setupAxiosRetry() {
...
 
 function compareResultByProvider(a, b) {
-    if (a['provider'] > b['provider']) {
+    if (a['provider'].toLowerCase() > b['provider'].toLowerCase()) {
         return 1
-    } else if (a['provider'] < b['provider']) {
+    } else if (a['provider'].toLowerCase() < b['provider'].toLowerCase()) {
+        return -1
+    }
+    return 0
```
pakb, line 62: we could `compareCaseInsensitive(a['provider'], b['provider'])` here, so that we re-use the case insensitive test instead of writting it twice 😉

pakb (approved):

PR #505, opened 2023-11-01
BGDIINF_SB-2890: Removed lightbasemap under pixelkarte in webmercartor

ltshb:
Both map doesn't fit at all together and it doesn't make sense to use them
as a background.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-2890-background/index.html)


pakb (approved):
no more custom code, yes 🙌

PR #512, opened 2023-11-01
BGDIINF_SB-2890 : fix 3d tests

pakb:
URL param parsing : processing projection before position, so that when we (cold) start the app with &3d=true, it correctly transition from the default center to the same one but re-projected in Mercator

Moving all "position" related reprojection from the store plugin to the store file (no need to separate it, as it is the responsibility of the store file to handle its own stuff)

In Cypress commands.js : only hard coding the center to the old value when 3D flag is not present (otherwise a center expressed in LV95 is added to the URL while it should function in Mercator when in 3D)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-2890_fix_3d_tests/index.html)


src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -111,8 +111,16 @@ const handleLegacyParam = (
...
 
-        // if no special work to do, we just copy past legacy params to the new viewer
+        case '3d':
+            // if the 3d flag is given without being placed behind the hash (meaning at startup),
+            // we need to make sure the projection is set to Mercator, otherwise the startup sequence
+            // ends up not working properly and center the view wrongly
+            if ((typeof legacyValue === 'string' && legacyValue === 'true') || legacyValue) {
+                newQuery['sr'] = WEBMERCATOR.epsgNumber
+            }
+        // no break so we go into default too
```
ltshb, line 121: `3d` is not a legacy parameter and we should never have it before the `#` of I'm not misteaken ?

src/store/modules/position.store.js
```diff
@@ -264,6 +259,37 @@ const actions = {
             )
         }
         if (matchingProjection) {
+            const oldProjection = state.projection
```
ltshb, line 262: Should we not check if the projection isalready correct before doing all this work ?

```
if (matchingProjection && matchingProjection.epsg != oldProjection.epsg) {
...
}
```

src/store/modules/position.store.js
```diff
@@ -264,6 +259,37 @@ const actions = {
...
+            if (
+                oldProjection instanceof CustomCoordinateSystem &&
+                matchingProjection instanceof StandardCoordinateSystem
+            ) {
+                commit('setZoom', oldProjection.transformCustomZoomLevelToStandard(state.zoom))
+            }
+            if (
+                oldProjection instanceof CustomCoordinateSystem &&
+                matchingProjection instanceof CustomCoordinateSystem &&
+                oldProjection.epsg !== matchingProjection.epsg
```
ltshb, line 282: See comment above, I guess if the projection did not change we don't need to change the center nor the zoom ?

src/store/plugins/reproject-everything-on-projection-change.plugin.js
```diff
@@ -32,9 +32,6 @@ const reprojectEverythingOnProjectionChangePlugin = (store) => {
                 const reprojectCoordinates = (coordinates = []) =>
                     proj4(oldProjection.epsg, newProjection.epsg, coordinates)
 
-                // reprojecting the center of the map
-                store.dispatch('setCenter', reprojectCoordinates(state.position.center))
```
ltshb, line 36: I would suggest to rename this plugin as it only reproject selected features and not everything.

tests/e2e-cypress/support/commands.js
```diff
@@ -132,7 +132,7 @@ Cypress.Commands.add(
         if (!queryParams.hasOwnProperty('lang')) {
             queryParams.lang = 'en'
         }
-        if (!queryParams.hasOwnProperty('center')) {
+        if (!queryParams.hasOwnProperty('center') && !queryParams.hasOwnProperty('3d')) {
```
ltshb, line 135: Here I would also not add the center in legacy url, if `withHash` == false

src/router/legacyPermalinkManagement.routerPlugin.js
pakb: In some of our tests, it is.
And even though it's not legacy, I think we should make sure (advanced) users don't have to write this # to make it work.

We could think about having a dedicated file for current URL params than need special care, that would make more sense

src/store/modules/position.store.js
pakb: might be a good idea, yes

src/router/legacyPermalinkManagement.routerPlugin.js
ltshb: As discussed we do this later on

ltshb (approved):

PR #514, opened 2023-11-02
BGDIINF_SB-3162, BGDIINF_SB-3163: Fixed background in lv95

ltshb:
Now the background are the regular old products, pixelkarte-grau and swissimage
in lv95 projection.

Unfortunately I could not easily add the lightbasemap and the world image
as other options in Web Mercator mode because the current BackgroundSelector
cannot be dynamic, it is made of lots of CSS magics with a hardcoded value
in css on how many background is available. So we cannot have a dynamic number
of available background without reworking the bg selector.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3162-lightbasemap/index.html)

### DIFF
--- a/src/modules/map/components/footer/MapFooterBackgroundSelector.vue
+++ b/src/modules/map/components/footer/MapFooterBackgroundSelector.vue
@@ -20,17 +20,14 @@
                 :data-cy="`background-selector-${background.getID()}`"
                 @click="selectBackground(background, $event)"
             >
-                <div class="bg-selector-button-label">
+                <div class="bg-selector-button-label text-wrap">
                     {{ $t(getLayerTranslationLabel(background)) }}
                 </div>
             </button>
         </div>
         <button
             class="bg-selector-trigger"
-            :class="[
-                { 'bigger-pulse': animateMainButton },
-                'bg-ch-swisstopo-leichte-basiskarte-imagery_world-vt',
-            ]"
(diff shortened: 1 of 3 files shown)


src/modules/map/components/footer/MapFooterBackgroundSelector.vue
```diff
@@ -20,17 +20,14 @@
                 :data-cy="`background-selector-${background.getID()}`"
                 @click="selectBackground(background, $event)"
             >
-                <div class="bg-selector-button-label">
+                <div class="bg-selector-button-label text-wrap">
```
pakb, line 23: We might want to modify a little bit the no background here.
See how it looks with your PR :
image

And on "prod"
image

I don't mind the wrap, but it changes the visual "logic". It looks a bit wrong to me this way. We could keep the wrap but maybe add a border around the empty rectangle.

pakb (approved):

src/modules/map/components/footer/MapFooterBackgroundSelector.vue
ltshb: @pakb The wrap is needed for the grey map in french otherwise we don't see anything. It is with the wrap like on the old viewer.
To be honest I don't like at all the style of the bg selector and the code is horribly static with lots of CSS magics, so you cannot add dynamically an additional background for example. So I propose to let this like this for now and to re-think and eventually rework the bg selector ui overall.

PR #516, opened 2023-11-03
Add vegetation and buildings in 3D (as opt-in)

pakb:
Adds a second flavor of Vector layer that we provide, as it functions not the same in Cesium as when using Maplibre (URL are different) Also removing the creation of vector layer from the backend layers config (we do not yet provide such config, the format is still TBD)

Adding two buttons in 3D, to opt-in for both new dataset (buildings and vegetation) There is still an issue with labels, only the white dot is loaded, further investigation is required

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_add_buildings_and_vegetation_to_3d/index.html)


pakb:
replaced by #520

PR #519, opened 2023-11-06
Fixed search bar category background

ltshb:
@pakb if you agree I would change the hover list item background color of other component to the new variable `$hover-selection` created here ?

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-search-ui/index.html)


pakb (approved):
Yep, this hover color as $grey-300 is perfect, thanks!

pakb:
I think you can even rename / overwrite the Bootsrap `$light` with this value, and we can then use `text-bg-light` further down the road

PR #521, opened 2023-11-06
BGDIINF_SB-3161 : constrains OL to only show map in LV95 extent

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3161_consrtaint_view_to_lv95/index.html)

### DIFF
--- a/tests/e2e-cypress/integration/mouseposition.cy.js
+++ b/tests/e2e-cypress/integration/mouseposition.cy.js
@@ -1,6 +1,6 @@
 /// <reference types="cypress" />
 
-import { DEFAULT_PROJECTION } from '@/config'
+import { BREAKPOINT_TABLET, DEFAULT_PROJECTION } from '@/config'
 import {
     LV03Format,
     LV95Format,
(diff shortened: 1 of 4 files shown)


pakb:
I need to rework the geodesic drawing test, as it's testing some drawing in each hemisphere but it is not impossible

ltshb (commented):
Code looks good to me but some issues in cypress test with the mouse position.

tests/e2e-cypress/integration/mouseposition.cy.js
```diff
@@ -66,241 +65,167 @@ function checkMousePositionNumberValue(expectedX, expectedY, parser) {
...
-            cy.readStoreValue('getters.isPhoneMode').then((isPhoneMode) => {
-                if (isPhoneMode) {
-                    cy.get('[data-cy="map"]').click('center')
-                    cy.waitUntilState((state) => !state.ui.fullscreenMode)
-                }
-            })
-        })
+    it('shows coordinate under the mouse cursor in the footer, according to the selected projection format', () => {
+        // the viewport needs to be at least tablet sized in order for the mouse cursor tracker to be visible
+        cy.viewport('ipad-2')
```
ltshb, line 84: Here and below when using the `cy.viewport('iphone-4)` I would use the mocha `skip()` functionality and only run the tests if the viewport is set to the correct view. Until now we have all tests that should work on all viewport and the ability to run separately all view port. I would keep this strategy but if some test can only be run on a specific viewport, then we should only run it on that viewport and not force the viewport. So a proposal here would be to put all viewport specific test together in a context as such

```javascript
context('Iphone 4 tests', () => {
  before(() => {
     if (viewport != 'iphone-4') {
       this.skip()
     }
  })
  beforeEach(() => {
     cy.goToMapVeiw()
  })
  it('', () => {})
})
```

PR #523, opened 2023-11-08
BGDIINF_SB-3185 : rework of the BG selector

pakb:
Component based approach, so that it is possible to change the list of BG dynamically (we want to swap the BG list when going 3D). For images, I removed the sprite approach and added one PNG per layer. The image is then loaded based on the layer ID.

Both style (squared and rounded) have been reproduced, with a bit more of a Vue approach than pure CSS regarding animations

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3185_bg_selector_rework/index.html)


ltshb (commented):
Concerning the look and feel it looks better. However the button are quite big now, I see that they are big enough to avoid word wrapping, but personally I would keep the same size as on the old map.
For word wrapping issue we should consider changing the French translation from `Cartes noir & blanc` to `Cartes N&B` in German we already use an abbreviation for black and white.
I would also keep the order of the bg like the old viewer; white, grey, color, image.
I'll review the code, later on.

ltshb:
By the way I like the fact that the `Background` button take the active background image 💯

pakb:
> we should consider changing the French translation from `Cartes noir & blanc` to `Cartes N&B` in German we already use an abbreviation for black and white.

I think that's a good point, and as these are `web-mapviewer` translation only, we can change that without altering the legacy viewer.


>  I would also keep the order of the bg like the old viewer; white, grey, color, image. I'll review the code, later on.

I keep them in the order I receive them from the backend, and I think we should stick to this. Maybe we could ask that the backend sort them in a certain way?

pakb:
image
image

I don't know if we can really go smaller than what is currently on the PR, see these two languages

ltshb:
I would keep the word wrap for RM, and focus on EN, DE, FR and IT. For IT we could change `Carta bianca e nera` to `Carta B/N` (see https://it.wikipedia.org/wiki/Bianco_e_nero)

ltshb:
@pakb great look and feel 👍🏼

src/modules/map/components/footer/backgroundSelector/BackgroundSelectorSquared.vue
```diff
@@ -82,7 +83,7 @@ const { show, animate, getImageForBackgroundLayer, toggleShowSelector, onSelectB
 @import './bg-selector';
 
 $main-element: '.bg-selector-squared';
-$square-button-width: 7rem;
```
ltshb: Just wonder if using `6rem` or `5rem` would have achieved to get close to the old viewer size of `98px` ?

ltshb (approved):
Thanks

src/modules/map/components/footer/backgroundSelector/BackgroundSelectorSquared.vue
pakb: I tried to play a little bit with this value, and at `96px` there would be some line-break in English... So I prefered having some hard values that we are sure won't produce this...

PR #524, opened 2023-11-08
BGDIINF_SB-3187: Reduce the number of tests for the search bar

ltshb:
All possible coordinates formats are already tested in unittest so no need to
test everything here again.

This will speed up e2e tests but also reduce the number of tests result which
are not cheap if we want to run the tests on cypress cloud.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3187-e2e-tests/index.html)


tests/e2e-cypress/integration/search/coordinates-search.cy.js
pakb, line 1: You've removed quite a lot of test cases here, no more testing with `,`, `/` or things like that. I don't know if it's a good idea

ltshb: All of these tests are already done in the unittests in https://github.com/geoadmin/web-mapviewer/blob/6ae102053b8ebd7af072b6cdbd13710e6fcc5178/src/utils/coordinates/__test__/coordinateExtractors.spec.js#L19 I don't think it make sense to test them in twice, once as unittest and once as e2e tests.
We should also always prefer unittest to e2e tests whenever possible. As they are cheaper (quicker and don't cost nothing in cypress)

pakb (approved):

PR #525, opened 2023-11-10
BGDIINF_SB-3187: Improved layers.api cypress tests

ltshb:
Reduce the number of it() to speed up cypress and reduce cost.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3187-e2e-tests-layers/index.html)


pakb (approved):

pakb:
but it seems that `layers.cy.js` now fails on the CI 🤔

PR #527, opened 2023-11-13
Cesium embellishments

pakb:
Add a Skybox/Stars when looking at the sky

Loading higher resolution tile earlier, so that the rendering is more pleasant to took at. Here's the same view, but with and without this change.
image


pakb:
would be better to check this out after https://github.com/geoadmin/web-mapviewer/pull/523 is merged (and this PR has been rebased). Otherwise, it's not possible to swap BG in 3D

ltshb (approved):

PR #529, opened 2023-11-14
BGDIINF_SB-3180: Fix external WMS layer attribution

ltshb:
The external WMS layer attribution was lost after a reload of the page.

This was because the attribution was not saved in the url.

@pakb what do you think of putting the parsing and stringify (getID()) inside the ExternalWMSLayer object ? If you are OK I would do the same for other external layers ?
@pakb about the way I put the attributions inside the URL parameter, I think we should discuss it in a tech talk, this PR is just as example on how we could do it and will serve as discussion base.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-3180-ext-attributions/index.html)


pakb:
I would maybe leverage the "extra param" feature here, such as how we use it to describe the current timestamp / year
`layers=WMS|wms_url|wms_layer_name|wms_version@attributions=xxx|xxx_url,yyy|yyy_url`

ltshb:
superseeded by #536

PR #530, opened 2023-11-14
Remove lib ogc-parser

pakb:
While updating libs, I had issues with ogc-parser as it is using very outdated dependencies. After checking the GitHub repo, I saw it was kind of staled, so I replaced it with OpenLayers function. I had to add some browser API equivalent to NodeJS in order to achieve this (see first lines of the file) as this is then used under the hood by OpenLayers to parse XMLs

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_remove_ogc_parser/index.html)

### DIFF
--- a/scripts/check-external-layers-providers.js
+++ b/scripts/check-external-layers-providers.js
@@ -6,9 +6,7 @@ import writeYamlFile from 'write-yaml-file'
 import axiosRetry from 'axios-retry'
 import yargs from 'yargs'
 import { hideBin } from 'yargs/helpers'
-import ogcParser from 'ogc-parser'
 import { exit } from 'process'
-
 import {
     transformUrl,
     isWmsGetCap,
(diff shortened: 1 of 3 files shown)


scripts/check-external-layers-providers.js
```diff
@@ -6,16 +6,22 @@ import writeYamlFile from 'write-yaml-file'
...
     isGpx,
 } from '@/modules/infobox/utils/external-provider'
+import { WMSCapabilities } from "ol/format";
+import WMTSCapabilities from "ol/format/WMTSCapabilities";
+import { JSDOM } from 'jsdom'
+
+// faking browser support, so that OpenLayers has what it requires to parse XMLs
+const dom = new JSDOM()
+global.DOMParser = dom.window.DOMParser
+global.Node = dom.window.Node
```
ltshb, line 24: Hey great I tried to fake the jsdom but couldn't, the example that I used was a bit different from this one and did not worked, that's why I used the ogc-parser lib.

ltshb (commented):
Great findings with the jsdom faking. 
Now that this faking works I would go a bit further and reuse the `getCapWMSLayers()` and `getCapWMTSLayers()` method.

scripts/check-external-layers-providers.js
pakb: maybe a one liner that pushes everything from `dom.window` to `global` is achievable here, would ease things if you require more than theses two API entries

ltshb (approved):
Forget about using the `getCapWMSLayers()` and `getCapWMTSLayers()` methods I'll rework them with the attribution issue.

PR #532, opened 2023-11-15
Update and cleaup libs

pakb:
removing :
- vue-loader : as it was a webpack module used to handle Vue (now we run Vite, no need for this package anymore)
- register-service-worker : was there by default when creating the Vue project, but as we do not use this API (yet?) we can remove it
- aws-sdk : was used when the deploy script was still hosted here, now that it is in another Github project, this package isn't required anymore
- canise-lite : was used by the removed npm target update:browserlist
- git-branch : was used by the deploy script (now moved to another GitHub repo)
- vite-plugin-required : was used to mitigate an issue with require images in the code after Vite migration, but we have since moved to another approach to import our images dynamically, so this is not needed anymore

[Test link](https://sys-map.dev.bgdi.ch/preview/update_and_cleanup_libs/index.html)


ltshb (approved):

pakb:
replaces #526

PR #533, opened 2023-11-15
BGDIINF_SB-3193 : topics are now group of layers

pakb:
preparation work so that we can feed an external getCap parsing to the LayerCatalog component

Removing all related TopicTree classes, and replacing them with the GeoAdminGroupOfLayers equivalent.
Adapting the menu to use this new GeoAdminGroupOfLayers to build the topic tree.

also fix : now implementing properly the opening of certain categories, if the backend tells to do so (switch to Aviation topic to have an example)

simplification of topics.cy.js and fix an issue when default topic is set at startup with some catalog node that should be opened, the catalog wasn't opened correctly because of race condition between URL param parsing and topic loading (hence the move of isFirstSetTopic into the callback function)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_bgdiinf_sb-3193_geoadmin_group_of_layers/index.html)


ltshb (approved):
Looks good just a minor comment.

I also found a bug but apparently not related to this PR, see https://jira.swisstopo.ch/browse/BGDIINF_SB-3197 the bug is already present on the current develop branch, but when doing the rework we might first consider this strange bug, maybe there is a use case that we are not aware of ?

PR #536, opened 2023-11-16
BGDIINF_SB-3180: External Layer refresh attributes on reload

ltshb:
TODO in follow up PRs:
- [x] Update ADR about URL param, see #537 
- [x] Reuse the new capabilities class in the ImportContent.vue component #538 
- [x] Reuse the new capabilities class in the external-layer check script #538
- [ ] Add unit tests
- [ ] Add e2e tests

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3180-ext-layers/index.html)

### DIFF
--- a/src/api/layers/AbstractLayer.class.js
+++ b/src/api/layers/AbstractLayer.class.js
@@ -17,6 +17,16 @@ export class LayerAttribution {
 }
 
 /**
+ * Get the default layer attributions based on URL
+ *
+ * @param {string} baseUrl Get Capabilities base URL
+ * @returns {LayerAttribution[]} Default list of layer attributions
+ */
+export function getDefaultAttribution(baseUrl) {
+    return [new LayerAttribution(new URL(baseUrl).hostname)]
+}
+
+/**
  * Base class for layers' config description, must be extended to a more specific flavor of Layer
  * (e.g. {@link GeoAdminWMTSLayer}, {@link GeoAdminWMSLayer}, {@link GeoAdminGeoJsonLayer},
  * {@link GeoAdminAggregateLayer} or {@link KMLLayer})
@@ -34,6 +44,8 @@ export default class AbstractLayer {
      * @param {Boolean} hasTooltip Define if this layer shows tooltip when clicked on
      * @param {Boolean} isExternal Define if this layer comes from our backend, or is from another
      *   (external) source
+     * @param {boolean} isLoading Set to true if some parts of the layer (e.g. metadata) are still
+     *   loading
      */
     constructor(
         name = '',
@@ -42,7 +54,8 @@ export default class AbstractLayer {
         visible = false,
         attributions = [],
         hasTooltip = false,
-        isExternal = false
+        isExternal = false,
+        isLoading = false
     ) {
         this.name = name
         this.type = type
@@ -51,6 +64,7 @@ export default class AbstractLayer {
         this.attributions = [...attributions]
         this.hasTooltip = hasTooltip
         this.isExternal = isExternal
+        this.isLoading = isLoading
     }
 
     /**
--- a/src/api/layers/AbstractLayer.class.js
+++ b/src/api/layers/AbstractLayer.class.js
@@ -17,16 +17,6 @@ export class LayerAttribution {
 }
 
 /**
- * Get the default layer attributions based on URL
- *
- * @param {string} baseUrl Get Capabilities base URL
- * @returns {LayerAttribution[]} Default list of layer attributions
- */
-export function getDefaultAttribution(baseUrl) {
-    return [new LayerAttribution(new URL(baseUrl).hostname)]
-}
-
-/**
  * Base class for layers' config description, must be extended to a more specific flavor of Layer
  * (e.g. {@link GeoAdminWMTSLayer}, {@link GeoAdminWMSLayer}, {@link GeoAdminGeoJsonLayer},
  * {@link GeoAdminAggregateLayer} or {@link KMLLayer})
--- a/src/api/layers/layers-external.api.js
+++ b/src/api/layers/layers-external.api.js
@@ -1,6 +1,6 @@
 import axios from 'axios'
-import WmsCapabilities from '@/api/layers/wms-capabilities.class'
-import WmtsCapabilities from '@/api/layers/wmts-capabilities.class'
+import WMSCapabilitiesParser from '@/api/layers/WMSCapabilitiesParser.class'
+import WMTSCapabilitiesParser from '@/api/layers/WMTSCapabilitiesParser.class'
 import log from '@/utils/logging'
 
 /** Timeout for accessing external server in [ms] */
@@ -10,7 +10,7 @@ const externalServerTimeout = 30000
  * Read and parse WMS GetCapabilities
  *
  * @param {string} baseUrl Base URL for the WMS server
- * @returns {Promise<WmsCapabilities | null>} WMS Capabilities
+ * @returns {Promise<WMSCapabilitiesParser | null>} WMS Capabilities
  */
 export async function readWmsCapabilities(baseUrl) {
     const url = new URL(baseUrl)
@@ -32,11 +32,11 @@ export async function readWmsCapabilities(baseUrl) {
  * @param {string} content Input content to parse
  * @param {string} originUrl Origin URL of the content, this is used as default GetCapabilities URL
  *   if not found in the Capabilities
- * @returns {WmsCapabilities} Get Capabilities object
+ * @returns {WMSCapabilitiesParser} Get Capabilities object
  */
 export function parseWmsCapabilities(content, originUrl) {
     try {
-        return new WmsCapabilities(content, originUrl)
+        return new WMSCapabilitiesParser(content, originUrl)
     } catch (error) {
         log.error(`Failed to parse WMS Get Capabilities`, error)
         return null
@@ -47,7 +47,7 @@ export function parseWmsCapabilities(content, originUrl) {
  * Read and parse WMTS GetCapabilities
  *
  * @param {string} baseUrl Base URL for the WMTS server
- * @returns {Promise<WmtsCapabilities | null>} WMTS Capabilities
+ * @returns {Promise<WMTSCapabilitiesParser | null>} WMTS Capabilities
  */
 export async function readWmtsCapabilities(baseUrl) {
     const url = new URL(baseUrl)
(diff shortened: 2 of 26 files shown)


src/api/layers/layers-external.api.js
```diff
@@ -0,0 +1,83 @@
...
+ * Read and parse WMS GetCapabilities
+ *
+ * @param {string} baseUrl Base URL for the WMS server
+ * @returns {Promise<WmsCapabilities | null>} WMS Capabilities
+ */
+export async function readWmsCapabilities(baseUrl) {
+    const url = new URL(baseUrl)
+    url.searchParams.set('SERVICE', 'WMS')
+    url.searchParams.set('REQUEST', 'GetCapabilities')
+    const response = await axios.get(url.toString(), { timeout: 20000 })
```
pakb, line 16: why the special timeout length?
might be good idea to set it as a constant on top of this file, if this is really needed, so that we declare this timeout only once

src/api/layers/layers-external.api.js
```diff
@@ -0,0 +1,83 @@
...
+
+    if (response.status !== 200) {
+        log.error(`Failed to read GetCapabilities from ${url.toJSON()}`, response)
```
pakb, line 19:
```suggestion
        log.error('Failed to read GetCapabilities from', url, response)
```
You can give as many extra params as you want to the log function, they will be outputted as Object in the console (and you will be able to look into them, as usual)

src/api/layers/layers-external.api.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+export async function readWmtsCapabilities(baseUrl) {
+    const url = new URL(baseUrl)
+    url.searchParams.set('SERVICE', 'WMTS')
+    url.searchParams.set('REQUEST', 'GetCapabilities')
+    try {
+        const response = await axios.get(url.toString(), { timeout: 20000 })
+
+        if (response.status !== 200) {
+            log.error(`Failed to read GetCapabilities from ${url.toJSON()}`, response)
```
pakb, line 57:
```suggestion
            log.error('Failed to read GetCapabilities from', url, response)
```

src/api/layers/wms-capabilities.class.js
```diff
@@ -0,0 +1,174 @@
+import { LayerAttribution } from '@/api/layers/AbstractLayer.class'
+import ExternalWMSLayer from '@/api/layers/ExternalWMSLayer.class'
+import ExternalGroupOfLayers from '@/api/layers/ExternalGroupOfLayers.class'
+import { WMSCapabilities as OlWmsCapabilities } from 'ol/format'
+import log from '@/utils/logging'
+
+/** Wrapper around the OpenLayer WMSCapabilities to add more functionalities */
+export default class WmsCapabilities {
```
pakb, line 8: Maybe a different name would help differentiate them?
Something like `ExtendedWMSCapabilitesParser` or something similar

src/api/layers/wmts-capabilities.class.js
```diff
@@ -0,0 +1,138 @@
+import { LayerAttribution } from '@/api/layers/AbstractLayer.class'
+import ExternalWMTSLayer from '@/api/layers/ExternalWMTSLayer.class'
+import OlWmtsCapabilities from 'ol/format/WMTSCapabilities'
+import log from '@/utils/logging'
+
+/** Wrapper around the OpenLayer WMSCapabilities to add more functionalities */
+export default class WmtsCapabilities {
```
pakb, line 7: same remark as with `WmsCapabilites` as only the lower cased `ms` are different from the OL counterpart, we could easily confuse them when importing

src/api/layers/wms-capabilities.class.js
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /**
+     * Find recursively in the capabilities the matching layer ID node
+     *
+     * @param {string} layerId Layer ID to search for
+     * @param {OlWmsCapabilities.Capability.Layer} startFrom Layer node to start from
+     * @returns {OlWmsCapabilities.Capability.Layer} Capability layer node
+     */
+    findLayer(layerId, startFrom = null) {
```
pakb, line 22: I'm not sure I'm a fan of outputting a new type of layer specific to this class/parser outside of this file, that is then checked and transformed into what we usually use through `getExternalLayerObject`.

I would maybe prefer that we transform this node directly in a `ExternalLayer` type (or return `null` of nothing was found), but that means that you have to receive the projection and other things here too...

ltshb: This method is used to search the layers inside the capabilities and cannot return an external layer object !

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -55,32 +56,42 @@ export function createLayerObject(parsedLayer) {
...
-            layerName,
+            wmsLayerId,
             parsedLayer.opacity,
             parsedLayer.visible,
             wmsServerBaseURL,
-            wmsLayerIds,
-            [new LayerAttribution(new URL(wmsServerBaseURL).hostname)],
-            wmsVersion
+            wmsLayerId,
+            [new LayerAttribution(new URL(wmsServerBaseURL).hostname)]
```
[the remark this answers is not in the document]
ltshb: Good point it can be removed here

src/api/layers/layers-external.api.js
ltshb: I added a timeout to avoid having the spinner forever if the server is not answering, if got the axios docu correct by default there is no timeout !

src/api/layers/wms-capabilities.class.js
ltshb: @pakb I'm not a big fan for such long name, but I could use `WmsCapabilitiesParser` ?

src/api/layers/AbstractLayer.class.js
```diff
@@ -16,6 +16,16 @@ export class LayerAttribution {
...
+/**
+ * Get the default layer attributions based on URL
+ *
+ * @param {string} baseUrl Get Capabilities base URL
+ * @returns {LayerAttribution[]} Default list of layer attributions
+ */
+export function getDefaultAttribution(baseUrl) {
+    return [new LayerAttribution(new URL(baseUrl).hostname)]
+}
+
```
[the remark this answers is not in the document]
ltshb: Done

src/api/layers/layers-external.api.js
ltshb: done

src/api/layers/wms-capabilities.class.js
ltshb: Moreover this class old a Capabilities object and it would be correct to have a `findLayer` member returning one of its layer part. Returning an `ExternalLayer` object would not be correct unless it is specified in the name of the method.
Also I added this method because during parsing of the URL we might need just one `ExternalLayer` out of the whole capabilities, returning all capabilities as `ExternalLayer` in this case would be overkill especially if the capabitilities holds several hundreds of layers.

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -24,10 +24,22 @@
...
             <span
                 class="menu-layer-item-name"
+                :class="{ 'text-body-tertiary fst-italic': layer.isLoading }"
                 :data-cy="`active-layer-name${id}`"
                 @click="onToggleLayerVisibility"
                 >{{ layer.name }}</span
             >
+            <button
+                v-if="layer.isLoading"
+                id="loadingButton"
```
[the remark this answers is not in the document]
ltshb: done

src/router/storeSync/LayerParamConfig.class.js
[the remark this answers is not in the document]
ltshb: Done

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -183,6 +196,25 @@ export default {
...
+    mounted() {
+        this.tippyInstances = tippy('#loadingButton', {
+            content: this.tooltipContent,
+            arrow: true,
+            placement: 'top',
+            touch: false,
+        })
+    },
+    beforeUnmount() {
+        this.tippyInstances?.forEach((tooltip) => tooltip.destroy())
```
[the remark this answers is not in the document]
ltshb: The issue with the `ref` approach is that we have this warning at first application load 
image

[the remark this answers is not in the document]
ltshb: Also using `ref` doesn't necessary return one instance because it is attached to a list of elements, to use a `ref` as for the third party disclaimer I would have to put the load button in a separate vue component.

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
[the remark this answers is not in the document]
ltshb: See comment https://github.com/geoadmin/web-mapviewer/pull/536#discussion_r1398696506

src/api/layers/wms-capabilities.class.js
ltshb: Done as `WMSCapabilitiesParser`

src/api/layers/wmts-capabilities.class.js
ltshb: Done

src/store/modules/layers.store.js
```diff
@@ -281,6 +281,12 @@ const actions = {
             log.error('Can not remove layer that is not yet added', layerIdOrObject)
         }
     },
+    updateLayer({ commit }, layer) {
+        if ((!layer) instanceof AbstractLayer) {
```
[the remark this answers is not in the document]
ltshb: Done

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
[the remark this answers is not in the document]
ltshb: Changed to the class instead of Id.

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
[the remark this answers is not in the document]
ltshb: changed to the class instead of id

pakb (approved):

PR #537, opened 2023-11-17
BGDIINF_SB-3180: Updated URL param ADR for the External Layers

ltshb:
This is based on discussion of this tech talk https://ltwiki.adr.admin.ch:8443/display/PB/2023-11-14  and implementation of #536 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3180-ext-layers-adr/index.html)

### DIFF
--- a/adr/2021_03_16_url_param_structure.md
+++ b/adr/2021_03_16_url_param_structure.md
@@ -4,6 +4,8 @@
 
 > Date: 16.03.2021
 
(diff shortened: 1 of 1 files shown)


ltshb:
@ltclm @rebert here I'm not 100% sure that removing the WMS and WMTS version from URL and relying on the Get Capabilities answer for that is a good idea,

PR #538, opened 2023-11-20
BGDIINF_SB-3180: Use new Capabilities Parser for import tool

ltshb:
Now the code of the WMS/WMTS parsing is reused in all places.

This also fixes some issues, see commits messages.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3180-ext-layers-import/index.html)


src/api/layers/layers-external.api.js
```diff
@@ -4,23 +4,47 @@ import WMTSCapabilitiesParser from '@/api/layers/WMTSCapabilitiesParser.class'
 import log from '@/utils/logging'
 
 /** Timeout for accessing external server in [ms] */
-const externalServerTimeout = 30000
+const EXTERNAL_SERVER_TIMEOUT = 30000
```
pakb, line 7: I'm surprised you can import that in the `scripts/check-external-layers-providers.js` without adding a `export ...` before the constant declaration

src/api/layers/layers-external.api.js
```diff
@@ -4,23 +4,47 @@ import WMTSCapabilitiesParser from '@/api/layers/WMTSCapabilitiesParser.class'
...
+
+/**
+ * Sets the WMS url parameters
+ *
+ * @param {URL} url Url to set
+ * @param {string} language Language to use
+ * @returns {URL} Url with wms parameter
+ */
+export function setWmsUrlParameters(url, language) {
```
pakb, line 16: that is a name a bit too generic for my taste, maybe something like `setGetCapWmsUrlParameters` (I know it's long...)
Or at least explain in the JSDoc that this is made to request the getCapabilites endpoint, and not other endpoint (such as `getMap` for tiles, etc...)

src/api/layers/layers-external.api.js
```diff
@@ -38,34 +62,53 @@ export function parseWmsCapabilities(content, originUrl) {
...
  *
- * @param {string} baseUrl Base URL for the WMTS server
- * @returns {Promise<WMTSCapabilitiesParser | null>} WMTS Capabilities
+ * @param {URL} url Url to set
+ * @param {string} language Language to use
+ * @returns {URL} Url with wmts parameters
  */
-export async function readWmtsCapabilities(baseUrl) {
-    const url = new URL(baseUrl)
+export function setWmtsUrlParameters(url, language) {
```
pakb, line 76: same as above, I think it would be better to state somehow that this is for getCapabilites request only (so that when the auto-complete of IDE provides it, we can know what this is for, and not use it for some other WM(T)S use cases)

src/api/layers/layers-external.api.js
ltshb: You are right but it works strangely. Anyway I'll add the `export`

ltshb: Done

src/api/layers/layers-external.api.js
ltshb: Done

src/api/layers/layers-external.api.js
ltshb: Done

pakb (approved):

PR #539, opened 2023-11-20
BGDIINF_SB-3199: Re-added lint rules and fixed linting issues

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-3199-linter/index.html)


tests/e2e-cypress/integration/menuTray.cy.js
pakb, line 1: Have you tried without the arbitrary waits?
IMO it might work without it, as the next `cy.get(...)` will wait for the default timeout duration before crashing (meaning more than 200ms)

pakb (approved):

tests/e2e-cypress/integration/menuTray.cy.js
ltshb: Yes I tried and it failed therefore I let the wait and added the lint disable rule.

PR #541, opened 2023-11-21
BGDIINF_SB-3108 cut top layer

ltkum:
Issue : we need to cut off the top layer where the slider is

Fix :
We add a getter that checks if the top layer changed and returns it
If there is a change, we bind a pre and post render event on that layer, cutting it where needed. We also remove the old binding if there was any.
We ensure that this works even if there are no layers selected (thus, no bind at all)

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-3108-cut-top-layer/index.html)

### DIFF
--- a/src/modules/map/components/CompareSlider.vue
+++ b/src/modules/map/components/CompareSlider.vue
@@ -93,6 +93,14 @@ function listenToMouseMove(event) {
     } else {
         currentPosition = event.clientX - compareSliderOffset.value
     }
+    // we ensure the slider can't get off the screen
+    if (currentPosition < 1) {
+        currentPosition = 1
+    }
+    // same on the other side, but with also the idea of keeping the cartes completely in the screen
+    if (currentPosition > clientWidth.value - 36) {
+        currentPosition = clientWidth.value - 36
+    }
     compareRatio.value = round(currentPosition / clientWidth.value, 3)
     getMap().render()
 }
--- a/src/modules/map/components/CompareSlider.vue
+++ b/src/modules/map/components/CompareSlider.vue
@@ -1,6 +1,6 @@
 <script setup>
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
-import { computed, defineProps, onMounted, ref, toRefs, watch } from 'vue'
+import { computed, defineProps, inject, onMounted, ref, toRefs, watch } from 'vue'
 import { useStore } from 'vuex'
 
 import { round } from '@/utils/numberUtils'
@@ -10,11 +10,12 @@ const props = defineProps({
         default: window.innerWidth,
     },
 })
+inject['getMap']
 
 const { clientWidth } = toRefs(props)
+const olMap = ref(null)
 
 const compareRatio = ref(-0.5)
-
 const store = useStore()
 const storeCompareRatio = computed(() => store.state.ui.compareRatio)
 const compareSliderPosition = computed(() => {
@@ -22,14 +23,50 @@ const compareSliderPosition = computed(() => {
         left: compareRatio.value * 100 + '%',
     }
 })
+const visibleLayerOnTop = computed(() => store.getters.visibleLayerOnTop)
 
 watch(storeCompareRatio, (newValue) => {
     compareRatio.value = newValue
 })
 
+watch(visibleLayerOnTop, () => {
+    $nextTick(slice())
+})
+
 onMounted(() => {
     compareRatio.value = storeCompareRatio.value
+    slice()
 })
+
+function slice() {
+    if (olMap.value) {
+        olMap.value.un('prerender', onPreCompose)
+        olMap.value.un('postrender', onPostCompose)
+    }
+    olMap.value = getMap()
+        .getAllLayers()
+        .find((layer) => {
+            return visibleLayerOnTop.value && layer.get('id') === visibleLayerOnTop.value.getID()
+        })
+    if (olMap.value) {
+        olMap.value.on('prerender', onPreCompose)
+
+        olMap.value.on('postrender', onPostCompose)
+    }
+}
+
+function onPreCompose(event) {
+    const ctx = event.context
+    const width = ctx.canvas.width * this.compareRatio + 20
+    ctx.save()
+    ctx.beginPath()
+    ctx.rect(0, 0, width, ctx.canvas.height)
+    ctx.clip()
+}
+
+function onPostCompose(event) {
+    event.context.restore()
+}
 
 function grabSlider() {
     window.addEventListener('mousemove', listenToMouseMove, { passive: true })
@@ -41,6 +78,7 @@ function grabSlider() {
 function listenToMouseMove(event) {
     const currentPosition = event.type === 'touchmove' ? event.touches[0].pageX : event.pageX
     compareRatio.value = round(currentPosition / clientWidth.value, 2)
+    this.getMap().render()
 }
 
 function releaseSlider() {
--- a/src/modules/map/components/CompareSlider.vue
+++ b/src/modules/map/components/CompareSlider.vue
@@ -1,23 +1,20 @@
 <script setup>
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
-import { computed, defineProps, inject, onMounted, ref, toRefs, watch } from 'vue'
+import { unByKey } from 'ol/Observable'
+import { computed, inject, nextTick, onMounted, ref, watch } from 'vue'
 import { useStore } from 'vuex'
 
 import { round } from '@/utils/numberUtils'
-const props = defineProps({
-    clientWidth: {
-        type: Number,
-        default: window.innerWidth,
-    },
-})
-inject['getMap']
+const getMap = inject('getMap')
 
-const { clientWidth } = toRefs(props)
-const olMap = ref(null)
+const preRenderKey = ref(null)
+const postRenderKey = ref(null)
+const compareSliderOffset = ref(0)
 
 const compareRatio = ref(-0.5)
 const store = useStore()
 const storeCompareRatio = computed(() => store.state.ui.compareRatio)
+const clientWidth = computed(() => store.state.ui.width)
 const compareSliderPosition = computed(() => {
     return {
         left: compareRatio.value * 100 + '%',
@@ -27,10 +24,11 @@ const visibleLayerOnTop = computed(() => store.getters.visibleLayerOnTop)
 
 watch(storeCompareRatio, (newValue) => {
     compareRatio.value = newValue
+    slice()
 })
 
 watch(visibleLayerOnTop, () => {
-    $nextTick(slice())
+    nextTick(slice)
 })
 
 onMounted(() => {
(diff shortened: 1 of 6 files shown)


src/modules/map/components/CompareSlider.vue
```diff
@@ -39,9 +41,40 @@ export default {
...
+                })
+            if (olLayer) {
+                olLayer.on('prerender', this.onPreCompose)
+
+                olLayer.on('postrender', this.onPostCompose)
+            }
+        },
+        onPreCompose(event) {
+            const ctx = event.context
+            const width = ctx.canvas.width * this.compareRatio + 20
```
pakb, line 68: it might be good to explain why this 20 (I imagine it's the half width of the element with the handle?)

src/modules/map/components/CompareSlider.vue
```diff
@@ -53,6 +86,7 @@ export default {
                 event.type === 'touchmove' ? event.touches[0].pageX : event.pageX
             // THIS DIES ON CLIENT SMALLER THAN SCREEN
```
pakb, line 87: what does that mean? shouldn't we address this?

src/modules/map/components/CompareSlider.vue
pakb, line 1: We are moving to Vue's Composition API for component description as of late, could you transform this one into a Composition API approach in the process of editing it?

See our README.md for more informations

src/modules/map/components/CompareSlider.vue
ltkum: I'll put a comment to inform about this and yes, this is to allow the layer to reach the red bar of the component.

src/modules/map/components/CompareSlider.vue
ltkum: I'm aware of it and I'll work on correcting it before merging. Here is a description of the problem :

When dragging and dropping the component, it takes the position of the cursor compared to the screen rather than to the page. So if your web browser is not occupying the whole screen, it works, but your cursor position is not where the component will end.

I've tried with event.pageX, screen and client and in each case it takes the screen as a whole rather than the width of the browser.

src/modules/map/components/CompareSlider.vue
ltkum: Sure, I can do that. If I have questions about it I'll come to you.

src/modules/map/components/CompareSlider.vue
ltkum: just for information : this has been solved

src/modules/map/components/CompareSlider.vue
```diff
@@ -10,27 +10,65 @@ const props = defineProps({
         default: window.innerWidth,
     },
 })
-
+const getMap = inject('getMap')
 const { clientWidth } = toRefs(props)
+const olMap = ref(null)
```
pakb, line 15: Do not put the map in a ref, that will wrap it around a Vue Reactivity proxy.
As the Map instance is quite big (and has a lot of attributes/functions) this will be very CPU intensive, and isn't needed.

Whenever you need the map instance, call the inject function and checks that you have a valid instance before doing stuff

If you need to keep track of things you have already linked to an event, a binding function typically returns an identifier, or callback function, so that you may remove it later (check https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#on )

src/modules/map/components/CompareSlider.vue
```diff
@@ -10,27 +10,65 @@ const props = defineProps({
...
     compareRatio.value = storeCompareRatio.value
+    slice()
 })
 
+function slice() {
+    if (olMap.value) {
+        olMap.value.un('prerender', onPreCompose)
+        olMap.value.un('postrender', onPostCompose)
+    }
+    olMap.value = getMap()
```
pakb, line 45: As said above, no storing of big object such as the map in a ref, that is a huge performance issue

src/router/storeSync/storeSync.config.js
```diff
@@ -104,6 +104,15 @@ const storeSyncConfig = [
...
 
+    new SimpleUrlParamConfig(
+        'compare_ratio',
+        'setCompareRatio',
+        'setCompareRatio',
+        (store) => store.state.ui.compareRatio,
+        false,
+        Number,
+        COMPARE_SLIDER_DEFAULT_VALUE
+    ),
```
pakb, line 115: It's already there, below, please do not add it twice

pakb (changes requested):

pakb (commented):
I also couldn't get a cut layer when I add the URL param. I need to move the slider so that it cuts the top layer.
Please check that you already apply the `slice()` whenever the component is mounted (or whenever you have a valid compare ratio)

src/modules/map/components/CompareSlider.vue
```diff
@@ -99,16 +100,23 @@ function onPostCompose(event) {
...
-function grabSlider() {
+function grabSlider(event) {
     window.addEventListener('mousemove', listenToMouseMove, { passive: true })
     window.addEventListener('touchmove', listenToMouseMove, { passive: true })
     window.addEventListener('mouseup', releaseSlider, { passive: true })
     window.addEventListener('touchend', releaseSlider, { passive: true })
+    compareSliderOffset.value =
+        event.type === 'touchstart'
+            ? event.touches[0].clientX - compareRatio.value * clientWidth.value
+            : event.clientX - compareRatio.value * clientWidth.value
```
pakb, line 111: could you split that into `if...else` statement? it is becoming quite big to be a ternary operator (hard to read what's happening)

src/modules/map/components/CompareSlider.vue
```diff
@@ -99,16 +100,23 @@ function onPostCompose(event) {
...
 }
 
 function listenToMouseMove(event) {
-    const currentPosition = event.type === 'touchmove' ? event.touches[0].clientX : event.clientX
-    compareRatio.value = round(currentPosition / clientWidth.value, 2)
+    const currentPosition =
+        event.type === 'touchmove'
+            ? event.touches[0].clientX - compareSliderOffset.value
+            : event.clientX - compareSliderOffset.value
```
pakb, line 118: same here

src/modules/map/components/CompareSlider.vue
```diff
@@ -10,44 +11,121 @@ const props = defineProps({
...
+
+        postComposeKey.value = getMap()
+            ?.getAllLayers()
+            .find((layer) => {
+                return (
+                    visibleLayerOnTop.value && layer.get('id') === visibleLayerOnTop.value.getID()
+                )
+            })
+            .on('postrender', onPostCompose)
+    }
```
pakb, line 79: you should create a variable with the result of this `getMap()?.getAllLayers().find(...)` as you are reusing it twice afterward

src/modules/map/components/CompareSlider.vue
```diff
@@ -10,44 +11,121 @@ const props = defineProps({
...
+}
+
+function onPreCompose(event) {
+    const ctx = event.context
+    // the offset is there to ensure we get to the slider line, and not the border of the element
+
+    const width =
+        compareRatio.value < 1.0 && compareRatio.value > 0.0
+            ? ctx.canvas.width * compareRatio.value + 20
+            : ctx.canvas.width
```
pakb, line 90: this could also be transformed into a `if...else` block, easier to read IMO

src/modules/map/components/CompareSlider.vue
```diff
@@ -10,44 +11,121 @@ const props = defineProps({
         default: window.innerWidth,
     },
 })
-
+const getMap = inject('getMap')
 const { clientWidth } = toRefs(props)
```
pakb, line 15: client width doesn't seem to be updated correctly, as if I'm resizing the window the slider behave in a weird way.
Maybe you could directly get this info from the store, by reading `state.ui.width`?

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,53 +1,109 @@
...
 
-function grabSlider() {
+function slice() {
+    if (preComposeKey.value != null && postComposeKey.value != null) {
+        unByKey(preComposeKey.value)
+        unByKey(postComposeKey.value)
+        preComposeKey.value = null
+        postComposeKey.value = null
+    }
+    const aTopLayerEXist = getMap()
```
pakb, line 46: that's not really a boolean, maybe the name should reflect its content better, e.g. `topVisibleLayer` or something like that

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,53 +1,109 @@
...
+            event.touches[0].clientX - compareRatio.value * clientWidth.value
+    } else {
+        compareSliderOffset.value = event.clientX - compareRatio.value * clientWidth.value
+    }
 }
 
 function listenToMouseMove(event) {
-    const currentPosition = event.type === 'touchmove' ? event.touches[0].pageX : event.pageX
-    compareRatio.value = round(currentPosition / clientWidth.value, 2)
+    var currentPosition
```
pakb, line 91: don't use `var` keyword anymore (that's old school JS), use `let` instead

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,53 +1,109 @@
...
+        postComposeKey.value = null
+    }
+    const aTopLayerEXist = getMap()
+        ?.getAllLayers()
+        .find((layer) => {
+            return visibleLayerOnTop.value && layer.get('id') === visibleLayerOnTop.value.getID()
+        })
+    if (aTopLayerEXist) {
+        preComposeKey.value = aTopLayerEXist.on('prerender', onPreCompose)
+        postComposeKey.value = aTopLayerEXist.on('postrender', onPostCompose)
```
pakb, line 53: you might want to match the event name with your function's name, meaning `onPreRender` and `onPostRender`

pakb (commented):
Can you clean up the git log, and squash all irrelevant / WIP git commits into one or two commits for this work?

Also please use the JIRA ticket identifier when you commit stuff that is related to a JIRA task

src/modules/map/components/CompareSlider.vue
ltkum: true

src/modules/map/components/CompareSlider.vue
ltkum: The issue with this is that `let` and `const` are both scoped, which means if I want to use let, I have to repeat the code twice within the if - else

We have three options here

1. Revert to a `const` with a ternary operator
2. Use `var`
3. add the code that use the variable in both the if and the else

pakb: if you use `let` that is the same thing as ` var`. `var` was the old way of describing a variable, without telling if it would later change or not. Hence the `let` and `const` distinction.

If you write the exact same code as there, but replace `var` with `let` it will function the same. I don't see the problem

ltkum: sorry, I was a bit dumb

let is indeed scoped, but not in the way I thought. It does indeed work like this

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,53 +1,109 @@
...
+
+    ctx.save()
+    ctx.beginPath()
+    ctx.rect(0, 0, width, ctx.canvas.height)
+    ctx.clip()
+}
+
+function onPostRender(event) {
+    event.context.restore()
+    getMap().render()
```
pakb, line 74: I'm not sure this force render is useful, as you are in a post render hook. That might re-render, and then re-trigger this hook

src/modules/map/components/CompareSlider.vue
```diff
@@ -1,53 +1,109 @@
...
 })
+const visibleLayerOnTop = computed(() => store.getters.visibleLayerOnTop)
 
 watch(storeCompareRatio, (newValue) => {
     compareRatio.value = newValue
+    slice()
+})
+
+watch(visibleLayerOnTop, () => {
+    nextTick(slice())
```
pakb, line 31: Not sure this works as intended if you call the slice instead of passing it's reference.
You maybe should change that to either `nextTick(slice)` or `nextTick(() => slice())`

pakb (commented):
There's a little bug when you drag a slider to the edges (on the width axis) of the map

- on the left side : the slider isn't taken into account starting a couple pixels from the side, if you release it it is then removed
- on the right side : it is possible to move the slider "out of range" and out of the visible map, if you release it it then disappear

There should be some hard limits on the value the slider can have when it is drag and dropped

src/modules/map/components/CompareSlider.vue
ltkum: I'll remove it and check if any problems arise

src/modules/map/components/CompareSlider.vue
ltkum: it seemed to work, but it works just as well with `nextTick(slice)`, so might as well use it the intended way.

pakb (approved):

PR #542, opened 2023-11-21
BGDIINF_SB-3109: add compare slider menu component

ltkum:
Issue : we needed a menu component to toggle the compare slider correctly

Fix : 
The menu component exists, is visible only when there is at least one layer selected, toggles the compare slider availability and slice the content as soon as it's activated.

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-3109-add-compare-slider-menu-component/index.html)

### DIFF
--- a/src/modules/map/components/CompareSlider.vue
+++ b/src/modules/map/components/CompareSlider.vue
@@ -69,7 +69,7 @@ function onPreRender(event) {
     // the offset is there to ensure we get to the slider line, and not the border of the element
     let width = ctx.canvas.width
     if (compareRatio.value < 1.0 && compareRatio.value > 0.0) {
-        width = ctx.canvas.width * compareRatio.value + 20
+        width = ctx.canvas.width * compareRatio.value
     }
 
     ctx.save()
@@ -103,12 +103,12 @@ function listenToMouseMove(event) {
         currentPosition = event.clientX - compareSliderOffset.value
     }
     // we ensure the slider can't get off the screen
-    if (currentPosition < 1) {
-        currentPosition = 1
+    if (currentPosition < 14) {
+        currentPosition = 14
     }
     // same on the other side, but with also the idea of keeping the cartes completely in the screen
-    if (currentPosition > clientWidth.value - 36) {
-        currentPosition = clientWidth.value - 36
+    if (currentPosition > clientWidth.value - 14) {
+        currentPosition = clientWidth.value - 14
     }
     compareRatio.value = round(currentPosition / clientWidth.value, 3)
     getMap().render()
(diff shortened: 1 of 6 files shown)


src/modules/map/components/CompareSlider.vue
```diff
@@ -28,6 +29,9 @@ watch(storeCompareRatio, (newValue) => {
 })
 
 watch(visibleLayerOnTop, () => {
+    if (!visibleLayerOnTop.value) {
```
ltkum, line 32: should we make this test "visibleLayerOnTop.value === null" ?

pakb (commented):
I would keep the "compare" visible in the menu, as it's the case now on map.geo.admin.ch

You can activate it "in advance" and as soon as a layer is added, the slider shows up and cuts the top layer. It then doesn't deactivate when the layer is removed, if you add back another it comes back up

I also tried a little bit to toy with it with a mobile device, and it doesn't respond to touch event. That might be addressed in another PR on your work branch though

ltkum:
I'll be checking this touch event not working. If I can get it to work before the reviews are over, I'll integrate it with this PR. Otherwise I'll do it after the merge.

ltkum:
I found the issue with the touch event and it seems to work on my own phone, feel free to test

Also, it seems that depending on where the compare slider is, on mobile, there is a tiny gap between the line and the right arrow. I'm keeping a note on it and I'll do a css cleanup on the work branch once everything that's functional has been done.

src/modules/menu/components/advancedTools/MenuAdvancedToolsList.vue
```diff
@@ -1,3 +1,52 @@
...
+}
+
+function onToggleImportCatalogue() {
+    store.dispatch('toggleImportCatalogue')
+}
+
+function onToggleCompareSlider() {
+    // this allows us to store the previous compare ratio while making
+    // the Compare Slider invisible.
+    store.dispatch('setCompareRatio', -storeCompareRatio.value)
```
pakb, line 37: you could give a `null` here, and in the store handle this and set it back to the default value (instead of null).
Otherwise the URL still contains the compare value, event if de-activated.

src/store/modules/ui.store.js
```diff
@@ -239,10 +239,11 @@ export default {
...
-                The logic is, if we try to set a value outside of the range at which the compare
-                slider would be visible, we set it to its default value instead
+                The logic is, we want the compare ratio to be either in its visible range,
+                which is 0.001 to 0.999, and it's "storage range" (-0.001 to -0.999). If
+                we are not within these bounds, we revert to the default value (-0.5)
             */
-            if (0.0 < value && value < 1.0) {
+            if (-1.0 < value && value < 1.0 && value != 0.0) {
                 commit('setCompareRatio', value)
             } else {
```
pakb, line 248: if you follow what I suggested above, you might want to revert this to checking between 0 and 1

pakb (approved):

src/modules/menu/components/advancedTools/MenuAdvancedToolsList.vue
ltkum: that's a good idea

PR #543, opened 2023-11-22
Improve mobile navigation in 3D

pakb:
I found that most (of not all) Cesium sandcastles (online examples) have this CSS rule to block touch action, it fixes all issues I experienced with panning while on a mobile device!

I also de-activated the location popup (right-click) on mobile, as it wasn't very consistent how to trigger a right click with a touch screen.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_cesium_mobile_nav_improvements/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumMap.vue
+++ b/src/modules/map/components/cesium/CesiumMap.vue
@@ -1,5 +1,16 @@
 <template>
-    <div v-if="isProjectionWebMercator" id="cesium" ref="viewer" data-cy="cesium-map">
+    <div
+        v-if="isProjectionWebMercator"
+        id="cesium"
+        ref="viewer"
+        class="cesium-widget"
+        data-cy="cesium-map"
+        @touchstart.passive="onTouchStart"
+        @touchmove.passive="clearLongPressTimer"
+        @touchend.passive="clearLongPressTimer"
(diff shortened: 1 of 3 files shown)


src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -528,6 +535,18 @@ export default {
...
             this.clearAllSelectedFeatures()
         },
+        onTouchStart(event) {
+            this.clearLongPressTimer()
+            if (event.touches.length === 1) {
+                this.contextMenuTimeoutId = setTimeout(() => {
+                    const touch = event.touches[0]
+                    this.onContextMenu(touch)
+                }, 500)
+            }
```
ltshb, line 545: Here it would be good to use a constant `LONG_PRESS_TIMEOUT = 500`

ltshb (commented):
The code looks good thanks. The 3d in mobile is definitively much better.
However le location popup in 3d has still some issues:
- sometimes it disappear on touch stop
- sometimes it also select the whole map

In the PR comment you added a notice that the location popup is disabled in 3d mobile, it is not but it might be the solution, or at least for the moment ?

ltshb (approved):
As discussed in daily, all good for now

PR #544, opened 2023-11-22
BGDIINF_SB-3180: Full support for external WMS and WMTS layer extent

ltshb:
Try to supports all possibles way to define a layer extent based on the
WMS and WMTS specifications. This is required by some of our providers.

Some final rework also on the WMS and WMTS Capabilities parsers as well as fixing WMS version.

Currently openlayer only supports WMS version 1.3.0.

The WMTSCapabilities in openlayer has an issue that affect the following provider:
- https://wmts.geo.bs.ch/1.0.0/WMTSCapabilities.xml

See issue and pr https://github.com/openlayers/openlayers/issues/15363


src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -11,96 +11,100 @@ import proj4 from 'proj4'
...
+                found.layer = layers[i]
+                found.parents = parents
             } else if (layers[i].Layer?.length > 0) {
-                layer = this.findLayer(layerId, layers[i].Layer)
+                found = this._findLayer(layerId, layers[i].Layer, [layers[i], ...parents])
             }
         }
-        return layer
+        return found
     }
```
pakb, line 51: it really not a big deal/improvement, but I see you do not use any `this.xyz` in this function, so if you really wanted to hide it, you could declare it above the class definition as a function. As this function will not be exported it will be hidden to other files 😉

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -140,15 +151,56 @@ export default class WMSCapabilitiesParser {
...
+        }
+
+        if (!this.version) {
+            throw new Error(`No WMS version found in Capabilities of ${this.originUrl.toString()}`)
+        }
+        if (!WMS_SUPPORTED_VERSIONS.includes(this.version)) {
+            throw new Error(
+                `WMS version ${this.version} of ${this.originUrl.toString()} not supported`
+            )
+        }
```
pakb, line 178: shouldn't `ignoreError = true` outputs the same behavior as above, meaning returning an empty `{}` instead of a `throw ...`?

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -171,15 +248,18 @@ export default class WMSCapabilitiesParser {
...
-    getLayerAttribution(layer) {
+    _getLayerAttribution(layer) {
         let title = null
         let url = null
         if (layer.Attribution || this.Capability.Layer.Attribution) {
             const attribution = layer.Attribution || this.Capability.Layer.Attribution
             url = attribution.OnlineResource
-            title = attribution.Title || new URL(attribution.OnlineResource).hostname
+            title =
+                attribution.Title || new URL(attribution.OnlineResource || this.originUrl).hostname
```
pakb, line 258: `new URL` may throw a `TypeError` if the given URL is malformed/invalid, it might be a good idea to drag the `ignoreError` from the parent call here, and deal with any error with a `try...catch` and decide depending on `ignoreError` what to do with them

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -171,15 +248,18 @@ export default class WMSCapabilitiesParser {
...
         } else {
-            title = this.Service.Title || new URL(this.Service.OnlineResource).hostname
+            title =
+                this.Service?.Title ||
+                new URL(this.Service?.OnlineResource || this.originUrl).hostname
```
pakb, line 262: same as above

src/api/layers/WMSCapabilitiesParser.class.js
pakb: or wrap the call to this function in a `try...catch` in the parent, if that make more sense business wise

src/api/layers/WMSCapabilitiesParser.class.js
ltshb: My idea with the `ignoreError = true` was to return as much as possible, e.g. if a capabilities is valid and only one layer element is invalid, then we still parse it and returns layer objects for all other layers. Here if the version is not valid I cannot return anything. But I see that is not 100% clear, I'll have a look if I can improve the code/logic.

src/api/layers/WMSCapabilitiesParser.class.js
ltshb: Done

src/api/layers/WMSCapabilitiesParser.class.js
ltshb: Done

src/api/layers/WMSCapabilitiesParser.class.js
ltshb: Done

src/api/layers/WMSCapabilitiesParser.class.js
ltshb: Done

pakb (approved):

PR #545, opened 2023-11-23
OpenLayers/Drawing components as Composition API

pakb:
breaking changes : 

- vue-i18n has now `legacy: false`, meaning it is no longer usable through `this.$i18n` but only through `useI18n()` composable function
- GeoJSON now load in the same manner as external layers would. Instead of loading the data in the dedicated OL or Cesium component, it is now loaded by a store plugin, whenever such a layer is added (and it as not yet been loaded proper)

Most of the logic of the old `OpenLayersMap` is now split into multiple composable chunks, with a focus on making as many non-mapping framework dependent composable chunks as possible (so that we can re-use them later when rewriting CesiumMap as composable too)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_ol_composable/index.html)


pakb:
will be replaced by smaller PR covering each one or two commits from this work branch

PR #547, opened 2023-11-27
New Release v1.4.0 - #minor

ltshb:
This will deploy on test.map.geo.admin.ch the same state as currently on sys-map.dev.bgdi.ch (the advance tool would be disabled on test.map.geo.admin.ch because the import tool is not yet production ready).
This deployment would totally remove the need of vectortiles that will not be available at the end of the month without a key.


hansmannj (approved):
Thanks :+1:

stebie (changes requested):
To merge on 30th of November

PR #558, opened 2023-12-04
Load KML data in the same fashion GeoJSON data/style are loaded

pakb:
meaning we load these data whenever such a layer is added to the map, through a store plugin.

I also moved the "metadata" attribute that was sitting in AbstractLayer into KMLLayer, as it is only used by this layer so far (and I don't see any use cases where it would be by any other layer type) This way, the addToMap attribute could also be removed and replaced by the same logic as the GeoJSON and external layer async management, meaning the flag isLoading

The logic to add, or not add, the KML layer to the map is now dependent if the KML layer is the one being edited by the drawing module, and if the drawing module is active (no need for the addToMap flag anymore, this information is derived from others we already have)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_load_kml_data_async/index.html)


src/api/files.api.js
```diff
@@ -254,33 +252,69 @@ export const getKmlMetadataByAdminId = (adminId) => {
 }
 
 /**
- * Get KML metadata by fileId
- *
- * @param {string} fileId KML ID
- * @param {string} adminId OPTIONAL KML admin ID
- * @returns {Promise<KmlMetadata>} KML metadata
+ * @param {KMLLayer} kmlLayer
+ * @returns {Promise<KmlMetadata>}
```
ltshb, line 256: A short description here would be good

src/modules/drawing/DrawingModule.vue
```diff
@@ -328,46 +308,45 @@ export default {
...
+                        kmlMetadata.id,
+                        kmlMetadata.adminId,
                         this.$t('draw_layer_label'),
-                        metadata,
+                        kmlData,
+                        kmlMetadata,
                         false, // external
                         // do not add the drawing to the openlayer map overlay yet
                         // to not interfer with the drawing overlay.
                         false // addToMap
```
ltshb, line 332: Update the comment here as `addToMap` doesn't exists anymore

ltshb (approved):
Looks good just two minor comment

PR #560, opened 2023-12-04
Fix an issue when reloading a KML with adminId

pakb:
with the new KML data approach, and as we do not have the metadata of the layer at app startup, the URL param parser couldn't figure out if the drawing module should be shown or not (because of a protection against legacy KML layer being loaded at startup)

So instead I've completely removed any adminId passthrough when parsing legacy KML layer, this way they are never given to the "modern" app, and will not be modified by web-mapviewer (that was the goal of this protection)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_fix_reloading_kml_with_adminid/index.html)

### DIFF
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -1,5 +1,6 @@
 import proj4 from 'proj4'
 
+import { DISABLE_DRAWING_MENU_FOR_LEGACY_ON_HOSTNAMES } from '@/config'
 import { transformLayerIntoUrlString } from '@/router/storeSync/LayerParamConfig.class'
 import { LV95, WEBMERCATOR, WGS84 } from '@/utils/coordinates/coordinateSystems'
 import CustomCoordinateSystem from '@/utils/coordinates/CustomCoordinateSystem.class'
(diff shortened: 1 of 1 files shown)


src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -176,7 +176,11 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
     const urlWithoutQueryParam = window.location.href.substr(0, window.location.href.indexOf('?'))
     window.history.replaceState(window.history.state, document.title, urlWithoutQueryParam)
 
-    if ('adminId' in legacyParams) {
+    // TODO BGDIINF_SB-2685: re-activate once on prod
+    //  we cannot let adminId get through right now because legacy KML will be broken by the new viewer if edited
+    //  (broken in a sense that they will not be usable by mf-geoadmin3 anymore)
+    const absolutelyNot = () => false
+    if (absolutelyNot() && 'adminId' in legacyParams) {
```
ltshb, line 183: Would be actually better to check the hostname, if `map.geo.admin.ch` then true, otherwise false, this way we have less risk forgetting this during the go live.

pakb: but if it doesn't fail, we might forget to do it, don't you think?

ltshb (approved):

PR #563, opened 2023-12-04
BGDIINF_SB-3194: Implement new UI for import external KML/GPX file.

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3194-import-tool-ui-p2/index.html)

### DIFF
--- a/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
+++ b/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
@@ -1,0 +1,71 @@
+<script setup>
+import { ref } from 'vue'
+
+import ImportFileOnlineTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue'
+import ImportFileLocalTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue'
+
+const selectedTab = ref('online')
+</script>
+
+<template>
+    <div data-cy="import-file-content">
+        <ul class="nav nav-tabs" role="tablist">
+            <li class="nav-item" role="presentation">
+                <button
+                    class="nav-link py-1"
+                    :class="{
+                        active: selectedTab === 'online',
+                    }"
+                    data-bs-toggle="tab"
+                    data-bs-target="#nav-online"
+                    type="button"
+                    role="tab"
+                    aria-controls="nav-online"
+                    :aria-selected="selectedTab === 'online'"
+                    data-cy="online-import-btn"
+                    @click="() => (selectedTab = 'online')"
+                >
+                    Online
+                </button>
+            </li>
+            <li class="nav-item" role="presentation">
+                <button
+                    class="nav-link py-1"
+                    :class="{
+                        active: selectedTab === 'local',
+                    }"
+                    data-bs-toggle="tab"
+                    data-bs-target="#nav-local"
+                    type="button"
+                    role="tab"
+                    aria-controls="nav-local"
+                    :aria-selected="selectedTab === 'local'"
+                    data-cy="local-import-btn"
+                    @click="() => (selectedTab = 'local')"
+                >
+                    Local
+                </button>
+            </li>
+        </ul>
+        <div class="tab-content mt-2">
+            <!-- Online Tab -->
+            <ImportFileOnlineTab
+                :class="{ active: selectedTab === 'online', show: selectedTab === 'online' }"
+            />
+            <!-- Local tab -->
+            <ImportFileLocalTab
+                :class="{ active: selectedTab === 'local', show: selectedTab === 'local' }"
+            />
+        </div>
+    </div>
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+.nav-tabs {
+    .nav-link.active {
+        background-color: $primary;
+        color: white;
+    }
+}
+</style>
--- a/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
+++ b/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
@@ -58,10 +58,4 @@ const selectedTab = ref('online')
 
 <style lang="scss" scoped>
 @import 'src/scss/webmapviewer-bootstrap-theme';
-.nav-tabs {
-    .nav-link.active {
-        background-color: $primary;
-        color: white;
-    }
-}
 </style>
--- a/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
+++ b/src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
@@ -1,8 +1,8 @@
 <script setup>
 import { ref } from 'vue'
 
-import ImportFileOnlineTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue'
 import ImportFileLocalTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue'
+import ImportFileOnlineTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue'
 
 const selectedTab = ref('online')
 </script>
@@ -16,14 +16,12 @@ const selectedTab = ref('online')
                     :class="{
                         active: selectedTab === 'online',
                     }"
-                    data-bs-toggle="tab"
-                    data-bs-target="#nav-online"
                     type="button"
                     role="tab"
                     aria-controls="nav-online"
                     :aria-selected="selectedTab === 'online'"
                     data-cy="online-import-btn"
-                    @click="() => (selectedTab = 'online')"
+                    @click="selectedTab = 'online'"
                 >
                     Online
                 </button>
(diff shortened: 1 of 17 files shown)


src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
```diff
@@ -0,0 +1,71 @@
...
+                        active: selectedTab === 'online',
+                    }"
+                    data-bs-toggle="tab"
+                    data-bs-target="#nav-online"
+                    type="button"
+                    role="tab"
+                    aria-controls="nav-online"
+                    :aria-selected="selectedTab === 'online'"
+                    data-cy="online-import-btn"
+                    @click="() => (selectedTab = 'online')"
```
pakb, line 26: I don't think you need this function wrapper, you can simply
```
@click="selectedTab = 'online'"
```

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
```diff
@@ -0,0 +1,71 @@
...
+    <div data-cy="import-file-content">
+        <ul class="nav nav-tabs" role="tablist">
+            <li class="nav-item" role="presentation">
+                <button
+                    class="nav-link py-1"
+                    :class="{
+                        active: selectedTab === 'online',
+                    }"
+                    data-bs-toggle="tab"
+                    data-bs-target="#nav-online"
```
pakb, line 20: these `data-bs-...` aren't really useful in your case, as you are not relying on the JS from Bootstrap to create the necessary logic to show or hide the tabs, you are doing it yourself by adding or removing the `show` class.
So this `data-bs-xxx` can be safely removed (and the ID on the child too)

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
```diff
@@ -0,0 +1,71 @@
...
+        <div class="tab-content mt-2">
+            <!-- Online Tab -->
+            <ImportFileOnlineTab
+                :class="{ active: selectedTab === 'online', show: selectedTab === 'online' }"
+            />
+            <!-- Local tab -->
+            <ImportFileLocalTab
+                :class="{ active: selectedTab === 'local', show: selectedTab === 'local' }"
+            />
+        </div>
```
pakb, line 59: Wouldn't it be better to only load the component matching the current selection? Something like

```
<script setup>
const tabComponent = computed(() => {
    if (selectedTab === 'online') {
        return ImportFileOnlineTab
    }
    return ImportFileLocalTab
})
</script>
<template>
    <component :is="tabComponent" class="active show" />
<template>
```

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
```diff
@@ -0,0 +1,71 @@
...
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+.nav-tabs {
+    .nav-link.active {
+        background-color: $primary;
+        color: white;
+    }
+}
```
pakb, line 70: It's usually better to add you own classes when you want to add something to some Bootstrap's elements, your scoped style is still applied to children, so it might break their styling down the line if you overwrite some Bootstrap rules here.

src/modules/menu/components/advancedTools/ImportFile/ImportFileButtons.vue
```diff
@@ -0,0 +1,74 @@
...
+        case 'default':
+        default:
+            return 'import'
+    }
+})
+const isLoading = computed(() => buttonState.value === 'loading')
+</script>
+
+<template>
+    <div class="mt-2 d-grid gap-2 d-md-flex justify-content-md-center">
```
pakb, line 44: Regarding the use of `mt-2` here : as a good rule when building components : It's better to have all the positioning and margin conundrum at the parent level (which is the one placing the component side by side or in a container) and let the component itself just declare whatever it wants to show in the UI.

This way, if it is then re-used somewhere else, there's no need to cancel some margin or padding added here, it's a "blank canvas" that can be shaped however you want in each places it's called/used

src/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue
```diff
@@ -0,0 +1,118 @@
...
+import { useI18n } from 'vue-i18n'
+
+import log from '@/utils/logging'
+import { handleFileContent } from '@/modules/menu/components/advancedTools/ImportFile/utils'
+import ImportFileButtons from '@/modules/menu/components/advancedTools/ImportFile/ImportFileButtons.vue'
+
+const BTN_RESET_TIMEOUT = 3000 // milliseconds
+
+const LOCAL_UPLOAD_ACCEPT = '.kml,.KML,.gpx,.GPX'
+const LOCAL_UPLOAD_MAX_SIZE = 250 * 1000 * 1000 // 250mb
```
pakb, line 12: I'm not sure we shouldn't be `250 * 1024 * 1024` here instead?

src/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue
```diff
@@ -0,0 +1,118 @@
...
+            />
+            <input
+                type="text"
+                class="form-control import-input rounded-end import-file-local-input"
+                :class="{ 'is-valid': isValid, 'is-invalid': isInvalid }"
+                :placeholder="i18n.t('no_file')"
+                :value="filePathInfo"
+                readonly
+                required
+                @click="() => $refs.importFileLocalInput.click()"
```
pakb, line 97: that might not work in the Composable world, I don't think `$refs` is still used

src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
```diff
@@ -0,0 +1,112 @@
...
+import { computed, ref } from 'vue'
+import { useI18n } from 'vue-i18n'
+import axios, { AxiosError } from 'axios'
+
+import { isValidUrl } from '@/utils/utils'
+import log from '@/utils/logging'
+import { handleFileContent } from '@/modules/menu/components/advancedTools/ImportFile/utils'
+import ImportFileButtons from '@/modules/menu/components/advancedTools/ImportFile/ImportFileButtons.vue'
+
+const BTN_RESET_TIMEOUT = 3000 // milliseconds
```
pakb, line 11: as this logic is also present in the other type of input, maybe it would be a good idea to use some composable to isolate it and describe it only once (or move this timeout management straight into the `ImportFileButtons` component)

src/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue
ltshb: Actually it works, but this is a copy paste from the old option api component. I'll change it to be more composable.

ltshb: Done

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
ltshb: I see your point, however I think we should change these for all nav tabs in order to be consistent with colors. So I've moved this in the /.src/scss/webmapviewer-bootstrap-theme.scss file

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
ltshb: Would this not be less performant when switching back and forth on the tabs ? Also now if you write a url in online tab and switch to file and back to online what you typed is kept which is kind of a nice feature I think

src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
```diff
@@ -0,0 +1,110 @@
...
+        <form class="input-group d-flex needs-validation">
+            <input
+                type="text"
+                class="form-control import-input"
+                :class="{
+                    'rounded-end': !fileUrl?.length,
+                    'is-valid': isValid,
+                    'is-invalid': isInvalid,
+                }"
+                :placeholder="$t('import_online_placeholder')"
```
pakb, line 86: that sort of work (I don't know how as we've deactivated the legacy mode), but it might better to replace it by `:placeholder="i18n.t('import_online_placeholder')"` so that we may not be suprised by a future update of `vue-i18n` that removes the `$t` support altogether

src/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue
```diff
@@ -0,0 +1,119 @@
...
+}
+</script>
+
+<template>
+    <div id="nav-local" class="tab-pane fade" role="tabpanel" aria-labelledby="nav-local-tab">
+        <form class="input-group rounded needs-validation">
+            <button
+                class="btn btn-outline-secondary"
+                type="button"
+                @click="() => $refs.importFileLocalInput.click()"
```
pakb, line 77: you can also change this one to be more composable if you want 😉

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
pakb: The memory of the URL is a nice feature yes, we can keep it this way

pakb (approved):

PR #565, opened 2023-12-06
Rework of some of the infobox module components into Composition API

pakb:
as I was trying to fix a couple issues :

- there was no more combo infobox (when the floating tooltip is disabled, but an editable feature was selected, the detail of the feature was not shown by the profile plot)
- some props were unnecessary as it could be grabbed through the state directly
- the look and feel of the title and description of features when readonly were not clear enough to signify they were read only (no changes visually), it is now different

I also took the opportunity to make a huge simplification/gathering of all drawing tests into one file, this should enhance browsing these tests, and their run speed too.

[Test link](https://sys-map.dev.bgdi.ch/preview/infobox_composable/index.html)


pakb:
There's still an issue with the Infobox's height when having a line/measure opened
The profile is half-hidden. The rest looks good IMO, and could be reviewed as it is

ltshb:
Issues found in testing

- [ ] After editing a drawing we cannot select anymore any features, I observed this error 
image. Removing the drawing solve the issue


Other issues that I noticed but not due on this PR, needs to discuss if fixing here or create a ticket.
- [ ] In 3d mode the selected feature is not marked, so with the infobox it is not possible to find out which feature is selected (issue already present before the PR, could be done in separate PR/ticket)
image

- [ ] Clicking on the header of the infobox minimize it, but from a user perspective this is not intuitive. If a user unwilligly click on the title bar, it is then minimize and the intuitive way to unminimize would be to click on the caret button but this put the infobox to floating. This issue is already present and can/should be done in separate PR/ticket

ltshb:
Another new issue

- [ ] The area is now undefined (new issue)
image

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -23,16 +136,19 @@
...
             <div v-if="isFeatureLine">
                 <font-awesome-icon :icon="['far', 'square']" />
-                {{ lengthRounded }} {{ length > 100 ? 'km' : 'm' }}
+                {{ roundValueIfGreaterThan(length, 100, 1000) }} {{ length > 100 ? 'km' : 'm' }}
             </div>
             <div v-if="isFeaturePolygon">
                 <font-awesome-icon :icon="['far', 'square']" class="bg-secondary text-secondary" />
-                {{ areaRounded }} {{ area > 10000 ? 'km' : 'm' }}<sup>2</sup>
+                {{ roundValueIfGreaterThan(area, 1000, 100000) }} {{ area > 10000 ? 'km' : 'm'
+                }}<sup>2</sup>
```
ltshb, line 151: `roundValueIfGreaterThan` could be integrated in the `length` and `area` computed property, might be more efficient and would be more readable in the template.

src/modules/map/components/LocationPopupCopySlot.vue
```diff
@@ -1,102 +1,121 @@
...
+let copyTooltip = null
+
+onMounted(() => {
+    copyTooltip = tippy(copyButton.value, {
+        arrow: true,
+        placement: 'right',
+        hideOnClick: false,
+        // no tooltip on mobile/touch
+        touch: false,
+    })
```
ltshb, line 53: The `allowHTML: true,` is required due the `&nbsp;` in the french translation of `copy_done`

ltshb (approved):
👍🏼

PR #574, opened 2023-12-11
Fixing error raised when displaying the Void layer

pakb:
as its value is "null", this was not taken into account for z-index calculation (it was assumed visible layers wouldn't hold a null value)

I also simplified the z-index calc function, removing the part that was checking in the BG layers first, as they are added to the computed/variable "visibleLayersHoldingAZIndex" anyway

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_zindex_calc_issue_with_void_layer/index.html)


ltshb (approved):
👍🏼

PR #576, opened 2023-12-12
Catalogue button circle-plus-minus solid 

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-bgdiinf_sb-3194-import-tool-ui-btn-2/index.html)


ltshb:
We decided to go for fontawesome pro  and use the solid version until we have the pro version the solid version is implemented in #573

PR #581, opened 2023-12-18
BGDIINF_SB-3211: Fix zoom after going back to 2D

pakb:
Also removing/simplifying the geolocation tracking/untracking when the map is dragged. As we now have a central composable that handles all map interaction, this state isMapBeingDragged was not necessary anymore, it could be handle in the mouse click composable.

The zoom level was not set properly when the view (dependent on the projection) was changed in the OL composable.

I also removed the spinner backdrop from the drawing module when it closes, replacing it by a spinner inside the close button (to give the user a feedback something is happening anyway)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-3211_fix_zoom_after_going_back_2d/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -10,13 +10,11 @@ import DrawingToolbox from '@/modules/drawing/components/DrawingToolbox.vue'
 import DrawingTooltip from '@/modules/drawing/components/DrawingTooltip.vue'
 import { DrawingState } from '@/modules/drawing/lib/export-utils'
 import useKmlDataManagement from '@/modules/drawing/useKmlDataManagement.composable'
-import BlackBackdrop from '@/modules/menu/components/BlackBackdrop.vue'
 import log from '@/utils/logging'
 
 const olMap = inject('olMap')
 
 const drawingInteractions = ref(null)
-const isClosing = ref(false)
 const isNewDrawing = ref(true)
 
 const store = useStore()
@@ -119,7 +117,7 @@ function removeLastPointOnDeleteKeyUp(event) {
 }
 
 async function closeDrawing() {
-    isClosing.value = true
+    await store.dispatch('setShowLoadingBar', true)
 
     log.debug(
         `Closing drawing menu: isModified=${isDrawingModified.value}, isNew=${isNewDrawing.value}, isEmpty=${isDrawingEmpty.value}`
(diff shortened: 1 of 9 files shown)


src/modules/drawing/DrawingModule.vue
```diff
@@ -119,7 +117,7 @@ function removeLastPointOnDeleteKeyUp(event) {
 }
 
 async function closeDrawing() {
-    isClosing.value = true
+    await store.dispatch('setShowLoadingBar', true)
```
ltshb, line 120: We should not have any spinner when closing the drawing like we used to have. We used to be able to close without any blocking action, this should still be then case. 
The original loading bar which was the red bar moving from right to left was much better as a spinner I think.

src/modules/drawing/components/DrawingHeader.vue
```diff
@@ -1,17 +1,32 @@
...
+        <button
+            class="drawing-header-close-button btn btn-dark"
+            :disabled="isClosing"
+            @click="onClose"
+        >
+            <FontAwesomeIcon v-if="isClosing" icon="spinner" spin />
+            <span v-else>
+                <FontAwesomeIcon class="icon me-2" :icon="['fas', 'arrow-left']" />
+                {{ i18n.t('draw_back') }}
+            </span>
```
ltshb, line 29: See question above we should remove totally the spiner when closing the menu, saving is done in background and should not be seen by the user. This provide a much better user experience.

ltshb (commented):
The zoom fix is good, however the closing of drawing is not good. This could/should be in a separate PR or document as separate commit.

src/modules/drawing/DrawingModule.vue
ltshb: I would separate this functionality from this PR and only push merge the zoom issue. And create a separate PR for the drawing spinner, for this one I think it requires some UI refinement/discussion. 
We should also try to keep the PR small and not mixing everything together, keeping bug fix in separate PR.

pakb: It got mixed because I intended to use this backdrop/sprinner when going back 2D, but in the end I used the loading bar (as we discussed)
I can separate this into two work package, although I think this version of the close drawing is better than what we have on `develop`

ltshb (approved):

PR #583, opened 2023-12-18
BGDIINF_SB-3153 : adding debug tools to see layer extents (and tiles)

pakb:
this should help debug our "zoomToExtent" function with some needed visual feedback

I've done a little bit of rework on zoomToExtent, I think it now works "OK", maybe you can give me a couple examples where it's not the case @ltshb ? (with the new debug tool to see extents it's quite easy to see if the zoom was totally off)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_bgdiinf_sb-3153_fix_zoom_to_extent/index.html)


ltshb (commented):
The zoom to extent seems at the limit and from a user perspective it would be quite better to have 1 zoom level out more.

I've seen that I broke the external layer, they are not shown anymore I'll fix it (I should do the e2e tests for external layer...).

src/store/modules/position.store.js
```diff
@@ -220,7 +220,12 @@ const actions = {
                     y: centerOfExtent[1],
                 })
             }
-            const newResolution = (extent[1][0] - extent[0][0]) / rootState.ui.width
+            // taking the biggest resolution calculated with either the width or height of the screen
+            // so that it fits to the screen
+            const newResolution = Math.max(
+                (extent[1][0] - extent[0][0]) / rootState.ui.width,
+                (extent[1][1] - extent[0][1]) / rootState.ui.height
+            )
```
ltshb, line 228: Maybe as a rule of thumb taking a resolution of one lower would be good, or removing the space taken by the menu in desktop view for the ui width.

ltshb:
Very nice tools for debugging 👍🏼

ltshb (approved):

PR #584, opened 2023-12-18
BGDIINF_SB-3194: Fixed external layer

ltshb:
This reverts commit 6a3ec3d1714e0561b1ed2d13e4a889cc8d797fd5.

Some external provider original URL return a 302 and are not value for the GetMap.

Therefore keep the URL provided by the GetCapabilities as base URL.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-3194-external-layer/index.html)

### DIFF
--- a/src/api/layers/ExternalGroupOfLayers.class.js
+++ b/src/api/layers/ExternalGroupOfLayers.class.js
@@ -58,7 +58,9 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
 
     getID() {
         // format coming from https://github.com/geoadmin/web-mapviewer/blob/develop/adr/2021_03_16_url_param_structure.md
-        return `GRP|${this.baseURL}|${this.externalLayerId}`
+        // NOTE we don't differentiate between group of layers and regular WMS layer. This differentiation was not
+        // done the legacy parameter and is not required.
+        return `WMS|${this.baseURL}|${this.externalLayerId}`
     }
     clone() {
         let clone = super.clone()
--- a/src/api/layers/WMSCapabilitiesParser.class.js
+++ b/src/api/layers/WMSCapabilitiesParser.class.js
@@ -35,7 +35,7 @@ export default class WMSCapabilitiesParser {
             throw new Error(`Failed to parse WMTS Capabilities: invalid content: ${error}`)
         }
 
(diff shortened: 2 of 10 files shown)


src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -58,7 +58,9 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
 
     getID() {
         // format coming from https://github.com/geoadmin/web-mapviewer/blob/develop/adr/2021_03_16_url_param_structure.md
-        return `GRP|${this.baseURL}|${this.externalLayerId}`
+        // NOTE we don't differentiate between group of layers and regular WMS layer. This differentiation was not
+        // done the legacy parameter and is not required.
+        return `WMS|${this.baseURL}|${this.externalLayerId}`
```
pakb, line 63: Why reverting the distinction between WSM and GRP of layers in the process?
Couldn't we, technically, also have a group of WMTS layer?

ltshb: On the old viewer we did not have any distinction, see comment above. Having the distinction would need a dirty hack to handle legacy group of layers, see commit https://github.com/geoadmin/web-mapviewer/pull/584/commits/1db3dce307efaa50ce8c442175d1dd2b35ed9029

pakb: Couldn't we let it as `isLoading` at startup, load the getCapabilites (we've got one even within the legacy param, right?) and check what it really is?

I'd really like to keep this concept of groups of layers, as we might move to other type of external layers (WMTS, vector?)

ltshb: But that would still require the dirty hack for legacy url that cannot be handle in the legacy params plugin then. I don't think we need the distinction in the url if this is a group of layer or not, unless we support in future mix of group of layers, but I don't this that would work, having a url that contains both WMTS and WMS.

ltshb: If you look at the code not having this distinction in the URL make everything simpler, less hacky and error prone in my opinion.

pakb (approved):

PR #586, opened 2023-12-20
BGDIINF_SB-3194: Improved error handling of external layer

ltshb:
Now if an external layer in a perma link is invalid, then the layer in the
active layer is marked as errored.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-bgdiinf_sb-3194-external-layer-err/index.html)


src/store/plugins/external-layers.plugin.js
```diff
@@ -62,13 +65,9 @@ async function updatedWMSLayerAttributes(externalLayer, projection) {
         externalLayer.externalLayerId,
         projection,
         externalLayer.opacity,
-        externalLayer.visible
+        externalLayer.visible,
+        false /* ignoreError */
```
pakb, line 69: might be better to tell here that you currently want error to be shown. Even though it's correct that this param is called ignoreError, it could be understood here that you are choosing to ignore errors there when you are instead doing the opposite

src/store/plugins/external-layers.plugin.js
```diff
@@ -78,12 +77,8 @@ async function updatedWMTSLayerAttributes(externalLayer, projection) {
...
```
pakb, line 81: same

pakb (approved):
little i18n fix with this French translation we talked about, and we are good to go

PR #589, opened 2024-01-02
PB-88: Import tool/Geo Catalogue checkbox action changed to add/remove layer

ltshb:
The checkbox in the geo catalogue and import tool has been change to add/remove
layer in active layers. Previously it did add the layer if it was not in active
layers and was toggling the layer visibility.

In order to keep a clean and intuitive user interface, the layer config is kept
in one place, which means in active layer, where the visibility, opacity and timestamp
can be configured. In the catalogue we can define which layer can be added or removed.

This also allow click on wrong layer to be removed from the same view which was
previously not possible.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-88-import-improvement/index.html)


tests/e2e-cypress/integration/legacyParamImport.cy.js
pakb, line 1: was that supposed to come with the search bar work? 🙈

pakb (approved):

PR #590, opened 2024-01-02
PB-73: Added geo catalogue and import tool search bar

ltshb:
Now we have a search bar for the geocatalog.

Also improved menu items colors

The active layers and geocatalog menu were always white which make it hard
to differentiate the menu title from the menu content.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-73-search-geo-catalogue/index.html)

### DIFF
--- a/src/modules/menu/components/LayerCatalogue.vue
+++ b/src/modules/menu/components/LayerCatalogue.vue
@@ -18,6 +18,7 @@ const { layerCatalogue, compact } = defineProps({
 const store = useStore()
 
 const searchText = ref('')
+const searchInput = ref(null)
 
 function clearPreviewLayer() {
     if (store.state.layers.previewLayer) {
@@ -35,6 +36,7 @@ function onSearchInput(event) {
 
 function clearSearchText() {
     searchText.value = ''
+    searchInput.value.focus()
 }
 </script>
 
@@ -45,6 +47,7 @@ function clearSearchText() {
                 <FontAwesomeIcon :icon="['fas', 'search']" />
             </span>
             <input
+                ref="searchInput"
                 type="text"
                 class="form-control"
                 :class="{ 'rounded-end': !searchText.length }"
--- a/src/modules/menu/components/LayerCatalogue.vue
+++ b/src/modules/menu/components/LayerCatalogue.vue
@@ -1,10 +1,10 @@
 <script setup>
-import { ref } from 'vue'
+import { ref, toRefs } from 'vue'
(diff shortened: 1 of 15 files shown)


ltclm:
@ltshb great improvement of the user experience.
one small remark after a few tests with the catalogue search.
especially on bigger catalogues like wms.geo.admin.ch the search is a bit laggy on the first characters. maybe it would be better to activate the search not directly but p.e. after the 3 character has been typed.

ltshb:
> @ltshb great improvement of the user experience. one small remark after a few tests with the catalogue search. especially on bigger catalogues like wms.geo.admin.ch the search is a bit laggy on the first characters. maybe it would be better to activate the search not directly but p.e. after the 3 character has been typed.

@ltclm I tried to improved the performance, can you try and let me know what you think ?

ltclm:
thank you @ltshb it is much better now :+1:

rebert (approved):
@ltshb I love it. As you mentioned, it would be a good idea when the search bar only appears when a wms (external source) has been loaded.

src/modules/menu/components/LayerCatalogue.vue
```diff
@@ -13,22 +14,73 @@ const { layerCatalogue, compact } = defineProps({
...
+        <div v-if="showSearchBar" class="input-group input-group-sm d-flex p-1">
+            <span id="searchCatalogueIcon" class="input-group-text">
+                <FontAwesomeIcon :icon="['fas', 'search']" />
+            </span>
+            <input
+                ref="searchInput"
+                type="text"
+                class="form-control"
+                :class="{ 'rounded-end': !searchText.length }"
+                :placeholder="$t('search_in_catalogue_placeholder')"
```
pakb, line 61: please replace `$t` by `i18n.t` and importing it as `const i18n = useI18n()` above 😉

src/utils/utils.js
pakb, line 1: no unit tests to cover this new function? 😉

pakb (approved):
Very cool! Helps greatly search for things in an external layer provider's list!

src/utils/utils.js
ltshb: I've introduced component testing in next PR https://github.com/geoadmin/web-mapviewer/pull/591 and do more testing in the PB-88 ticket

src/modules/menu/components/LayerCatalogue.vue
ltshb: My bad copy paste habits 😉

PR #592, opened 2024-01-03
PB-90: Reduce shortlink location popup calls

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-90-reduce_shortlink_location_popup_calls/index.html)

### DIFF
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -245,6 +245,7 @@ function toggleEmbedSharing() {
                 data-cy="menu-share-embed-button"
                 @click="toggleEmbedSharing"
             >
+                <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
                 <span class="ms-2">{{ $t('share_link') }}</span>
             </button>
             <CollapseTransition :duration="100">
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -44,6 +44,7 @@ const currentLang = computed(() => store.state.i18n.lang)
 const projection = computed(() => store.state.position.projection)
 const showIn3d = computed(() => store.state.cesium.active)
 var requestClipboard = false
+var buttonText = 'share_link'
 
 const mappingFrameworkSpecificPopup = computed(() => {
     if (showIn3d.value) {
@@ -170,11 +171,17 @@ async function copyValue() {
     try {
         if (showEmbedSharing.value && shareLinkUrlShorten.value != null) {
             await navigator.clipboard.writeText(shareLinkUrlShorten.value)
+            buttonText = 'copy_success'
+            setTimeout(resetButton, 1000)
         }
     } catch (error) {
         log.error(`Failed to copy to clipboard:`, error)
     }
 }
+
+async function resetButton() {
+            buttonText = 'share_link'
+        }
 
 async function updateQrCode(url) {
     try {
@@ -265,7 +272,7 @@ function toggleEmbedSharing() {
                 @click="copyValue(), (requestClipboard = true), toggleEmbedSharing()"
             >
                 <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
-                <span class="ms-2">{{ $t('share_link') }}</span>
+                <span class="ms-2">{{ $t(buttonText) }} </span>
             </button>
             <CollapseTransition :duration="100">
                 <div
--- a/src/modules/map/components/LocationPopup.vue
+++ b/src/modules/map/components/LocationPopup.vue
@@ -33,7 +33,7 @@ const height = ref(null)
 const qrCodeImageSrc = ref(null)
 const shareLinkUrlShorten = ref(null)
 const shareLinkUrl = ref(null)
-const showEmbedSharing = ref(true)
+const showEmbedSharing = ref(false)
 
 
 const i18n = useI18n()
@@ -93,7 +93,7 @@ watch(clickInfo, (newClickInfo) => {
     if (newClickInfo) {
         updateWhat3Word()
         updateHeight()
-        updateShareLink()
+        showEmbedSharing.value = false
     }
 })
 watch(currentLang, () => {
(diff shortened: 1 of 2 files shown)


ltshb (commented):
On mobile the button when open is wrapped 
IMG_3466

src/modules/map/components/LocationPopup.vue
```diff
@@ -89,7 +92,7 @@ watch(clickInfo, (newClickInfo) => {
     if (newClickInfo) {
         updateWhat3Word()
         updateHeight()
-        updateShareLink()
+        showEmbedSharing.value = false
     }
 })
 watch(currentLang, () => {
```
ltshb, line 98: Here the share link should only be updated if the embedsharing is shown

src/modules/map/components/LocationPopup.vue
```diff
@@ -98,6 +101,13 @@ watch(currentLang, () => {
 })
 watch(() => route.query, updateShareLink)
 
+watch(showEmbedSharing, () => {
+    if(showEmbedSharing.value)
+    {
+        updateShareLink()
+    }
+})
```
ltshb, line 109: Code formatting is not correct, run `npm run format`. This PR passed CI due tohttps://github.com/geoadmin/infra-terraform-bgdi-builder/pull/176 which has been applied but not #588.

src/modules/map/components/LocationPopup.vue
```diff
@@ -98,6 +101,13 @@ watch(currentLang, () => {
 })
 watch(() => route.query, updateShareLink)
```
ltshb, line 102: Same here it seems that the share link is updated each time the query parameter is updated, should be done only it is displayed.

src/modules/map/components/LocationPopup.vue
```diff
@@ -98,6 +101,13 @@ watch(currentLang, () => {
...
 
+watch(showEmbedSharing, () => {
+    if(showEmbedSharing.value)
+    {
+        updateShareLink()
+    }
+})
+
 function clearClick() {
     store.dispatch('clearClick')
```
ltshb, line 112: here it might be wise to clear the show share link flag

src/modules/map/components/LocationPopup.vue
```diff
@@ -215,23 +229,36 @@ async function updateQrCode(url) {
...
+                class="btn btn-light btn-sm embedded-button"
+                data-cy="menu-share-embed-button"
+                @click="toggleEmbedSharing"
+            >
+                <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
+                <span class="ms-2">{{ $t('share_link') }}</span>
+            </button>
+            <CollapseTransition :duration="100">
+                <div v-show="showEmbedSharing" class="p-2 card border-light bg-light">
+                    <!-- eslint-disable vue/no-v-html-->
```
ltshb, line 244: Why disabling the eslint n-v-html here? Should not be needed and could be dangerous

src/modules/map/components/LocationPopup.vue
```diff
@@ -247,6 +274,7 @@ async function updateQrCode(url) {
         grid-column-gap: 8px;
         font-size: 0.75rem;
         grid-row-gap: 2px;
+        padding-bottom: 0.5rem;
```
ltshb, line 277: Padding should be added via bootstrap class `pb-1` or `pb-2`

src/modules/map/components/LocationPopup.vue
```diff
@@ -255,10 +283,13 @@ async function updateQrCode(url) {
         display: flex;
         align-items: center;
         padding-top: 0.5rem;
+        padding-bottom: 0.5rem;
```
ltshb, line 286: Dito use bootstrap class

src/modules/map/components/LocationPopup.vue
```diff
@@ -255,10 +283,13 @@ async function updateQrCode(url) {
...
     }
     &-qrcode {
         display: none;
         text-align: center;
+        padding-top: 0.5rem;
+        padding-bottom: 0.5rem;
     }
```
ltshb, line 293: Dito use bootstrap class

src/modules/map/components/LocationPopup.vue
```diff
@@ -215,23 +229,36 @@ async function updateQrCode(url) {
...
+            <button
+                class="btn btn-light btn-sm embedded-button"
+                data-cy="menu-share-embed-button"
+                @click="toggleEmbedSharing"
+            >
+                <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
+                <span class="ms-2">{{ $t('share_link') }}</span>
+            </button>
+            <CollapseTransition :duration="100">
+                <div v-show="showEmbedSharing" class="p-2 card border-light bg-light">
```
ltshb, line 243: Here I would use `v-if` instead so we don't need to mount the element at all as long as no one click on the share button.

ltshb, line 243: Here when the collapse is open we should remove the rounded bottom corner of the button and also the rounded top right of the panel, so this will look nicer (especially if we change the light color in the future)

ltshb (commented):
In the `onMounted` lifecycle hook you also need to only update the share link when the embedsharing flag is true.

Otherwise overall good. Here below some user experience improvments:
- add a tooltip hover (using tippy, you will find some code examples) on the share button telling something like: "Open sharing and copy share link to clipboard". The tooltip should only appear after a short delay.
- When clicking on the share button to open it, also copy the share link to clipboard and display a short tooltip with "Share link copied"

LukasJoss:
> On mobile the button when open is wrapped IMG_3466

I cannot reproduce the wrapping. I tried via the mobile view from browsers, do I need to do it via phone to reproduce it?

src/modules/map/components/LocationPopup.vue
LukasJoss: I have use bootstrap classes for this for now but there seems to be no class available to only round one specific corner, only top or bottom ones.
Also MenuShareEmbed.vue also uses this button, should this be updated as well?

src/modules/map/components/LocationPopup.vue
LukasJoss: What is the intended behaviour here?
Should the share button upon opening a new popup window close or should it stay open and generate a new shortlink?

PR #593, opened 2024-01-04
PB-95: Added external KML support

ltshb:
To allow support of local KML file, the file name is used as URL and during the
parsing of the URL parameter if the url part of the layer doesn't starts with
http then it is ignored and not added to the layer list.

I also removed the `isLoading = true` during the parsing of the url parameter for
KML. I did this for 2 reasons:
 1. Avoid unnecessary spinner which makes the application feeling to be slow
 2. In the case of an external KML layer we cannot load the metadata and it would spin forever

Due to his `isLoading` changes I also updated the load kml plugin to only load
kml metadata when the kml is not external and it has missing kml metadata.


ltshb:
@pakb this PR is ready, it only has an issue with one drawing test that I haven't found yet the reason of this failing test. Maybe you can already have a look at my code and the test failure to see if you have some good input of the cause of this test failure ?

src/api/layers/__tests__/KMLLayer.class.spec.js
```diff
@@ -0,0 +1,104 @@
...
+        expect(layer.hasTooltip).toBe(false)
+
+        // Attribution-related tests
+        const defaultAttributionName = new URL(kmlFileUrl).hostname
+        expect(layer.attributions.length).toBe(1)
+        expect(layer.attributions[0]).toBeInstanceOf(LayerAttribution)
+        expect(layer.attributions[0].name).toBe(defaultAttributionName)
+    })
+
+    it('sopuld create a KMLLayer instance with custom values', () => {
```
pakb, line 28: small typo here `sopuld` => `should`

src/api/layers/__tests__/KMLLayer.class.spec.js
```diff
@@ -0,0 +1,104 @@
...
+import { LayerAttribution } from '@/api/layers/AbstractLayer.class'
+import KMLLayer from '@/api/layers/KMLLayer.class'
+
+describe('KMLLayer', () => {
+    it('should create a KMLlayer instance with default values', () => {
+        const fileId = '-uQyFMtTSCWC_9rZE3EJ6B'
+        const kmlFileUrl = `https://sys-public.dev.bgdi.ch/api/kml/files/${fileId}`
+        const layer = new KMLLayer(kmlFileUrl)
+        expect(layer.kmlFileUrl).toBe(kmlFileUrl)
+        expect(layer.visible).toBe(true)
```
pakb, line 12: with boolean there's a `to.be.true` and `to.be.false` if you want a shorthanded 😉

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -208,6 +209,11 @@ export default {
         tooltipContent() {
             return this.$t('loading_external_layer')
         },
+        showSpinner() {
+            // only show the spinner for external layer, for our layers the
+            // backend should be quick enough and don't require any spinner
+            return this.layer.isLoading && this.layer.isExternal && !this.layer.hasError
```
pakb, line 215: `isLoading` was removed, wasn't it? this might need to change to `!this.layer.kmlData` here

src/store/modules/layers.store.js
```diff
@@ -271,11 +271,27 @@ const actions = {
             log.error('Can not remove layer that is not yet added', layerIdOrObject)
         }
     },
-    updateLayer({ commit }, layer) {
-        if (!(layer instanceof AbstractLayer)) {
+    updateLayer({ commit, getters }, layer) {
+        if (layer instanceof AbstractLayer) {
+            commit('updateLayer', layer)
+        } else if (layer instanceof Object && layer.id) {
```
pakb, line 277: That might be good to document the different input possibilities in a JSDoc, especially the "partial update" part where an `id` is required, and then whatever we want to update, in the same JSON object

src/utils/components/TextInput.vue
```diff
@@ -0,0 +1,159 @@
...
+let components = 0
+</script>
+<script setup>
+/** Input with clear button component */
+import { computed, nextTick, ref, toRefs, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
+
+// On each component creation set the current component ID and increase the counter
+const clearButtonId = `button-addon-clear-${components}`
+components = components + 1
```
pakb, line 12: you might want to do that on the `onMounted` hook, I'm not sure this will be called many times if you instantiate more than one component with a single `import` statement 🤔

vite.config.mts
```diff
@@ -34,6 +34,10 @@ export default defineConfig(({ _, mode }) => {
                         isCustomElement: (tag) => tag === 'cesium-compass',
                     },
                 },
+                script: {
+                    // TODO remove this once updated vue to 3.4
```
pakb, line 38: did you try to update, with all the minor update that have been published lately? maybe the issue we saw back a week ago is somehow fixed now

pakb (commented):
very nicely done! couple questions/small improvements below

src/utils/components/TextInput.vue
ltshb: That's the idea and hack here, this ID is used as HTML ID which should be unique in the DOM if I'm correct, so with this hack each instance of `TextInput` has a different ID.

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
ltshb: No the `isLoading` is kept for external layers (WMS and WMTS), but actually for external KML it might be a good idea to kept it and to only remove it for our kmls (kml from our sys-public backend)

pakb: you can create a getter in the KML class, something like
```
get isLoading {
    return !this.kmlData
}
```

src/api/layers/__tests__/KMLLayer.class.spec.js
ltshb: done

src/api/layers/__tests__/KMLLayer.class.spec.js
ltshb: done

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
ltshb: Sounds like a very good idea, I'll have a deeper look

src/store/modules/layers.store.js
ltshb: Done

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
ltshb: getter and setter are a very bad idea as they are not reactive in Vue which leads to subtle bug hard to find. We should use only javascript plain object as reactive data (same for store data).
See https://vuejs.org/api/options-state.html#data

pakb (approved):
Looking good!

Two points that we can discuss :

- I'm not sure you can do differently, but I don't really like the store as param to some utils file with `updateKmlActiveLayer`, if it is possible to move this logic to a store plugin or router plugin, that would be neat.
- I find it can be hard to detect that the import UI has been added to the bottom of the screen when you are using a big screen. We could maybe (later or here?) add a little animation (swipe up?) to catch the eye, and make it easier to see that the UI is shown somewhere a bit "unexpected" in regards to the click location

PR #595, opened 2024-01-08
PB-95: KML Import UI - alternative with floating label

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-95-kml-import-2/index.html)


ltshb:
We don't need to pass the kml name in the form as it will be parsed from the kml content

PR #601, opened 2024-01-11
PB-92: setup fontawsome pro

rebert:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-92-setup-fontawsome-pro/index.html)

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,3 @@
 engine-strict=true
+@fortawesome:registry=https://npm.fontawesome.com/
+//npm.fontawesome.com/:_authToken=${FONTAWSOME_PRO_TOKEN}
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,1 @@
 engine-strict=true
-@fortawesome:registry=https://npm.fontawesome.com/
-//npm.fontawesome.com/:_authToken=${FONTAWSOME_PRO_TOKEN}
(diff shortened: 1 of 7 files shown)


rebert:
@pakb @ltshb 
With this solution it is possible to work locally without FONTAWESOME_PRO_TOKEN. The pro symbols simply don't show up when running `npm run start`. And there is a message being displayed that `@fortawesome/pro-regular-svg-icons` cannot be loaded.

To setup the CI, we will need to set the FONTAWESOME_PRO_TOKEN and running the command: `npm run build:fontawesome-pro`

ltshb:
On my first try without running `npm run build:fontawesome-pro` I got an error that the import for fontawesome could not be found, then I run the command above and it worked. Unfortunately as I've already run the `build:fontawesome-pro` once I cannot reproduce it ! @pakb could you try as well without the `build:fontawesome-pro` ?

pakb:
image
`npm i` worked fine

I got an error while running `npm run dev`
image

Then it's not possible to open the app in the browser with the given URL, as there's an error
image

So this way of doing it is not sufficiently robust, someone without a FontAwesome pro access can't use our project, and thus contribute

ltshb:
@pakb another issue is that the npm registry for fontawesome (all pro and free) changes when using the pro version and that the new  registry requires authentication even if we only use the free icons. The registry is also saved in the package-lock.json file ! We probably need to refine this during a tecktalk.

ltshb (commented):
As discused wait on tecktalk

PR #603, opened 2024-01-15
PB-90: Reduce shortlink location popup calls tabs version

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-90-reduce_shortlink_location_popup_calls_tabs_version/index.html)


tests/cypress/tests-e2e/mouseposition.cy.js
```diff
@@ -122,13 +122,11 @@ describe('Test mouse position and interactions', () => {
             cy.intercept(/^http[s]?:\/\/(sys-s\.\w+\.bgdi\.ch|s\.geo\.admin\.ch)\//, {
                 body: { shorturl: shortUrl, success: true },
             }).as('shortlink')
-            cy.intercept(`**/api/qrcode/generate**`, {
```
LukasJoss, line 125: I have removed the old qr code test since now the on from the menu is imported, should there be a new test added?

ltshb: Because the qrcode and the whole social sharing component is used twice it doesn't make much sense to test on both place. Check if the qrcode is tested in the share menu, if yes no need to add a test case here. If not we should consider adding a component test for the social media component, please create a jira ticket for this task.
In any case I would add simple assertions making sure that the social buttons are present and visible (no need to tests their action)

LukasJoss:
> 4\. `Share`

LukasJoss:
> I just did a quick functional test of the UI (no code review yet) and looks very promising. A few issues/adjustment to do are listed below:
> 
>     1. Once the popup is open in the share tab, if you move the map around or change the map zoom, the URL parameter are updated, but this update is not reflected to the share link. So changes in the url parameter should trigger an update of the share link
> 
>     2. The name of the share tab should be changed to `Share position` (the old viewer is using this traduction so you should find a traduction key)
> 
>     3. The size of the window should be reduced, with both tab we have some white space on the right hand side.
> 
>     4. I would add the `Share link` label to the short link input element like in the `Share` menu (personal preference)
> 
>     5. The title of the popup window could be updated to the title of the tab (just an idea that needs to be experienced if it looks good)

1.) I added the sharelink update upon zoom but did not add the update upon map movement since the shortlink generated will always be centered on the chosen location anyway(what should be the behaviour upon changing maps?)
3.) Instead of changing the windows size I made the tab buttons a constant size too keep consistency in the looks between the different language, is that good - though the copy sign might need to be removed?
4.) I didn't quite understand what you mean here.
5.)I temporarily added this change but I am not convinced of it

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -0,0 +1,192 @@
...
+}
+</script>
+
+<template>
+    <div
+        id="nav-local"
+        class="tab-pane fade"
+        role="tabpanel"
+        aria-labelledby="nav-local-tab"
+        data-cy="import-file-local-content"
```
ltshb, line 115: Wrong data-cy if not used remove it, otherwise rename it

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -0,0 +1,192 @@
...
+    >
+        <form class="input-group rounded needs-validation">
```
ltshb, line 117: Remove the `input-group need-validations` classes this is only needed for from. Also don't use a form but a simple div here. Check if `rounded` class is needed but probably not

ltshb: You might not need at all this html element as you already have a div below

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+    }
+    shareLinkUrl.value = `${location.origin}/#/map?${stringifyQuery(query)}`
+    shortenShareLink(shareLinkUrl.value)
+}
+async function shortenShareLink(url) {
+    try {
+        shareLinkUrlShorten.value = await createShortLink(url)
+        emit('shareLink', shareLinkUrlShorten.value)
+    } catch (error) {
+        shareLinkUrlShorten.value = null
```
ltshb, line 82: add an error log

src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -1,15 +1,21 @@
...
         >
             <FontAwesomeIcon :icon="`caret-${showEmbedSharing ? 'down' : 'right'}`" />
             <span class="ms-2">{{ $t('share_more') }}</span>
         </button>
         <CollapseTransition :duration="200">
-            <div v-show="showEmbedSharing" class="p-2 card border-light bg-light">
+            <div
+                v-show="showEmbedSharing"
+                class="p-2 rounded-0 rounded-bottom rounded-end card border-light bg-light"
+            >
```
ltshb, line 18: Those changes are obsolete with #614 they should be reverted in order to avoid merge conflict

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+<template>
+    <!-- Online Tab -->
+    <div
+        id="nav-online"
+        class="tab-pane fade"
+        role="tabpanel"
+        aria-labelledby="nav-online-tab"
+        data-cy="import-file-online-content"
+    >
+        <form class="input-group d-flex needs-validation">
```
ltshb, line 96: Same as above the `form` here is not needed and should be remove, same for its classes

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+                <MenuShareInputCopyButton
+                    :input-text="shareLinkUrlShorten"
+                    :copy-text="'copy_url'"
+                    :copied-text="'copy_success'"
+                    class="px-0 py-2"
+                    data-cy="location-popup-link-bowl-crosshair"
+                />
+            </div>
+        </form>
+        <ImportFileButtons class="mt-2" :button-state="buttonState" @load-file="loadFile" />
```
ltshb, line 108: This should be removed

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+    <!-- Online Tab -->
+    <div
+        id="nav-online"
+        class="tab-pane fade"
+        role="tabpanel"
+        aria-labelledby="nav-online-tab"
+        data-cy="import-file-online-content"
+    >
+        <form class="input-group d-flex needs-validation">
+            <div class="menu-share-embed">
```
ltshb, line 97: Removed unused class `menu-share-embed`

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+            </div>
+        </form>
+        <ImportFileButtons class="mt-2" :button-state="buttonState" @load-file="loadFile" />
+    </div>
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+.location-popup {
+    @extend .clear-no-ios-long-press;
```
ltshb, line 115: This has no effect as you don't have a `.location-popup` class in your template and is probably not needed as it is in the LocationPopup component, some for the `&-link`

src/modules/map/components/LocationPopup.vue
```diff
@@ -251,23 +209,5 @@ async function updateQrCode(url) {
             white-space: nowrap;
         }
     }
-    &-link {
```
ltshb, line 254: Also remove the `&-coordinates` above that has been removed from the template

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -0,0 +1,192 @@
...
+                </LocationPopupCopySlot>
+            </div>
+        </form>
+    </div>
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+.location-popup {
+    @extend .clear-no-ios-long-press;
```
ltshb, line 179: This class is not needed here, already part of the parent component

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+})
+
+watch(clickInfo, () => {
+    if (showEmbedSharing.value) {
+        updateShareLink()
+    }
+})
+watch(currentLang, () => {
+    if (showEmbedSharing.value) {
+        setTimeout(() => updateShareLink(), 1)
```
ltshb, line 54: Why the `setTimeout` here ? if you have a race condition it could be also solved using vue `nextTick` method

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,122 @@
...
+    }
+})
+watch(zoom, () => {
+    if (showEmbedSharing.value) {
+        setTimeout(() => updateShareLink(), 1)
```
ltshb, line 59: Dito as above

src/modules/map/components/LocationPopup.vue
```diff
@@ -153,86 +116,81 @@ async function updateQrCode(url) {
...
+                />
+                <!-- Share tab -->
+                <LocationPopupShare
+                    :class="{ active: selectedTab === 'share', show: selectedTab === 'share' }"
+                    :coordinate="coordinate"
+                    :click-info="clickInfo"
+                    :current-lang="currentLang"
+                    :show-embed-sharing="showEmbedSharing"
+                    @share-link="(i) => (shareLink = i)"
                 />
```
ltshb, line 191: I found the logic of the share link with the child event and the `show-embed-sharing` hard to follow, might be better to put the shortlink logic in the parent vue, meaning in this component `LocationPopup` and to pass the shortlink to the child component, might make the code simpler to follow.

ltshb (commented):
The results looks very nice.

Some code clean up and improvements are required see comment, but thanks for the effort.

src/modules/map/components/LocationPopup.vue
```diff
@@ -1,45 +1,39 @@
 <script setup>
 /** Right click pop up which shows the coordinates of the position under the cursor. */
 
-import proj4 from 'proj4'
-import { computed, onMounted, ref, watch } from 'vue'
+// importing directly the vue component, see https://github.com/ivanvermeyen/vue-collapse-transition/issues/5
```
ltshb, line 4: The comment is probably a copy past mistake

src/modules/map/components/LocationPopup.vue
```diff
@@ -1,45 +1,39 @@
...
-const route = useRoute()
 const store = useStore()
 
 const clickInfo = computed(() => store.state.map.clickInfo)
-const currentLang = computed(() => store.state.i18n.lang)
 const projection = computed(() => store.state.position.projection)
 const showIn3d = computed(() => store.state.cesium.active)
+const currentLang = computed(() => store.state.i18n.lang)
+
+const route = useRoute()
```
ltshb, line 27: Just code style I would keep the `useRoute` above with the other `use...`

src/modules/map/components/LocationPopup.vue
```diff
@@ -50,77 +44,65 @@ const mappingFrameworkSpecificPopup = computed(() => {
...
+watch(
+    () => route.query,
+    (newQuery, oldQuery) => {
+        //Cannot watch language and zoom directly due to the delayed url update
+        if (showEmbedSharing.value) {
+            if (oldQuery['lang'] != newQuery['lang']) {
+                updateShareLink()
+            }
+        }
+        console.log('Query parameters changed:', newQuery)
```
ltshb, line 82: No console, if you want to keep some debugging log, you need to use `log.debug`

src/modules/map/components/LocationPopup.vue
```diff
@@ -1,45 +1,39 @@
...
-const currentLang = computed(() => store.state.i18n.lang)
 const projection = computed(() => store.state.position.projection)
 const showIn3d = computed(() => store.state.cesium.active)
+const currentLang = computed(() => store.state.i18n.lang)
+
+const route = useRoute()
+const selectedTab = ref('position')
+const shareTabButton = ref(null)
+const newClickInfo = ref(true)
+const showEmbedSharing = ref(false)
```
ltshb, line 31: This could be simplified by using a computed property 
```
const showEmbedSharing = computed(()=> selectedTab.value === 'share')
```
So you don't have to set it and clear it below

src/modules/map/components/LocationPopup.vue
```diff
@@ -133,104 +115,123 @@ function updateShareLink() {
...
+    requestClipboard.value = false
+}
 </script>
 
 <template>
     <component
         :is="mappingFrameworkSpecificPopup"
         v-if="coordinate"
-        :title="$t('position')"
+        :title="selectedTab == 'position' ? $t('position') : $t('link_bowl_crosshair')"
```
ltshb, line 165: use the new style for translation `i18n.t` instead of `$t`

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -0,0 +1,182 @@
...
+                <a href="http://what3words.com/" target="_blank">what3words</a>
+            </LocationPopupCopySlot>
+
+            <LocationPopupCopySlot
+                v-if="height"
+                identifier="height"
+                :value="heightInMeter"
+                :extra-value="heightInFeet"
+            >
+                <a :href="$t('elevation_href')" target="_blank">{{ $t('elevation') }}</a>
```
ltshb, line 162: `i18n.t`

src/modules/map/components/LocationPopupShare.vue
```diff
@@ -0,0 +1,39 @@
...
+            :copied-text="'copy_success'"
+            class="px-0 py-2"
+            data-cy="location-popup-link-bowl-crosshair"
+        />
+    </div>
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
+</style>
```
ltshb, line 39: Because the style is not used it can be removed

ltshb (approved):
Very nice just a few minor comments but otherwise its all good.

PR #608, opened 2024-01-18
PB-95: Add zoom to KML extent in import file

ltshb:
Now when importing a KML file we automatically zoom to its extent

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-95-kml-zoom-to-extent/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -1,6 +1,7 @@
 import AbstractLayer from '@/api/layers/AbstractLayer.class'
 import ExternalGroupOfLayers from '@/api/layers/ExternalGroupOfLayers.class'
 import LayerTypes from '@/api/layers/LayerTypes.enum'
+import { getKmlExtent, getKmlExtentForProjection, parseKmlName } from '@/utils/kmlUtils'
 import { ActiveLayerConfig } from '@/utils/layerUtils'
 import log from '@/utils/logging'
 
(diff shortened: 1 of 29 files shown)


ltshb:
@pakb I don't understand why the cesium test is now failing ? The comment about empty collection seems also a bit strange 
```
// should be 3 (line, icon, text) but ol-cesium creates additional empty collection
expect(layerCollection.length).to.eq(4)
```
Do you know if the test is really correct ?

ltkum:
Hey, @ltshb 
I tried to test the failing test personally on your branch (basically : drawing a few lines, a marker, a text), and as soon as I get out of the drawing mode, everything except the line gets deleted. As soon as I get back into the drawing mode, they reappear. I think there is something that changed in the way we store the kmls that makes them not load in "normal mode" and still let them appear in drawing mode.

Since the failing test checks the drawings displayed once in 3d mode, my guess is it doesn't see the layers that haven't been loaded correctly.

ltshb:
@ltkum thanks for the hint I found the bug.

src/store/modules/layers.store.js
```diff
@@ -429,6 +430,48 @@ const actions = {
...
+        if (!currentLayer) {
+            throw new Error(
+                `Failed to update layer error key "${layerId}", layer not found in active layers`
+            )
+        }
+        const updatedLayer = currentLayer.clone()
+        updatedLayer.errorKey = errorKey
+        updatedLayer.hasError = !!errorKey
+        commit('updateLayer', updatedLayer)
+    },
```
pakb, line 445: Shouldn't that only be applied to KML layers?
If that's not what is intended, I would move the declaration of `errorKey` and `hasError` into `AbstractLayer.class.js` instead of `KMLLayer.class.js`

ltshb: `errorKey` and `hasError` were meant for `ExternalLayer` but because for KML we have only one class for internal and external I had to put the flag again in this class. But this is a good idea to move them into the abstract layer even if for the moment they are only used for external layer (internal layer should never have an error ;-) ).

pakb (approved):

PR #610, opened 2024-01-19
PB-95: add an animation when we open the import file popup

pakb:
using GSAP to achieve this, as it is much easier to set multiple CSS properties this way, and it has a built-in animation engine when changing the CSS style.

I couldn't find an easy solution to place this element on top of its parent (the menu "Import file") so I resorted to place it behind the menu when it opens up.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-95_animate_import_file_open/index.html)


ltshb:
Close this as we take the solution of #611 with a modal on top

PR #611, opened 2024-01-19
PB-95: Put the import file as modal on top

ltshb:
Alternative to #610

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-95-kml-modal-top/index.html)


ltshb:
@pakb here is also another idea to be discussed on monday which solution is the best modal on top or animation on footer.

src/modules/menu/components/advancedTools/ImportFile/ImportFile.vue
```diff
@@ -1,67 +1,55 @@
...
 import ImportFileOnlineTab from '@/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue'
-import SimpleWindow from '@/utils/components/SimpleWindow.vue'
 
 const selectedTab = ref('online')
-
-const store = useStore()
 </script>
 
 <template>
-    <teleport to="#map-footer-middle-1">
```
pakb, line 15: I think this teleport target was added for the sole purpose of this ImportFile component, so if we do not use it anymore, that might be a good idea to clear it too (and in the process simplify the structure of the InfoboxModule)

pakb (approved):

PR #585, opened 2023-12-18
PB-87:  Improving the search bar

ltshb:
See commits messages for details of this pr.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-search-bar/index.html)

PR #617, opened 2024-01-24
PB-203: Debounce edit text dispatch

Angi-Kinas:
In the drawing feature, when changing a text the store got updated after every keystroke, and with this a request was sent.

In this PR a debounce function for ~3~ 0.3 seconds after a keystroke was used to delay the update on the store and minimize the amount of requests being sent.

Ismail: I set the delay to 0.3 seconds since 3 seconds is too long.

Example of the delay:
Peek 2024-02-05 14-59

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-203-debounce-drawing-feat-text-edit/index.html)

### DIFF
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -34,14 +34,14 @@ const description = ref(feature.value.description)
 
 // Update the UI when the feature changes
 watch(
-    () => feature.title,
+    () => feature.value.title,
     (newTitle) => {
         title.value = newTitle
     }
 )
 
 watch(
-    () => feature.description,
+    () => feature.value.description,
     (newDescription) => {
         description.value = newDescription
     }
@@ -61,13 +61,16 @@ watch(description, (newDescription) => {
 
 function updateFeatureTitle(previousTitle) {
     if (previousTitle === title.value) {
-        store.dispatch('changeFeatureTitle', { feature, title: title.value })
+        store.dispatch('changeFeatureTitle', { feature: feature.value, title: title.value })
     }
 }
 
 function updateFeatureDecription(previousDescription) {
     if (previousDescription === description.value) {
-        store.dispatch('changeFeatureDescription', { feature, description: description.value })
+        store.dispatch('changeFeatureDescription', {
+            feature: feature.value,
+            description: description.value,
+        })
     }
 }
 
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -32,6 +32,21 @@ const { feature, readOnly } = toRefs(props)
 const title = ref(feature.value.title)
 const description = ref(feature.value.description)
 
+// Update the UI when the feature changes
+watch(
+    () => feature.title,
+    (newTitle) => {
+        title.value = newTitle
+    }
+)
+
+watch(
+    () => feature.description,
+    (newDescription) => {
+        description.value = newDescription
+    }
+)
+
 // The idea is watching the title and the description.
 // Put a debounce on the update of the feature so that we can compare with the current UI state
 // If the value is the same as in the UI, we can update the feature
--- a/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -50,28 +50,29 @@ watch(
 // The idea is watching the title and the description.
 // Put a debounce on the update of the feature so that we can compare with the current UI state
 // If the value is the same as in the UI, we can update the feature
-watch(title, (newTitle) => {
-    const debounceTitleUpdate = debounce(updateFeatureTitle, 300)
-    debounceTitleUpdate(newTitle)
+watch(title, () => {
+    debounceTitleUpdate(store)
 })
-watch(description, (newDescription) => {
-    const debounceDescriptionUpdate = debounce(updateFeatureDecription, 300)
-    debounceDescriptionUpdate(newDescription)
+watch(description, () => {
+    debounceDescriptionUpdate(store)
 })
 
-function updateFeatureTitle(previousTitle) {
-    if (previousTitle === title.value) {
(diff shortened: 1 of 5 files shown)


src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -32,15 +33,17 @@ const description = computed({
         return feature.description
     },
     set(value) {
-        store.dispatch('changeFeatureDescription', { feature, description: value })
+        this.description = value
+        debounce(store.dispatch('changeFeatureDescription', { feature, description: value }), 3000)
```
ltshb, line 37: The debounce helper is not working like this, it return a new function that would be debounced when calling, and after testing it effectively don't debounce.
You should do something like this
```javascript
const dispatch = debounce(store.dispatch, 500)
dispatch('changeFeatureDescription', { feature, description: value })
```
However I tried it quickly and it don't work well because the dispatch will overwrite what you has written in between the debouncing.
I think we should totally decoupled the local state from the store and use a normal ref. Then you can add a watch on the ref that would set the store (using eventually a debouncing only on the store). You can have a look at this PR/commit on how I solve the issue on the search bar https://github.com/geoadmin/web-mapviewer/pull/585/commits/4c4fe4618a02463cd65da87a1bb8d3fea8163d61

ltshb (changes requested):

src/modules/infobox/components/styling/FeatureStyleEdit.vue
ismailsunni: hi @ltshb 
I can't use the same approach, since I need to compare if the text on the UI stays the same or not.
My approach is watching the UI state, whenever it changes, it will create a debounce to call the update function in the next 0.3 seconds. The function will compare the parameter that it receives with the current state of the UI (when the function runs). If they are the same (it means the user stops typing), it will update the value of the store.

I chose 0.3 seconds since 3 seconds is too long to see the text updated on the map. More than 1 second seems like a lag for me.

What do you think?

ismailsunni:
I am sorry, I need to rebase the code because the changes in develop breaks my code. But all is good now.

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -28,22 +29,51 @@ const props = defineProps({
...
-    set(value) {
-        store.dispatch('changeFeatureTitle', { feature: feature.value, title: value })
-    },
+watch(description, (newDescription) => {
+    const debounceDescriptionUpdate = debounce(updateFeatureDecription, 300)
+    debounceDescriptionUpdate(newDescription)
 })
+
+function updateFeatureTitle(previousTitle) {
+    if (previousTitle === title.value) {
```
ltshb, line 63: If the debounce method is working fine this `if` should not be needed, I had once an issue with the debounce method where its closure was somehow not working and therefore not debouncing and I couldn't find out why the closure was not working and I ended up using my own debounce implementation instead of the debounce method, maybe you have the same issue here ?

ltshb: I could quickly checked that in your case the debounce method closure does not work like it should work in App.vue. I don't know why ? Would be great if you find out why otherwise you would need to create your own debounce implementation.

ismailsunni: But in the end, we still need to check if the text in the input field are not changing, right? That's why we will still need the `if` statement. If we don't check it, then the dispatch will be called as many as the text change, just with delay.

ltshb: No the idea of the debounce method is that if it is called again before the delay the previous call is dropped, see https://github.com/geoadmin/web-mapviewer/blob/develop/src/utils/debounce.js
Somehow if you define the debounce method inside a watcher, the javascript closure of the debounce method don't work and it doesn't debounce. In this case the debounce is not fully working and is just delaying the call, therefore you need the `if`

ltshb: Somehow my commit seems to break a cypress test on 3d, @ismailsunni could you try to rebase on top of develop and see if it fails again ?

ismailsunni: Sure!
And yes, after I read it again, you are right, I shouldn't need the `if`. Thanks for the explanation.

ismailsunni:
It passes locally, so I just put a wait and see how is it in the CI, and it seems work. I will need to check deeper to see where should the intercept happens. But since it passes locally, it's kind of tricky to make the CI passes.

ltshb (approved):

tests/cypress/tests-e2e/3d/layers.cy.js
```diff
@@ -184,13 +184,15 @@ describe('Test of layer handling in 3D', () => {
...
         cy.readWindowValue('cesiumViewer').then((viewer) => {
-            const mainCollection = viewer.scene.primitives.get(0)
-            expect(mainCollection.length).to.eq(1)
-            const layerCollection = mainCollection.get(0)
+            // main collection
+            cy.wrap(viewer.scene.primitives.get(0)).should('have.length', 1)
+            // layer collection
             // should be 3 (line, icon, text) but ol-cesium creates additional empty collection
-            expect(layerCollection.length).to.eq(4)
+            cy.wrap(viewer.scene.primitives.get(0).get(0), { timeout: 10000 }).should(
```
ismailsunni, line 192: so, it's kind of waiting until the timeout or the length is 4, right?

ltshb: @ismailsunni yes its correct using the `cy.wrap()` make use of cypress retriability and it a best practice in comparison to the `expect` depending on the use case. However it turned out that here that was not the issue. Yesterday I did a deeper debugging session with Pascal and it seems that we have a race condition in our code and that effectively sometimes on the cypress test the primitives of the KML are missing and the KML is not displayed. We could not find out the real reason and we skip that test for the moment. This not so critical as it works and for the go live this feature (display of KML in 3d is not a stopper as it is a new feature not available on the old viewer.
See #643

PR #618, opened 2024-01-25
PB-88: Completed E2E tests for external WMS/WMTS layers

ltshb:
Follow up of #615 
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-88-import-e2e-2/index.html)


ltkum (approved):
From what I was able to see, everything seems to be good.

PR #619, opened 2024-01-25
PB-246: bugfix legacy timestamps parameters not taken into account

ltkum:
Issue : When using legacy parameters with timestamps on time enabled layers, those parameters were not taken into account on the new mapviewer.

Fix: A typographical error stopped the timestamps parameters from being populated, and then we never entered the condition to apply those parameters to the layers. Once this was solved, it appeared we were trying to assign value to a protected element outside a mutation, thus we added a new setter to apply these changes to the layers, even if they're not active yet.
[Test link with legacy timeseries parameters](https://sys-map.dev.bgdi.ch/preview/pb-246-bugfix-legacy-timestamps/index.html?topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.swisstopo.swissimage-product&layers_timestamp=19951231,2002&lang=en&layers_opacity=0.45,0.7&layers_visibility=false,true)
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-246-bugfix-legacy-timestamps/index.html)

### DIFF
--- a/src/api/layers/GeoAdminWMSLayer.class.js
+++ b/src/api/layers/GeoAdminWMSLayer.class.js
@@ -79,6 +79,12 @@ export default class GeoAdminWMSLayer extends GeoAdminLayer {
         this.wmsVersion = wmsVersion
     }
 
+    clone() {
+        const clone = super.clone()
+        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
+        return clone
+    }
+
     getURL() {
         return this.baseURL ?? WMS_BASE_URL
     }
--- a/src/api/layers/GeoAdminWMSLayer.class.js
+++ b/src/api/layers/GeoAdminWMSLayer.class.js
@@ -81,7 +81,7 @@ export default class GeoAdminWMSLayer extends GeoAdminLayer {
 
     clone() {
         const clone = super.clone()
-        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
+        clone.timeConfig = this.timeConfig?.clone() ?? null
         return clone
     }
 
--- a/src/api/layers/GeoAdminWMTSLayer.class.js
+++ b/src/api/layers/GeoAdminWMTSLayer.class.js
@@ -68,6 +68,12 @@ export default class GeoAdminWMTSLayer extends GeoAdminLayer {
         this.hasMultipleTimestamps = this.timeConfig?.timeEntries?.length > 1
     }
 
+    clone() {
+        const clone = super.clone()
+        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
+        return clone
+    }
+
     /**
      * @param {Number} epsgNumber The EPSG number of the projection system to use (for instance,
      *   EPSG:2056 will require an input of 2056)
--- a/src/api/layers/GeoAdminWMTSLayer.class.js
+++ b/src/api/layers/GeoAdminWMTSLayer.class.js
@@ -70,7 +70,7 @@ export default class GeoAdminWMTSLayer extends GeoAdminLayer {
 
     clone() {
         const clone = super.clone()
-        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
+        clone.timeConfig = this.timeConfig?.clone() ?? null
         return clone
     }
 
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -117,8 +117,9 @@ const handleLegacyParam = (
         case 'layers_opacity':
         case 'layers_visibility':
         case 'layers_timestamp':
-            // we ignore those params as they are now obsolete
-            // see adr/2021_03_16_url_param_structure.md
+            // Those are checked within the `isLayersUrlParamLegacy` function, which is
+            // called under the `layers` case. We simply ensure here that they're not called
+            // multiple times for nothing
             break
 
         case '3d':
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -151,11 +151,11 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
     let latlongCoordinates = []
     let cameraPosition = []
 
-    Object.keys(legacyParams).forEach((param) => {
+    legacyParams.forEach((param_value, param_key) => {
         handleLegacyParam(
             legacyParams,
-            param,
-            legacyParams[param],
+            param_key,
+            param_value,
             store,
             newQuery,
             latlongCoordinates,
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -94,10 +94,6 @@ const handleLegacyParam = (
         // taking all layers related param aside so that they can be processed later (see below)
         // this only occurs if the syntax is recognized as a mf-geoadmin3 syntax (or legacy)
         case 'layers':
-            console.log('----------------------------')
-            console.log(legacyValue)
-            console.log(isLayersUrlParamLegacy(legacyValue))
-            console.log('----------------------------')
             if (isLayersUrlParamLegacy(legacyValue)) {
                 // for legacy layers param, we need to give the whole search query
                 // as it needs to look for layers, layers_visibility, layers_opacity and
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -14,7 +14,7 @@ import log from '@/utils/logging'
 
 const handleLegacyKmlAdminIdParam = async (legacyParams, newQuery) => {
     log.debug('Transforming legacy kml adminId, get KML ID from adminId...')
-    const kmlLayer = await getKmlLayerFromLegacyAdminIdParam(legacyParams['adminId'])
+    const kmlLayer = await getKmlLayerFromLegacyAdminIdParam(legacyParams.get('adminId'))
     log.debug('Adding KML layer from legacy kml adminId')
     if (newQuery.layers) {
         newQuery.layers = `${newQuery.layers};${kmlLayer.getID()}@adminId=${kmlLayer.adminId}`
@@ -198,8 +198,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
     // removing old query part (new ones will be added by vue-router after the /# part of the URL)
     const urlWithoutQueryParam = window.location.href.substr(0, window.location.href.indexOf('?'))
     window.history.replaceState(window.history.state, document.title, urlWithoutQueryParam)
-
-    if ('adminId' in legacyParams) {
+    if (legacyParams.get('adminId')) {
         // adminId legacy param cannot be handle above in the loop because it needs to add a layer
         // to the layers param, thats why we do handle after.
         handleLegacyKmlAdminIdParam(legacyParams, newQuery)
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -45,6 +45,7 @@ const handleLegacyKmlAdminIdParam = async (legacyParams, newQuery) => {
 }
 
 const handleLegacyParam = (
+    params,
     param,
     legacyValue,
     store,
@@ -96,11 +97,12 @@ const handleLegacyParam = (
             // for legacy layers param, we need to give the whole search query
             // as it needs to look for layers, layers_visibility, layers_opacity and
             // layers_timestamp param altogether
-
-            // HERE : use params to get visibility, opacity, timestamps
             newValue = getLayersFromLegacyUrlParams(
                 store.state.layers.config,
-                window.location.search
+                legacyValue,
+                params['layers_visibility'],
+                params['layers_opacity'],
+                params['layers_timestamp']
             )
                 .map((layer) => transformLayerIntoUrlString(layer))
                 .join(';')
@@ -167,6 +169,7 @@ const handleLegacyParams = (legacyParams, store, to, next) => {
 
     Object.keys(legacyParams).forEach((param) => {
         handleLegacyParam(
+            legacyParams,
             param,
             legacyParams[param],
             store,
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -75,12 +75,10 @@ const handleLegacyParam = (
             cameraPosition[1] = Number(legacyValue)
             break
 
-        // taking all layers related param aside so that they can be processed later (see below)
-        // this only occurs if the syntax is recognized as a mf-geoadmin3 syntax (or legacy)
         case 'layers':
-            // for legacy layers param, we need to give the whole search query
-            // as it needs to look for layers, layers_visibility, layers_opacity and
-            // layers_timestamp param altogether
+            // for legacy layers param, we need to give the layers visibility, opacity and timestamps,
+            // as they are combined into one value with the layer in the current 'layers' parameter
+            // implementation
             newValue = getLayersFromLegacyUrlParams(
                 store.state.layers.config,
                 legacyValue,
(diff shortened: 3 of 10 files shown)


ltshb (changes requested):
I manually tested it and it don't work, see https://sys-map.dev.bgdi.ch/preview/pb-246-bugfix-legacy-timestamps/index.html?topic=ech&lang=fr&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen&layers_timestamp=20131231 should have the time 2013 but it doesn't

src/store/modules/layers.store.js
```diff
@@ -527,6 +556,11 @@ const mutations = {
             timedLayer.timeConfig.getTimeEntryForYear(year)
         )
     },
+    setLegacyLayerYear(state, { timedLayer, year }) {
+        timedLayer.timeConfig.updateCurrentTimeEntry(
+            timedLayer.timeConfig.getTimeEntryForYear(year)
+        )
+    },
```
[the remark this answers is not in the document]
ltkum: `setLayerYear` checks the store for active layers using the layer Id,

In the case of layers that are not active by default (for example: swissimage), this means we can't find them in the store and it will refuse to update their time series. A better name would have been `setNonActiveLayerYear`

src/utils/legacyLayerParamUtils.js
```diff
@@ -123,9 +126,28 @@ export function getLayersFromLegacyUrlParams(layersConfig, legacyLayersParam) {
...
+                                    year: year,
+                                })
+                            } else {
+                                // this does tell us that the changes have been taken into account
+                                // and would be updated, without using a store that is not active
+                                // during unit tests
+                                layer.timeConfig.updateCurrentTimeEntry(
+                                    layer.timeConfig.getTimeEntryForYear(year)
+                                )
+                            }
```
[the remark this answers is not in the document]
ltkum: `layer.timeConfig.updateCurrentTimeEntry` creates a vue Error at runtime, telling us it can't modify the store outside of a mutation

the `if(store) ... else` part is quite hacky indeed, as the unit tests crashed, the store not being instantiated during unit tests.

I'll try to see if I can avoid the store completely during this. I feel like, with these layers being cloned to be modified before pushing them to be activated, we should be able to do so, but I failed to make it work.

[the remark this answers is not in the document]
ltshb: This should not, you should be able to use the `updateCurrentTimeEntry` on a layer that has been cloned you probably had another issue in your code.
Here you are in the context of a router plugin that will update the store with a new layer object, so you don't have to do any store dispatch.

ltkum:
for the manual test that did not go through :

When doing the check `isLayersUrlParamLegacy`, in the `Layers` case,
the `ch.swisstopo.zeitreihen` returns false, (external layer is false, regex is true) 
the `ch.swisstopo.swissimage-product` returns true (both are false, and we return the opposite)

and since the split is made with `;`, the regex is checked with `ch.swisstopo.zeitreihen,ch.swisstopo.swissimage-product` as an input when we put the two together

I'll be checking this regex and what it represents exactly with Pascal tomorrow morning, but the crux of the issue is that, when putting only the `ch.swisstopo.zeitreihen` layer, the checks specify that this doesn't need to transfer the layer to the `getLayersFromLegacyUrlParams` function, thus skipping the Layers_timestamp parameter

src/store/modules/layers.store.js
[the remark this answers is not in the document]
ltshb: Active layer means that they are in the layer list, but not necessarily visible

src/utils/legacyLayerParamUtils.js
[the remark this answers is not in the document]
ltshb: You might need to check that the layer clone method is cloning also the layer config this might be an issue and the reason of the vue failure

src/api/layers/GeoAdminWMSLayer.class.js
```diff
@@ -79,6 +79,12 @@ export default class GeoAdminWMSLayer extends GeoAdminLayer {
         this.wmsVersion = wmsVersion
     }
 
+    clone() {
+        const clone = super.clone()
+        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
```
ltshb, line 84: This line could be simplified to `clone.timeConfig = this.timeConfig?.clone() 
?? null`

src/api/layers/GeoAdminWMTSLayer.class.js
```diff
@@ -68,6 +68,12 @@ export default class GeoAdminWMTSLayer extends GeoAdminLayer {
         this.hasMultipleTimestamps = this.timeConfig?.timeEntries?.length > 1
     }
 
+    clone() {
+        const clone = super.clone()
+        clone.timeConfig = this.timeConfig ? this.timeConfig.clone() : null
```
ltshb, line 73: Dito

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -95,6 +95,10 @@ const handleLegacyParam = (
         // taking all layers related param aside so that they can be processed later (see below)
         // this only occurs if the syntax is recognized as a mf-geoadmin3 syntax (or legacy)
         case 'layers':
+            console.log('----------------------------')
+            console.log(legacyValue)
+            console.log(isLayersUrlParamLegacy(legacyValue))
+            console.log('----------------------------')
```
ltshb, line 101: No console.log

src/utils/legacyLayerParamUtils.js
```diff
@@ -15,8 +14,12 @@ function readUrlParamValue(url, paramName) {
...
 }
 
-const newLayerParamRegex = /^[\w.]+[@\w=]*[,ft]*[,?\d.]*$/
-
+/*
+ Return true if the layer is a layer WITH a specified non legacy parameter.
+ Layers without parameters will return false
+ */
+const newLayerParamRegex =
+    /^([\w.-]+)((@[\w=\d]+)+(,[ft,]+)*([,\d.]+)*|(@[\w=\d]+)*(,[ft,]+)+([,\d.]+)*|(@[\w=\d]+)*(,[ft,]+)*([,\d.]+)+)$/
```
ltshb, line 22: I don't get the idea of this regex here and the `isLayersUrlParamLegacy()` check below ? @pakb the `handleLegacyParam()` method is called if a legacy url has been detected by a simple check with the missing `#` test, so there is no need for further complex detection on the layer side ? 
However in the legacy url detection I see an issue in the regex, see https://github.com/geoadmin/web-mapviewer/blob/e5ed2ca7ea79f188f875b8a12e91467d8508d5b2/src/router/legacyPermalinkManagement.routerPlugin.js#L21 where we do a global search, this global search regex could match even if it is not containing a `?` at the beginning.
Here I would suggest such regex: `/^\?([^=]+)=([^&]+)/g` but to be honest I don't see the advantage of regex there also could be much simpler as follow:

```javascript
const parseLegacyParams = (search) => {
   if (search.startsWith('?')) {
       return search.split('&').map((p) => p.split('=')).reduce((params, [k, v]) => params[k] = v, {})
   }
   return {}
}
```

Of course this method should be tested with unit test and it would not work if the query parameters would contain any non encoded `&` or `=` in any parameter value, but that should not be the case if the old url are properly encoded.

src/utils/legacyLayerParamUtils.js
```diff
@@ -26,6 +29,11 @@ function isExternalLayer(layerId) {
...
 
 export function isLayersUrlParamLegacy(layersParamValue) {
+    // current error : layers thich finish in f/t are recognized as problematic
+    if (layersParamValue.split(';').length > 1) {
+        // if layers are separated by ;, this means we are not having legacy layers
+        return false
+    }
     return !layersParamValue.split(';').some((layer) => {
         return isExternalLayer(layer) || newLayerParamRegex.test(layer)
     })
```
ltshb, line 39: See comment above, here we don't need to check again for legacy param

src/utils/legacyLayerParamUtils.js
```diff
@@ -123,8 +130,17 @@ export function getLayersFromLegacyUrlParams(layersConfig, legacyLayersParam) {
...
                                 layer.timeConfig.getTimeEntryForTimestamp(layerTimestamps[index])
+                                    .year
+                            ) {
+                                year = layer.timeConfig.getTimeEntryForTimestamp(
+                                    layerTimestamps[index]
+                                ).year
+                            }
+                            layer.timeConfig.updateCurrentTimeEntry(
+                                layer.timeConfig.getTimeEntryForYear(year)
                             )
```
ltshb, line 144: If I'm not mistaken this code is doing exactly the same as before but is more complex, before it was more consise.

src/store/modules/layers.store.js
```diff
@@ -367,20 +367,6 @@ const actions = {
...
-            const currentIndex = state.activeLayers.indexOf(activeLayer)
-            if (currentIndex > 0) {
-                commit('moveActiveLayerFromIndexToIndex', {
-                    layer: activeLayer,
-                    startingIndex: currentIndex,
-                    endingIndex: currentIndex - 1,
-                })
-            }
-        }
-    },
```
ltshb, line 383: ?

ltshb:
@pakb @ltkum When already touching the legacy url param parsing I propose also to simplify the code. Currently the `layers`, `layers_visibility`, `layers_opacity` and `layers_timestamp` are parsed twice; first by `parseLegacyParams()` and then by `getLayersFromLegacyUrlParams()`, the latest function docu is also outdated. I propose to change the latest function simpler as such
```javascript
/**
 * Reads URL params :
 *
 * - Layers
 * - Layers_opacity
 * - Layers_visibility
 * - Layers_timestamp
 *
 * And returns a list of (copies of) layers that should be activated with the correct configuration
 * already set (visibility, opacity, etc...). Bear in mind that these layers are not yet dispatched
 * to the store (they are deep copy of what was given as layersConfig param, with opacity/visibility
 * set according to the legacyLayersParam)
 *
 * @param {GeoAdminLayer[]} layersConfig
 * @param {String} legacyLayers legacy layers value
 * @param {String} legacyVisibility legacy layers_visibility value
 * @param {String} legacyOpacity legacy layers_opacity value
 * @param {String} legacyTimestamp legady layers_timestamp value
 * @returns {AbstractLayer[]}
 */
export function getLayersFromLegacyUrlParams(layersConfig, legacyLayers, legacyVisiblity, legacyOpacity, legacyTimestamp) {
    const layersToBeActivated = []
    const layerIds = legacyLayers.split(',').map(decodeURIComponent) // decode external layers IDs which can be encoded due to special characters (e.g. WMS||https://testc.om/?k1=v1&k2=v2
    const layerVisibilities = legacyVisiblity.split(',')
    const layerOpacities = legacyOpacity.split(',')
    const layerTimestamps = legacyTimestamp.split(',')

    layerIds.forEach((layerId, index) => {
       // same code as before
       // But be more robust for the opacity, use a method `parseOpacity()` that would catch
       // error if the opacity is not a number, or is a bigger number than 1, returning 1 in case of any error
       ... 
    })
```

src/store/modules/layers.store.js
ltkum: I must have removed it at the same time as the function I wanted to remove by mistake. I'm putting this one back

src/utils/legacyLayerParamUtils.js
ltkum: Well, I think you're right : once we're in the legacy handler, we should consider that we are within the supposed use case: someone used an old mapviewer link. I would make the new regex and unit tests associated its own ticket, and I can tackle that once this current issue is resolved.

src/utils/legacyLayerParamUtils.js
ltkum: It seems that you're right

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -94,26 +78,26 @@ const handleLegacyParam = (
         // taking all layers related param aside so that they can be processed later (see below)
         // this only occurs if the syntax is recognized as a mf-geoadmin3 syntax (or legacy)
```
ltshb, line 79: Comment is not up to date, I would simply remove it

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -94,26 +78,26 @@ const handleLegacyParam = (
...
-                )
-                newValue = layers.map((layer) => transformLayerIntoUrlString(layer)).join(';')
-                log.debug('Importing legacy layers as', newValue)
-            } else {
-                // if not legacy, we let it go as it is
-                newValue = legacyValue
-            }
+            // for legacy layers param, we need to give the whole search query
+            // as it needs to look for layers, layers_visibility, layers_opacity and
+            // layers_timestamp param altogether
```
ltshb, line 83: Comment not 100% correct you don't pass the whole query but only the part that are needed

src/utils/legacyLayerParamUtils.js
```diff
@@ -15,22 +16,33 @@ function readUrlParamValue(url, paramName) {
...
-function isExternalLayer(layerId) {
-    return (
-        layerId &&
-        (layerId.startsWith('WMS|') || layerId.startsWith('WMTS|')) &&
-        layerId.indexOf('||') === -1
-    )
+/**
+ * @param {String} search The query made to the mapviewer
+ * @returns True if the query starts with ? or /?
+ */
```
ltshb, line 22: Add a short description here

src/utils/legacyLayerParamUtils.js
```diff
@@ -15,22 +16,33 @@ function readUrlParamValue(url, paramName) {
...
+export const isLegacyParams = (search) => {
+    const parts = search?.match(/^(\?|\/\?).*$/g) ?? null
+    return parts && Array.isArray(parts)
```
ltshb, line 25: The type check is not needed here you could simply return `return !!search?.match(/^(\?|\/\?).*$/g) ?? false`

src/utils/legacyLayerParamUtils.js
```diff
@@ -15,22 +16,33 @@ function readUrlParamValue(url, paramName) {
...
-export function isLayersUrlParamLegacy(layersParamValue) {
-    return !layersParamValue.split(';').some((layer) => {
-        return isExternalLayer(layer) || newLayerParamRegex.test(layer)
-    })
+/**
+ * Ensure opacity is within its bounds (0 and 1). Also assign a default value of 1 should the
+ * parameter be something unexpected
+ *
+ * @param {String} value
+ * @returns
```
ltshb, line 33: Missing returns type

src/utils/legacyLayerParamUtils.js
```diff
@@ -45,93 +57,75 @@ export function isLayersUrlParamLegacy(layersParamValue) {
...
+            // checking if visibility is set in URL
+            if (layerVisibilities.length > index) {
+                layer.visible = layerVisibilities[index] === 'true'
+            } else {
+                // if param layers_visibility is not present, it means all layers are visible
+                layer.visible = true
+            }
+            // checking if opacity is set in the URL
+            if (layerOpacities.length > index) {
+                layer.opacity = Number(layerOpacities[index])
```
ltshb, line 118: `Number` here is superflous as it is already in the parseOpacity method.

src/utils/__tests__/legacyLayerParamUtils.spec.js
```diff
@@ -8,8 +8,18 @@ import GeoAdminWMSLayer from '@/api/layers/GeoAdminWMSLayer.class'
...
+// TODO HERE : GET NEW SYNTAX FOR ALL TESTS
+/*
+newValue = getLayersFromLegacyUrlParams(
+                store.state.layers.config,
+                legacyValue,
+                params['layers_visibility'],
+                params['layers_opacity'],
+                params['layers_timestamp']
+            )
+            */
```
ltshb, line 22: I guess this TODO can be removed ?

src/utils/__tests__/legacyLayerParamUtils.spec.js
```diff
@@ -200,63 +244,21 @@ describe('Test parsing of legacy URL param into new params', () => {
...
+                    `WMS|${'https://wms.server.url?PARAM1=x&'}|layer.id|5.4.3|LayerName`,
+                    undefined,
+                    undefined,
+                    undefined
                 )
                 expect(wmsResult).to.be.an('Array').empty
             })
         })
     })
-    describe('test isLayersUrlParamLegacy', () => {
```
ltshb, line 214: Here it would be good to test the new `isLegacyParams()` and `parseOpacity` methods.

ltshb (approved):
Nice thanks just some small minor adjustements.

PR #621, opened 2024-01-30
PB-99: Added Print Map UI

ismailsunni:
Add UI for map printing function including:
- Show printing area based on layout and scale
- Code is mostly ported from the old geoadmin3, keeping the same behaviour
- Add checkboxex and button (without any functionality)

Preview:
image

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-99-print-ui-element/index.html)


src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
```diff
@@ -0,0 +1,207 @@
...
+
+        context.fillStyle = 'rgba(0, 5, 25, 0.75)'
+        context.fill()
+
+        context.restore()
+    }
+
+    // Compute the optimal scale based on the map size (layout), resolution,
+    // non-covered map vie (by header and menu tray)
+    function getOptimalScale() {
```
pakb, line 182: that also could become a computed properties

src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
```diff
@@ -0,0 +1,207 @@
...
+        var scaleHeight = (height * UNITS_RATIO * POINTS_PER_INCH) / layoutSize.height
+        var testScale = scaleWidth
+        if (scaleHeight < testScale) {
+            testScale = scaleHeight
+        }
+        var nextBiggest = null
+        var selectedLayoutScales = Array.from(store.state.print.selectedLayout.scales)
+        // Make sure it's sorted ascending
+        selectedLayoutScales.sort((a, b) => a - b)
+        selectedLayoutScales.forEach(function (scale) {
```
pakb, line 200: what you're doing here is about the same as `selectedLayoutScales.find((scale) => scale < testScale)` which will give you the first entry to match your predicate

src/store/modules/ui.store.js
```diff
@@ -253,5 +267,11 @@ export default {
         setImportFile(state, flagValue) {
             state.importFile = flagValue
         },
+        setHeaderHeight(state, height) {
+            state.headerHeight = height
+        },
+        setMenuTrayWidth(state, width) {
+            state.menuTrayWidth = width
+        },
```
pakb, line 275: you should also add the corresponding action above (and do all the type checking, `parseFloat` and what not in the action)

src/store/modules/print.store.js
```diff
@@ -18,5 +33,8 @@ export default {
     },
     mutations: {
         setPrintLayouts: (state, layouts) => (state.layouts = layouts),
+        setSelectedLayout: (state, layout) => (state.selectedLayout = layout),
+        setSelectedScale: (state, scale) => (state.selectedScale = scale),
+        setprintSectionShown: (state, isShown) => (state.printSectionShown = isShown),
```
pakb, line 38: little missing camelCase "P" in "Print" here

src/store/modules/print.store.js
```diff
@@ -4,8 +4,23 @@ import log from '@/utils/logging.js'
 export default {
     state: {
         layouts: [],
+        selectedLayout: {},
+        selectedScale: 0,
+        printSectionShown: false,
+    },
+    getters: {
+        mapSize(state) {
+            var mapAttributes = state.selectedLayout.attributes.find((atr) => atr.name === 'map')
```
pakb, line 13: might be good considering handling the case when `mapAttributes` is undefined (or has no `clientParams` etc...), maybe by returning `null` or `undefined` in these cases.

Also please switch to a `const` instead of `var`

ismailsunni: I already use `?`. It should be enough, to handle null or undefined cases. Right?

src/modules/menu/components/header/HeaderWithSearch.vue
```diff
@@ -60,18 +60,45 @@ export default {
...
+    },
     methods: {
         resetApp() {
             // an app reset means we keep the lang and the current topic but everything else is thrown away
             window.location = `${window.location.origin}?lang=${this.currentLang}&topic=${this.currentTopicId}`
         },
+        handleResize() {
+            // Update the height when the window is resized
+            this.currentHeight = this.$refs.header.clientHeight
+            this.emitHeightChangeSignal()
```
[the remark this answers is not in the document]
ismailsunni: ah ok, I thought to make the component independent. But I think you are right.

src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
ismailsunni: I need to sort descending with that code, but that's ok, I use it, since it's simpler

src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
ismailsunni: I am not sure about this one. The function has several references to the store. If we make it a computed property, it will give unnecessary computation. It's also only called once when a user activates the printArea (opening the print menu).

Or is there any other reason?

src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
```diff
@@ -0,0 +1,207 @@
...
+    function deactivatePrintArea() {
+        map.removeLayer(worldPolygon)
+        while (deregister.length > 0) {
+            var item = deregister.pop()
+            if (typeof item === 'function') {
+                item()
+            } else {
+                item.target.un(item.type, item.listener)
+            }
+        }
```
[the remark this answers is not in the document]
ismailsunni: I am not sure about using `.forEach` here since I also want to remove the element. Or do you mean empty the `deregister` (e.g. `deregister = []`) after `.forEach`?

src/modules/map/components/openlayers/utils/print-area-renderer.composable.js
pakb: no the main reason was that it would only be calculated when necessary (meaning on value change).
But if what you describe is correct, meaning there will be more changes to its dependencies / more recomputation as number of calls, let's keep it a function then

ismailsunni:
@pakb done, it was caused by the default polygon style in OL.

pakb (approved):

PR #623, opened 2024-01-30
BGDIINF_SB-3110: Compare Slider Tests

ltkum:
Goal : add e2e tests for the compare slider component and the compare toggle menu button.

[Test link](https://sys-map.dev.bgdi.ch/preview/bgdiinf_sb-3110-creating-compare-slider-tests/index.html)


ltkum:
The ticket has changed and is now : https://jira.swisstopo.ch/browse/PB-98 (no longer BGDIINF_SB-3110)

tests/cypress/tests-e2e/compareSlider.cy.js
```diff
@@ -0,0 +1,445 @@
...
+            })
+        })
+        /*
+            in the current implementation, cutting a layer doesn't stop it from being selected
+            (this is the exact same implementation here as in the old mapviewer). While the features
+            are not visible, they are still selectable. Once we have a compare Slider which stops features
+            from being selected when cut, we can reactivate the skipped tests here
+        */
+        context.skip('Cutting layers', () => {
+            it.skip('cuts from the correct spot until the end of the map, and only from the top layer', () => {
```
pakb, line 230: I think it would be better to only keep the `.skip` from the parent context, and not skip children. It will be easier in the future to re-activate the whole block this way.

tests/cypress/tests-e2e/compareSlider.cy.js
pakb, line 1: As a general comment on Cypress tests writing, after a couple iteration with Brice (and reading https://docs.cypress.io/guides/references/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion )
we've ended on this way of writing them :
Try not to load the app to only test one thing, but instead couple many tests (that are related) into one big `.it(..)`, even if it means the state of the previous "assertion" alters a bit what the result of the test is.

So in your case that would mean something like
```
load app
test initial state
click somewhere
test new state
click somewhere else
test new state
etc...
```
instead of loading + testing for each case (loading the app is quite costly and lengthy)

src/router/storeSync/CompareSliderParamConfig.class.js
```diff
@@ -0,0 +1,39 @@
...
+        // small check here: if the user is inserting a non valid value, we won't activate the slider
+        if (urlParamValue > 0.0 && urlParamValue < 1.0) {
+            promisesForAllDispatch.push(store.dispatch('setCompareSliderActive', true))
+        }
+    }
+    return Promise.all(promisesForAllDispatch)
+}
+
+function generateCompareSliderUrlParamFromStore(store) {
+    if (store.state.ui.compareRatio !== null && store.state.ui.isCompareSliderActive === true) {
```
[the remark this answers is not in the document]
ltkum: that's true

tests/cypress/tests-e2e/compareSlider.cy.js
ltkum: I've tried to consolidate a few tests in one instance of testing. If you feel we could still merge some tests, feel free to tell me.

pakb (approved):

PR #624, opened 2024-01-30
PB-239: Add form for reporting problem

Angi-Kinas:
- Adding report problem button and form
- Show the button on all host beside test.map.geo.admin.ch 
- Update send feedback api to upload attachment
- The report problem doesn't include the drawing tool (will be added later)

Ticket: https://jira.swisstopo.ch/browse/PB-239

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-239-add-report-problem/index.html)


tests/cypress/tests-e2e/reportProblem.cy.js
```diff
@@ -0,0 +1,170 @@
...
+            cy.get('[data-cy="report-problem-button"]').should('be.visible').click()
+            cy.get('[data-cy="report-problem-form"]').should('be.visible')
+        })
+        it('should be in the header as a link on desktop', () => {
+            cy.viewport(1920, 1080)
+
+            cy.get('[data-cy="report-problem-link-button"]').should('be.visible').click()
+            cy.get('[data-cy="report-problem-form"]').should('be.visible')
+        })
+    })
```
ltshb, line 27: Generally speaking the number of test case `it()` in cypress are expensive, for the cypress.io cloud we pay per it() and more `it()` are less performant (test duration is longer). So due to this we tend to have fewer `it` in cypress e2e tests. You should merge some of the `it` together. To separate logic in one `it` we use `cy.log()` to mention what will be tested to ease debugging in the case of test failure.
So you can keep the `context` as is but try to merge some of the it together. Note in one `it` you can do several `cy.gotToMapView`

ismailsunni: Ah ok, I have reduced the number of test case now (now only 5).

src/store/modules/ui.store.js
```diff
@@ -228,6 +233,26 @@ export default {
...
             return state.featureInfoPosition === FeatureInfoPositions.NONE
         },
+        /**
+         * Flag to display to display give feedback button/form. On localhost, it will always shown
+         * for testing purpose
+         */
+        hasGiveFeedbackButton(state) {
+            if (state.hostname.includes('localhost')) {
+                return true
+            }
```
pakb, line 243: Instead of only showing that on `localhost` it might be better to check that the value of `.hasDevSiteWarning` is true. So that on all our dev hosting both link will be visible (but will be different on either prod staging)

src/store/modules/ui.store.js
```diff
@@ -228,6 +233,26 @@ export default {
...
+            return GIVE_FEEDBACK_HOSTNAMES.some((hostname) => state.hostname.includes(hostname))
+        },
+        /**
+         * Flag to display to display report problem button/form. On localhost, it will always shown
+         * for testing purpose
+         */
+        hasReportProblemButton(state, getters) {
+            if (state.hostname.includes('localhost')) {
+                return true
+            }
```
pakb, line 253: same

tests/cypress/tests-e2e/reportProblem.cy.js
```diff
@@ -0,0 +1,173 @@
...
+            cy.log('should be in the header as a link on desktop')
+            cy.goToMapView()
+            cy.viewport(1920, 1080)
+            cy.get('[data-cy="report-problem-link-button"]').should('be.visible').click()
+            cy.get('[data-cy="report-problem-form"]').should('be.visible')
+            // Close the form
+            cy.get('[data-cy="modal-close-button"]').click()
+        })
+    })
+    context('From validation', () => {
```
pakb, line 35: little typo, `From`=>`Form`

tests/cypress/tests-e2e/reportProblem.cy.js
pakb, line 1: IMHO you can merge all `it` into one big test, they are quite small

ltshb:
image
Those two lines should have less space in between in my opinion
- Also the button give feedback should be removed, either we have "give feedback" or "report a problem", the first one should only be on "test.map.geo.admin.ch" host.
- Once the message has been sent, the focus should be on the close button which would allow to close by simply pressing enter.
- The input for files should not have tabindex, only the browse button should have tabindex. Having tabindex on the input is useless as the user cannot enter anything.
- For the email I would use the same translation key as for give feedback (adding the optional word)
- When sending a feedback as anonymous the version is missing in email 
```
Anonymous sent a feedback:
 test.

Permalink: https://sys-s.dev.bgdi.ch/0uifk7u4gxdd.

User-Agent:          Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36

Application Version: unknown
```

src/api/feedback.api.js
```diff
@@ -6,19 +6,19 @@ import { API_SERVICES_BASE_URL, APP_VERSION } from '@/config'
...
  */
 export default async function sendFeedback(
+    subject,
     text,
-    rating = null,
-    maxRating = null,
     kmlFileUrl = null,
-    email = null
+    email = null,
+    attachment = null
```
ltshb, line 21: When already changing this, it would be better to use an options object
`(subject, text, {email = null, kmlFileUrl = null, attachement = null})`
So below you don't have to pass optional argument and if you want to pass only one you can simply do `sendFeedback(subject, text, {email: 'hello@world.com'})`

src/modules/menu/components/menu/feedback/FeedbackButton.vue
```diff
@@ -160,12 +160,17 @@ export default {
...
+                subject += ' User feedback'
                 const feedbackSentSuccessfully = await sendFeedback(
+                    subject,
                     this.feedback.message,
-                    this.feedback.rating,
-                    this.maxRating,
                     this.activeKmlLayer?.kmlFileUrl,
-                    this.feedback.email
+                    this.feedback.email,
+                    undefined // file attachment
```
ltshb, line 170: See comment above. Also personally I found better to use `null` instead of `undefined` for optional argument, but if you use the optional signature above then no need to pass anything for attachement here.

src/modules/menu/components/menu/feedback/FeedbackToolbar.vue
```diff
@@ -5,25 +5,33 @@
             'me-2': showAsLinks,
         }"
     >
-        <FeedbackButton :show-as-link="showAsLinks" />
+        <FeedbackButton v-if="hasGiveFeedbackButton" :show-as-link="showAsLinks" />
+        <ReportProblemButton v-if="hasReportProblemButton" :show-as-link="showAsLinks" />
```
ltshb, line 9: Because this file is so small and you touch a lots of lines I would suggest to quickly change it to composition api

src/modules/menu/components/menu/reportProblem/ReportProblemButton.vue
```diff
@@ -0,0 +1,222 @@
...
+import sendFeedback from '@/api/feedback.api'
+import { createShortLink } from '@/api/shortlink.api'
+import ImportFileLocal from '@/modules/menu/components/advancedTools/ImportFile/ImportFileLocal.vue'
+import HeaderLink from '@/modules/menu/components/header/HeaderLink.vue'
+import ModalWithBackdrop from '@/utils/components/ModalWithBackdrop.vue'
+import log from '@/utils/logging'
+
+// comes from https://v2.vuejs.org/v2/cookbook/form-validation.html#Using-Custom-Validation
+const EMAIL_REGEX =
+    /^(([^<>()[\]\\.,;:\s@"]+(\.[^<>()[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/
```
ltshb: Instead of copying this regex from FeedbackButton, better create a util function in src/utils/utils.js `isEmailValid()` so it can be reused in both formular.

src/store/modules/ui.store.js
```diff
@@ -228,6 +233,26 @@ export default {
...
             return state.featureInfoPosition === FeatureInfoPositions.NONE
         },
+        /**
+         * Flag to display to display give feedback button/form. On localhost, it will always shown
+         * for testing purpose
+         */
+        hasGiveFeedbackButton(state, getters) {
+            if (getters.hasDevSiteWarning) {
+                return true
+            }
```
ltshb, line 243: Make no sense to have this if, if we want the feedback button on all dev site, we should add the host name in the list below, just keep one way of configuring the presence of this button.
That said on sys-map.dev.bgdi.ch make sense to have both button for testing purpose, while on sys-map.int|prod.bgdi.ch and map.geo.admin.ch we should only have the report a problem button

src/store/modules/ui.store.js
```diff
@@ -228,6 +233,26 @@ export default {
...
+        },
+        /**
+         * Flag to display to display report problem button/form. On localhost, it will always shown
+         * for testing purpose
+         */
+        hasReportProblemButton(state, getters) {
+            if (getters.hasDevSiteWarning) {
+                return true
+            }
+            return !getters.hasGiveFeedbackButton
```
ltshb, line 254: Based on other comment better to have a new config variable `REPORT_PROBLEM_HOSTNAMES`

src/store/modules/ui.store.js
ismailsunni: Ok, to sum up:
```js
export const GIVE_FEEDBACK_HOSTNAMES = ['localhost', 'sys-map.dev.bgdi.ch', 'test.map.geo.admin.ch']

export const REPORT_PROBLEM_HOSTNAMES = [
    'localhost',
    'sys-map.dev.bgdi.ch',
    'sys-map.int.bgdi.ch',
    'sys-map.prod.bgdi.ch',
    'map.geo.admin.ch',
]
```

*adding the `test.map.geo.admin.ch'` from the JIRA ticket.

ltshb:
@ismailsunni about the version I've seen that it seems to be a build/deployment issue as the version on your PR test link is not displayed as well. I'll investigate this, but it is very likely not related to this PR.

src/modules/menu/components/menu/common/EmailValidationField.vue
```diff
@@ -0,0 +1,45 @@
...
+})
+// Watches
+watch(email, () => {
+    emits('email-updated', email.value)
+})
+</script>
+
+<template>
+    <div>
+        <span><slot name="header"></slot></span>
```
ltshb, line 30: Why using a slot for the field label ? I think it would be better to use a props containing a translation key to use. The label should also use `<label></label>` tag using the `for` attribute. See bootstrap examples (https://getbootstrap.com/docs/4.0/components/forms/).
Instead of using `isRequestPending` props which is quite specific I would use `disabled` props, letting the user defining what event/action is disabling the component.
Finally this component should be moved to src/utils/components/ as it is very generic and could be used any where in the application.

src/modules/menu/components/menu/reportProblem/ImportFileLocal.vue
```diff
@@ -0,0 +1,109 @@
+<script setup>
+import { computed, ref, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
+
+const LOCAL_UPLOAD_ACCEPT = '.kml,.gpx,.pdf,.zip,.jpg,.jpeg,.kmz'
+const LOCAL_UPLOAD_MAX_SIZE = 250 * 1024 * 1024 // 250mb
+
+const i18n = useI18n()
+
+const emits = defineEmits(['file-selected'])
```
ltshb, line 10: Same here as it seems to be the same code as the import kml file local, this should move to src/utils/component and be renamed as `InputLocalFile`, the `LOCAL_UPLOAD_ACCEPT` and `LOCAL_UPLOAD_MAX_SIZE` should be properties that the user can define, which would allow to reuse this component for other input file type. Also rename variables and attributes accordingly (e.g. ImportFileLocalInput => InputLocalFile, data-cy="import-file-local-content" => data-cy="input-file-local-content")

ltshb:
The issue with the version is that your PR is too old and should be rebased to develop, due to this the version using `git describe --tag` cannot work as in the CI the clone is done with a limited depth, so rebasing the PR to develop should solve the version issue.

src/modules/menu/components/menu/reportProblem/ReportProblemButton.vue
```diff
@@ -0,0 +1,215 @@
...
+        request.value.failed = true
+    } finally {
+        request.value.pending = false
+    }
+    await nextTick()
+    // scrolling down to make sure the message with request results is visible to the user
+    if (request.value.failed) {
+        requestResults.value.scrollIntoView()
+    } else if (request.value.completed) {
+        reportProblemCloseSuccessful.value.focus()
```
ltshb, line 84: The focus on success doesn't work on the test link using chrome

ltshb (approved):
All good just an issue with the close button focus which don't seems  to work

src/modules/menu/components/menu/reportProblem/ReportProblemButton.vue
ismailsunni: Hmm, very strange, it works for me as seen below. I use keyboard to fill up the form, tab to move, and enter/space to press the button.
Peek 2024-04-04 15-53 - auto focus on close button

I tried on Chrome and Firefox, and I have a test case for that also: https://github.com/geoadmin/web-mapviewer/pull/624/files#diff-60740a7e3043624f1fc8b73fa41349cc88bdf6ad1131cf3f3575bcdf73cd7a25R100

PR #625, opened 2024-01-30
PB-114: Made KML retro compatible with mf-geoadmin3

ltshb:
Now the KML are retro compatible with mf-geadmin3, see test link below which import a KML created by web-mapviewer into mf-geoadmin3
- KML file generated by web-mapviewer: https://sys-public.dev.bgdi.ch/api/kml/files/UguLO2B4RqKGqLosIGOsxA
- Link to mf-geoadmin3 using the file above: https://map.geo.admin.ch/?lang=fr&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=KML%7C%7Chttps:%2F%2Fsys-public.dev.bgdi.ch%2Fapi%2Fkml%2Ffiles%2FUguLO2B4RqKGqLosIGOsxA&E=2623716.62&N=1232091.54&zoom=4

In theory the KML are aslo compatible with other KML viewer, but for this it requires first PB-264 to be done.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-114-kml/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -55,6 +55,7 @@ watch(featureIds, (next, last) => {
     }
 })
 
+let unsubscribe = null
 onMounted(() => {
     // if icons have not yet been loaded, we do so
     if (availableIconSets.value.length === 0) {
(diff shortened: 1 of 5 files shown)


ltshb:
@pakb this PR is ready to review. See the different commits I had first another approach for the `title` and `description` which where only set during exporting while now I tried to keep them in sync with the `EditableFeature`. I think the sync approach is better however I'm not very happy with the code that I used for the sync It is a bit more obscure on how it works. Maybe you have a better idea for that ? I first tried to put the store subscribe into the `useModifyInteraction.composabl.js` but then the subscribe would habe been done multiple times.
I could do a store plugin but it would need access to the `drawingLayer` global object...

src/modules/drawing/DrawingModule.vue
```diff
@@ -93,6 +93,8 @@ onBeforeUnmount(async () => {
     document.removeEventListener('contextmenu', removeLastPoint)
     document.removeEventListener('keyup', removeLastPointOnDeleteKeyUp)
 
+    log.debug(`unsubscribe feature mutation`)
+
```
pakb, line 97: a little left over of the store subscribe part 😉

pakb (approved):
I don't know if it would make sense to squash what you tried here so that it isn't added to the main branch's log, or if it add values to keep it 🤔

ltshb:
> I don't know if it would make sense to squash what you tried here so that it isn't added to the main branch's log, or if it add values to keep it 🤔

I thought about squashing but then due to the complexity of the whole drawing I think it might be wise to let the commits as documentation.

PR #626, opened 2024-01-30
PB-101 : zoom on GPX extent on file added

pakb:
re-using some parts that were made for KML

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-101_gpx-zoom-to-extent/index.html)

### DIFF
--- a/src/modules/menu/components/advancedTools/ImportFile/utils.js
+++ b/src/modules/menu/components/advancedTools/ImportFile/utils.js
@@ -6,6 +6,7 @@ import KMLLayer from '@/api/layers/KMLLayer.class'
 import { OutOfBoundsError } from '@/utils/coordinates/coordinateUtils'
 import { getExtentForProjection } from '@/utils/extentUtils.js'
 import GPX from '@/utils/GPX'
+import { EmptyGPXError } from '@/utils/gpxUtils.js'
 import { EmptyKMLError, getKmlExtent } from '@/utils/kmlUtils'
 
 /**
@@ -56,10 +57,14 @@ export function handleFileContent(store, content, source) {
         const metadata = gpxParser.readMetadata(content)
         const parseGpx = new DOMParser().parseFromString(content, 'text/xml')
         layer = new GPXLayer(source, true, 1.0, content, metadata)
-        const extent = getExtentForProjection(
-            store.state.position.projection,
-            bbox(gpxToGeoJSON(parseGpx))
-        )
+        const extent = bbox(gpxToGeoJSON(parseGpx))
+        if (!extent) {
+            throw new EmptyGPXError()
+        }
+        const projectedExtent = getExtentForProjection(store.state.position.projection, extent)
+        if (!projectedExtent) {
+            throw new OutOfBoundsError(`GPX out of projection bounds: ${extent}`)
+        }
         store.dispatch('zoomToExtent', extent)
         store.dispatch('addLayer', layer)
     } else {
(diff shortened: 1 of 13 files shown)


src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -50,8 +54,13 @@ export function handleFileContent(store, content, source) {
...
         const gpxParser = new GPX()
         const metadata = gpxParser.readMetadata(content)
+        const parseGpx = new DOMParser().parseFromString(content, 'text/xml')
         layer = new GPXLayer(source, true, 1.0, content, metadata)
-        // TODO : zoom to extent
+        const extent = getExtentForProjection(
+            store.state.position.projection,
+            bbox(gpxToGeoJSON(parseGpx))
+        )
+        store.dispatch('zoomToExtent', extent)
```
ltshb, line 63: I would also add the `!extent` check and throw `OutOfBoundsError` error here like in KML so we handle import of GPX tracks that are totally outside of CH. This is also a use case that I tested in my kml e2e tests

src/utils/__tests__/extentUtils.spec.js
```diff
@@ -0,0 +1,112 @@
+import { expect } from 'chai'
+import { describe, it } from 'vitest'
+
+import { LV95 } from '@/utils/coordinates/coordinateSystems'
+import { getExtentForProjection } from '@/utils/extentUtils'
+import { getKmlExtent } from '@/utils/kmlUtils.js'
+
+describe('Test extent utils', () => {
```
ltshb, line 8: Here we should also add edge test cases:

- Empty KML/GPX
- KML/GPX outside of projection bound

Or do you plan to do it in a separate PR?

src/store/modules/layers.store.js
```diff
@@ -465,7 +466,7 @@ const actions = {
             if (!extent) {
                 updatedLayer.errorKey = 'kml_gpx_file_empty'
                 updatedLayer.hasError = true
-            } else if (!getKmlExtentForProjection(rootState.position.projection, extent)) {
+            } else if (!getExtentForProjection(rootState.position.projection, extent)) {
```
ltshb, line 469: I've just seen that in your previous PR for GPX support it seems that you don't support GPX after a reload as the GPX data will not be set (no store subscriber to get the GPX data from url). This means that you would need to add that support and also need a similar logic in the layer store as here and have a `updateGpxLayer()` and test here again for the extent being in projection bound like for kml. So we would handle the case were an online GPX is updated and the new version is out of bound.

pakb: That's something I wanted to do in another PR

src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -50,8 +55,17 @@ export function handleFileContent(store, content, source) {
     } else if (isGpx(content)) {
         const gpxParser = new GPX()
         const metadata = gpxParser.readMetadata(content)
+        const parseGpx = new DOMParser().parseFromString(content, 'text/xml')
         layer = new GPXLayer(source, true, 1.0, content, metadata)
-        // TODO : zoom to extent
+        const extent = bbox(gpxToGeoJSON(parseGpx))
+        if (!extent) {
+            throw new EmptyGPXError()
```
ltshb, line 62: You also need to handle this error here https://github.com/geoadmin/web-mapviewer/blob/d9e436d7925739b78e7f3feee678ee6303da1841/src/modules/menu/components/advancedTools/ImportFile/ImportFileLocalTab.vue#L79

ltshb (approved):

PR #633, opened 2024-02-02
PB-101 : add support for GPX reloading and e2e tests

pakb:
Reworking the KML loading pipeline, which is nearly identical to GPX, to also support this new format
Moving the metadata/data loading parts from the import utils to a dedicated store plugin, so that these things can be loaded on app reload when the layer is added through the URL directly
Add e2e test to cover GPX support

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-101_e2e_test_and_url_support/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -2,6 +2,7 @@ import AbstractLayer from '@/api/layers/AbstractLayer.class'
 import ExternalGroupOfLayers from '@/api/layers/ExternalGroupOfLayers.class'
 import LayerTypes from '@/api/layers/LayerTypes.enum'
 import { getExtentForProjection } from '@/utils/extentUtils.js'
+import { getGpxExtent } from '@/utils/gpxUtils.js'
 import { getKmlExtent, parseKmlName } from '@/utils/kmlUtils'
 import { ActiveLayerConfig } from '@/utils/layerUtils'
 import log from '@/utils/logging'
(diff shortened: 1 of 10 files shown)


src/store/modules/layers.store.js
```diff
@@ -447,22 +448,30 @@ const actions = {
...
-        if (kmlData) {
-            updatedLayer.name = parseKmlName(kmlData) || 'KML'
-            updatedLayer.kmlData = kmlData
+        if (data) {
+            let extent
+            if (updatedLayer.type === LayerTypes.KML) {
+                updatedLayer.name = parseKmlName(data) || 'KML'
+                updatedLayer.kmlData = data
+                extent = getKmlExtent(data)
+            } else if (updatedLayer.type === LayerTypes.GPX) {
```
ltkum, line 468: I think there is no way to enter this function with anything other than KML or GPX layers, we could get rid of one check by turning this into a simple else

That being said, this is a bit more resilient (in case a non GPX / KML layer enter this function) and easier to read. I would add a 'else' statement to have a  default comportment when entering with an other type of Layer, to prevent side effects to go untracked for a while.

ltkum (approved):
I offered a small opinion on a change I think is not stupid. We can completely assume it's only going to be kmls and gpx layers sent to the function (and it should be the case), but if you want to merge without doing it, fine by me. Everything seems to be in order :+1

src/store/modules/layers.store.js
pakb: I thought about it too, but I think it's a better solution to keep the two `if`s so that a non GPX/KML layer would fail quickly and without going into some kind of default parsing

PR #636, opened 2024-02-09
PB-269 : GPX/KML generic feature identification

pakb:
Enhancing how the tooltip is placed for geometry based features, at the most southern point of the most southern feature (also work the same way for points, but that is less obvious)

Should also fix WMS/WMTS feature identification when returned features have a complex geometry instead of a single point.
Fixing a small bug with time sensitive layer where features from all years were requested at the same time, now filtering taking the selected year into account.

A good example to test all this would be `ch.swisstopo.swissboundaries3d-gemeinde-flaeche.fill` while having a drawing on top of it, and clicking on one of the drawing features

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-269_gpx_kml_feature_generic_identification/index.html)


ltshb:
You cannot edit anymore features in drawing mode !

src/utils/identifyOnVectorLayer.js
```diff
@@ -169,14 +150,15 @@ export function identifyKMLFeatureAt(kmlLayer, coordinate, projection, resolutio
...
-                    coordinates: reprojectCoordinates(feature.geometry.coordinates, projection),
-                    title: feature.properties.name,
-                    description: feature.properties.description,
-                    featureType: feature.properties.type.toUpperCase(),
-                })
+                return new LayerFeature(
+                    kmlLayer,
+                    feature.id,
+                    feature.properties.name,
+                    generateHtmlPopup(kmlLayer.name, feature.properties.description),
```
ltshb, line 157: This is way too dangerous and subject to XSS attack we cannot trust the feature description content that comes from users including geoadmin KML, user can enter their own html code in a kml description. We should rethink this component and use an HTML sanitizer like https://github.com/cure53/DOMPurify

src/utils/identifyOnVectorLayer.js
```diff
@@ -1,30 +1,36 @@
...
-        }
-        popup += '</div>'
+        popup += featureDescription
+    } else {
+        popup += t('no_more_information')
     }
-    popup += `</div>`
+    popup += '</div></div>'
     return popup
 }
```
ltshb, line 35: I think here we should find a more Vue component way of doing this.

ltshb (changes requested):

src/utils/identifyOnVectorLayer.js
pakb: I've changed how we render the HTML for features, and if we receive a string value from the feature I sanitize it before injecting it (with this DOMPurify lib you found).

src/api/features/EditableFeature.class.js
```diff
@@ -0,0 +1,221 @@
...
+        this._icon = newIcon
+        this.emitStylingChangeEvent('icon')
+    }
+
+    /** @returns {String} */
+    get iconUrl() {
+        // TODO PB-266 use a fix size for icon from the backend and uses the icon style scale instead
+        return this._icon?.generateURL(this.iconSize, this.fillColor)
+        // return this._icon?.generateURL(LARGE, this.fillColor)
+    }
```
ltshb, line 151: This will needs some rebase merge conflict solving with #641 😟

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -0,0 +1,48 @@
...
+
+const featureDataEntries = computed(() => {
+    if (hasFeatureStringData.value) {
+        return []
+    }
+    return Object.entries(feature.value.data).filter(([_, value]) => value)
+})
+function sanitizeHtml(htmlText) {
+    return DOMPurify.sanitize(htmlText)
+}
```
ltshb, line 29: Would it not make more sense to use a computed here instead of a method ? This would improve performance.

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -0,0 +1,48 @@
...
+                <div class="htmlpopup-content-title">{{ i18n.t(key) }} :</div>
+                <!-- eslint-disable-next-line vue/no-v-html-->
+                <div class="htmlpopup-content-data" v-html="sanitizeHtml(value)"></div>
+            </div>
+            <div v-if="featureDataEntries.length === 0">
+                {{ i18n.t('no_more_information') }}
+            </div>
+        </div>
+    </div>
+</template>
```
ltshb, line 48: Where are the css classes `htmlpopup-content`, `htmlpopup-header`, ... ?

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -0,0 +1,48 @@
...
+<template>
+    <!-- eslint-disable-next-line vue/no-v-html-->
+    <div v-if="hasFeatureStringData" v-html="sanitizeHtml(feature.data)" />
+    <div v-else class="htmlpopup-container">
+        <div class="htmlpopup-header">{{ feature.layer.name }}</div>
+        <div class="htmlpopup-content">
+            <div v-for="[key, value] in featureDataEntries" :key="key" class="d-flex gap-1">
+                <div class="htmlpopup-content-title">{{ i18n.t(key) }} :</div>
+                <!-- eslint-disable-next-line vue/no-v-html-->
+                <div class="htmlpopup-content-data" v-html="sanitizeHtml(value)"></div>
```
ltshb, line 41: The div here should have some CSS tweaking, see result below 
image
You can use the kml from this link for testing again https://sys-s.dev.bgdi.ch/jlxh371ioze1

ltshb (commented):
Looks promissing, I've found an issue that is not covered by the E2E test (might be good to add test cover):
- Create a drawing with an icon
- Reload the page (loosing the adminId)
- Reopen the drawing => should create a copy of the previous drawing
- select the existing icon => now its tooltip is not displayed, new icon have the tooltip but not copied icon.

src/modules/infobox/components/FeatureDetail.vue
pakb: Classes that are defined globally in FeatureList.vue (historically, I might move the here though...)
They are also used by our backend HTML code generator for popup

src/modules/infobox/components/FeatureDetail.vue
pakb: I changed a bit how it's done, I sanitize the content of JSON data in the computed, and when the data is a String but doesn't come from our (trusted) backends.
I don't think I can make a computed out of every use case, so I kept the function as it is here

ltshb (commented):
Very nice just a few minor comments

ltshb (approved):

src/api/features/LayerFeature.class.js
```diff
@@ -0,0 +1,34 @@
...
+    get id() {
+        return `${this.layer.getID()}-${this._id}`
+    }
+
+    get popupDataCanBeTrusted() {
+        // We can't trust the content of the popup data for external layers, and for KML layers.
+        // For KML, the issue is that user can create text-rich (HTML) description with links, and such.
+        // It would then be possible to do some XSS through this, so we need to sanitize this before showing it.
+        return !this.layer.isExternal && this.layer.type !== LayerTypes.KML
+    }
```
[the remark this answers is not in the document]
pakb: As discussed with you, we need to keep this distinction for the time being, as our backends sends us HTML directly, which sometimes contains `<iframe>` tags.
Meaning if we do not trust our backend anymore we will be break the HTML popup for such layers (i.e. public transportation stop table, wind direction, etc)

PR #637, opened 2024-02-12
PB: 274 toggle fullscreen mode f11

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-274-toggle_fullscreen_mode_f11/index.html)


LukasJoss:
The F11 key is protected in chrome while in fullscreen, therefore I opted to detect the window and screen size, I dont know if this option is sufficiently stable

PR #638, opened 2024-02-12
PB-276 : bugfix missing compare slider tooltip

ltkum:
Issue : tooltip was missing from compare Slider menu component

fix : we now have a tooltip. We also use the same method across the advanced menu to ask for translations, rather than a mix of ways.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-278-bugfix-missing-compare-slider-tooltip/index.html)

### DIFF
--- a/src/modules/menu/components/advancedTools/MenuAdvancedToolsList.vue
+++ b/src/modules/menu/components/advancedTools/MenuAdvancedToolsList.vue
@@ -1,5 +1,6 @@
 <script setup>
 import { computed, toRefs } from 'vue'
+import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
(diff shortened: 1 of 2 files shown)


src/modules/menu/components/advancedTools/MenuAdvancedToolsList.vue
```diff
@@ -64,7 +62,7 @@ function onToggleImportFile() {
         >
             <ModalWithBackdrop
                 v-if="showImportFile"
-                :title="i18n.t('import_file')"
+                title="import_file"
```
ltshb, line 65: This breaks the modal view translation

ltshb (approved):

PR #639, opened 2024-02-12
PB-123 : Change color of hovered feature

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-123_change_color_hovered_feature/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureList.vue
+++ b/src/modules/infobox/components/FeatureList.vue
@@ -25,6 +25,15 @@ function generateFeatureIdForList(feature, indexInList) {
     }
     return featureId
 }
+function highlightFeature(feature) {
+    store.dispatch('setHighlightedFeatureId', {
+        featureId: feature?.id,
+        dispatcher: 'FeatureList.vue',
+    })
+}
+function clearHighlightedFeature() {
+    store.dispatch('setHighlightedFeatureId', { featureId: null, dispatcher: 'FeatureList.vue' })
+}
 </script>
 
 <template>
@@ -38,18 +47,27 @@ function generateFeatureIdForList(feature, indexInList) {
             :key="generateFeatureIdForList(feature, index)"
             :feature="feature"
             class="feature-list-item"
+            @mouseenter.passive="highlightFeature(feature)"
+            @mouseleave.passive="clearHighlightedFeature"
         />
     </div>
 </template>
 
 <style lang="scss" scoped>
 @import 'src/scss/media-query.mixin';
+@import 'src/scss/variables-admin.module';
+
+$feature-padding: 5px;
 
 .feature-list {
     margin: 0;
-    padding: 0;
+    padding: $feature-padding;
     list-style: none;
 
+    &-item:hover {
+        box-sizing: border-box;
+        outline: $feature-padding solid $mocassin-to-red-1;
+    }
     &-row {
         display: grid;
         // on mobile (default size) only one column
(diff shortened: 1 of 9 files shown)


src/modules/map/components/WarningRibbon.vue
```diff
@@ -14,7 +14,7 @@ export default {
 
 <style lang="scss" scoped>
 @import 'src/scss/media-query.mixin';
-@import 'src/scss/variables-admin';
+@import 'src/scss/variables-admin.module';
```
ltshb, line 17: Why the `.module` here ?

ltshb (commented):
Code looks good however on the old viewer the feature on the tooltip overlay was also marked with a hover color, this would be also great to have this on the new viewer, not only changing the feature color but also the tooltip color.

src/modules/map/components/WarningRibbon.vue
pakb: so that Vite handles this SCSS file as something it can send to Javascript (that's what I read somewhere in Vite support/documentation)
Without this module, it won't export things to JS

pakb:
> Code looks good however on the old viewer the feature on the tooltip overlay was also marked with a hover color, this would be also great to have this on the new viewer, not only changing the feature color but also the tooltip color.

That might be challenging with the concept of Infobox. In this regards, I also saw that the Infobox doesn't trigger the hover mechanism, so I have to rethink how I do it.
I tried to keep it isolated to OpenLayersHighlightedFeatures.vue, but if we want to have this also triggered through the Infobox I'm afraid I have to store this hovered feature somewhere in the store

src/modules/infobox/components/FeatureList.vue
```diff
@@ -25,6 +25,15 @@ function generateFeatureIdForList(feature, indexInList) {
...
 }
+function highlightFeature(feature) {
+    store.dispatch('setHighlightedFeatureId', {
+        featureId: feature?.id,
+        dispatcher: 'FeatureList.vue',
+    })
+}
+function clearHighlightedFeature() {
+    store.dispatch('setHighlightedFeatureId', { featureId: null, dispatcher: 'FeatureList.vue' })
+}
```
ltshb, line 36: Based on your comment of #644 I would use `highlightedFeatureId` instead of `featureId`

src/store/modules/features.store.js
```diff
@@ -26,13 +27,24 @@ export default {
...
+            })
             if (Array.isArray(features)) {
                 commit('setSelectedFeatures', features)
             }
         },
         /** Removes all selected features from the map */
         clearAllSelectedFeatures({ commit }) {
             commit('setSelectedFeatures', [])
+            commit('setHighlightedFeatureId', {
+                featureId: null,
```
ltshb, line 42: dito

src/store/modules/features.store.js
```diff
@@ -26,13 +27,24 @@ export default {
...
+                dispatcher: 'clearAllSelectedFeatures',
+            })
+        },
+        setHighlightedFeatureId({ commit }, { featureId = null, dispatcher }) {
+            commit('setHighlightedFeatureId', { featureId, dispatcher })
```
ltshb, line 47: dito

src/store/modules/features.store.js
```diff
@@ -223,6 +235,9 @@ export default {
         setSelectedFeatures(state, features) {
             state.selectedFeatures = [...features]
         },
+        setHighlightedFeatureId(state, { featureId }) {
+            state.highlightedFeatureId = featureId
```
ltshb, line 239: dito

ltshb (approved):

PR #628, opened 2024-01-31
PB-266: Added ADR about drawing text and icon size

ltshb:
Based on discussion with @boecklic  and @pakb 

Work is done in #641 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-114-icon-size/index.html)

PR #641, opened 2024-02-13
PB-266: Corrected KML icons size

ltshb:
Now the KML icon size and text size is retro compatible with the old kml adding a new `extra large` size.
The icon scaling is a bit tricky due to a factor to normalize icons to 32 pixel introduced by open layer in version 6.7 (old viewer used an older openlayer version, see https://github.com/openlayers/openlayers/pull/12695).

Relates to #628 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-266-icon-size/index.html)

### DIFF
--- a/src/api/icon.api.js
+++ b/src/api/icon.api.js
@@ -1,7 +1,7 @@
 import axios from 'axios'
 
 import { API_SERVICES_BASE_URL } from '@/config'
-import { MEDIUM, RED } from '@/utils/featureStyleUtils'
+import { RED } from '@/utils/featureStyleUtils'
 import log from '@/utils/logging'
 
 /**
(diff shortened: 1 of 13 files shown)


src/api/icon.api.js
```diff
@@ -131,20 +131,20 @@ export class DrawingIcon {
...
      * applied. If no iconColor is given, red will be applied (if applicable, as non-colorable icons
      * will not have {r}, {g}, {b} part of their template URL)
      *
-     * @param {FeatureStyleSize} iconSize The size (or scale) to use for this icon's URL
      * @param {FeatureStyleColor} iconColor The color to use for this icon's URL
+     * @param {Number} iconScale The scale to use for this icon's URL
      * @returns {String} A full URL to this icon on the service-icons backend
      */
-    generateURL(iconSize = MEDIUM, iconColor = RED) {
+    generateURL(iconColor = RED, iconScale = 1) {
```
pakb, line 142: why not using a constant anymore? (SMALL, MEDIUM, etc...)

src/utils/kmlUtils.js
```diff
@@ -413,6 +402,15 @@ export function getEditableFeatureFromKmlFeature(kmlFeature, availableIconSets)
...
     const iconStyle = getIconStyle(style)
     const iconArgs = iconStyle ? parseIconUrl(iconStyle.getSrc()) : null
+    if (iconArgs?.isLegacy) {
+        // On the legacy drawing, openlayer used the scale from xml as is, but since openlayer
+        // version 6.7, the scale has been normalized to 32 pixels, therefore we need to add the
+        // 32 pixel scale factor below
+        // scale_factor := iconStyle.getSize()[0] / 32
+        // iconStyle.getSize()[0] = 48 (always 48 pixel on old viewer)
+        // scale_factor = 48/32 => 1.5
+        iconStyle.setScale(iconStyle.getScale() * 1.5)
```
pakb, line 412: it might be good to move the constant `ICON_XML_SCALE_FACTOR` declared in E2E test here, export it, and re-use it in the test too.

src/api/icon.api.js
ltshb: Because the icon scale of MEDIUM, ... is the open layer scale and here I want to be sure to get the icon at its scale 1 by default and we always takes icon from backend at the same scale and let openlayer do the scaling.

src/utils/kmlUtils.js
ltshb: good point

pakb (approved):

PR #642, opened 2024-02-13
PB-190: Implementing bod layer Id

ltkum:
Issue : We want to implement the {bod-layer-id} parameter functionality as follows :

parameter 1: {layer-id}={features,id,separated,by,commas}
	This should activate the layer, and select the features
parameter 2: showToolTip={true, false}
	If set to false, the features should be highlighted, but the
tooltip should remain hidden. This is only valid at startup.

We added a small router that should be called after the legacy Parameters router, whose goal is to get rid of the dynamic parameter in the URL, transfer it to the 'layers' param and ensure it is visible.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-190-implement-bodid-feature/index.html)

### DIFF
--- a/src/router/BodLayerId.routerPlugin.js
+++ b/src/router/BodLayerId.routerPlugin.js
@@ -1,33 +1,44 @@
-import layersStore from '@/store/modules/layers.store'
-
 import storeSyncConfig from './storeSync/storeSync.config'
 
 const standardsParams = storeSyncConfig.map((param) => {
-    param.urlParamName
+    return param.urlParamName
 })
 standardsParams.push('showToolTip')
-
-function parseBodLayerParamsFromQuery(query) {
+standardsParams.push('redirect')
+function parseBodLayerParamsFromQuery(query, layersConfig) {
+    console.log(query)
     if (!query) {
         return undefined
     }
-    const urlParams = new URLSearchParams(query)
+    // we remove everything until the question mark
+    const urlParams = new URLSearchParams(
+        decodeURIComponent(query.slice(query.indexOf('map?') + 3))
+    )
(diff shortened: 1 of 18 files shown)


src/router/BodLayerId.routerPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+    delete newQuery.showtooltip
+    // the center parameter is always overwritten
+    delete newQuery.center
+
+    retrieveAndDispatchFeaturesToStore(bodIdParams, !newQuery.z, store)
+
+    next({
+        name: 'MapView',
+        query: newQuery,
+    })
```
ltshb, line 96: I would not try add the layer in the `layers` url parameter here, that's the job of the `LayersParamParser`. Simply check if the layer is in the active layers in store and if not dispatch a `addLayer` with the layer ID. We don't need to force the visible flag as it is not forced on the old viewer when the layer is in active layer list but invisible.

src/router/BodLayerId.routerPlugin.js
```diff
@@ -0,0 +1,236 @@
...
+                }
+            } else {
+                changesMadeByThisPlugin = false
+                next()
+            }
+        } else {
+            changesMadeByThisPlugin = false
+            next()
+        }
+    })
```
ltshb, line 233: This should be simplified, not using the `next()`.
I would also remove the `changesMadeByThisPlugin` flag, this would mean that parameter lookup would be made twice, but it will be anyway done for every route changes. It make also sense to register this plugin after the router store sync, at the end.

ltshb (changes requested):

PR #643, opened 2024-02-14
Fix cypress e2e 3d tests

ltshb:
Avoid intercept confusion and also make sure that all tiles are loaded before
continuing cesium tests.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-e2e-3d/index.html)


tests/cypress/tests-e2e/3d/navigation.cy.js
pakb, line 1: why removing this waits here?

pakb (approved):

tests/cypress/tests-e2e/3d/navigation.cy.js
ltshb: Because of the additional waits in `waitUntilCesiumTilesLoaded`

PR #644, opened 2024-02-15
PB-285: Added `dispatcher` to  store actions and improve startup procedure (language  change)

ltshb:
In order to improve startup procedure I added a global new `dispatcher` attribute to all store actions related to the URL query parameter. This should also be done for all actions, but for the time being only added it to the url parameter related actions.

This new generic parameter gives the possibility to do different actions in the
store plugins depending on the source of the event. This avoid bulky/flaky tests
using local closure flags.

This PR already improved some performances:

- language changes performance improves as it don't re-add every layer one at a time, but in a bulk operation
- Don't load the layer config and topics at startup twice if the language is not the browser default
- Setting default topic layers is now about 400-500 ms faster due to bulk store mutation `setLayers`
- improve logging for debugging by introducing a debug log for every mutation

It also solved the following bugs:

- meteoschweiz and schneesport topic default layer reverse order (due to plConfig)
- Fix topics that only specified selectedLayers in backend (BFS, IVS, KGS, ...)

TODO in separate PR:
- [ ] readme about store mutations, see #645 
- [ ] bulk operation for adding layer from URL param and bug fixing url layers param order

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-259-performance/index.html)

### DIFF
--- a/src/modules/map/components/footer/backgroundSelector/BackgroundSelector.vue
+++ b/src/modules/map/components/footer/backgroundSelector/BackgroundSelector.vue
@@ -44,7 +44,10 @@ const sortedBackgroundLayersWithVoid = computed(() =>
 )
 
 function selectBackground(backgroundLayer) {
-    store.dispatch('setBackground', backgroundLayer)
+    store.dispatch('setBackground', {
+        value: backgroundLayer,
+        dispatcher: 'BackgroundSelector.vue',
+    })
 }
 </script>
 
--- a/src/modules/map/components/footer/backgroundSelector/BackgroundSelector.vue
+++ b/src/modules/map/components/footer/backgroundSelector/BackgroundSelector.vue
@@ -45,7 +45,7 @@ const sortedBackgroundLayersWithVoid = computed(() =>
 
 function selectBackground(backgroundLayer) {
     store.dispatch('setBackground', {
-        value: backgroundLayer,
+        bgLayer: backgroundLayer,
         dispatcher: 'BackgroundSelector.vue',
     })
 }
--- a/src/router/appLoadingManagement.routerPlugin.js
+++ b/src/router/appLoadingManagement.routerPlugin.js
@@ -1,4 +1,7 @@
 import log from '@/utils/logging'
+import { getUrlQuery } from '@/utils/utils'
+
+const STORE_DISPATCHER_APP_ROUTER_PLUGIN = 'appLoadingManagement.routerPlugin'
 
 /**
  * Listen to the store and wait for a certain set of conditions to be fulfilled. It then triggers
@@ -18,6 +21,21 @@ import log from '@/utils/logging'
 const appLoadingManagementRouterPlugin = (router, store) => {
     const unRegisterRouterHook = router.beforeEach((to) => {
         if (to.meta.requiresAppReady && !store.state.app.isReady) {
+            // Upon application startup we need to first get the language and
+            // topic from the URL in order to quickly load the layers config and
+            // topics. We do this as early as possible as we need topics and config to define
+            // the default application state.
+            const queryParams = getUrlQuery()
+            const lang = queryParams.get('lang') ?? store.state.i18n.lang
+            const topic = queryParams.get('topic') ?? store.state.topics.current
+            store.dispatch('changeTopic', {
+                value: topic,
+                dispatcher: STORE_DISPATCHER_APP_ROUTER_PLUGIN,
+            })
+            store.dispatch('setLang', {
+                value: lang,
+                dispatcher: STORE_DISPATCHER_APP_ROUTER_PLUGIN,
+            })
             log.debug(`App is not ready redirect to /#/startup?redirect=${to.fullPath}`)
             return { name: 'LoadingView', query: { redirect: to.fullPath }, replace: true }
         }
--- a/src/router/appLoadingManagement.routerPlugin.js
+++ b/src/router/appLoadingManagement.routerPlugin.js
@@ -29,11 +29,11 @@ const appLoadingManagementRouterPlugin = (router, store) => {
             const lang = queryParams.get('lang') ?? store.state.i18n.lang
             const topic = queryParams.get('topic') ?? store.state.topics.current
             store.dispatch('changeTopic', {
-                value: topic,
+                topicId: topic,
                 dispatcher: STORE_DISPATCHER_APP_ROUTER_PLUGIN,
             })
             store.dispatch('setLang', {
-                value: lang,
+                lang: lang,
                 dispatcher: STORE_DISPATCHER_APP_ROUTER_PLUGIN,
             })
             log.debug(`App is not ready redirect to /#/startup?redirect=${to.fullPath}`)
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -211,29 +211,25 @@ const actions = {
      */
     setLayerConfig({ commit, state, getters }, { config, dispatcher }) {
         const activeLayerBeforeConfigChange = [...state.activeLayers]
-        commit('clearLayers', { dispatcher })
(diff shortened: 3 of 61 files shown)


src/store/modules/topics.store.js
```diff
@@ -15,7 +13,7 @@ const state = {
      *
      * @type {Topic}
```
pakb, line 14: the type has changed here, isn't it?

src/store/modules/topics.store.js
```diff
@@ -32,56 +30,53 @@ const state = {
...
+    setTopicTree: ({ commit }, { layers, dispatcher }) => {
+        commit('setTopicTree', { layers: layers.map((layer) => layer.clone()), dispatcher })
     },
-    setTopicById: ({ commit, state }, topicId) => {
-        const topic = state.config.find((topic) => topic.id === topicId)
-        if (topic) {
-            commit(CHANGE_TOPIC_MUTATION, topic)
+    changeTopic: ({ commit, state }, { value, dispatcher }) => {
+        if (
+            state.config.find((topic) => topic.id === value) ||
```
pakb, line 53: to have a boolean response, you can instead use `.some(filter)`

src/store/modules/topics.store.js
```diff
@@ -32,56 +30,53 @@ const state = {
...
-            commit(
-                'setTopicTreeOpenedThemesIds',
-                themes.indexOf(',') !== -1 ? themes.split(',') : [themes]
-            )
-        } else if (Array.isArray(themes)) {
-            commit('setTopicTreeOpenedThemesIds', themes)
+    setTopicTreeOpenedThemesIds: ({ commit }, { value, dispatcher }) => {
+        if (typeof value === 'string') {
+            commit('setTopicTreeOpenedThemesIds', {
+                themes: value.indexOf(',') !== -1 ? value.split(',') : [value],
```
pakb, line 66: `.split` always returns an array (even if no match is found) so we could simplify this by doing
`themes: values.split(',')` (it's already 100% sure this is a string with the if above)

src/store/plugins/topic-change-management.plugin.js
```diff
@@ -24,47 +25,86 @@ let isFirstSetTopic = true
...
-            // otherwise, the URL param bgLayer is ignored/overwritten by the setTopic
-            // we set it anyway if the URL doesn't contain a bgLayer URL param
-            if (!isFirstSetTopic || window.location.href.indexOf('bgLayer=') === -1) {
+        if (
+            // During application startup we trigger a changeTopic before the topics are loaded
+            // therefore in this case we ignore the changeTopic event
+            (mutation.type === 'changeTopic' && store.state.topics.config.length > 0) ||
+            mutation.type === 'setTopics'
+        ) {
+            let startTime = performance.now()
```
pakb, line 36: I know this is quite helpful to debug and improve performance, but shouldn't we check on which ENV we are to start "profiling" this mutation?
We don't really want this kinds of logs to permeate to a production environment

I don't know if it wouldn't be a better solution to implement some kind of "start-stop" logging in the log utils directly, and tie the creation of the `performance.now()` to the env we are currently on.

src/utils/logging.js
```diff
@@ -24,8 +24,11 @@ export const LogLevel = {
  * @param {...any} message
  */
 const log = (level, ...message) => {
-    // In production we only log errors.
-    if (ENVIRONMENT === 'production' && level !== LogLevel.ERROR) {
+    // In production we don't log debug level
+    if (
+        ENVIRONMENT === 'production' &&
+        [LogLevel.ERROR, LogLevel.WARNING, LogLevel.INFO].includes(level)
+    ) {
```
pakb, line 31: I don't know if we want INFO level either...

pakb (commented):
As a general comment on this new way of key used for the payload, I think we should decide one of two possibilities :
- Have some kind of naming convention for the value's key, for example if the action is called `setSomething` then the payload should be `{ something: ..., dispatcher: ... }`
- Only use `value` as key, with the exception of places where it helps to now what kind of payload we are dealing with (such as `addLayer`)

But a mix of both might be confusing in the long run.

src/store/plugins/topic-change-management.plugin.js
ltshb: I wanted to remove this I forgot it 😉

src/modules/map/components/footer/backgroundSelector/BackgroundSelector.vue
```diff
@@ -44,7 +44,10 @@ const sortedBackgroundLayersWithVoid = computed(() =>
 )
 
 function selectBackground(backgroundLayer) {
-    store.dispatch('setBackground', backgroundLayer)
+    store.dispatch('setBackground', {
+        value: backgroundLayer,
```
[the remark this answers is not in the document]
ltshb: OK I replaced mostly by the name of the store parameter and I'll update the docu in #645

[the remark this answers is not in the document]
ltshb: done see https://github.com/geoadmin/web-mapviewer/pull/644/commits/75e1dbfd4d867c9e11d05889eaaaa00d359e84a1

src/router/appLoadingManagement.routerPlugin.js
```diff
@@ -18,6 +21,21 @@ import log from '@/utils/logging'
...
+            // topic from the URL in order to quickly load the layers config and
+            // topics. We do this as early as possible as we need topics and config to define
+            // the default application state.
+            const queryParams = getUrlQuery()
+            const lang = queryParams.get('lang') ?? store.state.i18n.lang
+            const topic = queryParams.get('topic') ?? store.state.topics.current
+            store.dispatch('changeTopic', {
+                value: topic,
+                dispatcher: STORE_DISPATCHER_APP_ROUTER_PLUGIN,
+            })
```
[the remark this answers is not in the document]
ltshb: @pakb But then the store would need to have knowledge on the URL parameter which I would avoid. Moreover removing this dispatch has no real impact on perf, I would prefer leave it here in the router which handles the url parameter

src/store/modules/layers.store.js
```diff
@@ -189,17 +188,18 @@ const actions = {
...
      */
-    setBackground({ commit, getters }, layerIdOrObject) {
+    setBackground({ commit, getters }, { value, dispatcher }) {
+        const layerIdOrObject = value
         let futureBackground
         if (typeof layerIdOrObject === 'string') {
             futureBackground = getters.getLayerConfigById(layerIdOrObject)
         } else if (layerIdOrObject instanceof AbstractLayer) {
             futureBackground = getters.getLayerConfigById(layerIdOrObject.getID())
         }
```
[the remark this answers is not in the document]
ltshb: @pakb I would prefer leaving it as is, this PR has already quite a lots of changes.

src/store/modules/position.store.js
```diff
@@ -192,27 +192,27 @@ const actions = {
      * @param commit
      * @param center The new center, either an array of two numbers, or an object with {x, y}
      *   properties
-     * @param source Source of this change, for debug purposes (won't be stored, will be in output
-     *   of the debug console)
+     * @param dispatcher Source of this change, for debug purposes (won't be stored, will be in
+     *   output of the debug console)
      */
-    setCenter: ({ commit }, { center, source = 'unknown' }) => {
+    setCenter: ({ commit }, { center, dispatcher = 'unknown' }) => {
```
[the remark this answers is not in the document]
ltshb: done

src/store/modules/topics.store.js
ltshb: done

src/utils/logging.js
ltshb: Done

pakb (approved):

PR #649, opened 2024-02-21
Fix race condition when exporting/saving kml

ltshb:
To export a kml, openlayer requires the icon size in order to compute the
scale. If the icon has not yet been loaded when generating the style then
the icon scale would be computed with a default size of 64px.

The last commit tried to solve this issue in cypress test, but unfortunately
waiting on the icon to be loaded did not always solved the race condition.

I could not find a better way to ensure this as settings the size when
generating the icon style.

This solve temporarly the issue, but in the future we should get the size
from the backend with the others icon metadata.

The icon size test did failed time to time. This was due to a race condition.
    
The icon scale in KML is set by openlayer and uses a scale factor based on the
icon size. Now if the icon has not yet be loaded during the KML write, openlayer
takes a default size of 64 pixel which defer from our icon size of 48 pixel.
Due to this the icon scale of a large icon was 2 instead of 1.5.

We solve this race condition by waiting on all icons being loaded before doing
the icon size changes.

    NOTE this is not an issue in real life as the save drawing is always delayed
    at least 2 seconds which let plenty of time to load the icons. In cypress this
    delay was reduced to speed up testing.

Also did small test improvements:

- Remove the useless wait on icon, changing icon size don't trigger any new
icon request anymore as all icon are loading from the backend using the same
size.

- Remove a wrong closing action, at this time of the test the icon size selector
is already closed.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-drawing-e2e-ci/index.html)

### DIFF
--- a/adr/2024_01_31_drawing_icon_size.md
+++ b/adr/2024_01_31_drawing_icon_size.md
@@ -91,8 +91,8 @@ With table and following the openlayer scale formulas our Icon size should be
 | --------- | ------------- | ----------- | -------------- |
 | small     | 0.5           | 0.75        | 24             |
 | medium    | 0.75          | 1.125       | 36             |
-| big       | 1             | 1.5         | 48             |
-| bigger    | 1.25          | 1.875       | 60             |
+| large     | 1             | 1.5         | 48             |
+| xlarge    | 1.25          | 1.875       | 60             |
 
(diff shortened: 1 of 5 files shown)


pakb (approved):
👍

PR #652, opened 2024-02-21
PB-293: Corrected external layer order at startup and improve a bit performance

ltshb:
The external layers were always put at the top during startup even if they
were in the middle of the layers parameter.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-293-ext-layer-order/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -4,12 +4,34 @@ import LayerTypes from '@/api/layers/LayerTypes.enum'
 import { getExtentForProjection } from '@/utils/extentUtils.js'
 import { getGpxExtent } from '@/utils/gpxUtils.js'
 import { getKmlExtent, parseKmlName } from '@/utils/kmlUtils'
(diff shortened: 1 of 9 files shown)


src/store/modules/layers.store.js
```diff
@@ -4,13 +4,35 @@ import LayerTypes from '@/api/layers/LayerTypes.enum'
...
 
+const cloneActiveLayerConfig = (getters, layer) => {
+    const clone = getters.getLayerConfigById(layer.id)?.clone() ?? null
+    if (clone) {
+        if (layer.visible != undefined) {
+            clone.visible = layer.visible
+        }
+        if (layer.opacity != undefined) {
+            clone.opacity = layer.opacity
+        }
```
pakb, line 23: `!=` => `!==` 😉

pakb (approved):

src/store/modules/layers.store.js
ltshb: No I want `!=` in order to also catch `!== null` 😉

PR #655, opened 2024-02-22
PB-259: Router vue performance

ltshb:


ltshb:
The test failing is solved in next PR #660

pakb (approved):

PR #656, opened 2024-02-22
PB-244: coordinate display drawing marker and text element

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-244-coordinate-display-drawing-market-and-text-element/index.html)


src/modules/infobox/components/styling/DrawingStylePositionCopy.vue
```diff
@@ -0,0 +1,72 @@
...
+                )
+                this.copied = true
+                // leaving the "Copied" text for the wanted delay, and then reverting to "Copy"
+                setTimeout(() => {
+                    this.copied = false
+                }, 1000)
+            } catch (error) {
+                log.error(`Failed to copy to clipboard:`, error)
+            }
+        },
```
pakb, line 69: that looks a lot like what you did for the `LocationPopup.vue` component, maybe it would make sense to try to isolate this logic into a single, non drawing/location popup specific component, and use this new component in both places instead of copying the logic here

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -49,6 +51,18 @@ function sanitizeHtml(htmlText) {
                 {{ i18n.t('no_more_information') }}
             </div>
         </div>
+        <div
+            v-if="feature.geometry.type == 'Point'"
```
pakb, line 55: shouldn't be too much of an issue here, but you typically should use a triple equals in JS, so that it also type-checks (and not only try to guess-work the values and compare them in all way possible)

pakb (commented):
Little bit of linting and you should be good to go (you might want to add Prettier on a save action on your IDE, or run a `npm run lint` before committing if you don't want that)

src/modules/map/components/openlayers/OpenLayersMouseTracker.vue
```diff
@@ -40,6 +40,9 @@ function showCoordinateLabel(displayedFormat) {
     return displayedFormat?.id === LV95Format.id || displayedFormat?.id === LV03Format.id
 }
 function setDisplayedFormatWithId() {
+    store.dispatch('setDisplayedFormatId', {
```
LukasJoss, line 43: @ltshb
I have decided to use the mouse tracker selection instead of the projection button to access different coordinate formats(though lat-lon might need some visual improvements), is that good like that?

src/store/modules/position.store.js
```diff
@@ -50,6 +50,13 @@ export function normalizeAngle(rotation) {
  */
 
 const state = {
+    /**
+     * The display format selected for the mousetracker
+     *
+     * @type String
+     */
+    displayedFormatId: 'LV95',
```
ltshb, line 58: For the default use `LV95Format.id` from `coordinateFormat.js`

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
 <template>
     <div id="nav-local" class="tab-pane fade" role="tabpanel" aria-labelledby="nav-local-tab">
         <div class="pb-2 location-popup-coordinates">
-            <LocationPopupCopySlot
-                identifier="lv95"
-                :value="LV95Format.format(coordinate, projection)"
+            <CoordinateCopySlot
+                identifier="location-popup-lv95"
+                :value="coordinate"
+                coordinate-format="LV95"
```
ltshb, line 114: use `LV95Format.id` from `coordinateFormat.js` instead of `'LV95'`

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot
-                identifier="lv03"
-                :value="LV03Format.format(coordinate, projection)"
+            <CoordinateCopySlot
+                identifier="location-popup-lv03"
+                :value="coordinate"
+                coordinate-format="LV03"
```
ltshb, line 124: Same as for LV95 use the format definition

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
                 </a>
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot
-                identifier="wgs84"
+            <CoordinateCopySlot
+                identifier="location-popup-wgs84"
                 :value="coordinateWGS84Plain"
+                coordinate-format="None"
```
ltshb, line 134: `=null` would be better

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot
-                identifier="utm"
-                :value="UTMFormat.format(coordinate, projection)"
+            <CoordinateCopySlot
+                identifier="location-popup-utm"
+                :value="coordinate"
+                coordinate-format="UTM"
```
ltshb, line 143: Same as for LV95 use the format definition

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot
-                identifier="mgrs"
-                :value="MGRSFormat.format(coordinate, projection)"
+            <CoordinateCopySlot
+                identifier="location-popup-mgrs"
+                :value="coordinate"
+                coordinate-format="MGRS"
```
ltshb, line 151: Same as for LV95 use the format definition

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
                 MGRS
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot v-if="what3Words" identifier="w3w" :value="what3Words">
+            <CoordinateCopySlot
+                v-if="what3Words"
+                identifier="location-popup-w3w"
+                :value="what3Words"
+                coordinate-format="None"
```
ltshb, line 160: `=null`

src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -109,58 +108,69 @@ async function updateHeight() {
...
-            </LocationPopupCopySlot>
+            </CoordinateCopySlot>
 
-            <LocationPopupCopySlot
+            <CoordinateCopySlot
                 v-if="height"
-                identifier="height"
+                identifier="location-popup-height"
                 :value="heightInMeter"
+                coordinate-format="None"
```
ltshb, line 169: `=null`

src/modules/map/components/openlayers/OpenLayersMouseTracker.vue
ltshb: Good idea 👍🏼

src/utils/components/CoordinateCopySlot.vue
```diff
@@ -23,15 +24,21 @@ const props = defineProps({
         type: Number,
         default: 1000,
     },
+    coordinateFormat: {
```
ltshb, line 27: Here either use the the format object as input or the id. I would use the object, which means you should have

```
 coordinateFormat: {
        type: CoordinateFormat,
        default: null,
    },
```

src/utils/components/CoordinateCopySlot.vue
```diff
@@ -70,9 +77,22 @@ function setTooltipContent() {
...
+    if (coordinateFormat.value === 'None') {
+        return coordinates
+    }
+    if (coordinateFormat.value) {
+        displayedFormat = allFormats.find((format) => format.id == coordinateFormat.value)
+    } else {
+        displayedFormat = allFormats.find((format) => format.id === displayedFormatId.value)
+    }
+    return displayedFormat.format(coordinates, projection.value, true)
+}
```
ltshb, line 92: Based on having `coordinateFormat` being of type `CoordinateFormat` you can simplify this function

```javascript
function display(coordinates) {
    if (coordinateFormat.value) {
        return coordinateFormat.value.format(coordinates, projection.value, true)
    }
    return coordinates
}
```

The format should be given in the FeatureStyleEdit.vue and FeatureDetail.vue

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -49,6 +50,18 @@ function sanitizeHtml(htmlText) {
...
         </div>
+        <div
+            v-if="feature.geometry.type === 'Point'"
+            class="d-flex pb-2 px-2 gap-1 justify-content-start align-items-center"
+        >
+            <CoordinateCopySlot
+                class="d-flex"
+                identifier="feature-detail-coordinate-copy"
+                :value="feature.geometry.coordinates.slice(0, 2)"
+            >
```
ltshb, line 61: Add the `coordinateFormat` from the store value

```javascript
const coordinateFormat = computed(() => {
    return allFormats.find((format) => format.id === store.state.position.displayedFormatId) ?? null
})

...

<CoordinateCopySlot
    class="d-flex"
    identifier="feature-detail-coordinate-copy"
    :value="feature.geometry.coordinates.slice(0, 2)"
    :coordinate-format="coordinateFormat"
>

```

src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -210,7 +211,19 @@ function onDelete() {
...
-        <div class="d-flex">
+        <div
+            v-if="isFeatureMarker || isFeatureText"
+            class="d-flex small gap-1 justify-content-start align-items-center"
+        >
+            <CoordinateCopySlot
+                class="d-flex"
+                identifier="feature-style-edit-coordinate-copy"
+                :value="feature.coordinates[0].slice(0, 2)"
+            >
```
ltshb, line 222: Same as for `FeatureDetail.vue`

src/utils/components/CoordinateCopySlot.vue
```diff
@@ -90,20 +110,21 @@ async function copyValue() {
...
-            <div v-if="extraValue" :data-cy="`location-popup-extra-value-${identifier}`">
+            <div v-if="extraValue" :data-cy="`${identifier}-extra-value`">
                 {{ extraValue }}
             </div>
         </div>
         <button
             ref="copyButton"
-            class="location-popup-copy-button btn btn-light text-black-50 d-none d-md-block"
+            :data-cy="`${identifier}-button`"
+            class="location-popup-copy-button btn btn-light text-black-50"
```
ltshb, line 123: a margin or padding between the coordinate and the button should be added for the FeatureDetail.vue and FeatureStyleEdit.vue case

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -50,7 +50,14 @@ describe('Drawing module tests', () => {
...
         }
-
+        function readCoordinateClipboard(name) {
+            cy.get(`[data-cy="${name}-button"]`).focus()
+            cy.get(`[data-cy="${name}-button"]`).realClick()
+            cy.get(`[data-cy="${name}-icon"]`).should('have.class', 'fa-check')
+            cy.readClipboardValue().then((clipboardText) => {
+                expect(clipboardText).to.be.equal("2'660'013.50, 1'185'172.00")
+            })
+        }
```
ltshb, line 60: In cypress I'm not a big fan of such function helper, if for example the second call of the function is failing it is hard (almost impossible) to know form cypress logs which call failed. So either don't use a helper, or add a `cy.log()` in the helper using the input argument to ease the traking of the function call.

ltshb: I would also provide the coordinate as argument, see other comment.

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -284,6 +302,17 @@ describe('Drawing module tests', () => {
...
+            cy.get('[data-cy="drawing-style-delete-button"]').click({ force: true })
+            cy.clickDrawingTool(EditableFeatureTypes.ANNOTATION)
+            cy.get('[data-cy="ol-map"]:visible').click()
+            readCoordinateClipboard('feature-style-edit-coordinate-copy')
+            cy.log('Coordinates for marker can be copied in drawing mode')
+            cy.get('[data-cy="drawing-toolbox-close-button"]').click()
+            cy.get('[data-cy="menu-button"]').click()
+            cy.get('[data-cy="ol-map"]').click('center')
+            readCoordinateClipboard('feature-detail-coordinate-copy')
+            cy.log('Coordinates for marker can be copied while not in drawing mode')
```
ltshb, line 315: This log should be place at the line 311

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -284,6 +302,17 @@ describe('Drawing module tests', () => {
                     new RegExp(`<LabelStyle><color>${KML_STYLE_BLACK}</color></LabelStyle>`),
                 ])
             })
+
+            cy.get('[data-cy="drawing-style-delete-button"]').click({ force: true })
+            cy.clickDrawingTool(EditableFeatureTypes.ANNOTATION)
+            cy.get('[data-cy="ol-map"]:visible').click()
+            readCoordinateClipboard('feature-style-edit-coordinate-copy')
+            cy.log('Coordinates for marker can be copied in drawing mode')
```
ltshb, line 310: The log should be place before the test take place otherwise if the test fails we don't see it

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -284,6 +302,17 @@ describe('Drawing module tests', () => {
                     new RegExp(`<LabelStyle><color>${KML_STYLE_BLACK}</color></LabelStyle>`),
                 ])
             })
+
+            cy.get('[data-cy="drawing-style-delete-button"]').click({ force: true })
+            cy.clickDrawingTool(EditableFeatureTypes.ANNOTATION)
+            cy.get('[data-cy="ol-map"]:visible').click()
+            readCoordinateClipboard('feature-style-edit-coordinate-copy')
```
ltshb, line 309: Here instead of deleted the previous feature and re-add a new one, I would leave the previous one, add a new at a different coordinate and test the copy of both feature making sure that the clipboard contains each time the correct coordinate. So we make sure the a the clipboard gets updated on a subsequent click (used to be a bug on the location popup)

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -230,6 +237,17 @@ describe('Drawing module tests', () => {
...
+            cy.get('[data-cy="drawing-style-delete-button"]').click({ force: true })
+            cy.clickDrawingTool(EditableFeatureTypes.MARKER)
+            cy.get('[data-cy="ol-map"]:visible').click()
+            readCoordinateClipboard('feature-style-edit-coordinate-copy')
+            cy.log('Coordinates for marker can be copied in drawing mode')
+            cy.get('[data-cy="drawing-toolbox-close-button"]').click()
+            cy.get('[data-cy="menu-button"]').click()
+            cy.get('[data-cy="ol-map"]').click('center')
+            readCoordinateClipboard('feature-detail-coordinate-copy')
+            cy.log('Coordinates for marker can be copied while not in drawing mode')
```
ltshb, line 250: See comment about annotation test

ltshb (commented):
Good some adjustements

ltshb (approved):
All good thanks.

I've just noticed an issue, you cannot select and copy the text, it seems that it comes from the `.no-ios-long-press-select` class in https://github.com/geoadmin/web-mapviewer/blob/develop/src/scss/main.scss, @pakb do you know why we want to prevent the text selection ?

pakb:
> @pakb do you know why we want to prevent the text selection ?

because iOs doesn't listen correctly to the meta tag in the header telling mobile device that test shouldn't be selectable. So I added a bit of CSS to cover that...

PR #657, opened 2024-02-23
PB-282: Location share Whats App

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-282-location_share_whats_app/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/shareShortLink.cy.js
+++ b/tests/cypress/tests-e2e/shareShortLink.cy.js
@@ -140,6 +140,12 @@ describe('Testing the share menu', () => {
                 `https://twitter.com/intent/tweet?text=&url=${encodeURIComponent(dummyShortLink)}`
             )
         })
+        it('opens a whatsapp popup with the short link as content', () => {
+            checkSocialNetworkType(
+                'whatsapp',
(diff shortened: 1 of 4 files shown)


tests/cypress/tests-e2e/shareShortLink.cy.js
```diff
@@ -140,6 +140,12 @@ describe('Testing the share menu', () => {
                 `https://twitter.com/intent/tweet?text=&url=${encodeURIComponent(dummyShortLink)}`
             )
         })
+        it('opens a whatsapp popup with the short link as content', () => {
+            checkSocialNetworkType(
+                'whatsapp',
+                `https://api.whatsapp.com/send?text=%0D%0A${encodeURIComponent(dummyShortLink)}`
+            )
+        })
```
ltshb, line 148: Cypress `it()` are quite costly in test performance and cost (if we would have a non open source software it would cost), so now we tend to do less `it()`, I would suggest instead of adding a new `it()` (test case) to merge all in `it()` of the social network context into one `it()`. Each part of the test should be separate by a `cy.log()` statement to ease debugging (e.g. `cy.log('opens a whatsapp popup ....')`

ltshb (approved):
Nice thanks, just a minor improvement in cypress test to do.

PR #660, opened 2024-02-26
Fix legacy parameter layer default opacity

ltshb:
If a layer has a matching default opacity in the layer url, in the new url
we should remove the opacity. This was not the case.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-259-legacy/index.html)


ltkum (approved):
Looks good to me, and the cypress test looks cleaner too :)

PR #661, opened 2024-02-26
PB-259: Improve startup performance by postponing the topic tree rendering

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-259-performance-menu/index.html)

PR #663, opened 2024-02-27
PB-259: Improve startup performance by rendering menu section only when the map is rendered

ltshb:
The rendering of catalogue is very costly and store events would re-trigger it
therefore wait that the map has been rendered before rendering the catalogue,
it is anyway not visible at startup until the user click on the menu entry.
It still needs to be always rendered once the application has been started, otherwise
the toggling of the menu would be too slow.

Also differ the rendering and opening of the active layers for the same reason
and also because if the menu contains external layer (e.g. kml, gpx, ...) the
application needs to load its metadata/data before rendering the correct name
in the list. For kml and GPX it would use at startup the default KML or GPX
name until the name has been loading. By defering the active layers list
we avoid having the name changing.

Several other options has been tried but they did not improve the performances:

- using shallow ref on the layer catalogue, but this had no impact. Shallow ref is not easy to use on store content, the store content needs to be anyway added as computed value in order to use shallow ref. This might be better and more efficient to use with pinia store.
- using new events instead of the mapIsReady, see #661 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-259-performance-menu-off/index.html)

### DIFF
--- a/tests/cypress/support/commands.js
+++ b/tests/cypress/support/commands.js
@@ -253,6 +253,14 @@ Cypress.Commands.add(
 )
 
 /**
+ * Wait until the map has been rendered and is ready. This is useful and needed during the
+ * application startup phase
+ */
+Cypress.Commands.add('waitMapIsReady', ({ timeout = 15000 } = {}) => {
+    cy.waitUntilState((state) => state.app.isMapReady, { timeout: timeout })
+})
+
+/**
  * Changes a URL parameter without reloading the app.
  *
  * Help when you want to change a value in the URL but don't want the whole app be reloaded from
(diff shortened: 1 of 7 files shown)


tests/cypress/support/commands.js
```diff
@@ -252,6 +252,14 @@ Cypress.Commands.add(
     }
 )
 
+/**
+ * Wait until the map has been rendered and is ready. This is useful and needed during the
+ * application startup phase
+ */
+Cypress.Commands.add('waitMapIsReady', ({ timeout = 15000 } = {}) => {
+    cy.waitUntilState((state) => state.app.isMapReady, { timeout: timeout })
+})
```
pakb, line 261: It might make sense to use this in the ubiquitous `.goToMapView` command, don't you think?

pakb (approved):

tests/cypress/support/commands.js
ltshb: It is already Part of it

PR #665, opened 2024-02-27
PB-277: The mouse wheel zoom constraint had some issues with different

ltshb:
devices/os

On some device and os the mouse zoom wheel was unusable (one scroll step would result to full zoom in or out). A user reported this per email and I had the issue on my private windows machine with its touchpad and with 2 different mouses. Strangely the same mouse were working fine with my HP Zbook on linux.

See also openlayer issue https://github.com/openlayers/openlayers/issues/15423


pakb:
I don't know if it is something we should tackle in the scope of this change :

when we zoom through a map gesture (be it mouse wheel on desktop or pinch gesture on mobile) I would not trigger the zoom rounding computation.
Currently you can feel like a last small step applied clearly after the mouse wheel event has been handled, and I find it a bit annoying.
I'd keep the "round to nearest integer zoom level" approach with zoom buttons though

pakb (approved):

PR #672, opened 2024-02-29
PB-207 : transform layer constructor to have one single arg

pakb:
and deconstruct its content when called, should shorten most calls to constructors that were using many default/blank/null values to reach one last argument and set it to something else

Adding the updateDelay param to GeoJson layer (it will be relevant only relevant for this type of layers)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-207_preparation_work/index.html)


src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -35,32 +35,32 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
      * @param {boolean} isLoading Set to true if some parts of the layer (e.g. metadata) are still
      *   loading
      */
-    constructor(
-        name,
-        opacity,
-        visible,
-        baseUrl,
-        layerId,
+    constructor({
```
ltshb, line 38: Name and base url should not have default values that too risky they are Mandatory parameter

src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -35,32 +35,32 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
...
+        name = null,
+        opacity = 1.0,
+        visible = true,
+        baseUrl = null,
```
ltshb, line 42: Dito mandatory param actually name could be made optional as it is set later on

src/api/layers/ExternalLayer.class.js
```diff
@@ -8,7 +8,10 @@ import { LayerAttribution } from '@/api/layers/AbstractLayer.class'
  * @returns {LayerAttribution[]} Default list of layer attributions
  */
 export function getDefaultAttribution(baseUrl) {
-    return [new LayerAttribution(new URL(baseUrl).hostname)]
+    if (baseUrl) {
```
ltshb, line 11: I think base url should mandatory otherwise we might have subtle bugs

src/api/layers/ExternalLayer.class.js
```diff
@@ -69,34 +72,34 @@ export default class ExternalLayer extends AbstractLayer {
...
-    constructor(
-        name,
-        layerType,
-        externalLayerId,
-        baseURL,
-        opacity,
-        visible,
+    constructor({
+        name = null,
+        layerType = null,
```
ltshb, line 77: Dito layertype layer I’d and base url should be mandatory

src/api/layers/ExternalGroupOfLayers.class.js
pakb: I thought I should maybe validate the inputs and raise exception if what was given isn't valid. This way I can keep the "optional" approach, but check that we have sufficient information to build the instance in the constructor.
What is your take on that? Good or bad idea?

pakb:
Somehow new commits weren't detected and applied to the PR anymore, so I've re-opened it here : https://github.com/geoadmin/web-mapviewer/pull/677

PR #674, opened 2024-03-01
PB-190 : Implement feature pre selection

ltkum:
We added an additional possible custom parameter which allows to select features which IDS are known from the URL, like the old {bod-layer-id} parameter.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-190_implement-preselectedfeature-feature/index.html)

### DIFF
--- a/src/router/storeSync/LayerParamConfig.class.js
+++ b/src/router/storeSync/LayerParamConfig.class.js
@@ -1,4 +1,4 @@
-import getFeature, { getExtentOfFeatures } from '@/api/features/features.api'
+import getFeature from '@/api/features/features.api'
 import ExternalWMSLayer from '@/api/layers/ExternalWMSLayer.class'
 import ExternalWMTSLayer from '@/api/layers/ExternalWMTSLayer.class'
 import GPXLayer from '@/api/layers/GPXLayer.class.js'
@@ -9,6 +9,7 @@ import AbstractParamConfig, {
     STORE_DISPATCHER_ROUTER_PLUGIN,
 } from '@/router/storeSync/abstractParamConfig.class'
 import { parseLayersParam, transformLayerIntoUrlString } from '@/router/storeSync/layersParamParser'
+import { getExtentOfGeometries } from '@/utils/geoJsonUtils'
 import log from '@/utils/logging'
 import { getUrlQuery } from '@/utils/utils'
 
@@ -132,31 +133,30 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
         store.dispatch('setLayers', { layers: layers, dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN })
     )
     if (featuresRequests.length > 0) {
-        getAndDispatchFeatures(featuresRequests, store) // all features after promise
+        promisesForAllDispatch.push(getAndDispatchFeatures(featuresRequests, store)) // all features after promise
     }
 
     return Promise.all(promisesForAllDispatch)
 }
 
 async function getAndDispatchFeatures(featuresPromise, store) {
-    const features = []
     try {
-        const featuresReceived = await Promise.all(featuresPromise)
-        if (featuresReceived.length > 0) {
-            featuresReceived.forEach((feature) => {
-                features.push(feature)
-            })
-
-            store.dispatch('setSelectedFeatures', {
+        const responses = await Promise.allSettled(featuresPromise)
+        const features = responses
+            .filter((response) => response.status === 'fulfilled')
+            .map((response) => response.value)
+        if (features.length > 0) {
+            await store.dispatch('setSelectedFeatures', {
                 features: features,
                 dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
             })
-            const extent = getExtentOfFeatures(features)
+
+            const extent = getExtentOfGeometries(features.map((feature) => feature.geometry))
             // If the zoom level has been specifically set to a level, we don't want to override that.
             // otherwise, we go to the zoom level which encompass all features
             const query = getUrlQuery()
             if (!query.z) {
-                store.dispatch('zoomToExtent', {
+                await store.dispatch('zoomToExtent', {
                     extent: extent,
                     maxZoom: 8,
                     dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
@@ -166,14 +166,14 @@ async function getAndDispatchFeatures(featuresPromise, store) {
                     [(extent[0][0] + extent[1][0]) / 2],
                     [(extent[0][1] + extent[1][1]) / 2],
                 ]
-                store.dispatch('setCenter', {
+                await store.dispatch('setCenter', {
                     center: center,
                     dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
                 })
             }
         }
     } catch (error) {
-        log.error(`Error while processing features in bod-layer-id router. error is ${error}`)
+        log.error(`Error while processing features in feature preselection. error is ${error}`)
     }
 }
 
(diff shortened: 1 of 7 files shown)


pakb:
The parsing is fine and works well, but I think you should add a zoomToExtent to the feature's extent (or at least zoom closer if there's no extent available). The goal is to show the feature "in isolation" while using an iframe

tests/cypress/tests-e2e/featurePreSelection.cy.js
```diff
@@ -0,0 +1,52 @@
...
+                    }
+
+                    const featuresIds = []
+
+                    for (let i = 0; i < nbFeatures; i++) {
+                        featuresIds.push(features[i].id.toString())
+                    }
+                    cy.goToMapView({ layers: `${layer}@features=${featuresIds.join(':')}` })
+                    checkFeatures(featuresIds)
+                })
```
pakb, line 48: I think this could be greatly simplified by using a bit of `filter` `map` `reduce` (`forEach`) approach, pseudo code bellow

```
cy.fixture('features.fixture.json').then((jsonResult) => {
    const features = [...jsonResult.results]
        // we intercept every feature we want to retrieve
        cy.intercept(`**/MapServer/${feature.layerBodId}/${feature.id}`, {
            results: [feature],
        })
    })
    const featureIds = features.map((feature) => feature.id)
    cy.goToMapView({ layers: `${layer}@features=${featuresIds.join(':')}` })
    checkFeatures(featuresIds)
})
```

tests/cypress/tests-e2e/featurePreSelection.cy.js
```diff
@@ -0,0 +1,52 @@
+/// <reference types="cypress" />
+
+describe('Testing the feature selection in the URL', () => {
+    function checkFeatures(featuresIds) {
+        cy.readStoreValue('state.features.selectedFeatures').then((features) => {
+            cy.wrap(features.length).should('be.equal', featuresIds.length)
+            features.forEach((feature) => {
+                cy.wrap(featuresIds.includes(feature._id)).should('be.true')
```
pakb, line 8: Do not use private `_id` but `id` instead, will be more robust the day we get rid of the private prop (which will happen for sure)

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+            }
         }
         return layerObject
     })
+
     promisesForAllDispatch.push(
         store.dispatch('setLayers', { layers: layers, dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN })
     )
+    if (featuresRequests.length > 0) {
+        getAndDispatchFeatures(featuresRequests, store) // all features after promise
```
pakb, line 135: I think you should return all dispatch promises you will trigger inside this and add them to `promisesForAllDispatch`, so that the next line `Promise.all` wait on them too. Otherwise we might create some "funny" race condition here

I see that you made this function asyc, so in essence it already returns a Promise (which you could add to the list). The only issue I see with that is that you do not `await` dispatch calls, so either `await` them, or bundle them together and return them here

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+    }
 
     return Promise.all(promisesForAllDispatch)
 }
 
+async function getAndDispatchFeatures(featuresPromise, store) {
+    const features = []
+    try {
+        const featuresReceived = await Promise.all(featuresPromise)
```
pakb, line 144: We should maybe be more resilient to bad inputs here, and have something like `Promise.allSettled` (see [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) ) which should help ignore bad IDs that user could give use, not breaking the whole loading process

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+        if (featuresReceived.length > 0) {
+            featuresReceived.forEach((feature) => {
+                features.push(feature)
+            })
```
pakb, line 148: why use a `features` const when you only use `featuresReceived` as you get it?

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+
+            store.dispatch('setSelectedFeatures', {
```
pakb, line 150: so either return the Promise (among others) of this dispatch, or add `await` so that the global Promise of this async func wait on the result

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+            const extent = getExtentOfFeatures(features)
+            // If the zoom level has been specifically set to a level, we don't want to override that.
+            // otherwise, we go to the zoom level which encompass all features
+            const query = getUrlQuery()
+            if (!query.z) {
+                store.dispatch('zoomToExtent', {
+                    extent: extent,
+                    maxZoom: 8,
+                    dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
+                })
```
pakb, line 163: same

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +109,74 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+            } else {
+                const center = [
+                    [(extent[0][0] + extent[1][0]) / 2],
+                    [(extent[0][1] + extent[1][1]) / 2],
+                ]
+                store.dispatch('setCenter', {
+                    center: center,
+                    dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
+                })
```
pakb, line 172: same

tests/cypress/tests-e2e/featurePreSelection.cy.js
ltkum: `.id` returns a mix between the layer id and the feature id (for example : `my.layer.id-8835` )
`._id` returns only the feature id
right now, I'm using the one which returns what I want :)

src/utils/geoJsonUtils.js
```diff
@@ -77,3 +81,13 @@ export function transformIntoTurfEquivalent(geoJsonData, fromProjection = WGS84)
     log.error('Unknown geometry type', geometryWGS84.type)
     return null
 }
+
+/**
+ * @param {Object[]} geometries An array of all geometries in features.
+ * @returns An extent which covers all geometries given as parameter
```
pakb, line 87: you could specify that the returned type is `[[Number,Number], [Number,Number]]`

sometimes extents are described inline (`[a,b,c,d]`) but here we intentionally normalized it to `[[a,b], [c,d]]`

src/utils/geoJsonUtils.js
```diff
@@ -12,6 +14,8 @@ import CoordinateSystem from '@/utils/coordinates/CoordinateSystem.class'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import log from '@/utils/logging'
 
+import { normalizeExtent } from './coordinates/coordinateUtils'
```
pakb, line 17: Try not using relative path, but `@/utils/coordinates/coordinateUtils` instead

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -110,16 +110,73 @@ function dispatchLayersFromUrlIntoStore(store, urlParamValue) {
...
+            }
         }
         return layerObject
     })
+
     promisesForAllDispatch.push(
         store.dispatch('setLayers', { layers: layers, dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN })
     )
+    if (featuresRequests.length > 0) {
+        promisesForAllDispatch.push(getAndDispatchFeatures(featuresRequests, store)) // all features after promise
```
pakb, line 136: the comment on this line doesn't make much sense to me.
could you rephrase it, or remove it?

pakb (approved):

PR #676, opened 2024-03-01
Fix layer drag and drop blocking transparency slider usage

pakb:
only react to drag and dropping when the title of the layer is handled, and not the settings area anymore

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_fix_layer_drag_drop_block_settings_events/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
@@ -38,6 +38,8 @@ onMounted(() => {
         delayOnTouchOnly: true,
         touchStartThreshold: 3,
         animation: 150,
+        direction: 'vertical',
+        handle: '.menu-layer-item-title',
         onStart: function () {
             aLayerIsDragged.value = true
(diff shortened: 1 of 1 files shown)


ltshb (commented):
Strangely this does not work on mobile (iphone) were we still have the old behavior ?

ltshb (approved):

pakb:
> Strangely this does not work on mobile (iphone) were we still have the old behavior ?

I tested on my own device (iOS 17.3.1, Safari) it worked well, but this PR wasn't necessary. As we are using a delay on mobile/touch, this wasn't an issue there, only on desktop without the delay

ltshb:
@pakb would be good to add an e2e test case for the the drag and drop of the opacity bar ?

pakb:
> @pakb would be good to add an e2e test case for the the drag and drop of the opacity bar ?

Yes, but with all the pain I experienced trying to implement the "simple" drag and drop layer test, I'm not sure if I want to go back there 🙃

PR #679, opened 2024-03-01
upgrade to OL v9.0.0 and update other libs

pakb:
OpenLayers 9.0.0 => deleting our temporary own copy of ol/format/GPX that we amended on the repo itself

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_update_libs/index.html)


pakb:
Cypress forgot to run 99% of our tests, I'll check on Monday what has changed in the config between the two minor versions

pakb:
The issue was that some typescript type check we not valid anymore with the newer OL 9 and our shallow copy of ol-cesium

ltshb (commented):
There is some unit test for WMTS that needs to be uncommented, search for `https://github.com/openlayers/openlayers/issues/15363` in code

ltshb (approved):
Did some manual tests and it looks good

PR #681, opened 2024-03-01
PB-102: Fix legacy external layers with coma in layer tilte

ltshb:
In the legacy viewer the layer title was part of the URL and some layer
could contain a coma in its title which broke the layers parsing as the layers
in URL are separated by coma.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-102-legacy-external-layer/index.html)


ltkum (approved):
Took the time to test it quickly, and it looks like it's working fine. Good work :+1:

PR #682, opened 2024-03-03
PB-190 : implement legacy support for bod layer id parameter

ltkum:
Translate the {bod-layer-id} parameter from the legacy viewer to the new 'selected features' layer parameter.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-190-legacy-param-for-bodlayerid/index.html)

### DIFF
--- a/src/utils/__tests__/legacyLayerParamUtils.spec.js
+++ b/src/utils/__tests__/legacyLayerParamUtils.spec.js
@@ -284,6 +284,33 @@ describe('Test parsing of legacy URL param into new params', () => {
             })
         })
         describe('ensure layers parameter handler for feature preselection works as intended', () => {
+            // This function deals mostly with the special parameters and features id order,
+            // As they are not important, but could break the tests if we made a simple string comparison
+            // layer.id@time=123@features=1,2 is the same end result as layer.id@features=2,1@time=123
+            function compareLayersStrings(layerString1, layerString2) {
+                const [layer1AndParams, visibility1, opacity1] = layerString1.split(',')
+                const [layer2AndParams, visibility2, opacity2] = layerString2.split(',')
+                expect(visibility1).to.eq(visibility2)
+                expect(opacity1).to.eq(opacity2)
+                const layer1Split = layer1AndParams.split('@')
+                const layer2Split = layer2AndParams.split('@')
+                layer1Split.sort()
+                layer2Split.sort()
+                expect(layer1Split.length).to.eq(layer2Split.length)
+                for (let i = 0; i < layer1Split.length; i++) {
+                    if (layer1Split[i].includes('features')) {
+                        expect(layer2Split[i].includes('features')).to.eq(true)
+                        const features_1 = layer1Split[i].split('=')[1].split(':')
+
+                        const features_2 = layer2Split[i].split('=')[1].split(':')
+                        features_1.sort()
+                        features_2.sort()
+                        expect(features_1.join(':')).to.eq(features_2.join(':'))
+                    } else {
+                        expect(layer1Split[i]).to.eq(layer2Split[i])
+                    }
+                }
+            }
             it('checks that all possible layers given as parameter return what we expect', () => {
                 const layersParams = [
                     'layer.id;layer.id2',
@@ -306,7 +333,10 @@ describe('Test parsing of legacy URL param into new params', () => {
                     )
                 })
                 for (let i = 0; i < results.length; i++) {
-                    expect(expectedResults[i]).to.eq(results[i])
+                    const [layer1, layer2] = results[i].split(';')
+                    const [expectedLayer1, expectedLayer2] = expectedResults[i].split(';')
+                    compareLayersStrings(layer1, expectedLayer1)
+                    compareLayersStrings(layer2, expectedLayer2)
                 }
             })
         })
--- a/src/utils/__tests__/legacyLayerParamUtils.spec.js
+++ b/src/utils/__tests__/legacyLayerParamUtils.spec.js
@@ -9,6 +9,7 @@ import GeoAdminWMTSLayer from '@/api/layers/GeoAdminWMTSLayer.class'
 import LayerTimeConfig from '@/api/layers/LayerTimeConfig.class'
 import LayerTimeConfigEntry from '@/api/layers/LayerTimeConfigEntry.class'
 import {
+    createLayersParamForFeaturePreselection,
     getLayersFromLegacyUrlParams,
     isLegacyParams,
     parseOpacity,
@@ -280,6 +281,33 @@ describe('Test parsing of legacy URL param into new params', () => {
                     expect(isLegacyParams('#?test=false')).to.equal(false)
                     expect(isLegacyParams('#/?test=false')).to.equal(false)
                 })
+            })
+        })
+        describe('ensure layers parameter handler for feature preselection works as intended', () => {
+            it('checks that all possible layers given as parameter return what we expect', () => {
+                const layersParams = [
+                    'layer.id;layer.id2',
+                    'layer.id,,0.3;layer.id2',
+                    'layer.id@time=1234;layer.id2',
+                    'layer.id@features=3:4:5;layer.id2',
+                    'layer.id@features=3:4:5@time=1234,f,0.2;layer.id2',
+                ]
+                const expectedResults = [
+                    'layer.id@features=1:2:3;layer.id2',
+                    'layer.id@features=1:2:3,,0.3;layer.id2',
+                    'layer.id@time=1234@features=1:2:3;layer.id2',
+                    'layer.id@features=1:2:3:4:5;layer.id2',
+                    'layer.id@features=1:2:3:4:5@time=1234,f,0.2;layer.id2',
+                ]
+                const results = []
+                layersParams.forEach((params) => {
+                    results.push(
+                        createLayersParamForFeaturePreselection('layer.id', '1,2,3', params)
+                    )
+                })
+                for (let i = 0; i < results.length; i++) {
+                    expect(expectedResults[i]).to.eq(results[i])
+                }
             })
         })
     })
--- a/src/utils/__tests__/legacyLayerParamUtils.spec.js
+++ b/src/utils/__tests__/legacyLayerParamUtils.spec.js
@@ -284,6 +284,14 @@ describe('Test parsing of legacy URL param into new params', () => {
             })
         })
         describe('ensure layers parameter handler for feature preselection works as intended', () => {
+            /**
+             * Small precision regarding the createLayersParamForFeaturePreselection, which is what
+             * we are testing here : This function is only called when : there is a layer-id
+             * parameter set and that layer is already within the query 'layers' parameters. This
+             * means we do not need to test if there is no param_key set, or if there is no layers
+             * set
+             */
+
             // This function deals mostly with the special parameters and features id order,
             // As they are not important, but could break the tests if we made a simple string comparison
             // layer.id@time=123@features=1,2 is the same end result as layer.id@features=2,1@time=123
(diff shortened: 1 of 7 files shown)


tests/cypress/support/commands.js
```diff
@@ -231,6 +231,19 @@ Cypress.Commands.add(
...
+                if (!withHash && active !== target) {
+                    let targetBodIdParameter = 0
+                    state.layers.activeLayers.forEach((layer) => {
+                        targetBodIdParameter += Boolean(
+                            layer.geoAdminID in queryParams &&
+                                !queryParams.layers?.includes(layer.geoAdminID) &&
+                                !getters.currentTopic?.layersToActivate.includes(layer.geoAdminID)
+                        )
+                    })
+                    target = +targetBodIdParameter
```
pakb, line 245: I suspect you wanted to write `target += targetBodIdParameter`? `= +` seems a bit off
You might want to cover this with a test, or expand the test you already wrote to cover this if block thoroughly. I bet this line breaks the logic you wanted to implement

src/utils/legacyLayerParamUtils.js
```diff
@@ -171,3 +171,22 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
         kmlMetaData
     )
 }
+
+export function handleBodLayerIdParam(params, store, newQuery) {
```
pakb, line 175: I think we should rename/rephrase this param, BOD layer ID isn't much self-explanatory.
What about : `handleLayerFeaturePreSelection`

src/utils/legacyLayerParamUtils.js
```diff
@@ -171,3 +171,22 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+    params.forEach((param_value, param_key) => {
+        let layer = null
+        layer = store.state.layers.config.find((layer) => layer.getID() === param_key)
```
pakb, line 178: inline const would work too
`const layer = store.state.layers.config.find((layer) => layer.getID() === param_key)`
and be careful here when rebasing, `getID()` has been replaced by `id` param (no function call anymore)

pakb: As you don't use this `layer` constant after the `if` block, you could even replace this by
`if (store.state.layers.config.some((layer) => layer.id === param_key)) {`
and get away without a constant

pakb: a little comment/reminder on why we get the info this way in a legacy URL might help too (in 6 month, we will have dumped the help, and won't have any trace left of this way of describing a feature pre-selection in our official doc)

src/utils/legacyLayerParamUtils.js
```diff
@@ -171,3 +171,22 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+        if (layer) {
+            const featuresIds = param_value.split(',').join(':')
+            if (newQuery.layers?.includes(param_key)) {
+                //if the layer already exist, we insert the features id parameter
+                const [layerIdWithCustomParams, visible, opacity] = newQuery.layers.split(',')
+                newQuery.layers = `${layerIdWithCustomParams}@features=${featuresIds},${visible},${opacity}`
```
pakb, line 184: So if I understand correctly, this will lead to something like
`https://.../#/map?layers=my.cool.layer@features=1@features=2,1,1` ?

If so, I think it doesn't make sense to have a plural word for the feature ID, a `@feature=...` would be better IMHO

src/utils/legacyLayerParamUtils.js
```diff
@@ -171,3 +171,22 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+            } else if (newQuery.layers) {
+                newQuery.layers = newQuery.layers + `;${param_key}@features=${featuresIds}`
```
pakb, line 186: you can but the `newQuery.layers` in the format string definition too

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +411,55 @@ describe('Test on legacy param import', () => {
...
+                        })
+                    }
+
+                    const featuresIds = []
+
+                    for (let i = 0; i < nbFeatures; i++) {
+                        featuresIds.push(features[i].id.toString())
+                    }
+                    const params = {}
+                    params[layer] = featuresIds.join(',')
```
pakb, line 458: Isn't the current (legacy) syntax doing something like
`https://.../?layer.id=1&layer.id=2&layer.id=3` instead of `https://.../?layer.id=1,2,3` ?

src/utils/legacyLayerParamUtils.js
ltkum: those are three valid points

src/utils/legacyLayerParamUtils.js
ltkum: I reworked this whole bit : 
I was doing something wrong, the goal now will be to transform this : 
`.../?layers=layer-id-1,layer-id-2&layer-id-1=1,3,5` into `/#map?layers=layer-id-1@features=1:3:5;layer-id-2`

it will also work with something along the line of this `.../?layer-id-1=1,3,5/#map/layers=layer-id-1@features=1:2:4` to turn it into `/#map?layers=layer-id-1@features=1:2:3:4:5`

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,82 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+ * be showing by default. In the current mapviewer, this is accomplished by giving the extra
+ * parameter `@features` on the layer
+ *
+ * @param {Query} params All parameters given to us through the query
+ * @param {Store} store The store
+ * @param {Query} newQuery The Query which will be sent to the current parameter parser
+ */
+export function handleLayerFeaturePreselection(params, store, newQuery) {
+    params.forEach((param_value, param_key) => {
+        if (store.state.layers.config.find((layer) => layer.id === param_key)) {
```
pakb, line 190: to make sure you receive something "boolean-like" you could use `.some()` instead of `.find()`

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,82 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+ *   string
+ * @param {String} layers The new Query layers parameter, a semicolon separated string
+ * @returns
+ */
+function createLayersParamForFeaturePreselection(layerId, featuresIds, layers) {
+    const featuresArray = featuresIds.split(',')
+    // we go through each layer
+    const layersArray = layers.split(';')
+    for (let layerIndex = 0; layerIndex < layersArray.length; layerIndex++) {
+        if (layersArray[layerIndex].includes(layerId)) {
```
pakb, line 219: you could `const` this `layersArray[layerIndex]` once and then reuse it bellow, might be easier to read

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,82 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+}
+
+/**
+ * @param {String} layerId The layer Id for which we have features
+ * @param {String} featuresIds The features ids we need to add as a parameter. This a coma-separated
+ *   string
+ * @param {String} layers The new Query layers parameter, a semicolon separated string
+ * @returns
+ */
+function createLayersParamForFeaturePreselection(layerId, featuresIds, layers) {
```
pakb, line 214: As you've isolated this logic in a neat function, I'd like to see some 🥁 unit test! covering this 🙃 

The logic bellow isn't that easy to understand at first glance, would be better to double check it is correct with tests

tests/cypress/support/commands.js
```diff
@@ -227,6 +227,21 @@ Cypress.Commands.add(
...
+                    let targetLegacyFeaturePreselection = 0
+                    state.layers.activeLayers.forEach((layer) => {
+                        // this is true if the layer is in the query parameters, and if
+                        // the layer is not already active
+                        targetLegacyFeaturePreselection += Boolean(
+                            Object.keys(queryParams).includes(layer.id) &&
+                                !queryParams.layers?.includes(layer.id)
+                        )
+                    })
+                    console.log(targetLegacyFeaturePreselection)
```
pakb, line 243: debugging leftover I guess

tests/cypress/support/commands.js
```diff
@@ -227,6 +227,21 @@ Cypress.Commands.add(
                 if (!withHash && 'adminId' in queryParams) {
                     // In legacy drawing with adminId the layer is not added to the layers parameter
                     target += 1
+                    // When handling a legacy parameter, the bod Layer Id might add extra
+                    // layers, thus the need to check if those extra layers have been added
```
pakb, line 231: you might want to rephrase this and change the "BOD layer ID" to "Feature preselection", it might make it clearer what this is all about

src/utils/legacyLayerParamUtils.js
ltkum: that's a perfectly reasonable request

src/utils/__tests__/legacyLayerParamUtils.spec.js
```diff
@@ -282,5 +283,32 @@ describe('Test parsing of legacy URL param into new params', () => {
...
+                const results = []
+                layersParams.forEach((params) => {
+                    results.push(
+                        createLayersParamForFeaturePreselection('layer.id', '1,2,3', params)
+                    )
+                })
+                for (let i = 0; i < results.length; i++) {
+                    expect(expectedResults[i]).to.eq(results[i])
+                }
+            })
```
pakb, line 311: we are not limited (or limiting ourselves) with `.it` calls while unit testing, i would prefer each of this case your are testing here would be its own `.it` with a description of what you are testing as input message

And the beauty of unit testing is that you can safely break stuff, so please also test some "unwanted" inputs to see how your code would react (users will do it for your if you don't do it here :wink: )

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +411,70 @@ describe('Test on legacy param import', () => {
...
+
+            beforeEach(() => {
+                // add intercept for layer config
+                cy.intercept('**/rest/services/all/MapServer/layersConfig**', {
+                    fixture: 'layers.fixture',
+                })
+                // add intercept for the html popup
+                cy.intercept('**/MapServer/**/htmlPopup**', {
+                    fixture: 'html-popup.fixture.html',
+                })
```
pakb, line 461: I bet these two are already part of the default interception suite deployed when calling `.goToMapView`, and if it is not the case, please add them there instead (and remove them from your beforeEach context)

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +411,70 @@ describe('Test on legacy param import', () => {
...
+
+        beforeEach(() => {
+            // add intercept for layer config
+            cy.intercept('**/rest/services/all/MapServer/layersConfig**', {
+                fixture: 'layers.fixture',
+            })
+            // add intercept for the html popup
+            cy.intercept('**/MapServer/**/htmlPopup**', {
+                fixture: 'html-popup.fixture.html',
+            })
```
pakb, line 433: same

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,94 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+ * @param {String} layers The new Query layers parameter, a semicolon separated string
+ * @returns
+ */
+export function createLayersParamForFeaturePreselection(layerId, featuresIds, layers) {
+    // if the features Ids are null, we replace them with an empty array
+    // we also remove all empty strings from the featuresIds
+    const featuresArray = (featuresIds ? featuresIds.split(',') : []).filter(
+        (featureId) => featureId !== ''
+    ) // we go through each layer
+    console.log(featuresArray)
```
ltshb, line 224: No console log, use log.debug

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,94 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+    const layersArray = layers.split(';')
+    for (let layerIndex = 0; layerIndex < layersArray.length; layerIndex++) {
```
ltshb, line 226: This should be simplified. This code has way too much indentation which make it hard to read/understand, it needs to be break down.
I don't think we need to loop over all layers we need to find the one that match `layerId` something like this

```javascript
const layer = layers.split(';').find(l => l.match(new RegExp(`^${layerId}(@.*)?$`)))
```

src/utils/legacyLayerParamUtils.js
```diff
@@ -173,3 +173,94 @@ export async function getKmlLayerFromLegacyAdminIdParam(adminId) {
...
+        const [layerIdWithCustomParams, visible, opacity] = layersArray[layerIndex].split(',')
+        if (layerIdWithCustomParams.split('@')[0] === layerId) {
+            // here, we manipulate the layer string for which we received a parameter
+
+            // we declare the string that will replace the current string
+            let layerString = layerIdWithCustomParams
+
+            if (layerIdWithCustomParams.includes('features')) {
+                // if there are features already declared, we need to do some extra work
```
ltshb, line 235: We don't need to handle mixing up of old and new way of declaring the feature id selection parameter, so I don't fully understand why we would already have a `@feature` ?

tests/cypress/support/commands.js
```diff
@@ -307,6 +314,19 @@ Cypress.Commands.add('waitAllLayersLoaded', ({ queryParams = {}, legacy = false
...
+            if (legacy && active !== target) {
+                let targetLegacyFeaturePreselection = 0
+                state.layers.activeLayers.forEach((layer) => {
+                    targetLegacyFeaturePreselection += Boolean(
+                        Object.keys(queryParams).includes(layer.id) &&
+                            !queryParams.layers?.includes(layer.geoAdminID)
+                    )
+                })
+                target += targetLegacyFeaturePreselection
+            }
```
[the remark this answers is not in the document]
ltkum: This is what the check is supposed to do : 
the `legacy && active !== target` is there to exclude non legacy queries (as they can't receive such parameter), and only checks if there is a disparity between the active layers and the targets
It parses trough each layer in "the active layers", and if that layer is inside the query's keys but not in the `layers` param, it adds 1

I will make some tests to check if I can break this, but otherwise it behaved as expected.

PR #683, opened 2024-03-04
PB-190: Refactor Infobox Module for Composition API

ltkum:
As a preliminary work for the showtooltip implementation, we have refactored the Infobox Module to the composition API. 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-190-showtooltip-implementation/index.html)


ltshb (approved):
All good. While testing I've find a small bug, the cursor is set to pointer on the infobox header (see line 162), it should only be pointer when hovering the header buttons but not the bar, would be nice to also quickly fix this.

PR #685, opened 2024-03-05
PB 190 : alter tooltip position behaviour

ltkum:
We are introducing a new url parameter : tooltipposition, with its store value in tooltipPosition.
It can take up four different values : 
 - none : wont display tooltips. Clicking on a feature should remove this behavior (this is not yet done)
 - fixed : always display the tooltip in the infobox
 - floating : always display the tooltip on the map
 - default : the default value, which is the current behaviour (fixed on mobile, floating on desktop)


src/store/modules/ui.store.js
```diff
@@ -276,6 +295,35 @@ export default {
...
+                commit('setTooltipPosition', {
+                    position: TooltipPositions.DEFAULT,
+                    dispatcher,
+                })
+            } else if (getters.floatingTooltip) {
+                commit('setTooltipPosition', { position: TooltipPositions.FIXED, dispatcher })
+            } else {
+                commit('setTooltipPosition', { position: TooltipPositions.FLOATING, dispatcher })
+            }
+        },
```
ltshb, line 326: Doesn't make sense to have a toggle action for a value that is not a boolean flag

src/store/plugins/click-on-map-management.plugin.js
```diff
@@ -82,6 +82,9 @@ const clickOnMapManagementPlugin = (store) => {
                         features: newSelectedFeatures,
                         ...dispatcher,
                     })
+                    if (store.getters.noTooltip && newSelectedFeatures.length > 0) {
+                        store.dispatch('toggleTooltipPosition', dispatcher)
```
ltshb, line 86: Here I would not use `toggle` but `set` action setting the value to `DEFAULT`

ltshb (commented):
I think we should rethink the wording here, `tooltip` is by definition a small info box that is displayed on mouse hovering (https://en.wikipedia.org/wiki/Tooltip)

@pakb @boecklic @ltkum here is a proposal (feel free to comment improve it 😉 ):
- URL parameter key := featureInfo
- URL parameter values
  - no parameter => no `featureInfo` => `"none"`
  - no value (but parameter present) => `"default"`
  - `"DEFAULT"` => base on UI size choose between `"tooltip"` and `"panel"
  - `"panel"`` => actual infoBox, panel on the bottom of the map
  - `"tooltip"` => actual floating tooltip

We could also use `bPanel` or `bottomPanel` instead of `panel`, so we could have other panel position in future.

This would also allow use to have `window` value for floating window (like in old viewer), or a `fullscreen` to have the info in fullscreen mode, but that's just thoughts for future.

Of course based on the acceptance and/or other proposal the code wordings needs to be adapted.

ltkum:
@ltshb, @boecklic, @pakb  : this seems like a good proposal to me. featureInfo seems like a good name.
For the future thoughts, I don't see any problems with the idea. I would prefer `bottompanel` rather than `bpanel` to make it easier to read and understand what it does, and I would also use this naming convention for the floating one, since this is technically not a `tooltip` by definition. I would go for :
`featureInfo` as a parameter, with  the code using `featureInfoPosition` rather than `tooltipPosition`
and for the currently supported values :
`bottompanel` --> display the features informations in the infobox at the bottom
`floatingpanel` --> display the features informations on the map at the features coordinates
`default` --> depending on UI size, act as `bottompanel` or `floatingpanel`
`none` --> we do not show the feature info

boecklic:
'featureInfo' makes sense to me too 👍🏻, maybe just one thought: we have kind of a "raw" featureInfo i.e. a json response from an API and a rendered featureInfo (aka htmlPopup). Currently we fully rely on the gtmlPopup, I could imagine that we'll simplify this in the future and just find a simple generic way to render directly the json in the popup...

src/store/modules/ui.store.js
ltkum: It made sense in my mind since we were usually 'toggling' between two states, but only using 'setFeatureInfoPosition' in the code will make it easier to understand, especially if one day we implement other ways to display this.

ltshb:
> @ltshb, @boecklic, @pakb : this seems like a good proposal to me. featureInfo seems like a good name. For the future thoughts, I don't see any problems with the idea. I would prefer `bottompanel` rather than `bpanel` to make it easier to read and understand what it does, and I would also use this naming convention for the floating one, since this is technically not a `tooltip` by definition. I would go for : `featureInfo` as a parameter, with the code using `featureInfoPosition` rather than `tooltipPosition` and for the currently supported values : `bottompanel` --> display the features informations in the infobox at the bottom `floatingpanel` --> display the features informations on the map at the features coordinates `default` --> depending on UI size, act as `bottompanel` or `floatingpanel` `none` --> we do not show the feature info

@ltkum the current `floatingTooltip` is in my opinion a tooltip (although not floating), as it is attached to the feature with an arrow, the only difference to tooltip is that it is displayed on click event instead of on hover, but its look and feel is like a tooltip. Therefore I would keep `tooltip` instead of `floatingpanel`, it is also not floating as you cannot move it like in the old viewer where we have a floating window. For the the bottom panel please use camelcase `bottomPanel`

src/store/modules/ui.store.js
```diff
@@ -200,6 +207,21 @@ export default {
                 state.hostname.includes(hostname)
             )
         },
+        floatingPanelFeatureInfo(state, getters) {
```
ltshb, line 210: Keep the same nomenclature everywhere `floatingPanel` => `tooltip` => `tooltipFeatureInfo`

src/store/modules/ui.store.js
```diff
@@ -276,6 +295,23 @@ export default {
...
+                commit('setFeatureInfoPosition', {
+                    position: FeatureInfoPositions[upCasePos],
+                    dispatcher: dispatcher,
+                })
+            } else if (!FeatureInfoPositions[upCasePos]) {
+                commit('setFeatureInfoPosition', {
+                    position: FeatureInfoPositions.DEFAULT,
+                    dispatcher: dispatcher,
+                })
+            }
```
ltshb, line 313: I would not silence invalid input here, simply do 

```
if (!FeatureInfoPositions[upCasePos]) {
  log.error(`Invalid feature info position: ${FeatureInfoPositions[upCasePos]}`)
  return // do nothing
}
// set the value
```

ltshb: The commit dispatcher could be simply
```javascript
commit('setFeatureInfoPosition', {
  position: FeatureInfoPositions[upCasePos],
  dispatcher
})
```

src/store/plugins/click-on-map-management.plugin.js
```diff
@@ -82,6 +84,12 @@ const clickOnMapManagementPlugin = (store) => {
                         features: newSelectedFeatures,
                         ...dispatcher,
                     })
+                    if (store.getters.noFeatureInfo && newSelectedFeatures.length > 0) {
```
ltshb, line 87: On a click I would always force the default behavior, otherwise the behavior will not be consistent !

src/modules/infobox/InfoboxModule.vue
```diff
@@ -17,40 +18,30 @@ const content = ref(null)
 const store = useStore()
 
 const selectedFeatures = computed(() => store.state.features.selectedFeatures)
-const floatingTooltip = computed(() => store.state.ui.floatingTooltip)
+const bottomPanelFeatureInfo = computed(() => store.getters.bottomPanelFeatureInfo)
+const floatingPanelFeatureInfo = computed(() => store.getters.floatingPanelFeatureInfo)
```
ltshb, line 22: `tooltipFeatureInfo`

src/modules/infobox/InfoboxModule.vue
```diff
@@ -91,7 +85,7 @@ function onClose() {
                     v-if="showFloatingToggle"
```
ltshb, line 85: Wording `showTooltipToggle`

src/modules/infobox/InfoboxModule.vue
```diff
@@ -64,8 +55,11 @@ watch(selectedFeatures, (features) => {
 function onToggleContent() {
     showContent.value = !showContent.value
 }
-function onToggleFloating() {
-    store.dispatch('toggleFloatingTooltip', dispatcher)
+function changeFeatureInfoPosition() {
```
ltshb, line 58: `setTooltipFeatureInfoPosition`

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -192,6 +192,9 @@ export default {
                 (l) => l instanceof GeoAdminWMTSLayer || l instanceof GeoAdminWMSLayer
             )
         },
+        isFeatureTooltipInFooter() {
```
ltshb, line 195: `isFeatureInfoInBottomPanel` or `isBottomPanelFeatureInfo`

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -602,6 +605,12 @@ export default {
                 })
             }
         },
+        changeFeatureInfoPosition() {
```
ltshb, line 608: `setBottomPanelFeatureInfoPosition`

src/modules/map/components/openlayers/OpenLayersHighlightedFeatures.vue
```diff
@@ -29,10 +30,10 @@ const dispatcher = { dispatcher: 'OpenLayersHighlightedFeatures.vue' }
 const store = useStore()
 const selectedFeatures = computed(() => store.state.features.selectedFeatures)
 const isCurrentlyDrawing = computed(() => store.state.ui.showDrawingOverlay)
-const isFloatingTooltip = computed(() => store.state.ui.floatingTooltip)
 const projection = computed(() => store.state.position.projection)
 const highlightedFeatureId = computed(() => store.state.features.highlightedFeatureId)
-
+const isFloatingTooltip = computed(() => store.getters.floatingPanelFeatureInfo)
+const isFixedTooltip = computed(() => store.getters.bottomPanelFeatureInfo)
```
ltshb, line 36: wordings

src/modules/map/components/openlayers/OpenLayersHighlightedFeatures.vue
```diff
@@ -111,8 +116,11 @@ useVectorLayer(
 function clearAllSelectedFeatures() {
     store.dispatch('clearAllSelectedFeatures', dispatcher)
 }
-function toggleFloatingTooltip() {
-    store.dispatch('toggleFloatingTooltip', dispatcher)
+function changeFeatureInfoPosition() {
```
ltshb, line 119: wording

src/store/modules/position.store.js
```diff
@@ -312,15 +312,18 @@ const actions = {
...
+                ? {
+                      x: position.x,
+                      y: position.y,
+                      z: position.z,
+                      // wrapping all angle-based values so that they do not exceed a full-circle value
+                      roll: wrapDegrees(position.roll),
+                      pitch: wrapDegrees(position.pitch),
+                      heading: wrapDegrees(position.heading),
+                  }
+                : null
```
ltshb, line 326: Why this ? I'm not sure that setting a null position is a very good idea ?

ltshb (changes requested):

src/store/plugins/click-on-map-management.plugin.js
ltshb: After testing I saw the point here, when selecting a feature, then put the info in bottom panel, then selecting another feature, in this case you still want the feature in bottom panel. So this is ok, but I would add a small comment here so for the code reader it is clearer.

src/store/modules/position.store.js
ltkum: null is the camera value everytime we are in 2d 
https://github.com/geoadmin/web-mapviewer/blob/7f7397ceef45213da51192bb710795321fe8309c/src/store/modules/position.store.js#L104-L111

I encountered a bug where, when switching from 3d to 2d, I got an error because there was no check to see if the camera received a valid input (it was trying to do `null.x` , which does not work.)

I saw that on `develop` there was an error too, but that error wasn't blocking. It was blocking locally on my branch, so I thought I could correct this. We can make this a separate bugfix if you'd prefer and add a comment in the function as well if you think it'd  be clearer as to why we are doing this.

src/modules/map/components/openlayers/OpenLayersHighlightedFeatures.vue
```diff
@@ -94,8 +95,12 @@ watch(nonEditableFeature, () => {
...
             ['Point', 'MultiPoint'].includes(feature.geometry?.type)
         ).length === nonEditableFeature.value.length
-    if (isFloatingTooltip.value && !containsOnlyPoints) {
-        toggleFloatingTooltip()
+    if (!isBottomPanelFeatureInfo.value && !containsOnlyPoints) {
+        // we need to also go to 'BOTTOMPANEL' if we are on 'NONE', here
+        store.dispatch('setFeatureInfoPosition', {
+            featureInfo: FeatureInfoPositions.BOTTOMPANEL,
+            dispatcher: dispatcher,
+        })
```
ltshb, line 103: Would this not be an issue (a contradiction) when the featureInfo parameter is not present or None ?

src/router/storeSync/storeSync.config.js
```diff
@@ -104,6 +105,15 @@ const storeSyncConfig = [
...
     new CompareSliderParamConfig(),
     new LayerParamConfig(),
+    new SimpleUrlParamConfig(
+        'featureInfo',
+        'setFeatureInfoPosition',
+        'setFeatureInfoPosition',
+        (store) => store.state.ui.featureInfoPosition,
+        false,
+        String,
+        FeatureInfoPositions.DEFAULT
```
ltshb, line 115: I suggest to set `NONE` as default, so it make easier to create a new permalink without info, but we can quickly discuss this in refinement.

src/store/modules/ui.store.js
```diff
@@ -77,12 +84,13 @@ export default {
...
-         * the feature's coordinates. If false, the tooltip will be displayed in the footer
+         * Expected position of the features tooltip position when selecting features.
          *
-         * @type Boolean
+         * The default position depends on the UI Modes and is checked on the fly
+         *
+         * @type String
          */
-        floatingTooltip: false, // Configured in screen-size-management.plugin.js
+        featureInfoPosition: FeatureInfoPositions.DEFAULT,
```
ltshb, line 93: See comment about we need to define the default behavior in refinement meeting, which means that if the default needs to change it also needs to change here, to be discussed.

ltshb (commented):
Do you plan to add e2e test in subsequent PR ?

src/router/storeSync/storeSync.config.js
ltkum: setting it to default would also simplify the legacy parameter, as it was the default behavior of 'showTooltip' . We can discuss this quickly tomorrow and see if anyone has another idea

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -192,13 +192,16 @@ export default {
                 (l) => l instanceof GeoAdminWMTSLayer || l instanceof GeoAdminWMSLayer
             )
         },
+        isFeatureInfoInBottomPanel() {
+            return !this.tooltipFeatureInfo
+        },
```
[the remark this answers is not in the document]
ltkum: Looks like I was a bit dumb here, as this computed only was used in showFeaturesPopover, and on its negation. I'm replacing it with isFeatureInfoInTooltip

I might have replaced the previous implementation which was set as a negative without thinking, here.

ltkum:
> Do you plan to add e2e test in subsequent PR ?

I'll be working on that right now

tests/cypress/tests-e2e/featurePreSelection.cy.js
```diff
@@ -26,24 +55,36 @@ describe('Testing the feature selection in the URL', () => {
...
+            })
+            it('Shows the tooltip in its correct position when set to default (bottom Panel on Phone)', function () {
+                goToMapViewWithFeatureSelection(FeatureInfoPositions.DEFAULT)
+                checkFeatures()
+                checkFeatureInfoPosition(FeatureInfoPositions.DEFAULT)
+            })
+            it('Shows the tooltip on the map when featureInfo is set to tooltip, and handle strange cases', function () {
+                goToMapViewWithFeatureSelection('TOoLtIp')
+                checkFeatures()
+                checkFeatureInfoPosition(FeatureInfoPositions.TOOLTIP)
```
ltshb, line 87: Do all the tests in one `it()`. `it()` are costly in term of cost in cypress cloud and performance

ltshb (approved):

PR #686, opened 2024-03-05
PB-244: Coordinate display css styling and text selectability

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-244-coordinate-display-css-styling-and-text-selectability/index.html)


ltshb:
@LukasJoss @pakb I think I've found the issue with the flaky test, the issue is due to the save drawing debouncing that is started when adding a feature and the close drawing. It seems to me that the debouncing save happens once the menu is closed and therefore save an empty drawing which replace the layer. I'll add a fix to this PR and let you know for the review. I think this issue has actually nothing to do with the changes of this PR but is a pre-existing race condition.

PR #689, opened 2024-03-06
PB-339: Fix layer z-index computation

ltshb:
The logic has two issues, one is that for group of layers we counted always
one index too much, and that for the on top of all layers we did not took the
group of layers into account.

This solve the issues in the e2e tests for the external layers.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-339-layers-zindex/index.html)


pakb (approved):
👍

PR #694, opened 2024-03-07
PB-314: Moved the &embed to #/embed

ltshb:
Now we use a new view for the embed instead of a query parameter.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-314-embed/index.html)

### DIFF
--- a/src/router/storeSync/storeSync.routerPlugin.js
+++ b/src/router/storeSync/storeSync.routerPlugin.js
@@ -3,8 +3,8 @@ import axios from 'axios'
 import { IS_TESTING_WITH_CYPRESS } from '@/config'
 import { STORE_DISPATCHER_ROUTER_PLUGIN } from '@/router/storeSync/abstractParamConfig.class'
 import storeSyncConfig from '@/router/storeSync/storeSync.config'
+import { MAP_VIEWS } from '@/router/viewNames'
 import log from '@/utils/logging'
-import { MAP_VIEWS } from '@/views/views'
 
 export const FAKE_URL_CALLED_AFTER_ROUTE_CHANGE = '/tell-cypress-route-has-changed'
 
--- a/src/store/modules/ui.store.js
+++ b/src/store/modules/ui.store.js
@@ -55,7 +55,7 @@ export default {
          *
          * @type Boolean
          */
-        embedMode: false,
+        embed: false,
         /**
          * Flag telling if a loading bar should be shown to tell the user something is on going
          *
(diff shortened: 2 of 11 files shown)


src/router/storeSync/storeSync.routerPlugin.js
pakb, line 1: As a global comment on this file, I'm not sure we shouldn't make our view names constant in `view/view.js` and never use a hard-coded string. It might be more resilient to future renaming of the views

src/store/modules/ui.store.js
```diff
@@ -289,7 +289,7 @@ export default {
             state.fullscreenMode = mode
         },
         setEmbeddedMode(state, { embed }) {
```
pakb, line 291: It would make sense to reflect the change `embedded` => `embed` also in the action/mutation name

src/utils/utils.js
```diff
@@ -188,3 +188,65 @@ export function getUrlQuery() {
...
+ *
+ * We use query in the hash fragment, the standard new URL doesn't parse the query from the hash,
+ * this function does and return the URL object with the hash fragment without query and the the
+ * query separated.
+ *
+ * @param {string} url
+ * @returns {{ urlObj: URL; hash: string; query: string }} Parsed url
+ */
+export function parseUrlHashQuery(url) {
+    const urlObj = new URL(url)
```
pakb, line 203: this might raise a `TypeError` if the given URL isn't valid, I don't know if it wouldn't be better to wrap all this in a `try...catch` and manage the error here (instead of giving it back to the caller)

Or we decide this is a purely internal function, we know what we are doing, and it is while developing that we have error in the console and all is "well"

src/views/EmbedView.vue
```diff
@@ -0,0 +1,44 @@
...
+</template>
+
+<style lang="scss" scoped>
+@import 'src/scss/variables.scss';
+#map-view {
+    position: absolute;
+    width: 100%;
+    height: 100%;
+    overflow: hidden;
+}
```
pakb, line 43: might make sense to add this styling as a class `.map-view` declared in `webmapviewer-bootstrap-theme.scss` (or another specialized SCSS file) so that we may share how our "main view" are styled across the board

ltshb:
> At first glance, by playing with the test link : great improvement! thanks
> 
> I saw that you filtered out some part of the UI that were forgotten (when they were added...) I think we should also remove the BG selector when in embedded (check an iframe example from mf-geoadmin3 here : https://codepen.io/geoadmin/pen/yOBzqM )
> 
> I'm totally OK not having the 3D button 👍

The footer clean up and button will be done in separate PR/tickets

src/utils/utils.js
ltshb: I can add a try catch, might be a good idea for future. Currently the function is only called with window.location.href which is by design always valid

src/store/modules/ui.store.js
ltshb: Done

src/utils/utils.js
ltshb: After looking close I think it is better to leave it as is. I think it would be wise to do the try catch if the input comes from a user (like for the external provider), but here it is for the moment used to programmatisch change the url from map to embed or vise versa and the url should always be valid. If it is not, then it is a code mistake and it is better to crash early with a backtrace as silently in my opinion.

src/views/EmbedView.vue
ltshb: Done

pakb (approved):

PR #698, opened 2024-03-12
PB 113: Legacy test coverage

ltkum:
We are adding tests for every official legacy parameter

Only missing now : 
`bod-layer-id` and `showTooltip` which are part of PB-190
`time` which is not yet implemented and will be part of PB-307

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-113-legacy-test-coverage/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/legacyParamImport.cy.js
+++ b/tests/cypress/tests-e2e/legacyParamImport.cy.js
@@ -3,7 +3,6 @@
 import proj4 from 'proj4'
 
 import { DEFAULT_PROJECTION } from '@/config'
-import { CrossHairs } from '@/store/modules/position.store'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 
 describe('Test on legacy param import', () => {
@@ -341,18 +340,6 @@ describe('Test on legacy param import', () => {
                 expect(query.toString()).to.equal(expectedQuery.toString())
             })
         })
-        it('Sets the background layer correctly', () => {
-            cy.goToMapView(
-                {
-                    bgLayer: 'test.background.layer2',
-                },
-                false
-            )
-            cy.readStoreValue('state.layers.currentBackgroundLayer').then((bgLayer) => {
-                expect(bgLayer).to.not.be.null
-                expect(bgLayer.id).to.eq('test.background.layer2')
-            })
-        })
     })
 
     context('3D import', () => {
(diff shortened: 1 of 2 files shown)


tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -324,6 +341,18 @@ describe('Test on legacy param import', () => {
...
+                {
+                    bgLayer: 'test.background.layer2',
+                },
+                false
+            )
+            cy.readStoreValue('state.layers.currentBackgroundLayer').then((bgLayer) => {
+                expect(bgLayer).to.not.be.null
+                expect(bgLayer.id).to.eq('test.background.layer2')
+            })
+        })
```
ltshb, line 355: bglayer is not a legacy parameter and we should not test it here, it is already tested in layers.cy.js

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +440,142 @@ describe('Test on legacy param import', () => {
             cy.readStoreValue('state.position.projection.epsgNumber').should('eq', 3857)
         })
     })
+
+    context('Extra Parameter Imports', () => {
+        it('sets the language correctly', () => {
+            cy.goToMapView({ lang: 'it' }, false)
+            cy.readStoreValue('state.i18n.lang').should('eq', 'it')
+        })
```
ltshb, line 448: Dito lang is not a legacy parameter and is already tested, we should remove it from here to keep the number of `it()` low

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +440,142 @@ describe('Test on legacy param import', () => {
...
+
+        it('shows the correct crosshair', () => {
+            function checkCrosshair(crossHair) {
+                cy.readStoreValue('state.position').then((positionStore) => {
```
ltshb, line 452: Dito crosshair is not a legacy param and is already tested.

ltshb, line 452: Generally we should avoid such construct in cypress `cy.readStoreValue('state.position').then()` this has for sure no retry and can lead to flaky test unless we are sure that the value is set before the call is done (due to other waits). Basically the following construct should be prefered
`cy.readStoreValue('state.position').should()` which will retry until the assertion in should is done (note we can do assertion in a callback in should as well)

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +440,142 @@ describe('Test on legacy param import', () => {
...
+                    layers: 'test-1.wms.layer',
+                    swipe_ratio: '0.3',
+                },
+                false
+            )
+            // initial slider position is width * 0.3 -20
+            cy.get('[data-cy="compare_slider"]').then((slider) => {
+                cy.readStoreValue('state.ui.width').then((width) => {
+                    cy.wrap(slider.position()['left']).should('eq', width * 0.3 - 20)
+                })
```
ltshb, line 497: Use `.should((width)=>{})` construct instead with `expect` this ensure retries until the assertion is true

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +440,142 @@ describe('Test on legacy param import', () => {
...
+
+            it("Doesn't prompt the user if geolocation has previously been authorized", () => {
+                cy.on('window:alert', () => {
+                    throw new Error('Should not prompt for geolocation API permission again')
+                })
+                cy.readStoreValue('state.geolocation.active').should('be.true')
+            })
+
+            it('Uses the values given by the Geolocation API to feed the store', () => {
+                cy.readStoreValue('state.geolocation.position').then((position) => {
```
ltshb, line 534: Dito use `should` instead of `then`

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -411,4 +440,142 @@ describe('Test on legacy param import', () => {
...
+        }
+    )
+    context('Topics Import', () => {
+        let mockupTopics = {}
+        beforeEach(() => {
+            cy.fixture('topics.fixture').then((topics) => {
+                mockupTopics = topics
+            })
+        })
+        it('shows the correct topic, with the correct layers, visibilities, opacities and background layers', () => {
```
ltshb, line 550: Same topic is not a legacy parameter and is already tested

tests/cypress/tests-e2e/legacyParamImport.cy.js
ltkum: bgLayer is part of the official parameters of the old viewer. I figured I would also test it through a legacy test to make sure it works well. If you feel this is not useful, I'll remove it.

ltkum: That being said, there is already a tests which checks that params that do not change are transferred as is. So we can indeed ignore the official old parameters which remained the same

ltshb (approved):

PR #701, opened 2024-03-12
PB-190: adding features param in URL when clicking

ltkum:
Issue : when clicking on features, the special params `@features=...` was not shown in the URL

Fix : we now add those features Ids to their layers special params

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-190-feat-add-featuresids-on-click/index.html)


ltshb (commented):
When selecting one feature of a layer and then selecting another feature, the former stays in url and in infobox but the selection diseapeared on the map ! 
It might be wise to support only selection of feature of a single layer for the moment and to remove the @feature from url when a feature is deselected. This would also avoid after closing the infobox and do a reload that the feature is selected again.
E2E test should also be updated to do more testing
- select feature => check url
- select another feature =Y check url
- select another feature from another layer (with timestamp) => check url
- reload => check that feature is selected
- close infobox and reload => check that feature is not selected
-

src/router/storeSync/layersParamParser.js
```diff
@@ -109,3 +113,22 @@ export function transformLayerIntoUrlString(layer, defaultLayerConfig) {
...
+ */
+export function orderFeaturesByLayers(selectedFeatures) {
+    const layersFeatures = {}
+
+    selectedFeatures.forEach((feature) => {
+        if (!layersFeatures[feature.layer.id]) {
+            layersFeatures[feature.layer.id] = []
+        }
+        // In some features, the 'id' get attribute is built by concatening {layer.id}-{feature.id}
+        // We remove this concatenation here.
```
ltshb, line 129: Here the comment seems out of date ? Or some code is missing ?

tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -0,0 +1,212 @@
+/// <reference types="cypress" />
+import { FeatureInfoPositions } from '@/store/modules/ui.store'
+
+describe('Testing the feature selection in the URL', () => {
+    function checkFeatures() {
+        cy.readStoreValue('state.features.selectedFeatures').then((features) => {
+            cy.get('@featuresIds').then((featuresIds) => {
```
ltshb, line 7: To avoid flakiness might make worse a try to put this the other way around

```javascript
// Read first the fixture that don't require retries
cy.get('@featuresIds').then((featuresIds) => {
 cy.readStoreValue('state.features.selectedFeatures').should((features) => {
  expect(features.length).to.be.equal(featuresIds.length)
    expect(feature)...
  })
  })
})
```

src/router/storeSync/layersParamParser.js
ltkum: It is indeed an outdated comment

tests/cypress/tests-e2e/featureSelection.cy.js
ltkum: good idea

ltshb (approved):

PR #702, opened 2024-03-13
PB-315: Handle legacy embed view

ltshb:
The legacy embed view was at hostname/embed.html. Unfortunately we cannot reroute
here to another path before the #, because the vue router using the createWebHashHistory
can only handle route after the hash. Therefore we have an external redirect
service  that will redirect to /embed.html to ?legacyEmbed.

We use this pseudo legacyEmbed to redirect to #/embed once the other legacy
parameters have been translated.

NOTE: I first tried to set the redirect service directly to #/embed but then it made the startup procedure difficult as we needed to detect if other legacy parameters had to be translated and wait on them first. So for simplicity we use a legacy parameter as redirect
which simplify a lot the startup procedure which is then identical for other legacy parameters.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-315-legacy-embed/index.html)

### DIFF
--- a/src/router/appLoadingManagement.routerPlugin.js
+++ b/src/router/appLoadingManagement.routerPlugin.js
@@ -1,6 +1,5 @@
 import { START_LOCATION } from 'vue-router'
 
-import { MAP_VIEWS } from '@/router/viewNames'
 import { isLegacyParams } from '@/utils/legacyLayerParamUtils'
 import log from '@/utils/logging'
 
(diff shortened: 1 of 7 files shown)


src/router/appLoadingManagement.routerPlugin.js
```diff
@@ -52,12 +47,12 @@ const appLoadingManagementRouterPlugin = (router, store) => {
                     isLegacyUrl,
                     ...dispatcher,
                 })
+                unRegisterRouterHook()
```
pakb, line 50: What made it not working well keeping it under in the store subsribe?

pakb (approved):
the redirect service has already been updated with this new rule?

src/router/appLoadingManagement.routerPlugin.js
ltshb: Nothing i just thought it would be wise to stop this plug-in sooner rather than later

ltshb:
@pakb yes the redirect is ready and applied on dev see https://github.com/geoadmin/infra-kubernetes/pull/508 with this pr merge I’ll do some test before closing the ticket.

PR #704, opened 2024-03-13
PB-325: Add styling to openlayers drag zoom

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-325-draw_rectangle_when_shift_select/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersMouseTracker.vue
+++ b/src/modules/map/components/openlayers/OpenLayersMouseTracker.vue
@@ -78,6 +78,7 @@ function setDisplayedFormatWithId() {
 </template>
 
 <style lang="scss" scoped>
+@import 'src/scss/webmapviewer-bootstrap-theme';
 .mouse-position {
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/OpenLayersMouseTracker.vue
```diff
@@ -84,6 +84,10 @@ function setDisplayedFormatWithId() {
     text-align: left;
     white-space: nowrap;
 }
+// Show selected area when shift click + drag on map
+:global(.ol-dragzoom) {
+    border: 2px solid red;
```
pakb, line 89: very "nitpicky" but could you use the SASS variable for red, so that we may change it later?
Either "$red" or "$primary" are fine (both can be imported by adding `@import 'src/scss/webmapviewer-bootstrap-theme';` at the top of the `<style>` section, if not there already)

pakb (approved):

PR #713, opened 2024-03-18
Improve compare slider styling

ltshb:
Use a bit bigger arrows and use rounding corner like the rest of the application.


pakb (approved):
Have you tried using classes `px-2` and `py-1` instead of using hard-coded padding? Might be better (but maybe too big for what you wanted to achieve)

ltshb:
> Have you tried using classes `px-2` and `py-1` instead of using hard-coded padding? Might be better (but maybe too big for what you wanted to achieve)

I tried but it would make the icons quite bigger that's why I used css with pixel

PR #714, opened 2024-03-18
PB-332: Allow duplicate layers in layer list

ltshb:
- [ ] TODO unit tests and e2e test for duplicate layer from search
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-332-duplicate-layer/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
@@ -1,3 +1,101 @@
+<script setup>
+import tippy from 'tippy.js'
+import { computed, onBeforeUnmount, onMounted, ref, toRefs } from 'vue'
+import { useI18n } from 'vue-i18n'
+import { useStore } from 'vuex'
+
+import LayerTimeConfig from '@/api/layers/LayerTimeConfig.class'
+import {
+    CURRENT_YEAR_WMTS_TIMESTAMP,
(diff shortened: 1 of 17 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -1,3 +1,101 @@
...
+    if (hasMultipleTimestamps.value) {
+        popover = tippy(timeSelectorButton.value, {
+            theme: 'popover-button light-border',
+            content: timeSelectorModal.value,
+            allowHTML: true,
+            placement: 'right',
+            interactive: true,
+            arrow: true,
+            trigger: 'click',
+        })
```
pakb, line 60: We could change the Tippy composable you made so that it can handle a `ref` as `data-tippy-content`... Just wondering if it would make sense to have that extracted as a tippy for future use with rich content...

pakb (approved):
Great!

I've tested a bit the compare slider and I think we need to address a couple bugs there, but that's out of the scope of what you did here.

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
ltshb: I also thought to change the tippy composable to use a ref, but it made it more complex and I think it is best to keep it simple to be used for simple use case and use directly the tippy for more special case like here for the timeselector button which is not a tooltip on hover but a tooltip interactive on click.

PR #715, opened 2024-03-19
PB-103 : arg constructor for SelectableLayers and children

pakb:
moving isDragged to EditableFeatures because it is used only in this context moving extent to base class, as we might want to use it across the board (also with drawing)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-103_arg_constructor_features/index.html)

### DIFF
--- a/src/api/features/EditableFeature.class.js
+++ b/src/api/features/EditableFeature.class.js
@@ -16,36 +16,37 @@ export const EditableFeatureTypes = {
 /** Describe a feature that can be edited by the user, such as feature from the current drawing */
 export default class EditableFeature extends SelectableFeature {
     /**
-     * @param {String | Number} id Unique identifier for this feature (unique in the context it
-     *   comes from, not for the whole app)
-     * @param {Number[][]} coordinates Coordinates [[x,y],[x2.y2],...] or [x,y] if point geometry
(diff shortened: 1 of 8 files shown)


src/api/features/EditableFeature.class.js
```diff
@@ -16,66 +16,45 @@ export const EditableFeatureTypes = {
...
-     * @param {EditableFeatureTypes} featureType Type of this editable feature
-     * @param {FeatureStyleColor} textColor Color for the text of this feature
-     * @param {FeatureStyleSize} textSize Size of the text for this feature
-     * @param {FeatureStyleColor} fillColor Color of the icon (if defined)
-     * @param {DrawingIcon} icon Icon that will be covering this feature, can be null
-     * @param {FeatureStyleSize} iconSize Size of the icon (if defined) that will be covering this
-     *   feature
+     * @param {String | Number} featureData.id Unique identifier for this feature (unique in the
+     *   context it comes from, not for the whole app)
+     * @param {Number[][]} featureData.coordinates Coordinates [[x,y],[x2.y2],...] or [x,y] if point
```
ltshb, line 21: Nice way to document options I might do something similar in the layers store.

ltshb (approved):

PR #716, opened 2024-03-19
PB-103 : add parsing of getFeatureInfo capabilities for external layers

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-103_getfeatureinfo_cap_parsing/index.html)

### DIFF
--- a/src/api/layers/ExternalLayer.class.js
+++ b/src/api/layers/ExternalLayer.class.js
@@ -1,5 +1,18 @@
 import AbstractLayer, { LayerAttribution } from '@/api/layers/AbstractLayer.class'
 import { InvalidLayerDataError } from '@/api/layers/InvalidLayerData.error'
+import { WGS84 } from '@/utils/coordinates/coordinateSystems'
+
+/**
+ * Information required to create a GetFeatureInfo request to this external WM(T)S server. This
+ * enables us to identify features at a mouse click.
+ *
+ * @property {String} baseUrl Base WMS URL to use to create the GetFeatureInfo WM(T)S request
+ *   (adding a ? and all necessary query params after the base URL)
+ * @property {'GET' | 'POST'} method HTTP method to use when requesting this server
+ * @property {String[]} formats Available formats in which the server can respond
+ *
+ * @typedef ExternalLayerGetFeatureInfoCapability
+ */
 
 /**
  * External Layer Legend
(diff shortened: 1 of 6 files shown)


src/api/layers/ExternalLayer.class.js
```diff
@@ -117,6 +140,11 @@ export default class ExternalLayer extends AbstractLayer {
         this.abstract = abstract
         this.extent = extent
         this.legends = legends
+        this.availableProjections = [...availableProjections]
+        if (this.availableProjections.length === 0) {
+            this.availableProjections.push(WGS84)
+        }
```
ltshb, line 146: Might be wise to add a log warning here that we have a fallback to WGS84

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -42,6 +42,43 @@ export default class WMSCapabilitiesParser {
...
+                baseUrl = httpElement.Post.OnlineResource
+            } else {
+                log.error(
+                    "Couldn't parse GetFeatureInfo data",
+                    this.Capability.Request.GetFeatureInfo
+                )
+                if (ignoreError) {
+                    return null
+                }
+                throw new CapabilitiesError('Invalid GetFeatureInfo data')
```
ltshb, line 71: Here I would add a translation key after the message

```javascript
throw new CapabilitiesError('Invalid GetFeatureInfo data', 'invalid_get_feature_info')
```

We could use it then later for displaying an error

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -42,6 +42,43 @@ export default class WMSCapabilitiesParser {
...
+            }
+            return {
+                baseUrl,
+                method,
+                formats: [...this.Capability.Request.GetFeatureInfo.Format],
```
ltshb, line 76: What if `Format` is undefined ? Maybe `...this.Capability.Request.GetFeatureInfo.Format ?? {}`

ltshb (approved):
Good just minor comments. 

You should also add some unitest for the new get feature info capabitlies

src/api/layers/WMSCapabilitiesParser.class.js
pakb: As I'm using this array later and wouldn't want "default" value such as an empty Object, I'll check that Format exists before pushing it into the formats array.
This way, the formats array will stay empty if no format is defined by the service

PR #718, opened 2024-03-19
Change FeatreAreaInfo to accept geom as input

pakb:
And use turfJS instead of OL to calculate the area of the geometry

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-74_geometry_as_input/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureAreaInfo.vue
+++ b/src/modules/infobox/components/FeatureAreaInfo.vue
@@ -1,12 +1,12 @@
 <script setup>
 import area from '@turf/area'
 import { computed, toRefs } from 'vue'
-import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
-import { WGS84 } from '@/utils/coordinates/coordinateSystems.js'
-import { reprojectGeoJsonData, transformIntoTurfEquivalent } from '@/utils/geoJsonUtils.js'
+import { WGS84 } from '@/utils/coordinates/coordinateSystems'
+import { reprojectGeoJsonData, transformIntoTurfEquivalent } from '@/utils/geoJsonUtils'
 import { round } from '@/utils/numberUtils'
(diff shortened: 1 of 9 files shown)


src/modules/infobox/components/FeatureAreaInfo.vue
```diff
@@ -38,22 +53,36 @@ function roundValueIfGreaterThan(value, threshold, divider) {
...
+    background: repeating-linear-gradient(
+        45deg,
+        // transparent stripe
+        $transparent,
+        $transparent $rectangle-line-width,
+        // gray stripe
+        $rectangle-line-color $rectangle-line-width,
+        // another transparent stripe, sandwiching the gray stripe
+        $transparent 2 * $rectangle-line-width
+    );
```
ltshb, line 85: Nice one 👍🏼

src/modules/infobox/components/FeatureAreaInfo.vue
```diff
@@ -38,22 +53,36 @@ function roundValueIfGreaterThan(value, threshold, divider) {
 </script>
 
 <template>
-    <div :title="$t('area')" class="d-flex align-items-center">
+    <div :title="i18n.t('area')" class="d-flex align-items-center">
```
ltshb, line 56: When already changing this would be good to use my new tippy composable, so here you would need to use `data-tippy-content="area"` instead of `:title...` and you would need to add also an id so you can add the tippy as follow
`useTippyTooltip('#areaInfo[data-tippy-content]')`

ltshb: NOTE the selector in tippy act on the whole DOM when using the composable, therefore to avoid confusion and duplicate tippy we need to be quite restrictive here.

ltshb (approved):
Just change to use the tippy composable

PR #719, opened 2024-03-19
Remove getURL function in layer classes

pakb:
and simplifying a bit OpenLayersWMTSLayer component OL source definition

This should put us one step away from moving to a pure JSON based layer config approach (no more classes)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_remove_geturl/index.html)


ltshb (approved):
All good 👍🏼

PR #721, opened 2024-03-20
PB-207 : GeoJSON data autoreload

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-207_geojson_data_autoreload/index.html)

### DIFF
--- a/src/store/plugins/load-geojson-style-and-data.plugin.js
+++ b/src/store/plugins/load-geojson-style-and-data.plugin.js
@@ -36,6 +36,7 @@ async function autoReloadData(store, geoJsonLayer) {
     // creating the new interval to reload this layer's data
     intervalsByLayerId[geoJsonLayer.id] = setInterval(async () => {
         try {
+            store.dispatch('setShowLoadingBar', { loading: true, ...dispatcher })
             const { data } = await load(geoJsonLayer.geoJsonUrl)
             const layerCopy = geoJsonLayer.clone()
             layerCopy.geoJsonData = data
@@ -45,6 +46,7 @@ async function autoReloadData(store, geoJsonLayer) {
                 layers: [layerCopy],
                 ...dispatcher,
             })
+            store.dispatch('setShowLoadingBar', { loading: false, ...dispatcher })
             log.debug(`Data reloaded according to updateDelay for layer ${geoJsonLayer.id}`)
         } catch (error) {
             log.error(`Error while reloading GeoJSON data for layer ${geoJsonLayer.id}`, error)
@@ -107,6 +109,7 @@ export default function loadGeojsonStyleAndData(store) {
                         self.indexOf(self.find((layer) => layer.id === geoJsonLayer.id)) === index
                 )
 
+            store.dispatch('setShowLoadingBar', { loading: true, ...dispatcher })
             const updatedLayers = await Promise.all(
                 geoJsonLayers
                     .filter((layer) => layer.isLoading)
@@ -115,6 +118,7 @@ export default function loadGeojsonStyleAndData(store) {
             if (updatedLayers.length > 0) {
                 store.dispatch('updateLayers', { layers: updatedLayers, ...dispatcher })
             }
+            store.dispatch('setShowLoadingBar', { loading: false, ...dispatcher })
 
             // after the initial load is done,
             // going through all layers that have an update delay and launching the routine to reload their data
--- a/src/store/plugins/load-geojson-style-and-data.plugin.js
+++ b/src/store/plugins/load-geojson-style-and-data.plugin.js
@@ -10,6 +10,8 @@ import log from '@/utils/logging'
 
 const dispatcher = { dispatcher: 'load-geojson-style-and-data.plugin' }
 
+const intervalsByLayerId = {}
+
 async function load(url) {
     try {
         return await axios.get(url)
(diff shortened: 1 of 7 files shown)


src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -54,20 +92,33 @@ async function loadDataAndStyle(geoJsonLayer) {
...
                 })
-            store.dispatch('updateLayers', { layers: updatedLayers, ...dispatcher })
         }
         if (mutation.type === 'addLayer') {
             addLayersSubscriber([mutation.payload.layer])
         }
         if (mutation.type === 'setLayers') {
             addLayersSubscriber(mutation.payload.layers)
         }
+        if (mutation.type === 'removeLayerWithId' && intervalsByLayerId[mutation.payload.layerId]) {
```
ltshb, line 115: Layer can also be removed by index, see `removeLayerByIndex` mutation

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -36,13 +68,19 @@ async function loadDataAndStyle(geoJsonLayer) {
...
     } catch (error) {
         log.error(`Error while fetching GeoJSON data/style for layer ${geoJsonLayer?.id}`, error)
         clone.isLoading = false
         clone.errorKey = 'loading_error_network_failure'
         clone.hasError = true
-        return clone
+        await store.dispatch('updateLayer', {
+            layer: clone,
+            ...dispatcher,
+        })
```
ltshb, line 83: Here I just modified the code to avoid multiple dispatch when we have more than one geojson layer, would be good to keep it for startup performance.

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -54,20 +92,33 @@ async function loadDataAndStyle(geoJsonLayer) {
...
+                    if (layer.isLoading) {
+                        log.debug(`Loading data/style for added GeoJSON layer`, layer)
+                        loadDataAndStyle(store, layer)
+                    }
+                    if (layer.updateDelay > 0) {
+                        log.debug('starting auto-reload of data for layer', layer)
+                        autoReloadData(store, layer)
+                    }
                 })
-            store.dispatch('updateLayers', { layers: updatedLayers, ...dispatcher })
```
ltshb, line 64: The single `updateLayers` dispatch should be kept for performance reason during startup.

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -54,20 +92,33 @@ async function loadDataAndStyle(geoJsonLayer) {
...
-                .filter((layer) => layer instanceof GeoAdminGeoJsonLayer && layer.isLoading)
-                .map((layer) => {
-                    log.debug(`Loading data/style for added GeoJSON layer`, layer)
-                    return loadDataAndStyle(store, layer)
+        const addLayersSubscriber = (layers) => {
+            layers
+                .filter((layer) => layer instanceof GeoAdminGeoJsonLayer)
+                .forEach((layer) => {
+                    if (layer.isLoading) {
+                        log.debug(`Loading data/style for added GeoJSON layer`, layer)
```
ltshb, line 100: Based on other comment, here you could simply keep the `.map()` and do the autoreload in it.

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -19,11 +21,41 @@ async function load(url) {
...
+        `Adding auto-reload of data for layer ${geoJsonLayer.id} according to its updateDelay`,
+        geoJsonLayer
+    )
+    // creating the new interval to reload this layer's data
+    intervalsByLayerId[geoJsonLayer.id] = setInterval(async () => {
+        try {
+            const { data } = await load(geoJsonLayer.geoJsonUrl)
+            const layerCopy = geoJsonLayer.clone()
+            layerCopy.geoJsonData = data
+            await store.dispatch('updateLayer', {
```
ltshb, line 42: await not needed here the dispatch is not async in this case

ltshb (changes requested):

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -19,11 +21,50 @@ async function load(url) {
...
+        try {
+            const { data } = await load(geoJsonLayer.geoJsonUrl)
+            const layerCopy = geoJsonLayer.clone()
+            layerCopy.geoJsonData = data
+            // we update through the action updateLayers, so that if multiple copies of the same GeoJSON layer are present,
+            // they will all be updated with the fresh data
+            store.dispatch('updateLayers', {
+                layers: [layerCopy],
+                ...dispatcher,
+            })
```
ltshb, line 47: Code is all good, but just as information, here instead of doing the clone you do 
```
store.dispatch('updateLayers', {layers: [{id: geoJsonLayer.id, geoJsonData: data}], ...dispatcher})
```

But no need to change both are ok and good in my opinion

ltshb, line 47: We should set the loading bar here

ltshb (approved):

PR #726, opened 2024-03-20
PB-332: Added duplicate layer button to time series layers

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-332-duplicate-layer-button/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -113,13 +113,13 @@ function duplicateLayer() {
         ref="menuLayerItem"
         class="menu-layer-item"
         :class="{ compact: compact }"
-        :data-cy="`menu-active-layer-${id}`"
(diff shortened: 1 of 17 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -65,6 +65,7 @@ const attributionName = computed(() =>
     layer.value.attributions.map((attribution) => attribution.name).join(', ')
 )
 const showLegendIcon = computed(() => layer.value.hasLegend)
+const hasTimeEntries = computed(() => layer.value.timeConfig?.timeEntries?.length > 0 ?? false)
```
pakb, line 68: You might get away by using `.hasMultipleTimestamps` from `GeoAdminLayer` class, it checks that there is more than 1 timestamp available (sometimes there's only one, that wouldn't make much sense to duplicate such a layer)

ltshb: Good to know, I changed it, thanks

PR #727, opened 2024-03-21
PB-111: Fix print spec send to the print service

ismailsunni:
- Exclude print rectangle layer
- Update mapfishprint from this PR https://github.com/geoblocks/mapfishprint/pull/28
   - Remove npm warning 
   - Fix WMS encoding

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-111-fix-print-spec/index.html)


tests/cypress/tests-e2e/print.cy.js
```diff
@@ -0,0 +1,803 @@
...
+                body: {
+                    app: 'default',
+                    layouts: [
+                        {
+                            name: '1. A4 landscape',
+                            attributes: [
+                                {
+                                    name: 'name',
+                                    type: 'String',
+                                    default: 'this is a name',
```
ltshb, line 26: Would be much better to put this response mockup in a fixture files as here inline

tests/cypress/tests-e2e/print.cy.js
```diff
@@ -0,0 +1,803 @@
...
+    function interceptPrintRequest() {
+        cy.intercept('POST', '**/report.pdf', (req) => {
+            req.reply({
+                body: {
+                    ref: printID,
+                    statusURL: '/print/status/' + printID,
+                    downloadURL: '/print/report/' + printID,
+                },
+                delay: 200,
+            })
```
ltshb, line 611: Small response mockup like here are fine inline as the opposite of the capabilities which should be in a fixture file

src/api/print.api.js
```diff
@@ -16,6 +16,23 @@ const PRINTING_DEFAULT_POLL_INTERVAL = 2000 // interval between each polling of
 const PRINTING_DEFAULT_POLL_TIMEOUT = 600000 // ms (10 minutes)
 
 const SERVICE_PRINT_URL = `${API_SERVICES_BASE_URL}print3/print/default`
+
+class GeoAdminCustomizer extends BaseCustomizer {
+    constructor(layerNamesToExclude) {
```
pakb, line 21: Could you add a little JSDoc to describe what type you expect with `layerNamesToExclude`?

src/api/print.api.js
```diff
@@ -16,6 +16,23 @@ const PRINTING_DEFAULT_POLL_INTERVAL = 2000 // interval between each polling of
...
+        super()
+        this.layerNamesToExclude = layerNamesToExclude
+        this.layerFilter = this.layerFilter.bind(this)
+    }
+
+    layerFilter(layerState) {
```
pakb, line 27: same here

src/api/print.api.js
```diff
@@ -169,6 +186,8 @@ export class PrintError extends Error {
  *   Default is `false`
  * @param {CoordinateSystem} [config.projection=null] The projection used by the map, necessary when
  *   the grid is to be printed (it can otherwise be null). Default is `null`
+ * @param {String[]} [config.excludedLayers=[]] List of layer names to exclude from the print.
+ *   Default is `[]`
```
pakb, line 190: I would either pass `AbstractLayer` instances, or rename this param `excludedLayerOlNames` to make it clear what it contains

If I understand correctly you will filter layers based on the name we have given them in the OpenLayers context, right?
Might be better to find a way to filter things out based on the layer ID we have in the store (we give each OL layer the same ID as the layer ID in the store, most of the time)

src/modules/menu/components/print/MenuPrintSection.vue
```diff
@@ -84,6 +84,7 @@ async function printMap() {
     try {
         const documentUrl = await print(printGrid.value, printLegend.value)
         if (documentUrl) {
+            console.log('documentUrl: ', documentUrl)
```
pakb, line 87: little console log left :wink:

tests/cypress/tests-e2e/print.cy.js
pakb, line 1: I would make more sense to move all fixtures in dedicated fixutre files, and add all the intercept you do here as global intercepts in `command.js`

src/api/print.api.js
ismailsunni: I change it to filter by ID, since I guess it's more common to have ID than name. So, I change the param with `excludedLayerIDs` to make it more clear. Is that ok?

ltshb:
@ismailsunni the legend seems not to be printed anymore, at least with the public transport layer and hiking closure layer.

pakb:
Maybe there's something wrong with how we build the legend attribute, if you have a look here https://github.com/geoadmin/service-print3/blob/develop/print-apps/mapviewer/requestData_legend.json it looks like it should be only one entry with all icons in an array.

We do that instead :
```javascript
{
	"legends": {
		"classes": [
			{
				"icons": [
					"https://sys-api3.dev.bgdi.ch/static/images/legends/ch.bav.haltestellen-oev_en.png"
				],
				"name": "Public transport stops"
			},
			{
				"icons": [
					"https://sys-api3.dev.bgdi.ch/static/images/legends/ch.swisstopo.swisstlm3d-wanderwege_en.png"
				],
				"name": "Hiking trails"
			}
		],
		"name": "Legend"
	}
}
```

ismailsunni:
@pakb that's strange. I will check it why and add a unit test for it also. But I am pretty sure it was working fine with more than one legend.

ltshb:
@ismailsunni @pakb @hansmannj the legend issue with the print might be also due to the service-print3 deployment of this morning, @hansmannj what is the difference of you local service-print3 image and the master and develop latest state ?

ltshb:
> @ismailsunni @pakb @hansmannj the legend issue with the print might be also due to the service-print3 deployment of this morning, @hansmannj what is the difference of you local service-print3 image and the master and develop latest state ?

I just tried to revert on dev the deployment of service-print3 but the issue with the legend is still present.

ismailsunni:
@ltshb just curious, is it possible to revert the service-print deployment to some time last week? Just to make sure. Because the payloads sent are the same as before, even after refactoring.

pakb:
> @ltshb just curious, is it possible to revert the service-print deployment to some time last week? Just to make sure. Because the payloads sent are the same as before, even after refactoring.

As said, the issue arise when there's more than one layer with legend. If you, by any chance, didn't test that, but only added one layer extra (with legend) before printing, it might have gone under your radar (as mine :upside_down_face: )

hansmannj:
I'll check on the train in about half an hour.
If I remember correctly, we did not really touch any of the service configurations itself. The only difference should be, that the image is based on a different image tag of c2c.
Our new service-print3 release should be based on an c2c image, which includes some bug fixes which where discovered due to our load testing.
The service config files should not have changed.

Please correct me, if I have it wrongly in my mind @LukasJoss

hansmannj:
And yes, probably we never had a test case so far, with more than one layer plus legend, might well be.

hansmannj:
Probably it is in the print payload.
I guess, this is how it should look like, when there's more than one legend entry: https://github.com/geoadmin/service-print3/blob/c69347869eb7f284886f5352a64aa8042e5f6979/print-apps/mapviewer/requestData_WMTS_legend.json#L226-L237

hansmannj:
Testing locally with the payload form the link in my last comment works and results in a PDF with two legends. So maybe there's something wrong with the payload sent by the viewer?

ismailsunni:
@hansmannj @pakb @ltshb 
I found the issue. It's `legend` vs `legends`. The correct one should be `legend`. It's indeed very tricky to find, I need to compare with the one that has legend on the older version. I have added the test also for the legend spec. Below is a sample report with the test link above.
[mapfish-print-report (49).pdf](https://github.com/geoadmin/web-mapviewer/files/14800874/mapfish-print-report.49.pdf)

ltshb (commented):
One last issue that I've noticed, the scale and the disclaimer is always in german in the PDF, I could not find anything related to this in the print payload, @hansmannj is it something that has been hardcoded on the server side ? 
image

ismailsunni:
@ltshb I guess that one is related to the service-print. There is this file that contains the disclaimer: https://github.com/geoadmin/service-print3/blob/c69347869eb7f284886f5352a64aa8042e5f6979/print-apps/mapviewer/Report.properties

ltshb:
> @ltshb I guess that one is related to the service-print. There is this file that contains the disclaimer: https://github.com/geoadmin/service-print3/blob/c69347869eb7f284886f5352a64aa8042e5f6979/print-apps/mapviewer/Report.properties

@hansmannj how is it to translate the scale and disclaimer on the print ?

ltshb (approved):
Ok from my side this is good to go. The issue with the disclaimer and scale translation can be done later in a separate PR as it might require some backend work.

pakb (approved):
LGTM as we will be dealing with the translation of the footer separately.
Also let's keep in mind we need to revert back to a more "stable" versioning of the geoblocks lib ASAP (as soon as it is published)

PR #730, opened 2024-03-21
PB-166 : use OpenLayers to detect vector features

pakb:
instead of using our own grown functions, it was very hard to detect shapes such as image from a marker (whereas OL has that built-in right away)

We will have to rework how Cesium does it too

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-166_vector_feature_selection_with_ol/index.html)


ltkum (approved):
Everything looks good to me here :)

ltshb (approved):

PR #731, opened 2024-03-22
PB-330 : rectangle selection

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-330_rectangle_selection/index.html)


src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
```diff
@@ -18,13 +17,13 @@ export default function useMapInteractions(map) {
     // NOTE: we cannot use the {constraintResolution: true} as it has zooming issue with some devices and/or os
     const freeMouseWheelInteraction = new MouseWheelZoom()
 
+    const { dragBoxSelect } = useDragBoxSelect()
+
     // Make it possible to rotate the map with ctrl+drag (in addition to OpenLayers default Alt+Shift+Drag).
```
LukasJoss, line 22: This comment should be updated for the new functionality

LukasJoss (commented):
There seems to be a bug with clicking on features in 3d mode.
I get an "invalid coordinate" error

ltshb:
@pakb I did a quick test and it seems that you have a limit when you try to select a lots of feature, which can make sense to have. This limit is also present on the old viewer, however you have in the old viewer a `Load more result` button to load more results.
Do you plan to implement this in a separate PR ?

pakb:
> Do you plan to implement this in a separate PR ?

I forgot this button/feature :see_no_evil: 
Although I think we should think if we want to postpone that to after go-live, because that would require some UI rework (I thought it would be "less-effort" to re-use our current tooltip for now)

I think we should also try to rework this tooltip when more than 4ish features are shown together, by maybe minimizing the content and only leaving the title of each features, or something similar.

ltshb (approved):
This is good for now but I think we should have the limit feature (show more) for the go live, we can discuss the UI tomorrow during refinement.

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -552,24 +552,23 @@ export default {
                 coordinates = proj4(this.projection.epsg, WEBMERCATOR.epsg, featureCoords)
             }
             this.click({
-                clickInfo: new ClickInfo(
+                clickInfo: new ClickInfo({
                     coordinates,
```
ltkum, line 556: For the feature selection bug in 3d : the field in `ClickInfo` is called `coordinate` and not `coordinates`, which means this can't translate it to `coordinate: value`.

Either we change this to `coordinate: coordinates` or we change `coordinates` to `coordinate` everywhere in this file.

src/modules/map/components/cesium/CesiumMap.vue
```diff
@@ -552,24 +552,23 @@ export default {
...
+                }),
                 ...dispatcher,
             })
         },
         onContextMenu(event) {
             const coordinates = this.getCoordinateAtScreenCoordinate(event.clientX, event.clientY)
             this.click({
-                clickInfo: new ClickInfo(
+                clickInfo: new ClickInfo({
                     coordinates,
```
ltkum, line 568: same comment as before

ltkum (changes requested):
As Lukas said, there is a bug when selecting features in 3d. This is due to a nomenclature difference between the cesium ClickInfo constructor call, and the other calls.

I would suggest changing `coordinates` to `coordinate` within `CesiumMap.vue`, to be consistent with the rest of the code.

ltkum (approved):
:+1:

PR #732, opened 2024-03-22
PB-331 : search layer features

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-331_search_layer_features/index.html)


src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -38,11 +36,17 @@ function gotToLayerCategory() {
...
-        locationCategory.value.focusFirstEntry()
-    } else if (results.value.layerResults?.length) {
-        layerCategory.value.focusFirstEntry()
+    if (locationResults.value.length) {
+        gotToLocationCategory()
+    } else if (layerResults.value.length) {
+        gotToLayerCategory()
+    } else if (layerFeatureResults.value.length) {
+        gotToLayerFeaturesCategory()
     }
```
ltshb, line 50: This logic broke the keyboard navigation, I think we should rename the function above `gotToLocationCategory` to `gotToLocationCategoryLastEntry`

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -38,11 +36,17 @@ function gotToLayerCategory() {
...
+    layerFeatureCategory.value.focusFirstEntry()
+}
+
 function focusFirstEntry() {
-    if (results.value.locationResults?.length) {
-        locationCategory.value.focusFirstEntry()
-    } else if (results.value.layerResults?.length) {
-        layerCategory.value.focusFirstEntry()
+    if (locationResults.value.length) {
+        gotToLocationCategory()
```
ltshb, line 45: Currently  `gotToLocationCategory` go to last entry !

src/modules/menu/components/search/SearchResultList.vue
```diff
@@ -65,31 +69,32 @@ defineExpose({ focusFirstEntry })
...
                 />
+                <SearchResultCategory
+                    v-show="layerFeatureResults.length > 0"
+                    ref="layerFeatureResults"
+                    :title="i18n.t('featuresearch_results_header')"
+                    :results="layerFeatureResults"
+                    data-cy="search-results-layer-features"
+                    @first-entry-reached="gotToLayerFeaturesCategory()"
+                />
             </div>
```
ltshb, line 96: Here the idea of the `first-entry-reached` and `last-entry-reached` is for the keyboard navigation so that for the user it can go up and down with the arrow key as if it had only one list, with this modification this is not the case anymore.
You should have something like this
- `first-entry-reached` => go to last entry of previous category or the search input field when no previous category
- `last-entry-reached` => got to next category first entry or nothing if no further category

src/modules/menu/components/search/SearchResultListEntry.vue
```diff
@@ -1,8 +1,103 @@
...
+function goToFirst() {
+    item.value.parentElement.firstElementChild?.focus()
+}
+function goToPrevious() {
+    if (item.value.previousElementSibling) {
+        item.value.previousElementSibling.focus()
+    } else {
+        emits('firstEntryReached')
+    }
+}
```
ltshb, line 57: Code style: blank lines between function declaration

src/modules/menu/components/search/SearchResultListEntry.vue
```diff
@@ -1,8 +1,103 @@
...
+            ...dispatcher,
+        })
+    } else if (entry.value.coordinate) {
+        store.dispatch('setPreviewedPinnedLocation', {
+            coordinates: entry.value.coordinate,
+            ...dispatcher,
+        })
+    }
+}
+function stopResultPreview() {
```
ltshb, line 83: Blank line

tests/cypress/tests-e2e/search/search-results.cy.js
```diff
@@ -293,24 +298,55 @@ describe('Test the search bar result handling', () => {
...
-    })
-    it('shows the results once again if the user clicks back on the search input', () => {
-        cy.goToMapView()
-        cy.get(searchbarSelector).paste('test')
-        cy.wait(`@search-locations`)
+
+        cy.log('Clicking on the first entry to test handling of zoom/extent/position')
+        cy.get('[data-cy="search-result-entry-location"]').first().click()
+        // checking that the view has centered on the feature
+        cy.readStoreValue('state.position.center').then((center) =>
```
ltshb, line 305: replace `then` by `should`

ltshb (changes requested):

src/store/modules/search.store.js
```diff
@@ -117,12 +125,13 @@ const actions = {
                     })
                 }
                 break
-            case RESULT_TYPE.LOCATION:
+            case SearchResultTypes.LOCATION:
+            case SearchResultTypes.FEATURE:
                 if (entry.extent.length === 2) {
                     dispatch('zoomToExtent', { extent: entry.extent, dispatcher })
```
ltkum, line 131: I would add a maxZoom parameter here, especially for features, as some features (for example : public transports) don't have icons defined on the closest zoom levels.

src/store/modules/search.store.js
```diff
@@ -131,22 +140,21 @@ const actions = {
                     })
                 }
                 dispatch('setPinnedLocation', {
-                    coordinates: entry.coordinates,
+                    coordinates: entry.coordinate,
                     dispatcher: dipsatcherSelectResultEntry,
```
ltkum, line 144: In the old mapviewer, getting a feature through the search would select the feature and show its feature Info box. Here we simply zoom to it and add a pin on the map.
Do we want to keep the old mapviewer behavior, and if so, do we plan on doing this at a later date ?

ltkum (commented):
I have a suggestion and a question, but nothing that should block a merge :)

src/store/modules/search.store.js
pakb: That might be a good idea to use `setSelectedFeatures`... I'll look into it

src/store/modules/search.store.js
pakb: There should already be some safeguard in the selection of the zoom level so that it doesn't go overboard, but I see your point.
I just don't know if we can achieve that without hard-coding a value (which I want to avoid at all cost)

ltkum: there is a maxZoom parameter in the 'zoomToExtent', which zoom to the extent in order to see the whole extent, but stops zooming if it's too close. (For example, in the `@features` special param, it sets the maxZoom to zoom level 8, but it will zoom less if needed).

That being said, it is true it's not pretty having a hard coded value.

if we want to simply make sure there are no issues with the layer shown, we could try to check for the max Zoom supported for the layer (for example `https://sys-wmts.dev.bgdi.ch/1.0.0/ch.bav.haltestellen-oev/default/current/2056/28/9363/6240.png` gives us a 400 error, but tells us the max Zoom level supported is 26. I'm fairly sure the 'maxZoom' supported for a layer should exist somewhere and we could use that)

ltshb (approved):

PR #734, opened 2024-03-25
PB-324: Show layer info button without legend

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-324-show-layer-info-button-without-legend/index.html)

### DIFF
--- a/src/api/layers/GeoAdminLayer.class.js
+++ b/src/api/layers/GeoAdminLayer.class.js
@@ -53,6 +53,9 @@ export default class GeoAdminLayer extends AbstractLayer {
      *   metadata) are still loading. Default is `false`
      * @param {LayerTimeConfig | null} [layerData.timeConfig=null] Time series config (if
      *   available). Default is `null`
+     * @param {Boolean} [layerData.hasDescription=true] Define if this layer has a description that
+     *   can be shown to users to explain its content. All internal layer (should) have a
+     *   description. TODO: update the backend API to return this value. Default is `true`
      * @param {Boolean} [layerData.hasLegend=false] Define if this layer has a legend that can be
      *   shown to users to explain its content. Default is `false`
      * @param {Boolean} [layerData.searchable=false] Define if this layer's features can be searched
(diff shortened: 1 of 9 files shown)


src/modules/menu/components/LayerCatalogueItem.vue
```diff
@@ -67,7 +67,7 @@ const activeLayers = computed(() => store.state.layers.activeLayers)
 const openThemesIds = computed(() => store.state.topics.openedTreeThemesIds)
 
 const hasChildren = computed(() => item.value?.layers?.length > 0)
-const hasLegend = computed(() => canBeAddedToTheMap.value && item.value?.hasLegend)
+const hasLegend = computed(() => canBeAddedToTheMap.value && item.value?.hasDescription)
```
LukasJoss, line 70: @pakb is hasDescription a good variable name? In that case I should probably update all the variable names which referer to legend instead of description.

pakb: I can't think of a better name if we want to keep it short... So let's go with that :wink:

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -64,7 +64,7 @@ const id = computed(() => layer.value.id)
 const attributionName = computed(() =>
     layer.value.attributions.map((attribution) => attribution.name).join(', ')
 )
-const showLegendIcon = computed(() => layer.value.hasLegend)
+const showLegendIcon = computed(() => layer.value.hasDescription)
```
pakb, line 67: We could maybe go all the way and rename this to `showLayerDescriptionIcon` and also rename the `LayerLegendPopup` to `LayerDescriptionPopup` so that we are thorough with our renaming

pakb (approved):

src/api/layers/GeoAdminLayer.class.js
```diff
@@ -53,6 +53,9 @@ export default class GeoAdminLayer extends AbstractLayer {
      *   metadata) are still loading. Default is `false`
      * @param {LayerTimeConfig | null} [layerData.timeConfig=null] Time series config (if
      *   available). Default is `null`
+     * @param {Boolean} [layerData.hasDescription=true] Define if this layer has a description that
```
LukasJoss, line 56: @pakb  does there already exist a ticket for this or should this remain hard coded?

PR #736, opened 2024-03-25
Divers UI bugs and improvments

ltshb:
### Fix tippy on long layer entry in external layer and geo catalogue

On the external layer catalogue as well as the geo catalogue, long entry list
where not truncated and no tippy was added.

Also changed this tippy theme to use the default one and put it on the right.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-divers-ui/index.html)


ltkum (approved):
lgtm

PR #658, opened 2024-02-26
PB-111: Send print request

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-111-send-print-request/index.html)

- Create print spec with geoblocks/mapfishprint
  - handling grid
  - handling legend
  - handling qrcode url
  - handling copyright with attributions
  - handling WMTS with different projection (2056, 3857)
- Send the spec with pooling (already in https://github.com/geoblocks/mapfishprint)
- Cancel the printing
- Still using a specific commit of [geoblocks/mapfishprint](https://github.com/geoblocks/mapfishprint), I will update it once the new version of the node package is released

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-111-send-print-request/index.html)

PR #737, opened 2024-03-25
Move "print" job logic and management out of area composable

pakb:
The print area preview composable was doing the preview, the printing and also coffee. It is now only responsible for the preview area. I also took the opportunity to remove from the store things that weren't making much sense being there, as no one else as the print menu section was using them (and as the menu section is now the one triggering the print job, it's much clearer this way)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_split_print_functions/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/usePrint.composable.js
+++ b/src/modules/map/components/openlayers/utils/usePrint.composable.js
@@ -1,0 +1,99 @@
+import { computed, ref } from 'vue'
+import { useStore } from 'vuex'
+
+import { abortPrintJob, createPrintJob, waitForPrintJobCompletion } from '@/api/print.api.js'
+import { getGenerateQRCodeUrl } from '@/api/qrcode.api.js'
+import log from '@/utils/logging'
+
+const dispatcher = { dispatcher: 'usePrint.composable' }
+
+/** @enum */
+export const PrintStatus = {
+    IDLE: 'IDLE',
+    PRINTING: 'PRINTING',
+    FINISHED_SUCCESSFULLY: 'FINISHED_SUCCESSFULLY',
+    FINISHED_FAILED: 'FINISHED_FAILED',
+}
+
+/**
+ * Gathering of all the logic that will trigger and manage a print request to service-print3
+ *
+ * @param {Map} map
+ */
+export function usePrint(map) {
+    const currentJobReference = ref(null)
+    /** @type {PrintStatus} */
+    const printStatus = ref(PrintStatus.IDLE)
+
+    const store = useStore()
+    const printLayout = computed(() => store.state.print.selectedLayout)
+    const printScale = computed(() => store.state.print.selectedScale)
+    const attributions = computed(() =>
+        store.getters.visibleLayers
+            .map((layer) => layer.attributions)
+            .map((attribution) => attribution.name)
+            .filter((attribution, index, self) => self.indexOf(attribution) === index)
+    )
+    const layersWithLegends = computed(() =>
+        store.getters.visibleLayers.filter((layer) => layer.hasLegend)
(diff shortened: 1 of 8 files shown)


ltshb (commented):
This will probably enter in conflict with #727 and it might make more sense that ismail review it

src/store/plugins/loading-bar-management.plugin.js
```diff
@@ -10,7 +10,7 @@ const loadingBarManagementPlugin = (store) => {
         const isLoadingBarVisible = state.ui.showLoadingBar
         // for now, we only check if app is ready and printing status
         // TODO: any pending request to the backend should trigger the loading bar
-        const shouldLoadingBarBeVisible = !state.app.isReady || state.print.printingStatus
+        const shouldLoadingBarBeVisible = !state.app.isReady
```
ismailsunni, line 13: If the printStatus is removed, how will the loading bar be shown?

src/modules/map/components/openlayers/utils/usePrint.composable.js
```diff
@@ -0,0 +1,99 @@
...
+        store.getters.visibleLayers
+            .map((layer) => layer.attributions)
+            .map((attribution) => attribution.name)
+            .filter((attribution, index, self) => self.indexOf(attribution) === index)
+    )
+    const layersWithLegends = computed(() =>
+        store.getters.visibleLayers.filter((layer) => layer.hasLegend)
+    )
+    const lang = computed(() => store.state.i18n.lang)
+    const projection = computed(() => store.state.map.projection)
```
ismailsunni, line 41: For these computed, should we just retrieve once when we call the function to prevent impact on performancce? e.g. only retrieve from store when we call `createPrintJob`?

src/modules/map/components/openlayers/utils/usePrint.composable.js
```diff
@@ -0,0 +1,99 @@
...
+     *   true or false
+     * @returns {Promise<String | null>}
+     */
+    async function print(printGrid = false, printLegend = false) {
+        try {
+            if (currentJobReference.value) {
+                await abortCurrentJob()
+            }
+            printStatus.value = PrintStatus.PRINTING
+            await store.dispatch('generateShortLinks', dispatcher)
```
ismailsunni, line 55: Previously, @ltshb suggested me to call the function directy without saving the result to avoid a dispatch. https://github.com/geoadmin/web-mapviewer/pull/658#discussion_r1505965268

ismailsunni (approved):
Really much better. Thanks!. Only some small questions.

src/modules/map/components/openlayers/utils/usePrint.composable.js
pakb: That is sound, then I'll move the part of the logic that decide to remove geolocation out of the store, in some function in `shortLink.api.js` so that we can re-use it.
I used the dispatch so that this part of the code wasn't duplicated (removing the geolocation URL param and so on)

src/modules/map/components/openlayers/utils/usePrint.composable.js
pakb: That might be a good idea yeah, I'll try and see if it works as intended

src/store/plugins/loading-bar-management.plugin.js
pakb: I should add a todo in the usePrint composable to tackle that.
The idea is that the composable itself will dispatch something (the source it is waiting on) and then subsequently dispatch once more when it is done. The loading bar will be shown whenever a source is not done loading, meaning we will be removing this loading bar management plugin altogether.

PR #738, opened 2024-03-27
Fix import tool local file tabindex

ltshb:
On the import tool local file, the input had a tabindex but the user could
do nothing with its focus because this field is readonly. Moreover due to the
form, when pressing enter on this input field it reloaded the page.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-import-file-tool/index.html)


ismailsunni (approved):
LGTM, tested good.

PR #739, opened 2024-03-27
Use caret instead of plus/minus icons

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-caret-instead-of-plus/index.html)


pakb (commented):
I like this look more than the plus/minus one, I vote for it :+1:

ismailsunni:
I prefer the caret, looks more natural for me. Like a table of content. FYI, here is the comparison to make other easier to compare:
**Caret:**
image
**Plus/Minus**
image

Angi-Kinas:
I also like it better with the caret, since it does the same thing as the other open/close-sections.
I also noticed that the circle/border around the caret is gone when being clicked. I really like it! It looks much better, like with the up/down-arrows.

pakb (approved):

PR #742, opened 2024-03-27
PB-318: refactor time slider to composition API

ltkum:
since we are working on the time Slider implementation, we might as refactor it with the composition API

[Test link](https://sys-map.dev.bgdi.ch/preview/compose-api-timeslider/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -1,4 +1,3 @@
-
 <script setup>
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
@@ -57,10 +56,10 @@ function findMostRecentCommonYear(timeConfigs) {
 const sliderWidth = ref(0)
 const allYears = ref(ALL_YEARS)
 const currentYear = ref(YOUNGEST_YEAR)
-const cursorX = ref(0)
-const playYearsWithData = ref(false)
-const yearCursorIsGrabbed = ref(false)
-const playYearInterval = ref(null)
+let cursorX = 0
+let playYearsWithData = false
+let yearCursorIsGrabbed = false
+let playYearInterval = null
 
 // refs to dom elements
 const yearCursor = ref(undefined)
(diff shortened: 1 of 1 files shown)


src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -1,3 +1,305 @@
...
+}
+
+// dynamic internal data
+const sliderWidth = ref(0)
+const allYears = ref(ALL_YEARS)
+const currentYear = ref(YOUNGEST_YEAR)
+const cursorX = ref(0)
+const playYearsWithData = ref(false)
+const yearCursorIsGrabbed = ref(false)
+const playYearInterval = ref(null)
```
pakb, line 63: I don't think you should store that in a ref, a simple `const` or `let` would do the trick

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -1,3 +1,305 @@
...
+        const [onlyLayerLeft] = newLayers
+        if (!onlyLayerLeft.timeConfig.years.includes(currentYear.value)) {
+            setCurrentYearAndDispatchToStore(onlyLayerLeft.timeConfig.years[0])
+        }
+    }
+})
+
+// we can't watch currentYear and dispatch changes to the store here, otherwise the store gets
+// dispatch too many times when the user is moving the time slider (we wait for mouseup our
+// touchend to commit the change)
```
pakb, line 163: That's good enough, but now that we have some "experience" and some utils to debounce actions, you might want to plan switching to a debounce based store dispatch approach.

Relying on both mouseup and touchend can be a bit tricky on some devices that fire both

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -1,3 +1,305 @@
...
+}
+
+function setSliderWidth() {
+    // 19px of padding (7.5 on both side of the container with p-2 class and 4 with px-1)
+
+    sliderWidth.value = sliderContainer.value.clientWidth - 19 - PLAY_BUTTON_SIZE
+}
+
+function positionNodeLabel(year) {
+    const timestampIndex = ALL_YEARS.indexOf(year) || 1
```
pakb, line 208: if you see this kinds of "default assign" through `||` you can translate them into newer JS with `??` operator instead

pakb (approved):

src/modules/map/components/toolbox/TimeSlider.vue
ltkum: My goal was mostly to make a 1 to 1 translation. I'll keep that in mind when I'm developing the new features, especially with the datas that are distinct.

PR #743, opened 2024-03-27
PB-324: show layer info button without legend rename variables

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-324-show-layer-info-button-without-legend-rename-variables/index.html)


src/modules/menu/components/LayerDescriptionPopup.vue
```diff
@@ -38,47 +38,50 @@ const isExternal = computed(() => layer.value?.isExternal ?? false)
...
 
 watch(layer, async (newLayer) => {
-    htmlContent.value = await getLayerLegend(currentLang.value, newLayer.id)
+    htmlContent.value = await getLayerDescription(currentLang.value, newLayer.id)
 })
 
 watch(layerId, async (newLayerId) => {
-    htmlContent.value = await getLayerLegend(currentLang.value, newLayerId)
+    htmlContent.value = await getLayerDescription(currentLang.value, newLayerId)
 })
```
ltshb, line 46: Here we should also check for non null value of `newLayerId` and for non external layer like in `onMounted`. Although it is not directly related to this PR, but it would be wise to change it, to improve the robustness of the code.

ltshb (approved):
Just a minor improvement, thanks

pakb (approved):

PR #745, opened 2024-03-27
PB-368 : selected features tooltip/infobox rework

pakb:
Group features by layer

show profile for all features with compatible geometry 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-368_selected_features_rework/index.html)


pakb:
that's an already pretty big chunk of changes...
I plan to improve the UI that displays the profile so that it takes less space, but thought I'd better stop for now seeing the size of this PR

ltshb:
There is a double scroll bar on chrome

src/modules/infobox/ZoomToExtentButton.vue
```diff
@@ -0,0 +1,41 @@
+<script setup>
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import { toRefs } from 'vue'
+import { useStore } from 'vuex'
```
ltshb, line 4: This component is very generic and could be moved into src/utils/components

ltshb (commented):
A first quick glance at the code looks all good. Some comment about the UI to discuss in refinements:
1. Perso I would put the profile below the description
2. When opening the profile we should be able to close it without loosing the selection/feature selection
3. the icon for showing the profile is a bit small and many user won't see it find it. Maybe a button `show profile` on bottom of description is more appropriate. I don't think it is needed to have it next to the zoom to extent button. On mobile it might be good to keep the icon button to spare space, but I would put it also below the description, or on bottom right corner of the description.
4. The print button is now not working with the new layout

src/router/storeSync/storeSync.config.js
```diff
@@ -108,15 +108,7 @@ const storeSyncConfig = [
...
-    new SimpleUrlParamConfig(
-        'featureInfo',
-        'setFeatureInfoPosition',
-        'setFeatureInfoPosition',
-        (store) => store.state.ui.featureInfoPosition,
-        false,
-        String,
-        FeatureInfoPositions.NONE
-    ),
+    new FeatureInfoPositionParamConfig(),
```
ltshb, line 111: Does this solve the issue https://jira.swisstopo.ch/browse/PB-358 ? Seems not but if reworking it here maybe have a quick look to see if you can quickly fix it here otherwise this ticket would probably need to wait on this PR to avoid conflict ?

ltshb (commented):
Will continue the review tomorrow

ltshb (commented):
Ok looks good just some minor improvements, see comments in code

src/modules/infobox/components/FeatureListCategoryItem.vue
```diff
@@ -0,0 +1,158 @@
...
+    showContentByDefault: {
+        type: Boolean,
+        default: false,
+    },
+})
+
+const { name, item, showContentByDefault } = toRefs(props)
+
+const content = ref(null)
+const showContent = ref(!!showContentByDefault.value)
```
ltkum, line 35: since we enforce that it should be a `Boolean`, wouldn't it be better to have `showContentByDefault.value`

ltkum (commented):
Couldn't find any problem while testing, and the code seems solid. A small comment on my part in `FeatureListCategoryItem.vue`.

src/modules/infobox/components/FeatureListCategoryItem.vue
pakb: The goal here is to snapshot the value, but then forget about `showContentByDefault` below (if the ref `showContentByDefault` is later changed, `showContent` is unaffected)

src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -74,13 +73,13 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
             )
         }
         super({
+            id,
             name,
             // format coming from https://github.com/geoadmin/web-mapviewer/blob/develop/adr/2021_03_16_url_param_structure.md
             // NOTE we don't differentiate between group of layers and regular WMS layer. This differentiation was not
             // done the legacy parameter and is not required.
-            id: `WMS|${baseUrl}|${externalLayerId}`,
+            urlId: `WMS|${baseUrl}|${id}`,
```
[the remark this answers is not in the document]
pakb: I've reworked this part of the app, now all the layer ID to URL logic is stored in utilities linked to URL param parsing, nothing is present anywhere else in the codebase. Should be a bit cleaner to work with (also covered most of it with some unit test, I'm sure there's still some blind spot, but the basis is there to be built upon if needed)

src/router/storeSync/storeSync.config.js
pakb: was fixed by https://github.com/geoadmin/web-mapviewer/pull/775

src/api/layers/AbstractLayer.class.js
```diff
@@ -45,6 +45,9 @@ export default class AbstractLayer {
      * @param {String} layerData.name Name of this layer in the current lang
      * @param {String} layerData.id The unique ID of this layer that will be used in the URL to
      *   identify it (and also in subsequent backend services for GeoAdmin layers)
+     * @param {String} layerData.urlId The unique ID in the context of the layers URL param, it
```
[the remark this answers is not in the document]
pakb: see comment bellow

ltshb (commented):
There is an overlap between the column in the infobox 
image
- When showing the profile of a geometry I would not do a zoom to extent, for this you have the zoom button
- I would put a bit bigger padding between the caret and the text 
image
- Same here and also the caret seems higher than the text, not aligned
image
- feature selection of external layer doesn't show any title in the infobox 
image

hansmannj (approved):
Very nice work, thanks! :+1:
Did not have a detailed look at the code, but did some testing using the test link.
Did not find issues beside those things, you are already addressing in another PR.
So, LGTM :rocket:

PR #747, opened 2024-03-28
PB-281: Removed print layout enumaration and removed dev site guard on the menu

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-281-print-enum/index.html)


ismailsunni (approved):
I will need to update my e2e test for the print. But I think this PR is good to be merged first.

PR #749, opened 2024-03-28
PB-89: Fix preview of geojson layers

ltshb:


pakb (commented):
Small issue I see with this approach, the `await` means that the GeoJSON will be added as preview layer wherever is the mouse when this finishes.
There's a feature of axios called cancel token that should be leverage here, so that if the mouse goes out of the GeoJSON layer we cancel the pending request.

If you want to see this issue "live", search for "Temperature", quickly go trough "Temperature, 2m min" and end your move on any other layer (that is no GeoJSON) in the searchbar, after 1sec or more, the preview layer will be forced to the GeoJSON instead of the one currently under the mouse

ltshb:
> Small issue I see with this approach, the `await` means that the GeoJSON will be added as preview layer wherever is the mouse when this finishes. There's a feature of axios called cancel token that should be leverage here, so that if the mouse goes out of the GeoJSON layer we cancel the pending request.
> 
> If you want to see this issue "live", search for "Temperature", quickly go trough "Temperature, 2m min" and end your move on any other layer (that is no GeoJSON) in the searchbar, after 1sec or more, the preview layer will be forced to the GeoJSON instead of the one currently under the mouse

Yea I just saw this issue, thanks for the axios tip I'll have a look

src/store/plugins/load-geojson-style-and-data.plugin.js
```diff
@@ -63,29 +64,62 @@ function clearAutoReload(layerId) {
...
+                clone.hasError = true
+                return clone
+            }),
+    }
+}
+
+let pendingPreviewLayer = null
+
+async function loadAndUpdatePreviewLayer(store, layer) {
+    cancelLoadPreviewLayer()
```
pakb, line 104: I'm wondering if we shouldn't check which layer we are loading. Because if we are hover a GeoJSON and then clicking it very quickly, this means this will cancel the initial load of data/style for the layer, and we will fire it once more to show it definitively on the map...
I don't know if this would bring too much complexity here, your thoughts?

pakb (approved):
Small question, but overall looking good as it is

src/store/plugins/load-geojson-style-and-data.plugin.js
ltshb: @pakb good point, however this would add much complexity. Currently the preview layer object and the active layer object are two different object, therefore when the layer is added it reload the geojson data and style for the new object. We could in theory reuse the preview object instead of creating a new object, but I think this would complexify the code and could bring more trouble. If the geojson style and data have proper cache-control header (I check a few meteoswiss layers) then we don't have any performance issue as the second load is done by the browser cache.

PR #753, opened 2024-04-02
PB-362: Improved loading bar which can now be used by other components

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-362-loading-bar/index.html)


src/store/modules/ui.store.js
```diff
@@ -246,11 +253,15 @@ export default {
         setEmbed({ commit }, { embed, dispatcher }) {
             commit('setEmbed', { embed: !!embed, dispatcher })
         },
-        setShowLoadingBar({ commit }, { loading, dispatcher }) {
-            commit('setShowLoadingBar', { loading, dispatcher })
+        setShowLoadingBar({ commit }, { requester, loading, dispatcher }) {
```
pakb, line 256: I don't know if I like this name for what it does.
Maybe something like `addLoadingBarSource` or something in this veins would help us remember that we need to pass more than the `loading` flag (`setShow` translate as, "give me a flag" to me, no necessarily give me a requester or a source that is loading)

Or maybe even split this functionality into two actions :
  - `addLoadingRequester`
  - `addLoadedRequester`

src/store/modules/app.store.js
```diff
@@ -19,8 +19,9 @@ export default {
         setAppIsReady: ({ commit }, { dispatcher }) => {
             commit('setAppIsReady', { dispatcher })
         },
-        mapModuleReady: ({ commit }, { dispatcher }) => {
+        mapModuleReady: ({ commit, dispatch }, { dispatcher }) => {
             commit('mapModuleReady', { dispatcher })
+            dispatch('clearLoadingBar4MapLoading', { dispatcher })
```
pakb, line 24: as you are dispatching something that doesn't belong to this store "module", the best course of action would be to do that in a store plugin.
Maybe the one that already manages the app readiness would be a good candidate?

src/store/modules/ui.store.js
ltshb: @pakb Good point the naming is not 100% good, what do you think of two actions:
- `dispatch('setLoadingBarRequester', {requester: 'printing-map', dispatcher})`
- `dispatch('clearLoadingBarRequester', {requester: 'printing-map', dispatcher})`
?

pakb (approved):

PR #754, opened 2024-04-02
PB-318: feature : timeslider over incomplete data

ltkum:
We : 
 - Allow users to use the time slider across the whole slider bar, even if there is no data
 - Add a specific styling for years with incomplete data

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-318-timeslider-over-incomplete-data/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -350,10 +350,13 @@ function togglePlayYearsWithData() {
 </template>
 
 <style lang="scss">
-@import 'src/scss/webmapviewer-bootstrap-theme';
+@use 'sass:color';
 
+@import 'src/scss/webmapviewer-bootstrap-theme';
+$time-slider-color-background: color.adjust($white, $alpha: -0.1);
+$time-slider-color-has-data: color.adjust($primary, $lightness: 30%);
 .time-slider {
-    background: rgba(255, 255, 255, 0.9) !important;
+    background: $time-slider-color-background !important;
     &:not(.grabbed) &-bar-cursor {
         cursor: grab;
     }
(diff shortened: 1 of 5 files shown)


src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -273,30 +233,52 @@ function releaseCursor() {
...
+function timeSliderBarInnerStepClassObject(year, index) {
+    const classObject = {
+        'has-joint-data': yearsWithData.value.yearsJoint.includes(year),
+        'big-tick': year % 50 === 0,
+        'medium-tick': year % 25 === 0,
+        'small-tick': year % 5 === 0,
+    }
+    classObject[`has-separate-data-${index % 2}`] = yearsWithData.value.yearsSeparate.includes(year)
+    return classObject
+}
```
pakb, line 281: I'd prefer we keep CSS/HTML classes defined in the HTML template if possible
Here I see that you moved that here to be able to use the `index % 2` inside the function, maybe switch to a something like
```
{
// ...
'separate-data': yearsWithData.yearsSeparate.includes(year),
'odd': index % 2
}
```
and then in the (S)CSS
```
.separate-data {
    background: // ...
    &.odd {
       background: // ... will override above rule
    }
}
```

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -150,4 +153,8 @@ function hidePopover() {
         white-space: nowrap;
     }
 }
+
+.w-13 {
+    width: 13%;
+}
```
pakb, line 159: you could move that to the bootstrap theme SCSS file, so that it will be available to the whole app

pakb (approved):
as soon as you have found the "right time" so that the CI drawing test want to pass, all good to go :upside_down_face: 
(we really need to fix this flakiness...)

PR #755, opened 2024-04-02
PB-295: CatalogueNodes parameter is now persistent

ltshb:
When opening catalogue nodes they are persistent upon reload which allows
to create permalink with pre-opened catalogue nodes.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-295-catalog-nodes/index.html)


src/store/plugins/topic-change-management.plugin.js
```diff
@@ -92,13 +92,32 @@ const topicChangeManagementPlugin = (store) => {
                     layers: topicTree.layers,
                     ...dispatcher,
                 })
-                // checking that no values were set in the URL at app startup, otherwise we might
-                // overwrite them here
-                if (store.state.topics.openedTreeThemesIds.length === 0) {
+                // Here we have different behavior possible
+                if (
+                    mutation.payload.dispatcher === 'MenuTopicSection.vue' ||
```
pakb, line 97: I'm not sure it's a good idea to use dispatcher outside of debug purposes (I've seen it's already twice the case in the code though)
Is there any other way to know the state you want to achieve other than using the dispatcher name?

If not, might be a good idea to export the dispatcher name as a constant instead of hard-coding it, might lead to bad refactor behavior in the future (and if a circular dep is created by having this const declared in `MenuTopicSection.vue` we should maybe consider having a "config file" exporting any dispatcher name used across the app outside of debug)

pakb (commented):
I'm not sure we shouldn't close children category when closing the parent, if you open a nested category and close the parent, the nested ID stays in the URL (which doesn't make much sense on its own)

ltshb:
@pakb I first also thought of closing all children when closing a parent, but the old viewer does not and I think this might be more anoying for user as helpful. A user might want to close one parent without loosing the open children (toggling the parent). We might consider in future a folding all button to achieve this.

src/store/plugins/topic-change-management.plugin.js
ltshb: To be honest that was the goal of the dispatcher, that we can make decision based on the dispatcher:
- ignoring self dispatching
- different action based on who is dispatching:
  - dispatched by url parser
  - dispatched by user action (button click)

In my opinion this goal was the trigger of the dispatcher changes and not the deguging purpose which is also nice of course.

I would also not put the dispatcher in an export variable as we will have a lots of dependencies in code. A store plugin should be able to run without the presence of a vue component (e.g. in embed mode you want the topic plugin without the code of the menu, the topic can only be changed by the url). I find using a text as dispatcher as we defined it a nice feature, of course by refactoring we need to change this text everywhere which is very simple with a search and replace, we just need not to forgot about it. In refactoring we need anyway to do a search and replace for every file import (as the dispatcher has the file name it works similar to file import).

pakb (approved):

PR #761, opened 2024-04-04
PB-360: Remove circular reference and unused properties from KML layer.

ismailsunni:


ltshb (commented):
I think it would be better to use the Customizer to remove those properties, see https://geoblocks.github.io/mapfishprint/docs/classes/BaseCustomizer.default.html#feature

src/api/print.api.js
```diff
@@ -39,6 +39,20 @@ class GeoAdminCustomizer extends BaseCustomizer {
...
+     * can cause issues with mapfishprint
+     *
+     * @param {State} layerState
+     * @param {GeoJSONFeature} feature Manipulated feature
+     */
+    feature(layerState, feature) {
+        // cause circular reference issues
+        delete feature.properties.geodesic
+        // unnecessary properties for printing and cause mapfishprint to throw an error
+        delete feature.properties.editableFeature
```
ltshb, line 54: This could lead to a type error if `feature.properties` does not exists, maybe use `delete feature.properties?.geodesic`

ltshb (commented):
I've got failures in the console when printing a kml with a measure, 
image
And also they was no error feedback, so without opening the console we could not see any error; You can use this shortlink to reproduce the issue

ltshb:
@ismailsunni I allow myself to rebase and if it works I'll merge this.

ltshb (approved):
Code looks good, however I found another issue (that could be fixed in a separate PR). For the KML measure, the total distance is not printed, like on the old viewer:
- old viewer
image

- new viewer
image

ltshb:
Also on the old viewer the drawing looks better (smaller lines) as on the new one that have thicker lines bolder text

PR #762, opened 2024-04-04
Fix GPX local file import

ltshb:
Some GPX file might not have metadata and it is ok to have it null. Due to this GPX file without metadata could not be loaded correctly. When no metadata was present, even if the file was a local file, the load gpx data plugin was triggered. On a deployment we could see an error on HTTP where the plugin tried to load GPX file on https://sys-map.dev.bgdi.ch/FILE_NAME while on localhost it tried to do on localhost which gave a differant answer.
    
So this triggered some weird bug. Now we allow GPX without metadata and only trigger the GPX load plugin when no data is present regardless of the metadata
    
Also make sure to set the layer error in case of GPX load failure

Also Avoid unnecessary store dispatch in layers plugins

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-gpx-local/index.html)

### DIFF
--- a/src/api/layers/GPXLayer.class.js
+++ b/src/api/layers/GPXLayer.class.js
@@ -46,7 +46,7 @@ export default class GPXLayer extends AbstractLayer {
             hasDescription: false,
             hasLegend: false,
             isExternal: true,
-            isLoading: !!gpxData && !!gpxMetadata,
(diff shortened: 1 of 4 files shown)


ismailsunni (approved):
LGTM, tested locally :+1:

PR #765, opened 2024-04-05
PB-376: Re-organized the settings menu

ltshb:
Now all component related to the settings menu are under the same folder and
also renamed some components.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-376-menu-settings/index.html)


src/modules/menu/components/settings/MoreInfo.vue
```diff
@@ -0,0 +1,31 @@
...
+function openCmsLink() {
+    window.open(cmsUrl.value, '_blank', 'noreferrer')
+}
+</script>
+
+<template>
+    <HeaderLink v-if="isDesktopMode" @click="openCmsLink">
+        {{ $t('cms_link_button_title') }}
+    </HeaderLink>
+    <button
```
pakb, line 23: image

there's a little gap missing between the two "groups" of button (same as the one between the two button of the first line)
Maybe we should place this button inside the same "container" as the two other button, so that if there's enough space (depending on device and selected lang) they can be all three displayed on the same line

pakb (approved):

PR #766, opened 2024-04-08
PB-390, PB-391: Change the pdf print file name and fixed print lang parameter

ltshb:
Now we set the pdf file name to HOSTNAME_DATETIME.pdf

Also fixed lang parameter that was not sent to print server. This parameter is
used to translate the disclaimer see also PB-383.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-390-file-name-pdf/index.html)


hansmannj (approved):
Thanks, LGTM :+1:

hansmannj (commented):
Just FYI:
- `config.lang`: Not sure, if that could ever be the case, but in case it is `null`, the backend would default to English.
- `config.attributions`: The backend defaults to `(c) swisstopo`, in case no attributions are passed in the print request's payload.

PR #769, opened 2024-04-08
PB-398: Improve print UI user feedback

ltshb:
Now provide a success, aborted or failure feedback

NOTE currently the failure feedback can be tested using a KML layer with a measure which is currently failing.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-398-print-error-mgmt/index.html)


src/modules/i18n/locales/de.json
```diff
@@ -653,5 +653,9 @@
     "kml_gpx_file_empty": "Die KML/GPX-Datei ist leer",
     "large_size": "Gross",
     "extra_large_size": "Extra Gross",
-    "duplicate_layer": "Karte duplizieren"
+    "duplicate_layer": "Karte duplizieren",
+    "feedback_empty_warning": "Die Berichtsnachricht darf nicht leer sein",
+    "operation_successful": "Erfolg! ",
```
hansmannj, line 658: Is the extra whitespace at the end intentional? Or might it break some formatting?

hansmannj (approved):
Thanks, LGTM! :1st_place_medal:

src/modules/i18n/locales/de.json
ltkum: I believe this might be a whitespace that was unseen in the google sheets.

ltkum (approved):
I would remove the extra whitespace, but apart from that LGTM and everything seems to be working as expected.

PR #772, opened 2024-04-09
PB-397: Revert back print backend to correct staging on DEV

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-397-tempo-print-int-backend-revert/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -7,7 +7,7 @@ import {
 } from '@geoblocks/mapfishprint'
 import axios from 'axios'
 
-import { API_BASE_URL, WMS_BASE_URL } from '@/config'
+import { API_BASE_URL, API_SERVICES_BASE_URL, WMS_BASE_URL } from '@/config'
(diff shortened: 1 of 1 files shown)


hansmannj (commented):
Just noticed something:
service-print3 seems to always have some kind of `default` setting. Not sure, if it can be deleted/suppressed. Maybe @marionb knows?

In our case, `default` seems to match our specific config, as there is only one. But it might be better ot explicilty use our configuration. Hence the endpoint would be:
`/print3/print/mapviewer/` instead of `/print3/print/default/`

(also see: https://github.com/geoadmin/service-print3/tree/develop/print-apps/mapviewer)

ltshb:
The DEV staging seems to work again, no idea why it hangs in the past needs to keep an eye on it. I also enabled the e2e tests for print see https://github.com/geoadmin/infra-e2e-tests/pull/279

LukasJoss (approved):

PR #773, opened 2024-04-09
PB-368 : fluid UI rework

pakb:
Redo the UI from the ground up with the idea that a component shouldn't place itself in the "grid", but simply display what it has to show. Its parent is the one responsible to place it at the correct position.

The map now takes up the space available between the header and the footer (or between the header and the infobox, if the infobox is shown). Meaning it won't be "hidden" under both of these UI component anymore (even though it was fancy with some transparency).
This should greatly improve the experience of "Zoom to extent" on a selected feature, because now the feature will be center in the map viewport in any case.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-368_infobox_ui_rework/index.html)


ltshb (approved):
Review will be done in #808

pakb:
Diff has been saved as "snapshot" in this PR for further review
https://github.com/geoadmin/web-mapviewer/pull/808

tests/cypress/support/commands.js
```diff
@@ -686,10 +643,10 @@ Cypress.Commands.add('checkOlLayer', (args = null) => {
             if (!l.id) {
                 throw new Error(`Invalid layer object ${l}: don't have an id`)
             }
-            if (l.visible == undefined) {
+            if (l.visible === undefined) {
```
ltshb, line 646: Here I would not put a `===` because it won't catch `null`, same below for opacity

tests/cypress/support/commands.js
```diff
@@ -740,3 +721,63 @@ Cypress.Commands.add('checkOlLayer', (args = null) => {
...
+                '), cannot click on these coordinate'
+            )
+            cy.wrap(false).should('be.true')
+        } else {
+            cy.log('Clicking at coordinate', coordinate, 'on pixel', pixel, 'with options', options)
+            if (doubleClick) {
+                cy.get('[data-cy="ol-map"]').dblclick(pixel[0], pixel[1])
+            } else {
+                cy.get('[data-cy="ol-map"]').click(pixel[0], pixel[1])
+            }
```
ltshb, line 780: Here the `:visible` selector should be added. Also this is kind of dangerous, as the computed pixel might be out of the ol map viewport !
We should move the map to the center of the coordinate before by modifying the url center, then you can simply do a click('center')

tests/cypress/support/commands.js
```diff
@@ -740,3 +721,63 @@ Cypress.Commands.add('checkOlLayer', (args = null) => {
...
+            cy.waitUntil(() => cesiumlayer.ready, {
+                description: `[${layer.id}] waitUntil layer.ready`,
+                errorMsg: `[${layer.id}] layer.ready is not true`,
+            })
+        })
+    })
+})
+
+Cypress.Commands.add('clickOlMapAtCoordinate', (coordinate, options = {}) => {
+    const { doubleClick = false } = options
```
ltshb, line 762: The options should be passed to the click

tests/cypress/support/commands.js
pakb: As discussed, I've switch to a logic summarized by "If there's no `visible` or `opacity` key set to the JSON object, I set some default values"

src/modules/map/components/toolbox/ZoomInButton.vue
```diff
@@ -0,0 +1,29 @@
+<script setup>
+import { useStore } from 'vuex'
+
+import { useTippyTooltip } from '@/utils/useTippyTooltip'
+
+const dispatcher = { dispatcher: 'ZoomOutButton.vue' }
```
ismailsunni, line 6: Should it be `ZoomInButton.vue`?

PR #774, opened 2024-04-09
PB-320 : Add timeSlider URL param

ltkum:
includes PB-319 : CSS changes to time selector when time slider is active and the layer's year is different than the preview year

Add the timeSlider URL parameter which allow to activate the time slider on startup
there should be at least one time enabled layer
the year should be between 1844 and the current year -1
Also store the preview year and ties the timeSlider activation to a boolean, rather than to the timestamp

also includes PB-321 : legacy support and tests


src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -129,8 +129,16 @@ function hidePopover() {
...
+                    'btn-primary': isTimeSliderActive
+                        ? previewYear === timeEntry.year
+                        : timeConfig.currentTimestamp === timeEntry.timestamp,
+                    'btn-outline-primary':
+                        isTimeSliderActive &&
+                        timeConfig.currentTimestamp === timeEntry.timestamp &&
+                        timeEntry.year !== previewYear,
+                    'btn-light':
+                        timeEntry.timestamp !== timeConfig.currentTimestamp &&
+                        (!isTimeSliderActive || previewYear !== timeEntry.year),
```
ltshb, line 141: This should be simplified with a computed value for each 3 use cases. Something like this

```javascript
'btn-primary': isSelected,
'btn-outline-primary': isSecondarySelected,
'btn-light': !isSelecte && !isSecondarySelected
```

However I'm not sure anymore did we not talked that the timeslider would also changes the time configuration of all visible layers with time config instead of a preview ?

tests/cypress/tests-e2e/legacyParamImport.cy.js
```diff
@@ -450,7 +451,16 @@ describe('Test on legacy param import', () => {
             cy.get('[data-cy="compareSlider"]').should('be.visible')
         })
     })
-
+    it.skip('should show the time slider on startup when setting it in the URL', () => {
```
ltshb, line 454: Why skipping here ?

tests/cypress/tests-e2e/timeSlider.cy.js
```diff
@@ -30,16 +30,18 @@ describe('Testing the time slider', () => {
...
         })
-        it('should change the year if the user drags the tooltip on the right with the mouse', () => {
+        it('should show the slider on startup when setting it in the URL, and change the year if the user drags the tooltip on the right with the mouse', () => {
+            cy.goToMapView({
+                layers: `test.timeenabled.wmts.layer`,
+                timeSlider: preSelectedYear,
+            })
             moveSlider(200)
             cy.get('[data-cy="time-slider-current-year"]').should('not.contain', preSelectedYear)
         })
```
ltshb, line 47: This should be/could be merged together. Also you should add more test case in e2e tests for the timeslider. Same when moving the slider we should make sure that the time is in the active layer list selected, and that non visible layer are not changed. Also make sure that the timeSlider url param has been changed, etc..

ltshb (commented):
When changing the time slider with the ch.astra.unfaelle-personenschaeden_getoetete layer set to invisible, in the timeslider it seems only the other time layer are visible but moving the slider change the time of all layer. In my opinion the slider should only change the visible layers, like it does on old viewer.
For this PR mimic the same behavior as the old viewer, which is fine for this one and once the comments are done can be merged.
However if I remember well we decided to change a bit the behavior, where the timeslider would change the configuration (removing the preview year) so we could either configure each layer independently or all together with the slider.

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
ltkum: Hey, I discussed this quickly with Pascal. we did indeed decide to alter the behavior of the time slider to have it change the years to all layers, but this change was not reflected in the tickets written after the discussion. We're keeping this for the post go live.

We made a [ticket](https://jira.swisstopo.ch/browse/PB-411) to discuss what we want to do with the time slider after the go live. Changing the time of layers is one of them, but we had a few suggestions and ideas that could be nice to discuss and refine.

On another note : the computed approach would make this easier to read. I'll change it.

src/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue
```diff
@@ -40,9 +40,12 @@ const timeSelectorModal = ref(null)
 
 const previewYear = computed(() => store.state.layers.previewYear)
 const hasMultipleTimestamps = computed(() => timeConfig.value.timeEntries.length > 1)
-
+const isTimeSliderActive = computed(() => store.state.ui.isTimeSliderActive)
+const isLayerVisible = computed(
+    () => !!store.getters.visibleLayers.find((layer) => layer.id === layerId.value)
+)
```
ltshb, line 46: This don't work as expected you can have duplicate layer don't search by id but by index.
The simplest is 
```javascript
const isLayerVisible = computed(() => store.state.layers.activeLayers[layerIndex.value].visible)
```

ltshb: This breaks if you have a duplicate layer, one which is visible and the other one not.

ltshb: Would be good to cover this in the e2e tests

ltkum: those are all good suggestions, and it should be done :)

ltshb (approved):

PR #775, opened 2024-04-09
PB-358: Fix Complex geometry feature infobox default display

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-358-complex-geometry-feature-infobox-default-display/index.html)


ltkum (approved):
This works as expected :+1:

PR #776, opened 2024-04-09
PB-396: Make search result dropdown disappear when going to location

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-396-make-search-result-dropdown-disappear/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchBar.vue
+++ b/src/modules/menu/components/search/SearchBar.vue
@@ -68,7 +68,6 @@ const clearSearchQuery = () => {
 
 const closeSearchResults = () => {
     showResults.value = false
-    searchInput.value.focus()
 }
 
 const goToFirstResult = () => {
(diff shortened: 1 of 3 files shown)


ltkum (approved):
LGTM

src/modules/menu/components/search/SearchBar.vue
```diff
@@ -68,7 +68,6 @@ const clearSearchQuery = () => {
 
 const closeSearchResults = () => {
     showResults.value = false
-    searchInput.value.focus()
```
ltshb, line 71: @LukasJoss this should not have been removed, for user that use the keyboard if they go to a loaction the focus should stay on the search input so that they cans still change the search result

ltshb (commented):
This broke the keyboard navigation

src/modules/menu/components/search/SearchBar.vue
LukasJoss: Currently focusing the search bar while containing text immediately opens the dropdown again, hiding the position. What approach would you suggest to fix that (or should it be a post prod ticket since the current implementation is as in the old viewer?)

PR #779, opened 2024-04-10
Revert "PB-296: Add failOnError to all layers"

ltshb:
Reverts geoadmin/web-mapviewer#768 due to mapfish print bug https://github.com/mapfish/mapfish-print/issues/3255

[Test link](https://sys-map.dev.bgdi.ch/preview/revert-768-feat-pb-296-print-flag/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -180,23 +180,6 @@ export class PrintError extends Error {
 }
 
 /**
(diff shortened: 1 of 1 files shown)


hansmannj (approved):
:+1: Probably we can re-think this and in mid-term find a nice way to inform the user, if one of multiple layers is not printed.

PR #781, opened 2024-04-11
PB-404 bugfix: removing error on drawing selection

ltkum:
Issue : When selecting a drawing feature outside of the drawing mode, an error log would appear in the console explaining there were no coordinates.

Second Issue : when creating a drawing feature, an error log would appear in the console log explaining the same issue

Fix : We use 'getCoordinates()' instead of 'coordinates' in the map interactions composable.
Second Fix : when creating a Feature with no coordinates (undefined), we set them to null in the constructor.
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-404-log-error-on-selecting-drawing-feature/index.html)

### DIFF
--- a/src/api/features/SelectableFeature.class.js
+++ b/src/api/features/SelectableFeature.class.js
@@ -42,7 +42,9 @@ export default class SelectableFeature extends EventEmitter {
         } = featureData
         this.id = id
         // using the setter for coordinate (see below)
-        this.coordinates = coordinates
+
+        // if coordinates are not defined, we set them to null
+        this.coordinates = coordinates ? coordinates : null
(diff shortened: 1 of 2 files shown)


LukasJoss (commented):
There is another instance where this bug occurs, when you go to drawing mode and place a marker:
When dragging the marker on the map in useDrawingModeInteraction.composable.js in the function onAddFeature there was
`args.coordinates = null` in a earlier version of the code, now it is undefined and thus fails.

ltshb (commented):
Same comment as Lukas

LukasJoss (approved):

PR #782, opened 2024-04-12
PB-424: Time slider year entry button

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-424-time-slider-year-entry-button/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -304,10 +304,12 @@ function setYearToInputIfValid() {
                     data-cy="times-slider-cursor"
                     class="time-slider-bar-cursor py-1 user-select-none d-flex gap-1 bg-body border rounded"
                     :style="cursorPosition"
-                    @touchstart.passive="grabCursor"
-                    @mousedown.passive="grabCursor"
                 >
-                    <div class="px-2 border-end d-flex align-items-center">
+                    <div
+                        class="px-2 border-end d-flex align-items-center"
+                        @touchstart.passive="grabCursor"
+                        @mousedown.passive="grabCursor"
+                    >
                         <FontAwesomeIcon icon="grip-lines-vertical" />
                     </div>
                     <input
@@ -320,7 +322,11 @@ function setYearToInputIfValid() {
                         onkeypress="return event.charCode >= 48 && event.charCode <= 57"
                         @input="setYearToInputIfValid"
                     />
-                    <div class="px-2 border-start d-flex align-items-center">
+                    <div
+                        class="px-2 border-start d-flex align-items-center"
+                        @touchstart.passive="grabCursor"
+                        @mousedown.passive="grabCursor"
+                    >
                         <FontAwesomeIcon icon="grip-lines-vertical" />
                     </div>
                 </div>
@@ -396,7 +402,7 @@ $time-slider-color-has-data: color.adjust($primary, $lightness: 30%);
             position: absolute;
             top: 0.75 * $spacer;
             height: $cursor-height;
-            width: 92px;
+            width: 100px;
             &-year {
                 position: relative;
                 top: 1px;
@@ -488,13 +494,13 @@ $time-slider-color-has-data: color.adjust($primary, $lightness: 30%);
     resize: none;
 }
 .form-control.time-slider-bar-cursor-year {
-    padding: 0 !important;
+    padding: 3px !important;
 }
 .form-control.is-invalid.time-slider-bar-cursor-year {
-    background-size: 0 !important;
+    background-size: 0;
 }
 .form-control:focus.time-slider-bar-cursor-year {
-    outline: none !important;
-    box-shadow: none !important;
+    outline: none;
+    box-shadow: none;
 }
 </style>
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -283,8 +283,7 @@ function setYearToInputIfValid() {
         currentYear.value != displayedYear.value &&
         allYears.value.includes(parseInt(displayedYear.value))
     ) {
-        currentYear.value = parseInt(displayedYear.value)
-        setCurrentYearAndDispatchToStore(currentYear.value)
+        setCurrentYearAndDispatchToStore(parseInt(displayedYear.value))
     }
 }
 </script>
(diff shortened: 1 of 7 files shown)


src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -291,15 +304,21 @@ function togglePlayYearsWithData() {
...
-                        {{ currentYear }}
-                    </div>
+                    <input
+                        v-model="displayedYear"
+                        data-cy="time-slider-bar-cursor-year"
+                        maxlength="4"
+                        class="time-slider-bar-cursor-year"
+                        type="text"
+                        @input="setYearToInputIfValid"
+                    />
```
ltshb, line 314: The input should only allow numbers and a valid year between the min an max defined. If an invalid year is provided the input box should be marked as invalid in red using the bootstrap validation (see https://getbootstrap.com/docs/5.0/forms/validation/ and some example in the code).
Pressing enter should validate the year and exit the input editing.
On the year input we should not allow to grab and move the cursor as it make almost impossible to select the year by click and select.

ltshb (changes requested):

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -268,6 +272,15 @@ function togglePlayYearsWithData() {
         playYearInterval = null
     }
 }
+function setYearToInputIfValid() {
+    if (
+        currentYear.value != displayedYear.value &&
+        allYears.value.includes(parseInt(displayedYear.value))
+    ) {
+        currentYear.value = parseInt(displayedYear.value)
+        setCurrentYearAndDispatchToStore(currentYear.value)
```
ltkum, line 281: currentYear is already set within 'setCurrentYearAndDispatchToStore', there is no need to assign its value here.

ltkum (commented):
I was not so sure about the watcher, but in the end, it's a durable approach if we start modifying currentYear in multiple part of the file.

After testing a bit (and checking the legacy viewer), there is something that's not very user friendly : when selecting the input field and keeping the mouse pressed, we both move the time slider and the selection within the input field. 
On the old mapviewer, only the sides of the time slider cursor allowed it to move, and the middle was only an input field. I would also keep it like that : 
clicking on the input field : does not call the 'onMouseDown' event which allows the time slider to move and to set a year.
clicking on the sides : same behavior as now.

this would also reduce a bit the commits sent by the timeSlider

@ltshb what do you think ? (we can totally do this in a second time)

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -452,4 +485,18 @@ $time-slider-color-has-data: color.adjust($primary, $lightness: 30%);
         text-align: center;
     }
 }
+.time-slider-bar-cursor-year {
+    &.form-control {
```
LukasJoss, line 489: @ltshb I opted to only show a red border in case of an invalid input since the space is very limited, should there be something more like a popup?

src/modules/map/components/toolbox/TimeSlider.vue
LukasJoss: Is there a specific reason why the year should only be evaluated after pressing enter? I personally prefer if it auto validates after the input  contains four characters

src/modules/map/components/toolbox/TimeSlider.vue
ltshb: You could simply add a tippy like this

```diff
diff --git a/src/modules/map/components/toolbox/TimeSlider.vue b/src/modules/map/components/toolbox/TimeSlider.vue
index 6321cc9c..e13a8875 100644
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -4,9 +4,15 @@ import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
 import { useStore } from 'vuex'
 
 import { round } from '@/utils/numberUtils'
+import { useTippyTooltip } from '@/utils/useTippyTooltip'
 
 const dispatcher = { dispatcher: 'TimeSlider.vue' }
 
+const { refreshTippyAttachment } = useTippyTooltip(
+    '.time-slider-bar-cursor-year[data-tippy-content]',
+    { theme: 'danger' }
+)
+
 /**
  * The oldest year in our system is from the layer Journey Through Time (ch.swisstopo.zeitreihen)
  * which has data from the year 1844
@@ -284,6 +290,8 @@ function setYearToInputIfValid() {
         allYears.value.includes(parseInt(displayedYear.value))
     ) {
         setCurrentYearAndDispatchToStore(parseInt(displayedYear.value))
+    } else {
+        refreshTippyAttachment()
     }
 }
 </script>
@@ -316,6 +324,7 @@ function setYearToInputIfValid() {
                         v-model="displayedYear"
                         class="form-control time-slider-bar-cursor-year"
                         :class="{ 'is-invalid': invalidYear }"
+                        :data-tippy-content="invalidYear ? 'invalid_slider_year' : null"
                         data-cy="time-slider-bar-cursor-year"
                         maxlength="4"
                         type="text"
```

You need to create a new translation key for the invalid year, the content should be kind of: `Year should be between $X and $Y` and you could then replace X and Y by the min max year configured.

LukasJoss: This will probably look nicer than using the form-control native tooltip

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -152,11 +175,26 @@ onMounted(() => {
...
+            '{' +
+            ALL_YEARS[0] +
+            '-' +
+            ALL_YEARS[ALL_YEARS.length - 1] +
+            '}',
+        arrow: true,
+        hideOnClick: false,
+        placement: 'bottom',
+        trigger: 'manual',
+        theme: 'danger',
```
ltshb, line 190: You need to manage the content also when the language changes

ltshb (commented):
Code is good just about the language changes in the tippy. Also the translation needs to be updated, you can send me the translation by slack and I add them to the spreadsheet.

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -152,13 +177,37 @@ onMounted(() => {
...
 })
 
+function tooltipContent() {
+    return (
+        i18n.t('outside_valid_year_range') +
+        '{' +
+        ALL_YEARS[0] +
+        '-' +
+        ALL_YEARS[ALL_YEARS.length - 1] +
+        '}'
```
ltshb, line 203: Perso I would not use the `{}` simply

```javascript
return `${i18n.t('outside_valid_year_range')} ${ALL_YEARS[0]}-${ALL_YEARS[ALL_YEARS.length - 1]}`
```

The output will look nicer less coding.

ltshb (approved):
Just a slight tooltip formating improvment

PR #786, opened 2024-04-15
PB-396: Updated search bar e2e tests

ltshb:
When clicking on a search bar result make sure to set the focus back on the search bar, so the user can quickly edit the search if needed.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-396-search-focus/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchBar.vue
+++ b/src/modules/menu/components/search/SearchBar.vue
@@ -52,8 +52,12 @@ const updateSearchQuery = (event) => {
     }, 100)
 }
 
-const onSearchInputFocus = () => {
-    if (hasResults.value) {
+const onSearchInputFocus = (event) => {
+    // When the focus event is due to a programatic focus event, the relatedTarget is not null
+    // and in this case we don't want to show the result. For example when selecting a result value
+    // we want to close the result and focus on the input, so that the user can directly change
+    // the search.
+    if (!event.relatedTarget && hasResults.value) {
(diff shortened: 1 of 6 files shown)


src/modules/menu/components/search/SearchBar.vue
```diff
@@ -130,7 +141,8 @@ const focusSearchInput = () => {
             @input="updateSearchQuery"
             @focus="onSearchInputFocus"
             @keydown.down.prevent="goToFirstResult"
-            @keydown.esc.prevent="closeSearchResults"
+            @keydown.esc.prevent="toggleResults"
```
LukasJoss, line 144: The escape toggle is only working when focus is on the search bar and not when it is focused on a search bar.

LukasJoss (approved):

src/modules/menu/components/search/SearchBar.vue
ltshb: Yes I know but this would make the code a bit more complicated as we would need to pass the event up a few hierarchy and we would also need to correctly manage focus.

ltshb: @LukasJoss  Actually on old viewer the ESC was clearing the input which make more sense what do you think ?

LukasJoss: I prefer the toggle but its up to you

PR #789, opened 2024-04-17
PB-360: Fix GPX layer printing issue

ismailsunni:
## Issue
- If we upload GPX layer with a point, it will be styled as white circle (should be red with black border as in the old viewer)
- If the uploaded GPX layer has two different geometry, it will be styled with the same style (white circle)

### The issue
Shown in the viewer:
image
Print result:
image

### After Fix
Shown in the viewer:
image
Print result:
image

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-360-gpx-print-style/index.html)


src/utils/styleUtils.js
```diff
@@ -50,7 +50,7 @@ export const pointStyle = {
 export const circleStyle = new Circle({
     ...pointStyle,
     fill: new Fill({
-        color: hexToRgba(white),
+        color: hexToRgba(red),
```
ltshb, line 53: This broke the style of the drawing feature selection which is white and should stay white. I suggest to create new redCircleStyle and rename this to whiteCircleStyle

src/utils/gpxUtils.js
```diff
@@ -41,7 +41,7 @@ export function parseGpx(gpxData, projection) {
         featureProjection: projection.epsg,
     })
     features.forEach((feature) => {
-        feature.setStyle(gpxStyle)
+        feature.setStyle(gpxStyles[feature.getGeometry().getType()] || gpxStyle)
```
ltshb, line 44: I would remove the fallback and make sure that the `gpxStyles` contain all geometry types. Then I would also remove the `gpxStyle` for better code clarity

ltshb (approved):

PR #791, opened 2024-04-18
Pb 378 temporary block profiles too big

ltkum:
Issue : our backend might not be able to process profiles request that have too many points. In the future, we will improve the service to handle it, but for now we need to avoid hitting the service with unnecessary queries.

Fix : When the number of coordinates in the profile is higher than a certain number set in the configuration, we bypass the service alti calls and act as if there was no data. We also log an error saying the request was temporarily blocked because of its size.

small TODO : we need to discuss to see what is the maximum coordinates the service can currently handle.

The limit has been set to 10000 here for now.


ltshb (commented):
See ticket comment https://jira.swisstopo.ch/browse/PB-378, to be discussed next week in refinement

PR #792, opened 2024-04-18
PB-418: bugfix opacity in WMS layers

ltkum:
Issue : when modyfing the opacity of an external WMS layer and setting it to 0, it would instead revert to its default config value (1 in most cases)

Cause : Opacity of a layer could be set by its parent (parentLayerOpacity). When this was set to 0, the computed check `parentLayerOpacity.value || wmsLayerConfig.value.opacity` would consider the 0 to be a 'false' and use the config value instead.

Fix : We switch the || operator to the ?? operator, which will only use the config value if the parentLayerOpacity is null. We did this change in all layers type to be on the safe side.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-418-bugfix-transparency-external-layers/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/layers.cy.js
+++ b/tests/cypress/tests-e2e/layers.cy.js
@@ -282,6 +282,36 @@ describe('Test of layer handling', () => {
                         WEBMERCATOR.epsg,
                     ])
                 })
+                cy.openMenuIfMobile()
+                // we play with the transparency to ensure nothing goes wrong
+                cy.log('We ensure transparency works as expected for external layers too')
+                cy.openLayerSettings(fakeWmsLayerId1)
+
+                cy.get(`[data-cy^="slider-opacity-layer-${fakeWmsLayerId1}-"]`)
+                    .should('be.visible')
+                    .realClick({ position: 'left' })
+                cy.openLayerSettings(fakeWmsLayerId4)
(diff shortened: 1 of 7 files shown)


tests/cypress/tests-e2e/layers.cy.js
```diff
@@ -754,6 +754,20 @@ describe('Test of layer handling', () => {
...
+                visibleLayerIds.forEach((layerId, index) => {
+                    cy.openLayerSettings(layerId)
+                    cy.get(`[data-cy="slider-opacity-layer-${layerId}-${index}"]`)
+                        .should('be.visible')
+                        .realClick({ position: 'left' })
+                    cy.readStoreValue('getters.visibleLayers').should((visibleLayers) => {
+                        const layer = visibleLayers.find((layer) => layer.id === layerId)
+                        expect(layer.opacity).to.eq(0.0)
+                    })
+                })
```
ltshb, line 770: Although this test is good if I understand the issue correctly it still don't test the issue, the layers here are simple WMS and WMTS, the issue describe in ticket talk about group of layer, so it need to be tested in the external layers where you have group of WMS layers (geoadmin don't have group of layer in my connaissance)

ltshb (approved):

PR #793, opened 2024-04-19
PB-389: Fix print attributions and duplicate legends

ltshb:
The attribution (copyright) on the pdf was empty.
    
Also add the background layer attribution if any background.

Fix duplicate print legends. We could duplicate some layers with different opacity, but when printing we need to make sure to have their legend only once.

Added attribution, language and filename tests. Also tested duplicate layer
in print.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-389-print-attributions/index.html)


ismailsunni (approved):
LGTM

PR #794, opened 2024-04-19
PB-422: Fix social network sharing and added linkedin

ltshb:
The social network sharing suffer from a DOM exception. The main issue was
that we used a library that wasn't anymore maintained since 2 years and not
officially supported vue3.

I changed the library by a new one and added linkedin as well.

Note the new library had no support for qrcode.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-422-social-sharing/index.html)

### DIFF
--- a/src/modules/i18n/locales/de.json
+++ b/src/modules/i18n/locales/de.json
@@ -660,6 +660,5 @@
     "operation_aborted": "Operation abgebrochen",
     "outside_valid_year_range": "Das eingegebene Jahr liegt außerhalb des verfügbaren Bereichs:",
     "share_map_title": "Link zum Geoportal der Schweizerischen Eidgenossenschaft",
-    "link_description": "Link Beschreibung (optional)",
-    "media_disclaimer": "Enthält Inhalte von Drittanbietern"
+    "linkedin_tooltip": "Teilen Sie diese Karte auf LinkedIn"
 }
--- a/src/modules/i18n/locales/de.json
+++ b/src/modules/i18n/locales/de.json
@@ -649,7 +649,7 @@
     "no_result": "Kein Ergebnis",
     "loading": "Laden",
     "search_in_catalogue_placeholder": "Suche in importierten Karten",
-    "kml_gpx_file_out_of_bounds": "Der Dateiinhalt liegt außerhalb der Projektionsgrenzen",
+    "kml_gpx_file_out_of_bounds": "Der Dateiinhalt liegt ausserhalb der Projektionsgrenzen",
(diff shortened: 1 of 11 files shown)


ltshb:
@stebie @hansmannj I added the share on LinkedIn on the mapviewer, what do you think it is a good or bad idea ?

src/modules/i18n/locales/de.json
```diff
@@ -658,5 +658,7 @@
     "operation_successful": "Erfolg!",
     "operation_failed": "Hoppla! Etwas ist schiefgegangen. Bitte versuchen Sie es erneut.",
     "operation_aborted": "Operation abgebrochen",
-    "outside_valid_year_range": "Das eingegebene Jahr liegt außerhalb des verfügbaren Bereichs:"
+    "outside_valid_year_range": "Das eingegebene Jahr liegt außerhalb des verfügbaren Bereichs:",
+    "linkedin_tooltip": "Teilen Sie diese Karte auf LinkedIn",
+    "share_map_title": "Link zum Geoportal der Schweizerischen Eidgenossenschaft"
```
ltshb, line 663: @stebie @hansmannj @mariokeusen for the share by email button, on windows with outlook it seems that we need to have a subject otherwise the body is set in the subject as follow: `&body=https://s.geo.admin.ch/asdfasdf`. That's why I created a generic email subject for the sharing, could you review it (in german and english) ? Or do you have a better proposal as default email subject ?

src/modules/i18n/locales/de.json
```diff
@@ -658,5 +658,7 @@
     "operation_successful": "Erfolg!",
     "operation_failed": "Hoppla! Etwas ist schiefgegangen. Bitte versuchen Sie es erneut.",
     "operation_aborted": "Operation abgebrochen",
-    "outside_valid_year_range": "Das eingegebene Jahr liegt außerhalb des verfügbaren Bereichs:"
+    "outside_valid_year_range": "Das eingegebene Jahr liegt außerhalb des verfügbaren Bereichs:",
```
hansmannj, line 661: In CH-german, please use a double `s` instead of `ß` in `außerhalb`

hansmannj (approved):
Thanks, LGTM! :rocket:
Only one little improvement for a CH-german translation.
And what do you mean with "no support for qr-code"? Using the test link, there still is a qrcode for sharing shown.

src/modules/i18n/locales/de.json
ltshb: I've changed this in the translation table, but will update the code in a separate PR as this translation is not part of this PR and that other open PR have changed translation.

PR #795, opened 2024-04-22
PB-421: Fix map footer attribution interactivity

ltshb:
The attributions could not be clicked and did not had any pointer cursor.
This has been now fixed and and e2e tests changed to test this functionality.


ltshb:
@pakb for the failing test it is an zindex issue with the tooltip which is below the map attribution we need to discussed what we do with this kind of issue.
image

pakb:
> @pakb for the failing test it is an zindex issue with the tooltip which is below the map attribution we need to discussed what we do with this kind of issue.

Testing the floating tooltip on such a small viewport is bound to create problems, maybe we should increase the viewport's height when doing such test cases.

PR #797, opened 2024-04-22
PB-360: Update gpx line style and adjust line stroke width for printing.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-360-gpx-print-style/index.html)

### OL comparison:
**Old Geoadmin:**
image
**New Geoadmin:**
image

### Print comparison
**Old Geoadmin:**
image
**New Geoadmin:**
image

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-360-gpx-print-style/index.html)

### DIFF
--- a/src/utils/styleUtils.js
+++ b/src/utils/styleUtils.js
@@ -136,13 +136,10 @@ export const highlightPointStyle = new Style({
     }),
 })
 
-// Change a distance according to the change of DPI
+// Change a distance according to the change of DPI (from the old geoadmin)
 export function adjustDistance(dist, dpi) {
     if (!dist) {
         return
     }
-    // Taken from the old geoadmin where the distance is calculated based on the DPI used.
-    // In the old geoadmin we use (dist * 90) / dpi. With dpi either 150 (A3) or 254 (A4).
-    // I use 34 as a constant to have a similar result (i.e. 90/254 * 96 = 34)
-    return (dist * 34) / dpi
+    return (dist * 90) / dpi
 }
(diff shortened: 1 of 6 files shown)


ismailsunni:
I will update this PR one we have the mapfishprint service updated

ismailsunni:
I have updated the PR with the DPI from the capabilities.

src/utils/styleUtils.js
```diff
@@ -132,3 +135,11 @@ export const highlightPointStyle = new Style({
         stroke: highlightedStroke,
     }),
 })
+
+// Change a distance according to the change of DPI (from the old geoadmin)
+export function adjustDistance(dist, dpi) {
+    if (!dist) {
+        return
+    }
+    return (dist * 90) / dpi
```
ltshb, line 144: Do you know the reason of the `90` factor ?

ltshb (approved):
Looks good, it seems to solve also PB-441

src/utils/styleUtils.js
ismailsunni: After some digging, the first time it's introduced is here https://github.com/geoadmin/mf-geoadmin3/pull/3280

It seems because we use different DPI for WMS and WMTS (using 96, although 90.7 is the WMTS standard) and in the print we use 150 or 254.

But for some unknown reason, it use 90 instead 96.

Should I change it to 96?

ltshb: @ismailsunni It is worse to try with 96 to see how it looks. In any case it would be good to comment this factor in code.

ltshb: @ismailsunni it might be also good to rename this function, from my understanding it is more like a width adjustment as a distance

PR #798, opened 2024-04-22
PB-440: Allow prefixed KML

ltshb:
Based on the KML standard the kml tags can be with `kml:` prefixed. This
issue has been mentioned on the old viewer. see https://github.com/geoadmin/mf-geoadmin3/issues/5310

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-440-prefixed-kml/index.html)


ltkum (commented):
I tested it with the files given by the issuer in the old geoadmin ticket, and it seemed to work. 

Maybe a small e2e test to check that both syntaxes are supported would be good.

ltkum (approved):
All good, thanks :)

PR #801, opened 2024-04-23
PB-432: Allow text selection in layer info popup

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-432-layer-legend-2/index.html)


LukasJoss (approved):
Wouldn't `@extend .clear-no-ios-long-press;` do the same?

ltshb:
> clear-no-ios-long-press

good point I did not thought about this class I'll change it.

PR #802, opened 2024-04-23
PB-443: Set year on timeslider close

ltkum:
When closing the timeslider, if it was modified in any way manually by the user, we save the year for all current layers. Also removed the numbers in the tests so that we don't have to re-enumerate each time we make a modification within the tests.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-443-set-year-on-timeslider-close/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -229,7 +229,6 @@ function positionNodeLabel(year) {
 
 function grabCursor(event) {
     yearCursorIsGrabbed = true
-    isPristine.value = false
     if (event.type === 'touchstart') {
         // for touch events we have to select which touch we want to get the screen position
         // (there can be multiple fingers gestures)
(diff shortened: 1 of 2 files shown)


src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -180,6 +182,20 @@ onMounted(() => {
     })
 })
 
+onUnmounted(() => {
+    if (!isPristine.value) {
+        activeLayers.value.forEach((layer, index) => {
```
pakb, line 187: you may re-use the ref `layersWithTimestamps` from above, it already filters all the relevant layers

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -213,6 +229,7 @@ function positionNodeLabel(year) {
 
 function grabCursor(event) {
     yearCursorIsGrabbed = true
+    isPristine.value = false
```
pakb, line 232: I would wait for the value to be altered before setting pristine to false, so that if the user grab but then releases the cursor without changing anything (miss-click) it doesn't block its settings with the preview year

src/modules/map/components/toolbox/TimeSlider.vue
ltkum: this is true, but we commit changes to layers through their indexex for timed layers, which `layersWithTimestamps` can't give.
I could also create a new way to commit changes which gives directly the layer for the change, but I would prefer to keep only one way to alter a timed layer.

pakb (approved):

PR #803, opened 2024-04-23
PB-304: Treasurehunt layer not displayed

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-304-treasurehunt-layer-not-displayed/index.html)


LukasJoss:
> Looks like this fixes the issue, nice!

I still would like to now whats different, I did not find any difference in the initialization, and in the json file the value is defined as empty string, somehow in the old viewer that must have been deleted to get an undefined

pakb (approved):
What's different I suspect is the way Javascript interprets a "falsy" value, if the `key` received here is `null` or `false` the previous test wouldn't trigger, but with the way you now check for "falsy-ness" (with `!`) now all those values (`undefined`, `null`, `false`, `0`) are considered false

PR #805, opened 2024-04-24
PB-375: Added drawing to the report a problem

ltshb:
- [x] TODO the title of the drawing overlay when entering from the report a problem is not always nicely displayed depending on the size of the viewport, see will be probably fixed with #773 PR, so once this PR is finish we need to check it again. 
image
Note with normal size it works good and on mobile the title is anyway not displayed.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-375-drawing-report/index.html)


src/store/modules/drawing.store.js
```diff
@@ -61,6 +94,30 @@ export default {
         setDrawingFeatures({ commit }, { featureIds, dispatcher }) {
             commit('setDrawingFeatures', { featureIds, dispatcher })
         },
+        toggleDrawingOverlay(
+            { commit, state },
+            { dispatcher, online = null, kmlId = null, title = null }
```
pakb, line 99: maybe `online = true` would be better here as default value

src/store/modules/drawing.store.js
```diff
@@ -69,5 +126,11 @@ export default {
         deleteDrawingFeature: (state, { featureId }) =>
             (state.featureIds = state.featureIds.filter((featId) => featId !== featureId)),
         setDrawingFeatures: (state, { featureIds }) => (state.featureIds = featureIds),
+        setShowDrawingOverlay(state, { show, online, kmlId, title }) {
+            state.drawingOverlay.show = show
+            state.drawingOverlay.title = title
+            state.drawingOverlay.online = online
+            state.drawingOverlay.kmlId = kmlId
```
pakb, line 133: there two do not match what you have described above in the state, it should be 
```
state.online = online
state.temporaryKmlId = kmlId
```
from what I understand

src/modules/drawing/DrawingModule.vue
```diff
@@ -28,14 +28,24 @@ const activeKmlLayer = computed(() => store.getters.activeKmlLayer)
 const featureIds = computed(() => store.state.drawing.featureIds)
 const isDrawingEmpty = computed(() => store.getters.isDrawingEmpty)
 const noFeatureInfo = computed(() => store.getters.noFeatureInfo)
+const online = computed(() => store.state.drawing.drawingOverlay.online)
```
pakb, line 31: here same thing, it should map `store.state.drawing.online` instead

src/modules/drawing/DrawingModule.vue
```diff
@@ -28,14 +28,24 @@ const activeKmlLayer = computed(() => store.getters.activeKmlLayer)
...
+const hasKml = computed(() => {
+    if (online.value) {
+        return !!activeKmlLayer.value
+    }
+    return !!store.state.layers.systemLayers.find(
+        (l) => l.id === store.state.drawing.drawingOverlay.kmlId
```
pakb, line 37: should be mapping `store.state.drawing.temporaryKmlId`

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -57,6 +57,7 @@ const drawingStateMessage = computed(() => {
             return null
     }
 })
+const online = computed(() => store.state.drawing.drawingOverlay.online)
```
pakb, line 60: should be mapping `store.state.drawing.online`

src/modules/drawing/useKmlDataManagement.composable.js
```diff
@@ -18,19 +18,33 @@ export default function useSaveKmlOnChange(drawingLayerDirectReference) {
     const drawingLayer = inject('drawingLayer', drawingLayerDirectReference)
 
     const store = useStore()
+    const online = computed(() => store.state.drawing.drawingOverlay.online)
```
pakb, line 21: dito

src/modules/drawing/useKmlDataManagement.composable.js
```diff
@@ -18,19 +18,33 @@ export default function useSaveKmlOnChange(drawingLayerDirectReference) {
...
     const projection = computed(() => store.state.position.projection)
     const activeKmlLayer = computed(() => store.getters.activeKmlLayer)
     const availableIconSets = computed(() => store.state.drawing.iconSets)
+    const temporaryKmlId = computed(() => store.state.drawing.drawingOverlay.kmlId)
```
pakb, line 25: dito

hansmannj (commented):
Thanks, did not check the code, but only the test link. 

Two minor things:

- when clicking on "Draw on the map" as a first action after clicking on "report a problem", i.e. the text field is still empty, sometimes the app does not switch into the drawing mode but is stuck and complains "Report message can not be empty.". Sometimes it switches into the drawing mode, but the red focus and message still shortly appears.
- Perhaps the text "indicate the appropriate location on the map" is not optimal, as it is not always about a certain location. Maybe something like "Please create the drawing you want to attach" or similar might be better. Picky, I know, sorry! :wink;

src/modules/drawing/useKmlDataManagement.composable.js
ltshb: I should have read my code after the refactoring of the showDrawingOverlay from ui store into drawing store 🙈.
I think I got it right know

ltshb:
> Thanks, did not check the code, but only the test link.
> 
> Two minor things:
> 
> * when clicking on "Draw on the map" as a first action after clicking on "report a problem", i.e. the text field is still empty, sometimes the app does not switch into the drawing mode but is stuck and complains "Report message can not be empty.". Sometimes it switches into the drawing mode, but the red focus and message still shortly appears.
> * Perhaps the text "indicate the appropriate location on the map" is not optimal, as it is not always about a certain location. Maybe something like "Please create the drawing you want to attach" or similar might be better. Picky, I know, sorry! 😉

@hansmannj for the first point regarding field validation I propose to do it in a separate PR as we should change the field validation behaviour globally not only here. The issue is that we validate the field on typing and focus out. 
For the translation I took the one from the old viewer if you have a better proposal I can change it.

pakb (approved):
Minor detail, otherwise looking good!

PR #806, opened 2024-04-25
PB-414: marker overlapping title

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-414-marker-overlapping-title/index.html)

### DIFF
--- a/src/api/features/EditableFeature.class.js
+++ b/src/api/features/EditableFeature.class.js
@@ -3,13 +3,8 @@ import { Icon as olIcon } from 'ol/style'
 import { extractOlFeatureGeodesicCoordinates } from '@/api/features/features.api'
 import SelectableFeature from '@/api/features/SelectableFeature.class'
 import { DEFAULT_ICON_URL_PARAMS } from '@/api/icon.api'
-import {
-    allStylingColors,
-    allStylingSizes,
-    DEFAULT_OFFSET,
-    MEDIUM,
-    RED,
-} from '@/utils/featureStyleUtils'
+import { DEFAULT_TEXT_OFFSET } from '@/api/icon.api'
+import { allStylingColors, allStylingSizes, MEDIUM, RED } from '@/utils/featureStyleUtils'
 
 /** @enum */
 export const EditableFeatureTypes = {
@@ -47,7 +42,7 @@ export default class EditableFeature extends SelectableFeature {
             title = '',
             description = '',
             featureType,
-            textOffset = DEFAULT_OFFSET,
+            textOffset = DEFAULT_TEXT_OFFSET,
             textColor = RED,
             textSize = MEDIUM,
             fillColor = RED,
--- a/src/api/features/EditableFeature.class.js
+++ b/src/api/features/EditableFeature.class.js
@@ -25,7 +25,7 @@ export default class EditableFeature extends SelectableFeature {
      * @param {String} featureData.description A description of this feature, can not be HTML
      *   content (only text)
      * @param {EditableFeatureTypes} featureData.featureType Type of this editable feature
-     * @param {Array} featureData.textAlign Alignment for the text of this feature
+     * @param {Array} featureData.textOffset Alignment for the text of this feature
      * @param {FeatureStyleColor} featureData.textColor Color for the text of this feature
      * @param {FeatureStyleSize} featureData.textSize Size of the text for this feature
      * @param {FeatureStyleColor} featureData.fillColor Color of the icon (if defined)
@@ -41,7 +41,7 @@ export default class EditableFeature extends SelectableFeature {
             title = '',
             description = '',
             featureType,
-            textAlign = [0, -41],
+            textOffset = [0, -41],
             textColor = RED,
             textSize = MEDIUM,
             fillColor = RED,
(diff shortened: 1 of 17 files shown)


LukasJoss:
Exaple kml:
[https://sys-s.dev.bgdi.ch/mr7dwvf1zanw](https://sys-s.dev.bgdi.ch/mr7dwvf1zanw)
[https://sys-s.dev.bgdi.ch/9wyu8ahpjpzh](https://sys-s.dev.bgdi.ch/9wyu8ahpjpzh)

src/utils/featureStyleUtils.js
```diff
@@ -203,6 +203,62 @@ export function getTextColor(style) {
...
+ * @param {Array} anchor Relative position of Anchor
+ * @param {String} iconName Name of icon
+ * @returns {Array | null} Returns the feature label offset
+ */
+export function calculateTextOffset(textScale, iconScale, anchor, iconName) {
+    if (!iconScale) {
+        return [0, 0]
+    }
+
+    const iconSize = [32, 32]
```
ltshb, line 239: The icon size should not be hardcoded. It is planned to have it from the backend see https://jira.swisstopo.ch/browse/PB-303. 
But why 32 pixel here and not 48 pixel ? Is it because of openlayer using 32 pixel as basis ?

ltshb (commented):
Personally I prefer the text to be below the marker, especially for the position marker having the text below is more intuitive.

src/utils/featureStyleUtils.js
LukasJoss: You are right it is 48 by 48, I should have used a scaling factor instead of reducing the extent. Since all our icons have quite a big transparency border their apparent size is much smaller, causing the gap between icon and text to grow with increasing icon size, but I have removed that scaling for now.

LukasJoss:
> Personally I prefer the text to be below the marker, especially for the position marker having the text below is more intuitive.

We can discuss that tomorrow, maybe somebody else has a strong opinion about it.

src/utils/featureStyleUtils.js
```diff
@@ -203,6 +204,59 @@ export function getTextColor(style) {
...
+
+    return [0, -(defaultOffset + iconOffset + textOffset)]
+}
+
+/** Offset of the marker with default styling and anchor on bottom of icon */
+export const DEFAULT_OFFSET = calculateTextOffset(
+    MEDIUM.textScale,
+    MEDIUM.iconScale,
+    [0, 1],
+    [48, 48]
```
ltshb, line 257: Dito use The `DEFAULT_ICON_URL_PARAMS.size` here

src/utils/kmlUtils.js
```diff
@@ -450,13 +455,14 @@ export function getEditableFeatureFromKmlFeature(kmlFeature, availableIconSets)
  * @param {DrawingIconSet[]} iconSets Icon sets to use for EditabeFeature deserialization
  * @returns {ol/Feature[]} List of OL Features
  */
-export function parseKml(kmlData, projection, iconSets) {
+export function parseKml(kmlLayer, projection, iconSets) {
```
ltshb, line 458: Update the js doc above

src/utils/kmlUtils.js
```diff
@@ -399,6 +401,8 @@ export function getEditableFeatureFromKmlFeature(kmlFeature, availableIconSets)
     const iconSize = iconStyle ? getIconSize(iconStyle) : null
     const fillColor = getFillColor(style, kmlFeature.getGeometry().getType(), iconArgs)
 
+    const textOffset =
+        iconSize && !kmlLayer.isLegacy() ? getTextOffset(textScale, iconSize, iconStyle) : [0, 0]
```
ltshb, line 405: Would make more sense to have something like this
```javascript
let textOffset = [0, 0]
if (!kmlLayer.isLegacy() && iconSize && featureType === EditableFeatureTypes.MARKER) {
  textOffset = getTextOffset(...)
}
```
Instead of checking the type in the EditableFeature, this would keep the last one very generic.

src/api/features/EditableFeature.class.js
```diff
@@ -40,6 +47,7 @@ export default class EditableFeature extends SelectableFeature {
             title = '',
             description = '',
             featureType,
+            textOffset = DEFAULT_OFFSET,
```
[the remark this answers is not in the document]
LukasJoss: I would need to call updateTextOffset somewhere doing the initialization as otherwise the markers would also have a default of [0,0] in drawing mode

src/utils/featureStyleUtils.js
ltshb: But maybe rename it to `DEFAULT_TEXT_OFFSET`, it is then clearer

LukasJoss:
@ltshb currently when exporting and then reimporting a kml file, there will be no offset since it is considered to be a legacy layer as there is no metedata when importing it from a file [https://github.com/geoadmin/web-mapviewer/blob/4c44e28db41740192f25fd4748aadcf6ca10cac6/src/api/layers/KMLLayer.class.js#L89-L91|https://github.com/geoadmin/web-mapviewer/blob/4c44e28db41740192f25fd4748aadcf6ca10cac6/src/api/layers/KMLLayer.class.js#L89-L91] should there be added some metadata in the kml file itself that we produce so we can detect that?

pakb (approved):
Looking good, thanks!

ltshb (approved):
Looks very thanks a lot.
Now that the offset is in KML it might make sense to test the presence/absence of the offset in KML in e2e test, this should be easily done ?

PR #810, opened 2024-04-29
Bugfix PB-128: implement client side zoom

ltkum:
Issue : when zooming further than the maximum resolution available, the application would try to fetch non existing tiles. Upon reloading, nothing would be displayed, at there was no tile already present.

Fix : we get the resolutions from the layer config endpoint and determine the maximum resolution of each layer, with a default identical to the one in the endpoint (0.5 m/pixel) Then, when creating TileGrid parameters for layers, we limit the resolutions to what is within range. We then let openLayers work its magic with it.

This also means that going to zoom levels higher than its possible max won't make new queries to the backend, limiting the number of requests and errors.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-128-client-side-zoom/index.html)

### DIFF
--- a/src/api/layers/ExternalGroupOfLayers.class.js
+++ b/src/api/layers/ExternalGroupOfLayers.class.js
@@ -1,7 +1,6 @@
 import ExternalLayer from '@/api/layers/ExternalLayer.class'
 import { InvalidLayerDataError } from '@/api/layers/InvalidLayerData.error'
 import LayerTypes from '@/api/layers/LayerTypes.enum'
-import { DEFAULT_MAX_GEOADMIN_RESOLUTION } from '@/config'
 
 /**
  * Description of a group of layers, that could be added altogether or separately, that stems from a
@@ -46,8 +45,6 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
      *   Default is `null`
      * @param {CoordinateSystem[]} [externalLayerData.availableProjections=[]] All projection that
      *   can be used to request this layer. Default is `[]`
-     * @param {Number} [externalLayerData.maxResolution=DEFAULT_MAX_GEOADMIN_RESOLUTION] Define the
-     *   maximum resolution the layer can reach. Default is `DEFAULT_MAX_GEOADMIN_RESOLUTION`
      * @throws InvalidLayerDataError if no `externalLayerData` is given or if it is invalid
      */
     constructor(externalLayerData) {
@@ -68,7 +65,6 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
             isLoading = true,
             availableProjections = [],
             getFeatureInfoCapability = null,
-            maxResolution = DEFAULT_MAX_GEOADMIN_RESOLUTION,
         } = externalLayerData
         if (!layers?.length > 0) {
             throw new InvalidLayerDataError(
@@ -92,7 +88,6 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
             // if one of the sublayers is not queryable, this group is then not queryable/has no tooltip
             hasTooltip: !layers.some((layer) => !layer.hasTooltip),
             getFeatureInfoCapability,
-            maxResolution,
         })
         this.layers = [...layers]
     }
--- a/src/api/layers/ExternalGroupOfLayers.class.js
+++ b/src/api/layers/ExternalGroupOfLayers.class.js
@@ -1,6 +1,7 @@
 import ExternalLayer from '@/api/layers/ExternalLayer.class'
 import { InvalidLayerDataError } from '@/api/layers/InvalidLayerData.error'
 import LayerTypes from '@/api/layers/LayerTypes.enum'
+import { DEFAULT_MAX_GEOADMIN_RESOLUTION } from '@/config'
 
 /**
  * Description of a group of layers, that could be added altogether or separately, that stems from a
@@ -45,6 +46,8 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
      *   Default is `null`
      * @param {CoordinateSystem[]} [externalLayerData.availableProjections=[]] All projection that
      *   can be used to request this layer. Default is `[]`
+     * @param {Number} [externalLayerData.maxResolution=DEFAULT_MAX_GEOADMIN_RESOLUTION] Define the
+     *   maximum resolution the layer can reach. Default is `DEFAULT_MAX_GEOADMIN_RESOLUTION`
      * @throws InvalidLayerDataError if no `externalLayerData` is given or if it is invalid
      */
     constructor(externalLayerData) {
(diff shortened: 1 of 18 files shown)


src/api/layers/ExternalGroupOfLayers.class.js
```diff
@@ -65,6 +68,7 @@ export default class ExternalGroupOfLayers extends ExternalLayer {
             isLoading = true,
             availableProjections = [],
             getFeatureInfoCapability = null,
+            maxResolution = DEFAULT_MAX_GEOADMIN_RESOLUTION,
```
ltshb, line 71: Can we not get the max resolution from the GetCapabilities of the external layers ? Otherwise this might not work for external layers.

src/api/layers/GPXLayer.class.js
```diff
@@ -27,6 +30,7 @@ export default class GPXLayer extends AbstractLayer {
             opacity = 1.0,
             gpxData = null,
             gpxMetadata = null,
+            maxResolution = DEFAULT_MAX_GEOADMIN_RESOLUTION,
```
ltshb, line 33: For GPX and KML we should not have any issue if I'm correct ? If yes then we should not have a max resolution for those layers and the AbstractLayer should have a `null` value per default

src/api/layers/GeoAdminGeoJsonLayer.class.js
```diff
@@ -44,6 +47,7 @@ export default class GeoAdminGeoJsonLayer extends GeoAdminLayer {
             geoJsonUrl = null,
             styleUrl = null,
             updateDelay = null,
+            maxResolution = DEFAULT_MAX_GEOADMIN_RESOLUTION,
```
ltshb, line 50: Same as for KML and GPX is this really needed here do we have a resolution limitation ?

src/api/layers/layers.api.js
```diff
@@ -29,6 +29,8 @@ import log from '@/utils/logging'
 const generateClassForLayerConfig = (layerConfig, id, allOtherLayers, lang) => {
     let layer = undefined
     if (layerConfig) {
+        const maxResolution =
+            layerConfig.resolutions?.slice(-1)[0] ?? DEFAULT_MAX_GEOADMIN_RESOLUTION
```
ltshb, line 33: Do we currently have layer without max resolution ? Make it really sense to use a default value? 
Here a danger we might hide a bug in the backend, if currently all layer have a resolutions list, then we should either let the application crash in the case it is missing or remove the layer from the config with an error so we can quickly identify the issue.

src/config.js
```diff
@@ -178,6 +178,14 @@ export const WMS_TILE_SIZE = 512 // px
 
 export const VIEW_MIN_RESOLUTION = 0.1 // meters/pixel
 
+/**
+ * Map default max resolution as it is set in mf-chsdi. When the layer has this as a max resolution,
+ * the layersConfig endpoint does not give the resolutions available.
+ * https://github.com/geoadmin/mf-chsdi3/blob/0236814544a6bf2df86598889b81ee4023494325/chsdi/models/bod.py#L119-L123
```
ltshb, line 184: @pakb @rebert @ltclm what does it mean when a layer don't give the resolutions in mf-chsdi3 ? Why do we have this use case at all ? Or should all layers define their resolutions ? From what I've seen/tested Geojson/GPX/KML layers don't require the max resolution, but WMS and WMTS layer do. What about Vector layer is it the same as KML layer I guess ?

ltshb (commented):
Some open questions/comment about the default value for the max resolution, if I'm not misttaken this max resolution is not needed for all layer and should be by default null. For external layer (WMS/WMTS) we should try to take the max resolution from the GetCapabilities.

src/config.js
ltclm: `minResolution/maxResolution` settings in such layers are used in mf-geoadmin3 for the scale dependant request of different layers from the backend, this is used in so called aggregated layers. this can be combinations of type wms/wms or wmts/wms.

```json
"ch.swisstopo-vd.geometa-gemeinde_wms": {
"gutter": 15,
"timeEnabled": false,
"searchable": false,
"type": "wms",
"tooltip": false,
"minResolution": 0,
"topics": "api,cadastre,ech,inspire,service-wms,swisstopo",
"background": false,
"maxResolution": 20,
"wmsLayers": "ch.swisstopo-vd.geometa-gemeinde",
"label": "ch.swisstopo-vd.geometa-gemeinde_wms",
"parentLayerId": "ch.swisstopo-vd.geometa-gemeinde",
"attribution": "AV",
"chargeable": false,
"hasLegend": true,
"serverLayerName": "ch.swisstopo-vd.geometa-gemeinde",
"highlightable": true,
"format": "png",
"singleTile": false,
"opacity": 0.75,
"wmsUrl": "https://wms.geo.admin.ch/",
"attributionUrl": "https://www.cadastre.ch/de/home.html"
},
"ch.swisstopo-vd.geometa-gemeinde_wmstile": {
"timeEnabled": false,
"searchable": false,
"type": "wms",
"tooltip": false,
"minResolution": 20,
"topics": "api,cadastre,ech,inspire,service-wms,swisstopo",
"background": false,
"wmsLayers": "ch.swisstopo-vd.geometa-gemeinde",
"label": "ch.swisstopo-vd.geometa-gemeinde_wmstile",
"parentLayerId": "ch.swisstopo-vd.geometa-gemeinde",
"attribution": "AV",
"chargeable": false,
"hasLegend": true,
"serverLayerName": "ch.swisstopo-vd.geometa-gemeinde",
"highlightable": true,
"format": "png",
"singleTile": false,
"opacity": 0.75,
"wmsUrl": "https://wms.geo.admin.ch/",
"attributionUrl": "https://www.cadastre.ch/de/home.html"
},

```

ltkum: @ltclm thanks for the insight. We checked with @ltshb  and it seems aggregate layers are not showing at all in the new mapviewer, and you were right regarding wms layers. I'll correct this PR to only affect  WMTS layers, then I'll investigate our aggregate layers problem.

ltshb (commented):
The bug is still present when testing on the test link

src/config.js
pakb: If I remember what I've discussed with @ltclm : only resolution that aren't "default" are given in the config, the one you have set as the default as constant is to be used otherwise

src/api/layers/ExternalWMSLayer.class.js
```diff
@@ -68,6 +68,7 @@ export default class ExternalWMSLayer extends ExternalLayer {
             hasTooltip = false,
             getFeatureInfoCapability = null,
         } = externalWmsData
+
```
pakb, line 71: left over of some changes? doesn't really matter, but if you have to push once again maybe clean that up :wink:

src/config.js
pakb, line 184: you can prefix this URL with a `@see`

src/config.js
```diff
@@ -178,6 +178,14 @@ export const WMS_TILE_SIZE = 512 // px
...
+ *
+ * @type {Number}
+ */
+export const DEFAULT_MAX_GEOADMIN_RESOLUTION = 0.5 // meters/pixel
```
pakb, line 188: I would maybe switch words GEOADMIN and MAX here, and add WMTS somewhere.
Something like `DEFAULT_GEOADMIN_MAX_WMTS_RESOLUTION` (as we now know it is only applicable to WMTS)

pakb (approved):
Works fine on the test link, nice!

PR #811, opened 2024-04-29
PB-378: handle profiles request with too many points

ltkum:
Issue : the profile backend has a query limit of 5000 coordinates. This will be updated in the future, but in the meantime we need to inform users which reach this limit why their query is failing.

Fix : Upon receiving a 413 error from the backend which tells us that the coordinates limit has been reached, we send a profile error with a translation key for the message.


src/config.js
```diff
@@ -69,6 +69,14 @@ export const API_SERVICE_ALTI_BASE_URL = enforceEndingSlashInUrl(
...
 )
 
+/**
+ * Queries from service alti are currently capped at 5000 points. We need to make some error
+ * processing when there is more points. This will be removed when we get rid of the limit in the
+ * backend.
+ *
+ * @type Integer
+ */
+export const PROFILE_MAX_POINTS = 5000
```
ltshb, line 79: Do not use hardcoded limit here in code

src/api/profile/profile.api.js
```diff
@@ -72,6 +72,22 @@ async function getProfileDataForChunk(chunk, startingPoint, startingDist, output
                 )
             }
         } catch (err) {
+            if (
+                err.response &&
+                err.response.status === 413 &&
+                err.response.data.error?.message ===
+                    `Request Geometry contains too many points. Maximum number of points allowed: ${PROFILE_MAX_POINTS}, found ${chunk.coordinates.length}`
```
ltshb, line 79: Here simply check for the start of the message to avoid having the limit hardcoded

src/api/profile/profile.api.js
```diff
@@ -72,6 +72,22 @@ async function getProfileDataForChunk(chunk, startingPoint, startingDist, output
...
+            ) {
+                log.error(
+                    'Requesting too many points for a profile, request could not be processed',
+                    err
+                )
+                throw new ProfileError(
+                    'Error requesting profile with too many points',
+                    'too_many_points_profile_error'
```
ltshb, line 87: The translation key is missing

ltshb (commented):
I've tested it and there is no message shown to the user, you need to handle the exception and display an error message instead of the profile.
An e2e test checking that an error message is displayed instead of the profile would be good as well.
I've attached in the ticket a KML file with a big red line.

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -1161,6 +1161,27 @@ describe('Drawing module tests', () => {
...
+            cy.get('[data-cy="ol-map"]').click(200, 200)
+            cy.get('[data-cy="ol-map"]').dblclick(150, 200)
+            cy.wait('@error-profile-too-long')
+            cy.get('[data-cy="profile-popup-content"]').should('be.visible')
+            cy.get('[data-cy="profile-error-message"]')
+                .should('be.visible')
+                .should('satisfy', (element) => {
+                    const classList = Array.from(element[0].classList)
+                    return classList.includes('text-danger')
+                })
```
ltshb, line 1183: Could be much simpler `.should('have.class', 'text-danger')`

ltshb (approved):
Just one minor test code improvement

PR #813, opened 2024-05-01
PB-475 : Activate a couple post-processing shader in Cesium

pakb:
this should enhance the look and feel of buildings. Removing config that was set to better place/set shadows, as we won't be using this approach to "color" building facades (too costly performance wise).

Also adding a FPS counter when not on PROD staging.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-475_add_post_processing/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumMap.vue
+++ b/src/modules/map/components/cesium/CesiumMap.vue
@@ -87,8 +87,8 @@ import {
     Color,
     defined,
     Ellipsoid,
-    JulianDate,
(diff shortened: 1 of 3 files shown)


pakb:
Post processing activated :
- Ambient occlusion
- FXAA (anti-aliasing)

Before
image

After
image

Buildings are a bit less jagged, rough

ltshb (approved):
Did not see much changes but code looks good as far as I can tell

PR #815, opened 2024-05-01
PB-471 : add floating mode for tooltip

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-471_floating_tooltip_part_2/index.html)


ltkum (commented):
looks pretty good to me. Quick question: 
When switching from floating to fixed, then to floating again, it doesn't keep the previous position. Same with reloading the page.
Do we want, in the future, to allow people to have the floating tooltip position be something they can share ?

It seems to work and I don't think I've seen anything problematic in the code, but I would just make a quick e2e test to see if we can move the infobox.

src/modules/map/components/LocationPopup.vue
```diff
@@ -159,9 +160,10 @@ function clearClick() {
     <component
         :is="mappingFrameworkSpecificPopup"
         v-if="coordinate"
-        :title="selectedTab == 'position' ? i18n.t('position') : i18n.t('link_bowl_crosshair')"
+        :title="selectedTab === 'position' ? i18n.t('position') : i18n.t('link_bowl_crosshair')"
         :coordinates="coordinate"
         :projection="projection"
+        :mode="MapPopoverMode.FEATURE_TOOLTIP"
```
ltshb, line 166: In FEATURE_TOOLTIP mode, when collapsing the feature it change slightly the size of the window and make the center of the tooltip moving, which is not good with the location popup. The later has already a bug when changing to sharing tab which also change it size.

src/modules/map/components/MapPopover.vue
```diff
@@ -1,24 +1,151 @@
...
+    }
+    return {
+        top,
+        bottom: isPhoneMode.value ? 0 : footerHeight,
+        left: 0,
+        right: 0,
+    }
+})
+
+useTippyTooltip('.map-popover-header .print-button[data-tippy-content]')
```
ltshb, line 87: Would simply do `useTippyTooltip('.map-popover-header [data-tippy-content]')` and add tippy to other components like the close and collapse buttons

src/modules/map/components/MapPopover.vue
```diff
@@ -1,24 +1,151 @@
...
                 </button>
+                <button
+                    class="btn btn-sm btn-light d-flex align-items-center"
+                    data-cy="map-popover-close-button"
+                    @click="showContent = !showContent"
+                >
+                    <FontAwesomeIcon
+                        icon="caret-down"
+                        :rotation="showContent ? 0 : 90"
+                        class="animate-everything"
```
ltshb, line 146: Rotate to the other side or use caret-right

src/modules/infobox/InfoboxModule.vue
```diff
@@ -82,8 +82,11 @@ function onClose() {
                 data-cy="infobox-minimize-maximize"
                 @click="onToggleContent"
             >
-                <FontAwesomeIcon v-if="!showContent" icon="window-maximize" />
-                <FontAwesomeIcon v-if="showContent" icon="window-minimize" />
+                <FontAwesomeIcon
+                    class="animate-everything"
+                    icon="caret-down"
+                    :rotation="!showContent ? 90 : 0"
```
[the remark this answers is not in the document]
pakb: I was trying to animate the change, but if you didn't notice it, then I'll simplify that by changing the icon :wink:

src/modules/map/components/LocationPopup.vue
pakb: I found a way to fix the location popup issue, which was mainly due to network share buttons overflowing. They now adapt to the available space they have, and wrap if needed (CSS Grid instead of CSS Flex)

ltshb: Note that on my PR #837 I set a fix width to solve the location popup arrow moving

ltkum (commented):
A few things I noticed while using the tooltip :

-  When moving around, the tooltip follows, but when reducing the size of the window, the tooltip stays where it was. This means that if we reduce the window enough to hide the tooltip, we no longer have a tooltip shown, even when clicking on new features.
- I had one instance where I wasn't able to drop the tooltip after dragging it. It lasted 5-6 seconds, and I think it's most certainly a local issue with my computer, but mentioning it here in the case someone else encounter the same issue and it turns out it's not a computer issue.
- It could be interesting (maybe more a post prod element) to try to position the tooltip the first time out of the selection. If someone selects a feature on the top right corner, we could move the tooltip slightly to show it, and if it's not possible (for example : someone selects the whole map), we simply put it in its default position
- Do we want to be able to remember the tooltip position ? Right now, if I press the 'move to footer' button by accident and want to get my tooltip back, I'll need to manually move it to the previous position after making it a floating tooltip again. And if I want to share a link to someone, I might be thinking the position of the tooltip is shared too. I think we can keep this idea for a post prod discussion.

src/modules/infobox/components/FeatureList.vue
```diff
@@ -103,7 +103,7 @@ function loadMoreResultForLayer(layerId) {
 @import '@/scss/variables.module';
 .feature-list {
     display: grid;
-    grid-template-columns: repeat(auto-fill, minmax($overlay-width, 1fr));
+    grid-template-columns: repeat(auto-fit, minmax($overlay-width, 1fr));
```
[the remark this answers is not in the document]
ltshb: Ok I see that this grid template is required when displayed in the infobox, but then we might simply add this grid template on the infobox or when the width of the feature-list is above 340px

ltshb (commented):
Some small issues:
 - Cursor when hovering the floating window title we should have the move cursor
 - When hovering an item in the list of features we should have the pointer cursor showing that it is clickable (issue already present before)
image
- Show profile is not very intuitive and have several issue that we might want at best discuss offline and fix in separate PR.

src/modules/infobox/components/FeatureList.vue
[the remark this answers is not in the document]
pakb: That would make sense to move that to the Infobox (it's meant to fix the use of space here anyway) if this is what caused the scrollbar issue on crhome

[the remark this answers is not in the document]
pakb: I ended up activating this "grid" layout only outside of the "fluid" mode, as I couldn't add this CSS only for the InfoboxModule without changing drastically how the HTML was built (or making InfoboxModule aware of the innards of the FeatureList component, which I would like to avoid)

Should be a "good-enough" solution

ltshb (approved):

PR #816, opened 2024-05-02
PB-448: Make sure print area layer always on top.

ismailsunni:
I use a simple approach here to set the zindex to something very high. I have checked the suggested approach in the ticket (using systemLayers), but it's not handling the addition to the map. Is it intended for this kind of use case?

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-448-missing-print-area-layer/index.html)


pakb (approved):

pakb:
You could also directly set the zIndex value at the creation of the VectorLayer, directly in the constructor args/option

Using the system layer approach could be a good idea if you were using the store (and a proper layer config) to set your crop area, but you are using directly the OL instance, so I don't think it might be compatible

PR #819, opened 2024-05-02
PB-4564: Form validation

ltshb:
Now forms are only validated after user action. This provide a better user
feedback especially when the form has several fields, the form is only validated
when the user click on the action button.

This has been done for all forms:
- report a problem
- import file tool
- drawing media link
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-454-form-validation/index.html)


src/utils/components/EmailInput.vue
```diff
@@ -0,0 +1,218 @@
...
+</script>
+
+<template>
+    <div class="form-group has-validation">
+        <label
+            v-if="label"
+            class="mb-2"
+            :class="{ 'fw-bolder': required }"
+            :for="inputEmailId"
+            :data-cy="`${dataCyPrefix}-label`"
```
pakb, line 183: It might be better (more comprehensible) to add static/fixed `data-cy="..."` here, and then add another one whenever the component is used
```
<EmailInput data-cy="some-identifier" />
```
It will be applied to the div wrapping the label and input, and we can still access them in Cypress test with `cy.get('[data-cy="some-identifier"] [data-cy="..."]')`

src/utils/composables/useFieldValidation.js
```diff
@@ -0,0 +1,132 @@
...
+    props,
+    model,
+    emits,
+    {
+        customValidate = () => {
+            return { valid: true, invalidMessage: '' }
+        },
+        requiredInvalidMessage = 'field_required',
+    } = {}
+) {
```
pakb, line 39: Could you add a bit of documentation that describe what inputs are expected and what they do?

src/utils/composables/useFieldValidation.js
```diff
@@ -0,0 +1,132 @@
...
+        !(returnObject.invalidMessage instanceof 'String')
+    ) {
+        return false
+    }
+    return true
+}
+
+export function useFieldValidation(
+    props,
+    model,
```
pakb, line 31: I don't know if I like that you send the whole `props` Object here, as normally you should send refs or getters to a composable
https://vuejs.org/guide/reusability/composables#conventions-and-best-practices
and use `toValue` to make sure you get the value if it is a ref you got

The less `ref(...)` you declare here, the better it will react to changes in the inputs

src/utils/composables/useFieldValidation.js
ltshb: Done

src/utils/composables/useFieldValidation.js
ltshb: As discussed we leave it but I documented it

src/utils/components/EmailInput.vue
ltshb: Done, hopefully it was easy with search and replace, will see if the test passes 🤞🏼

pakb (approved):

PR #820, opened 2024-05-03
PB-368 : add "load more features" button in feature list

pakb:
Also load 50 features when selecting a rectangle on the map (instead of 10)

The piece of code that was triggering the identification of feature was a store plugin, I thought it would make more sense to move all this to the feature store. Especially as I needed to write a second "get more feature" way of loading things. So now the store plugins only triggers a dispatch instead of loading the feature itself.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-368_load_more_features_button/index.html)


ltshb:
There is a black background around the tooltip of the zoom to extent, this looks quite ugly 
image
what is the reason of this ?

ltshb:
After selecting a bunch of features, clicking outside the selection box where there is nothing to select, the features stay selected when it should clear the selection.

src/api/features/features.api.js
```diff
@@ -118,11 +136,14 @@ async function identifyOnGeomAdminLayer({
                 imageDisplay: `${screenWidth},${screenHeight},96`,
                 geometryFormat: 'geojson',
                 geometryType: `esriGeometry${coordinate.length === 2 ? 'Point' : 'Envelope'}`,
-                limit: featureCount,
+                // there's a hard limit of 50 on our backend
+                // see https://api3.geo.admin.ch/services/sdiservices.html#id10
+                limit: Math.min(featureCount, 50),
```
ltshb, line 141: I would not do this here, we might change the backend in future to allow more than 50 and that will make the code hard to understand. Would be better to document the hard limit of 50 in the jsdoc of this function but that's all, if we update the limit to 100, then the backend will just fail and that's fine.

src/modules/infobox/components/FeatureList.vue
```diff
@@ -26,12 +28,26 @@ const layerFeatureCategories = ref([])
...
+const canLoadMore = computed(() => (layerId) => {
+    // if the app was loaded with pre-selected features there won't be a clickInfo to latch upon,
+    // so we won't be able to load more features where the user has previously selected these features (we don't know where it was)
+    return (
+        lastClick.value &&
+        selectedFeaturesByLayerId.value.find(
+            (featuresForLayer) => featuresForLayer.layerId === layerId
+        )?.featureCountForMoreData > 0
+    )
+})
```
ltshb, line 47: This would make more sense to move this to `FeatureListCategory.vue` with the `loadMoreResultForLayer()` function below. This would avoid to do the the search in active layer twice

src/store/modules/features.store.js
```diff
@@ -1,8 +1,15 @@
...
+import { identify, identifyOnGeomAdminLayer } from '@/api/features/features.api'
+import LayerFeature from '@/api/features/LayerFeature.class'
+import getProfile from '@/api/profile/profile.api'
+import { flattenExtent } from '@/utils/coordinates/coordinateUtils'
 import { allStylingColors, allStylingSizes } from '@/utils/featureStyleUtils'
-import log from '@/utils/logging.js'
+import log from '@/utils/logging'
+
+const DEFAULT_FEATURE_COUNT_SINGLE_POINT = 10
+const DEFAULT_FEATURE_COUNT_RECTANGLE_SELECTION = 50
```
ltshb, line 12: Just wondering if a max of 25 would make more sense and be more performant ? Did you test locally with the prod backend to see if it is faster ? Currently using 50 is quite slow in comparison to the old map geoadmin.

src/store/modules/features.store.js
```diff
@@ -20,11 +27,120 @@ const getEditableFeatureWithId = (state, featureId) => {
...
+ *   either a single point, or an extent (expressed as [minX, maxY, minY, maxY])
+ * @param {Number} config.resolution The current map resolution, in meter/pixel
+ * @param {[Number, Number, Number, Number]} config.mapExtent The current map extent, described as
+ *   [minX, maxX, minY, maxY]
+ * @param {Number} config.screenWidth Current screen width (map width) in pixel
+ * @param {Number} config.screenHeight Current screen height (map height) in pixel
+ * @param {String} config.lang Current lang ISO code
+ * @param {CoordinateSystem} config.projection Wanted projection with which to request the backend
+ * @param {Number} [config.featureCount] How many features should be requested. If not given, will
+ *   default to 10.
```
ltshb, line 50: Wrong default, it depends on the coordinate

src/store/modules/features.store.js
```diff
@@ -20,11 +27,120 @@ const getEditableFeatureWithId = (state, featureId) => {
...
+            .then((responses) => {
+                responses.forEach((response) => {
+                    if (response.status === 'fulfilled') {
+                        allFeatures.push(...response.value)
+                    } else {
+                        log.error('Error while identifying features', response.reason?.message)
+                        // no reject, so that we may see at least the result of requests that have been fulfilled
+                    }
+                })
+                // filtering out doppelgangers
```
ltshb, line 119: doppelgangers => duplicates 😉

pakb:
> There is a black background around the tooltip of the zoom to extent, this looks quite ugly !
what is the reason of this ?

Seems like it was always there, but having 50 features increase the side-effects of the bug. The tippy selector used in the ZoomToExtentButton was re-applying the same tooltip over and over, with 10 it was still "subtle" shadows, but with 50... well you saw

src/api/features/features.api.js
pakb: I'll move this to an app config constant (in `src/config.js`) it should be then easier to adjust this value later on

src/store/modules/features.store.js
pakb: about 100 to 200ms to get 50 features from the PROD backend, then loading each features' detail takes about 500ms (so in less than 1sec, you get your 50 features rendered on the map).
That's "good enough" but the fact that we need to request each feature's detail adds quite a lot of time to the processing (we could maybe parallelize that in the code)

src/modules/infobox/components/FeatureList.vue
pakb: I placed it here because in the Category component we already lost track of which kinds of "item/feature" this is, meaning there could be layer features (those where we want to load more) and vector feature (GeoJSON, KML, etc... for which it isn't possible to load more)

ltkum: We could have the `canLoadMore()` here, as it makes sense to have it where the informations are present. Then, we could call `loadMoreResultForLayer()` in the category, since this is where the button is. This would avoid defining an emitter and search the active layers twice.

ltshb (approved):

PR #821, opened 2024-05-03
PB-482: Elevation profile not updated when moved

LukasJoss:


ltshb (approved):

src/modules/drawing/components/useModifyInteraction.composable.js
```diff
@@ -102,6 +102,12 @@ export default function useModifyInteraction(features) {
                 feature: storeFeature,
                 geometry: new GeoJSON().writeGeometryObject(feature.getGeometry()),
             })
+            if (
+                storeFeature.geometry.type === 'Polygon' ||
+                storeFeature.geometry.type === 'LineString'
+            ) {
+                store.dispatch('setProfileFeature', { feature: storeFeature, ...dispatcher })
+            }
```
pakb, line 110: I'd maybe move that to `features.store.js` where there's already a `canFeatureShowProfile()` helper function,
and so that it is triggered whenever the geometry changes (not only through this modify interaction

pakb: As Lukas is "under the flags" for the foreseeable month, I'll do the change myself

PR #824, opened 2024-05-07
PB-437: Fix print scale

ltshb:
The print scale on mobile was empty. This was due to the fact that we took the menu width in the computation for the scale.

Now we don't take the menu width for the scale computation and we move the print center a bit down due to the header overlapping the map.

Also did some renaming to make the code clearer

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-437-print-scale/index.html)


pakb (commented):
Code looks good but I think you need to edit the print Cypress test to accommodate to your changes

ismailsunni (approved):
LGTM

PR #825, opened 2024-05-07
PB-447: legacy parser not switching to map view

ltkum:
Issue : in some occasions, the router plugins would process the legacy query correctly, but not switch to the map view once it was done, staying stuck in the legacy view forever and not showing anything.

This happened when there was no change to make to the query due to default parameters not being in the URL, but still needing a redirect to the map view.

Fix : If we change the store and the view should change, we ask for a query change too

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-447-broken-legacy-parser/index.html)

### DIFF
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -11,11 +11,11 @@ import {
     MAP_VIEW,
 } from '@/router/viewNames'
 import store from '@/store'
+import log from '@/utils/logging'
 import { parseQuery, stringifyQuery } from '@/utils/url-router'
 import EmbedView from '@/views/EmbedView.vue'
 import LegacyParamsView from '@/views/LegacyParamsView.vue'
 import MapView from '@/views/MapView.vue'
-import log from '@/utils/logging'
 
 const history = createWebHashHistory()
 
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -71,7 +71,7 @@ const router = createRouter({
 })
 
 router.onError((error) => {
-    log.error(error)
+    log.error('[Router error] :', error)
 })
 
 appLoadingManagementRouterPlugin(router, store)
--- a/src/router/index.js
+++ b/src/router/index.js
@@ -15,6 +15,7 @@ import { parseQuery, stringifyQuery } from '@/utils/url-router'
 import EmbedView from '@/views/EmbedView.vue'
 import LegacyParamsView from '@/views/LegacyParamsView.vue'
 import MapView from '@/views/MapView.vue'
+import log from '@/utils/logging'
 
 const history = createWebHashHistory()
 
@@ -67,6 +68,10 @@ const router = createRouter({
     ],
     parseQuery: parseQuery,
     stringifyQuery: stringifyQuery,
+})
+
+router.onError((error) => {
+    log.error(error)
 })
 
 appLoadingManagementRouterPlugin(router, store)
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -23,8 +23,6 @@ import {
 } from '@/utils/legacyLayerParamUtils'
 import log from '@/utils/logging'
 
-const dispatcher = { dispatcher: 'legacyPermalinkManagement.routerPlugin' }
-
 const handleLegacyKmlAdminIdParam = async (legacyParams, newQuery) => {
     log.debug('Transforming legacy kml adminId, get KML ID from adminId...')
     const kmlLayer = await getKmlLayerFromLegacyAdminIdParam(legacyParams.get('adminId'))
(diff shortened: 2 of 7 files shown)


src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -39,7 +39,8 @@ const isRoutePushNeeded = (store, currentRoute) => {
 
 // flag to distinguish URL change originated by this module or by another source
 let routeChangeIsTriggeredByThisModule = false
-
+// flag to tell us if we need to enforce a query redirection when coming from the legacy router
+let needRouteChangeBecauseOfLegacy = true
```
pakb, line 43: I'm not sure it's a good idea to set it to `true` by default.
Maybe it should be set to `true` the first time we enter the LegacyView. If I understand correctly the goal here is to make sure we exit this view, so it would make more sense this way (here I fear it will trigger something unwanted when entering the normal MapView...)

ltkum: That would make more sense. Functionally there is no real issue (the map view will go through the router one additional time, but it won't crash or go infinite), but it would be cleaner to only do it when we went through the legacy router first.

src/router/index.js
```diff
@@ -69,6 +70,10 @@ const router = createRouter({
     stringifyQuery: stringifyQuery,
 })
 
+router.onError((error) => {
+    log.error(error)
```
ltshb, line 74: Maybe add a prefix to the log error `log.error(`Router error: `, error)`

src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -281,6 +281,10 @@ const legacyPermalinkManagementRouterPlugin = (router, store) => {
         )
         let unSubscribeStoreMutation = null
         const unsubscribeRouter = router.beforeEach(async (to, from) => {
+            log.debug('[Legacy URL] entry into the legacy router')
+            log.debug('[Legacy URL] we came from the following route', from)
+            log.debug('[Legacy URL] We should reach the following route', to)
```
ltshb, line 286: Those logs should/could be concatenated into one log

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -100,15 +102,17 @@ function storeMutationWatcher(store, mutation, router) {
...
     if (routeChangeIsTriggeredByThisModule) {
         log.debug(`[Router store plugin] Url query watcher triggered by itself ignore it`, to)
         // Only sync route params when the route change has not been
         // triggered by the sync from store mutations watcher above.
         routeChangeIsTriggeredByThisModule = false
         return undefined
     }
-    log.debug(`[URL query watcher]`, to.query)
+    log.debug('[URL query watcher] coming from :', from?.query)
+    log.debug(`[URL query watcher] going to :`, to.query)
```
ltshb, line 115: Concatenate into one log

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -168,7 +173,10 @@ function urlQueryWatcher(store, to) {
             })
         })
     }
+
     if (requireQueryUpdate) {
+        // if there is a query update, we won't need to ask the viewer to reload because
+        // of going through the legacy parser.
```
ltshb, line 179: Probably a comment left over from you previous solution ?

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -191,9 +199,39 @@ function urlQueryWatcher(store, to) {
...
+
+        log.debug('[Router store plugin] exiting navigation guard with the following value', retVal)
+        // Note we return undefined to validate the route, see Vue Router documentation
+        return retVal
+    })
+    router.afterEach((to, from) => {
         if (MAP_VIEWS.includes(to.name) && !unsubscribeStoreMutation) {
             log.info('[Router store plugin] Entering MapView, register store mutation watcher')
+            log.warn(router.currentRoute.value.name)
+            log.warn('Router Plugin temporary log ', to.name, from.name)
```
ltshb, line 234: Remove this

src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -191,9 +199,39 @@ function urlQueryWatcher(store, to) {
...
+            // is ready
+            retVal = urlQueryWatcher(store, to, from)
+        }
+
+        log.debug('[Router store plugin] exiting navigation guard with the following value', retVal)
+        // Note we return undefined to validate the route, see Vue Router documentation
+        return retVal
+    })
+    router.afterEach((to, from) => {
         if (MAP_VIEWS.includes(to.name) && !unsubscribeStoreMutation) {
```
ltshb, line 231: A blank line before and a comment about why we need to do this afterEach would be good.

ltshb (approved):
Just some minor comments.

PR #827, opened 2024-05-07
PB-481: bugfix apply transparency instead of opacity

ltkum:
Issue : On all labels for layers, we mention that we're modifying the transparency, when we are currently modifying the opacity.

fix : we reverse the value on the transparency slider, committing `1 - value` to the layer's opacity rather than the value.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-481-reverse-transparency-slider/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -93,7 +93,6 @@ function onToggleLayerVisibility() {
 }
 
 function onOpacityChange(e) {
-    
     store.dispatch('setLayerOpacity', {
         index: index.value,
         opacity: 1.0 - e.target.value,
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -205,7 +205,7 @@ function duplicateLayer() {
                 max="1.0"
                 step="0.01"
                 :value="1.0 - layer.opacity"
-                :data-cy="`slider-opacity-layer-${id}-${index}`"
+                :data-cy="`slider-transparency-layer-${id}-${index}`"
                 @change="onTransparencyChange"
             />
             <button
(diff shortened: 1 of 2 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -95,7 +95,7 @@ function onToggleLayerVisibility() {
 function onOpacityChange(e) {
     store.dispatch('setLayerOpacity', {
         index: index.value,
-        opacity: e.target.value,
+        opacity: 1.0 - e.target.value,
         ...dispatcher,
     })
 }
```
ltshb, line 101: The name of the function needs to be changed as it is not opacity but transparency, otherwise the computation don't make sense for reader.

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -204,7 +204,7 @@ function duplicateLayer() {
                 min="0.0"
                 max="1.0"
                 step="0.01"
-                :value="layer.opacity"
+                :value="1.0 - layer.opacity"
                 :data-cy="`slider-opacity-layer-${id}-${index}`"
                 @change="onOpacityChange"
```
ltshb, line 209: See comment above. Also while testing I noticed that a small margin before and also maybe after the slider would be great, could you add it ?
image

ltshb: Testing with `ms-2 me-4` looks better in my opinion 
image

tests/cypress/tests-e2e/layers.cy.js
```diff
@@ -885,9 +885,9 @@ describe('Test of layer handling', () => {
                     .should('be.visible')
                     .click()
                 // change the opacity to check later on that the new layer as the non default opacity
-                cy.get(`[data-cy="slider-opacity-layer-${timedLayerId}-2"]`)
+                cy.get(`[data-cy="slider-transparency-layer-${timedLayerId}-2"]`)
                     .should('be.visible')
-                    .realClick()
+                    .realClick({ position: 'right', force: true })
```
ltshb, line 890: Force true is not a good idea why do you need it ?

ltkum: that's a leftover from when I was testing. I thought that maybe something was blocking the click. I'm removing it right now.

ltshb (approved):

PR #828, opened 2024-05-13
PB-496: External WMTS layer in 3d

ltshb:
TODO in separate PR:
- [ ] time series support for external WMS
- [ ] Proper time series update (see /PB-518)
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-496-external-wmts-3d/index.html)

### DIFF
--- a/src/api/layers/ExternalLayer.class.js
+++ b/src/api/layers/ExternalLayer.class.js
@@ -94,6 +94,11 @@ export default class ExternalLayer extends AbstractLayer {
      * @param {ExternalLayerGetFeatureInfoCapability | null} [externalLayerData.getFeatureInfoCapability=null]
      *   Configuration describing how to request this layer's server to get feature information.
      *   Default is `null`
+     * @param {LayerTimeConfig | null} [externalLayerData.timeConfig=null] Time series config (if
+     *   available). Default is `null`
+     * @param {Number} [externalWmtsData.currentYear=null] Current year of the time series config to
+     *   use. This parameter is needed as it is set in the URL while the timeConfig parameter is not
+     *   yet available and parse later on from the GetCapabilities. Default is `null`
      * @throws InvalidLayerDataError if no `externalLayerData` is given or if it is invalid
      */
     constructor(externalLayerData) {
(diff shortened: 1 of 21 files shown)


src/api/layers/ExternalLayer.class.js
```diff
@@ -148,5 +156,9 @@ export default class ExternalLayer extends AbstractLayer {
             this.availableProjections.push(WGS84)
         }
         this.getFeatureInfoCapability = getFeatureInfoCapability
+        this.currentYear = currentYear
```
pakb, line 159: does it really make sense to store this value, if we do not react/update the time config whenever it changes?

src/modules/map/components/openlayers/OpenLayersExternalWMTSLayer.vue
```diff
@@ -29,11 +30,39 @@ const { externalWmtsLayerConfig, parentLayerOpacity, zIndex } = toRefs(props)
...
+const previewYear = computed(() => store.state.layers.previewYear)
+const isTimeSliderActive = computed(() => store.state.ui.isTimeSliderActive)
 
 // extracting useful info from what we've linked so far
 const layerId = computed(() => externalWmtsLayerConfig.value.id)
 const opacity = computed(() => parentLayerOpacity.value ?? externalWmtsLayerConfig.value.opacity)
-const options = computed(() => externalWmtsLayerConfig.value.options)
+const options = computed(() => {
+    const options = { ...externalWmtsLayerConfig.value.options }
+    options.dimensions = { ...options.dimensions }
```
pakb, line 41: Might make sense that we add some cloneDeep functions from Lodash or other library for such cases

pakb (approved):
That was a lot to add to achieve the wanted result :open_mouth:

src/api/layers/ExternalLayer.class.js
ltshb: It is used by the external layer plugin once the capabilities has been loaded, so it is required

ltshb:
> That was a lot to add to achieve the wanted result 😮

I went a bit further by enabling time layer in external import as well 😉

PR #830, opened 2024-05-13
PB-493: Fixed Time slider input cutoff on 3d Maps

schtibe:
**RFC**
This could be one possible solution. It only addresses the font and the margin of the time slider though. There could potentially be other things that break with the cesium style. (1)

Another solution could be to re-order the css. For that, we'd have to prematurely load Cesium's `widgets.css` I guess, before our other rules are applied. The drawback would be that there's always cesium style around even if the 3d map isn't used. 

Open for other suggestions.

(1)
Although, when scanning through the `Cesium/Widgets/widgets.css`, the only other global style apart from `input` is `ul`. Everything else is a `.cesium-` prefixed class. So maybe it won't break too much.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-jira-pb-493-time-slider-cutoff/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -528,6 +528,9 @@ $time-slider-color-has-data: color.adjust($primary, $lightness: 30%);
         width: 32px;
         text-align: center;
     }
+    input {
+        margin: 0;
+    }
 }
 .time-slider-bar-cursor-year {
     &.form-control {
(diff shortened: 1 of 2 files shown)


ltshb (approved):
Result looks good to me and the proposed solution is good enough for the go live in my opinion, @pakb ?

pakb (approved):
Yep, with the comment on why we have this `#cesium` selector, that's quite good enough for a go-live without touching too much how we import Cesium's CSS :+1:

PR #833, opened 2024-05-13
PB-468 Fix arrow shift of time slider on the edges

schtibe:
The time band was so wide that the cursor was rendered weirdly when dragged to the very end of the time band:

image

On suggestion of Brice, I narrowed the band so that the cursor will always have the necessary
space to be displayed properly:

Screenshot from 2024-05-13 16-47-48

In the process, I also fixed a little display bug, where the arrow wasn't really in the middle of the cursor (or rather, the cursor wasn't centered above the arrow):

image

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-468-time-slider-arrow-shift/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -76,22 +76,16 @@ const innerBarStyle = computed(() => {
     return { width: `${sliderWidth.value}px` }
 })
 const yearPositionOnSlider = computed(
-    () => (1 + ALL_YEARS.indexOf(currentYear.value)) * distanceBetweenLabels.value
+    () => (1 + ALL_YEARS.indexOf(currentYear.value)) * distanceBetweenLabels.value + 42
 )
 const cursorPosition = computed(() => {
     const yearCursorWidth = yearCursor.value?.clientWidth || 0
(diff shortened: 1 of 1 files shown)


ltkum (commented):
It works quite well, but we encounter an issue where we can reach the end of the slider on the left, but not on the right. The smaller the width, the more visible it is.

After trying it on the current implementation, it looks like this issue is also present, event if it's a lot less visible, since the margins are smaller.

Screenshot from 2024-05-13 17-08-27

As we are here in the middle of reworking the CSS, it could make sense to change the CSS to be more dynamic (perhaps in another PR). The old mapviewer used a dropdown menu to avoid the issue completely.

That being said, the arrow seem centered and keep being in place. So, this looks good to me.

I'd like to see with @ltshb what we should plan for the CSS issue.

ltshb (approved):
Looks good, there are other issues as @ltkum said and also on the right the bar is too long but those issues are not related to this change which makes the bar better.
In my opinion for the issue reported by @ltkum we need to use a dropdown and the issue is only present for very small device, Iphone SE is OK (although for iphone a dropdown could make more sense). The dropdown will be done after go live.

PR #834, opened 2024-05-13
PB-491: Added 3D menu

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-491-3d-menu/index.html)


ltkum (commented):
Seems like good work, I did not manage to find something to complain about in the code.

Regarding the failing tests:

I think most of your 3d layers test fail currently because of the fact that in the previous implementation, labels were displayed by default (which is no longer the case)

The menuTray test might be failing because we added a new menu section, changing the size of the menu
The test on external wms layers fails because the menu is now too big, and we would need to scroll to reach the transparency slider.

src/modules/menu/components/menu/MenuTray.vue
```diff
@@ -1,29 +1,116 @@
...
+    toClose.forEach((section) => refs.value[section]?.close())
+}
+
+function onCloseMenuSection(id) {
+    if (['drawSection', 'toolsSection'].includes(id)) {
+        refs.value.activeLayersSection.open()
+    }
+}
+
+function updateRef(el) {
```
schtibe, line 77: Maybe this could benefit of a littledocumentation as it's not directly clear what it's doing

ltshb:
We decided to use #835

PR #835, opened 2024-05-14
PB-491: 3D menu only when active

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-491-3d-menu-2/index.html)


src/modules/menu/components/menu/MenuTray.vue
```diff
@@ -1,10 +1,102 @@
...
+const multiMenuSections = ref(['topicsSection', 'activeLayersSection'])
+// singleModeSections means that those section cannot be open together with other
+// sections and would therefore toggle other sections automatically.
+const singleModeSections = ref([
+    'drawSection',
+    'settingsSection',
+    'shareSection',
+    'toolsSection',
+    'printSection',
+    '3dSection',
```
pakb, line 39: I think we should let this be in the `multiMenuSections`, I can see users wanting to de-activate vegetation/building while setting up or ordering up layers, and be bothered that "Map displayed" has been closed when changing 3D settings

src/modules/menu/components/menu/MenuTray.vue
```diff
@@ -1,10 +1,102 @@
...
+    }
+}
+</script>
+
 <template>
     <div data-cy="menu-tray-inner" :class="[{ 'menu-tray-compact': compact }, 'menu-tray-inner']">
         <MenuSection
             id="settingsSection"
-            ref="settingsSection"
+            :ref="addRefById"
```
pakb, line 97: it would also be possible to directly give the correct "ref" name here (`multiMenuSections` or the single equivalent), having the same ref on multiple element makes them available as an array of ref in the `setup`

src/setup-fontawesome.js
```diff
@@ -66,6 +66,7 @@ import {
     faSpinner,
     faStar,
     faT,
+    faTags,
```
pakb, line 69: I don't think we are using it after the switch to text/checkbox instead of icon/button, might be possible to remove it

src/utils/composables/useTippyTooltip.js
```diff
@@ -93,7 +93,8 @@ export function useTippyTooltip(
         tooltips = tippy(selector, {
             ...options,
             arrow: true,
-            touch: false,
+            // Show tippy on long touch for mobile device
+            touch: ['hold', 500], // 500ms delay,
```
pakb, line 97: :+1:

src/modules/menu/components/menu/MenuTray.vue
ltshb: But then you have an array and not an object and close/open a specific section you will need to do a search in the array by id.

src/modules/i18n/locales/en.json
```diff
@@ -123,8 +123,8 @@
     "draw_next_measure": "Click to add another point (double-click to finish)",
     "draw_panel_title": "Draw & Measure on map",
     "draw_point": "Point",
-    "draw_point_description": "Create point",
-    "draw_point_instructions": "Create point",
+    "draw_point_description": "",
+    "draw_point_instructions": "import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'",
```
pakb, line 127: While working on `mf-geoadmin3` I saw this weird change coming there too, I updated/reverted this change in the spreadsheet, you might want to update the translation on your PR too :wink:

ltshb: Done

pakb (commented):
image
The height of each element doesn't match between the two, could you shrink the 3D element a little so that it matches the height of the "Map displayed' one?

pakb (approved):

PR #836, opened 2024-05-14
PB-486: Fix print area when map is rotated

ismailsunni:
The previous code from the old geoadmin was working by chance and when there is an internal change in the OpenLayers, it breaks. See the further discussion https://github.com/openlayers/openlayers/issues/15820

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-486-rotating-map-print-area/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/printConstants.js
+++ b/src/modules/map/components/openlayers/utils/printConstants.js
@@ -1,6 +1,0 @@
-/**
- * The name of the OpenLayer layer that is used to render the print area.
- *
- * @type {String}
- */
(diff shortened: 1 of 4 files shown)


ltshb (commented):
Looks good, just one issue
1. Open print menu
2. Switch to 3d
3. Rotate the map
4. Exit 3d
5. Go back to print => no print layout
7. close print and reopen again => print layer
Maybe force the closing of the print menu when entering 3d.

ismailsunni:
@ltshb done

ltshb (approved):

PR #842, opened 2024-05-14
PB-507: Fix logging in production build

ltshb:
In production build debug and info messages were printed when they shouldn't

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-507-log/index.html)

### DIFF
--- a/src/utils/logging.js
+++ b/src/utils/logging.js
@@ -24,8 +24,8 @@ export const LogLevel = {
  * @param {...any} message
  */
 const log = (level, ...message) => {
-    // In production we don't log debug level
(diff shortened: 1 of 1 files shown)


ismailsunni (commented):
LGTM, although I need to re-read to make sure my logic is correct.

I was wondering why not write it as
```
    if (ENVIRONMENT === 'production' && [LogLevel.DEBUG, LogLevel.INFO].includes(level)) {
        return
    }
```

Also, in line 21, it says that we only calls with `LogLevel.Error`, which different compare to the updated code.

pakb (approved):

PR #846, opened 2024-05-16
PB-501 Show dropdown for year selection in the time slider on tablet screens

schtibe:
On small screens (but not mobile) show a Dropdown instead of the time slider.

- Add Dropdown on the time slider for small devices such as tablets
- Fix the play button icon
- Refactor the data flow in the time slider
- Prevent from updating layers with invalid year data
- Debounce the store updating

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-501-dropdown-timeslider-on-tablets/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -344,7 +344,7 @@ function setYearToInputIfValid() {
         class="time-slider card"
         :class="{ grabbed: yearCursorIsGrabbed }"
     >
-        <div class="p-2 d-flex">
+        <div class="p-2 d-flex align-items-center justify-content-between">
             <div class="time-slider-bar px-5">
                 <div
                     ref="yearCursor"
@@ -418,6 +418,19 @@ function setYearToInputIfValid() {
                     </small>
                 </div>
             </div>
+
+            <div class="time-slider-dropdown">
+                <select
+                    class="form-select ml-2"
+                    :value="currentYear"
+                    @change="setCurrentYearAndDispatchToStore($event.target.value, true)"
+                >
+                    <option v-for="year in allYears" :key="year" :value="year">
+                        {{ year }}
+                    </option>
+                </select>
+            </div>
+
             <button
                 ref="playButton"
                 data-cy="time-slider-play-button"
@@ -450,6 +463,8 @@ function setYearToInputIfValid() {
 
 <style lang="scss">
 @use 'sass:color';
+
+@import '@/scss/media-query.mixin';
 @import '@/scss/webmapviewer-bootstrap-theme';
 
 $time-slider-color-background: color.adjust($white, $alpha: -0.1);
@@ -488,6 +503,47 @@ $time-slider-color-partial-data: color.adjust($primary, $lightness: 45%);
     border-color: $time-slider-color-has-data;
     background: $time-slider-color-has-data;
 }
+
+// Display the dropdown instead of the time slider on small screens (tablets),
+// but not on mobile
+.time-slider-dropdown {
+    display: none;
+}
+
+.time-slider {
+    width: 100%;
+}
+
+@include respond-above(sm) {
+    // dropdown mode
+    .time-slider-dropdown {
+        display: block;
+    }
+
+    .time-slider-bar {
+        display: none;
+    }
+
+    .time-slider {
+        width: auto;
+    }
+}
+
+@include respond-above(lg) {
+    // time slider mode
+    .time-slider-bar {
+        display: block;
+    }
+
+    .time-slider-dropdown {
+        display: none;
+    }
+
+    .time-slider {
+        width: 100%;
+    }
+}
+//
 
 .time-slider {
     background: $time-slider-color-background !important;
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -29,7 +29,7 @@ const currentYear = ref(YOUNGEST_YEAR)
 const displayedYear = ref(YOUNGEST_YEAR)
 const isPristine = ref(true)
 let cursorX = 0
-let playYearsWithData = false
+const playYearsWithData = ref(false)
 let yearCursorIsGrabbed = false
 let playYearInterval = null
 
@@ -292,8 +292,8 @@ function releaseCursor() {
 }
 
 function togglePlayYearsWithData() {
-    playYearsWithData = !playYearsWithData
-    if (playYearsWithData) {
+    playYearsWithData.value = !playYearsWithData.value
+    if (playYearsWithData.value) {
         let yearsWithDataForPlayer = ALL_YEARS.filter(
             (year) =>
                 yearsWithData.value.yearsJoint.includes(year) ||
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -68,6 +68,26 @@ const { tippyInstance: tippyOutsideRange, updateTippyContent } = useRangeTippy(
 )
 
 /**
+ * Debounce the input year a bit
+ *
+ * With this, the error won't be shown immediately while the user is still typing a year, but rather
+ * only when they have finished typing
+ */
+const updateInputYear = debounce((value) => {
+    value = parseInt(value)
+    // only if the year is valid we write this to the property
+    // otherwise we show errors
+    if (!allYears.value.includes(value)) {
+        isInputYearValid.value = false
+        falseYear.value = value || ''
+    } else {
+        isInputYearValid.value = true
+        currentYear.value = parseInt(value)
+        falseYear.value = null
+    }
+}, 500)
+
+/**
  * Used for the year in the input field Validate the input from the user. In case it's invalid, we
  * don't propagate the value to the state, but instead save it to an intermediate state variable to
  * be displayed along with an error message. This is important so that the cursor doesn't jump
@@ -81,18 +101,7 @@ const inputYear = computed({
         return currentYear.value
     },
     set(value) {
-        value = parseInt(value)
-
-        // only if the year is valid we write this to the property
-        // otherwise we show errors
-        if (!allYears.value.includes(value)) {
-            isInputYearValid.value = false
-            falseYear.value = value || ''
-        } else {
-            isInputYearValid.value = true
-            currentYear.value = parseInt(value)
-            falseYear.value = null
-        }
+        updateInputYear(value)
     },
 })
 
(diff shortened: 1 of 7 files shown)


ltshb:
Should be a bit shrinked 
image

schtibe:
@ltshb Done.

I was also thinking: the dropdown probably isn't very user friendly when it contains 150+ entries. Maybe we can (some time) search for other solutions.

Some spontaneous ideas:
1) Have a dropdown where you can also enter the year (a searchable dropdown component or so)
2) Have the time slider at the bottom of the screen. There's much more space there.
3) Have a vertical time slider...

(speaking of entering numbers: should we use `<input type="number">` in the time slider input?)

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -49,12 +49,56 @@ const layersWithTimestamps = computed(() =>
...
+ * don't propagate the value to the state, but instead save it to an intermediate state variable to
+ * be displayed along with an error message. This is important so that the cursor doesn't jump
+ * around to invalid years
+ */
+const inputYear = computed({
+    get() {
+        return falseYear.value || currentYear.value
+    },
+    set(value) {
+        if (value.length < 4) return value
```
ltshb, line 66: From our guideline we always use `if()` with brackets
```javascript
if (...) {
   ...
}
```

ltshb: Moreover this is an issue, if you enter a too small number, like 200, the validation is not done and you don't have a user feedback, even if you press enter.

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -49,12 +49,56 @@ const layersWithTimestamps = computed(() =>
...
+        return _currentYear.value
+    },
+    set(value) {
+        _currentYear.value = value
+
+        // we're currently dragging the cursor, don't update the store
+        if (yearCursorIsGrabbed) return
+
+        dispatchCurrentYearToStore()
+    },
```
ltshb, line 100: For the discussion about early return, I think here is a good example of where the early returns make the function harder to read as you have a return in a middle and having such 

```javascript
    set(value) {
        _currentYear.value = value
        // we're currently dragging the cursor, don't update the store
        if (!yearCursorIsGrabbed) {
            dispatchCurrentYearToStore()
        }
    },
```
Especially here you don't have any issue with indentation because you have everything in a function call.
Moreover always use `if (...)` with brackets

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -174,20 +211,20 @@ onMounted(() => {
...
-        })
-    }
+    activeLayers.value.forEach((layer, index) => {
+        if (layer.hasMultipleTimestamps) {
+            store.dispatch('setTimedLayerCurrentYear', {
+                index,
+                year: previewYear.value,
+                ...dispatcher,
+            })
+        }
```
ltshb, line 225: This was probably already a bug before, but here you cannot add the previewYear to all layer, as you might have layer with invalid timestamp, like in this example https://sys-s.dev.bgdi.ch/rdutl0i1t7wz

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -198,11 +235,7 @@ function setTooltipContent() {
     tippyInstance.setContent(tooltipContent)
 }
 
-function setCurrentYearAndDispatchToStore(year, removePristineStatus = false) {
-    if (removePristineStatus) {
-        isPristine.value = false
-    }
-    currentYear.value = year
+function dispatchCurrentYearToStore() {
```
ltshb, line 238: If already refactoring I would change this function name to `dispatchPreviewYear` or even better `setPreviewYear`.

Note that `previewYear` is used for the time slider and saved in url.

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -268,21 +302,25 @@ function listenToMouseMove(event) {
...
-        isPristine.value = false
-    }
-    store.dispatch('setPreviewYear', { year: currentYear.value, ...dispatcher })
+
+    // We call this explicitly here. So far, the store wasn't updated,
+    // as this was prevented by the yearCursorIsGrabbed switch.
+    dispatchCurrentYearToStore()
+    // another cool way would be to use a debounce on the dragging. Then
+    // the map gets updated when the user stops dragging, and not just when
+    // they release the mouse
```
ltshb, line 318: The debouncing is a very good remark and good idea ! On the old viewer if you grab the slider move it and keep it dragged then the year is changed while now on the view you have to let go to set the year. Note on currently we have a flickering issue with the slider that I'm fixing with PB-518. So it would be very nice to do the deboucing.

src/modules/map/components/toolbox/TimeSliderButton.vue
```diff
@@ -54,7 +54,9 @@ function toggleTimeSlider() {
                 'dev-disclaimer-present': hasDevSiteWarning,
             }"
         >
-            <TimeSlider v-if="isTimeSliderActive" />
+            <div class="d-flex justify-content-center">
+                <TimeSlider v-if="isTimeSliderActive" />
+            </div>
```
ltshb, line 59: the dropdown is not displayed correclty (year is cut off) on small device 
image

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -409,6 +437,15 @@ function setYearToInputIfValid() {
...
                 </div>
             </div>
+
+            <div class="time-slider-dropdown" data-cy="time-slider-dropdown">
+                <select v-model="currentYear" class="form-select ml-2">
+                    <option v-for="year in allYears" :key="year" :value="year">
+                        {{ year }}
+                    </option>
+                </select>
+            </div>
```
ltshb, line 447: Here it would be better/nicer to have an input-group with the play button something similar to 
image

About the search feature for the year it is a good idea and we can use the same logic as for the external layer provider dropdown, so the user could enter the year and that would filter, see how the external provider list is done.

ltshb: I don't think we would need a clear button for this use case though.

src/modules/map/components/toolbox/TimeSlider.vue
schtibe: Damn. Good Catch.

src/modules/map/components/toolbox/TimeSlider.vue
schtibe: Totally agree, I've done that in a later commit (I'll rebase them today)

src/modules/map/components/toolbox/TimeSlider.vue
schtibe: Hmm, definitely true :+1: . I think this is a leftover from where there was more in the function before I changed stuff. Now it doesn't make sense anymore.

src/modules/map/components/toolbox/TimeSlider.vue
schtibe: Althought we want to display the error only when the user finished entering... otherwise, say you start entering 184 you already get an error, but in case you wanted to type 1845, which would be OK, then the error is too early 🤔

ltshb: Do the validation on out of focus or when the user press enter 😉

ltshb: See TextInput component on how validation is done, maybe we could reuse this component.

src/modules/map/components/toolbox/TimeSlider.vue
schtibe: Fixed it, it doesn't update the layers that won't have the year available. I can't push `null` at this point, this would need further development.

src/modules/menu/components/common/SearchableDropdown.vue
```diff
@@ -0,0 +1,128 @@
...
+                <input
+                    :value="displayEntry"
+                    class="form-control rounded-end-0"
+                    @input="inputValue = $event.target.value"
+                    @focusin="openList"
+                    @keydown.esc.prevent="closeList"
+                    @keydown.down.prevent="focusSearchlist"
+                />
+
+                <div class="input-group-append btn-group btn-group-sm">
```
ltshb, line 102: Might be a better idea to use bigger button for touch device, so either the btn-group or btn-group-lg. Now especially the dropdown button is a bit small and difficult to touch on mobile.

ltshb (commented):
It kind of work on the tablet, but the buttons are a bit small.
One general issue is that when typing a number in the dropdown and press enter it action the play button instead of validating the entry. This happens when you enter an invalid number like 100. In this case the field should be marked as invalid as on the slider.

src/modules/map/components/toolbox/TimeSliderDropdown.vue
```diff
@@ -0,0 +1,184 @@
...
+                    @input="inputValue = $event.target.value"
+                    @focusin="openList"
+                    @keydown.esc.prevent="closeList"
+                    @keydown.down.prevent="focusSearchlist"
+                    @keydown.enter.prevent="onEnter"
+                />
+
+                <div class="input-group-append btn-group btn-group">
+                    <button
+                        class="btn border btn-light d-flex align-items-center rounded-start-0 rounded-end-0"
```
ltshb, line 152: For code consistency you should use `btn-outline-group` instead of `btn-light` for button groups.

ltshb (approved):
Very nice, just two minor issues:
- Button class consistency
- When entering an invalid year input, the tippy for invalid year is shown (which is correct), but then if you change the language the tippy lang is updated and stays display but the year is reseted to the previous selected year which is valid. In this case we need either to not reset to the previous selected year or also reset the tippy

src/modules/map/components/toolbox/TimeSlider.vue
```diff
@@ -194,29 +247,36 @@ onMounted(() => {
...
+            layer.timeConfig.getTimeEntryForYear(year)
+        ) {
+            store.dispatch('setTimedLayerCurrentYear', {
+                index,
+                year,
+                ...dispatcher,
+            })
+        }
+    })
+    tippyTimeSliderInfo?.destroy()
```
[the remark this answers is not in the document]
schtibe: Oh yes, that's a rebasing error, thanks!

src/modules/map/components/toolbox/TimeSliderDropdown.vue
schtibe: OK! The name is a bit misleading ;)

PR #847, opened 2024-05-16
PB-506 : embed sends information to parent about selected features

pakb:
using postMessage API to send the feature ID and layer ID of selected features to parent

Can be tested in the this codepen (by changing the URL from test.map.geo.admin.ch to this PR's testlink) [codepen here](https://codepen.io/geoadmin/pen/dyEXVKo)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-506_iframe_feature_selection_event/index.html)

### DIFF
--- a/src/views/EmbedView.vue
+++ b/src/views/EmbedView.vue
@@ -1,7 +1,8 @@
 <script setup>
-import { computed, onBeforeMount, onMounted } from 'vue'
+import { computed, onBeforeMount, onMounted, watch } from 'vue'
 import { useStore } from 'vuex'
 
+import { sendFeatureInformationToIFrameParent } from '@/api/iframeFeatureEvent.api'
 import I18nModule from '@/modules/i18n/I18nModule.vue'
 import InfoboxModule from '@/modules/infobox/InfoboxModule.vue'
 import MapFooter from '@/modules/map/components/footer/MapFooter.vue'
@@ -17,14 +18,22 @@ const dispatcher = { dispatcher: 'EmbedView.vue' }
 const store = useStore()
 
 const is3DActive = computed(() => store.state.cesium.active)
+const selectedLayerFeatures = computed(() => store.getters.selectedLayerFeatures)
 
 onBeforeMount(() => {
     store.dispatch('setEmbed', { embed: true, ...dispatcher })
 })
 
 onMounted(() => {
-    log.info(`Map view mounted`)
+    log.info(`Embedded map view mounted`)
 })
+
+// as described by this example on our documentation : https://codepen.io/geoadmin/pen/yOBzqM?editors=0010
+// our embed view should send a message (see https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)
+// when a feature is selected while in embedded mode, so that the parent can get the selected feature(s) ID(s)
+watch(selectedLayerFeatures, () =>
+    sendFeatureInformationToIFrameParent(selectedLayerFeatures.value)
+)
 </script>
 
 <template>
--- a/src/api/iframeFeatureEvent.api.js
+++ b/src/api/iframeFeatureEvent.api.js
@@ -1,0 +1,45 @@
+import log from '@/utils/logging.js'
+
+/**
+ * Sends information to the iFrame's parent about features, through the use of the postMessage
+ * HTML/Javascript API.
+ *
+ * @param {LayerFeature[]} features List of features for which we want to send information to the
+ *   iFrame's parent
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
+ */
+export function sendFeatureInformationToIFrameParent(features) {
+    const targetWindow = parent ?? window.parent ?? window.opener ?? window.top
+    if (!targetWindow) {
+        log.debug(
+            'Embed view loaded as root document of a browser tab, cannot communicate with opener/parent'
+        )
+        return
+    }
+    // if no features are given, nothing to do
+    if (features?.length === 0) {
+        return
+    }
+    log.debug('sending information about selected features to iframe parent')
+    // from what I can understand from the codepen, one event is fired per feature with a structured response
+    features.forEach((feature) => {
+        targetWindow.postMessage(
+            {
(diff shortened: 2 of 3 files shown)


src/views/EmbedView.vue
```diff
@@ -25,6 +26,31 @@ onBeforeMount(() => {
 onMounted(() => {
     log.info(`Map view mounted`)
 })
+
+// as described by this example on our documentation : https://codepen.io/geoadmin/pen/yOBzqM?editors=0010
+// our embed view should send a message (see https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage)
+// when a feature is selected while in embedded mode, so that the parent can get the selected feature(s) ID(s)
+watch(selectedLayerFeatures, () => {
+    if (selectedLayerFeatures.value.length > 0) {
```
schtibe, line 34: Just a minor thing, a general idea: we used to use the concept of early returns. This way, the function becomes a bit more readable, in my opinion. Less indentation.

so instead of having
```
function foo(param) {
   if (isGood(param)) {
       // do all the things
   }
}
```
we proposed to do it like this
```
function foo(param) {
   if (!isGood(param)) {
     return
   }
   
   // do all the things here
}
```
Just an idea I want to drop here :)

schtibe (commented):
It doesn't work on the test link, or maybe I don't do it correctly (I selected public transport stops). I'm gonna check out the branch tomorrow and test it locally again!

src/views/EmbedView.vue
pakb: I also quite like this approach, some other don't, so I complied and didn't use it much :wink: 
but you have my vote!

pakb:
> It doesn't work on the test link, or maybe I don't do it correctly (I selected public transport stops). I'm gonna check out the branch tomorrow and test it locally again!

right... I have not added a relative URL to the debug view but a localhost one... Will update that tomorrow

src/views/EmbedView.vue
schtibe: OK Cool :star_struck: Let's see if it changes some time :wink:

src/views/EmbedView.vue
```diff
@@ -17,13 +17,53 @@ const dispatcher = { dispatcher: 'EmbedView.vue' }
...
+                    featureId: feature.id,
+                    // if we want to expose more stuff from our features (EGID, EWID, etc...), it should come here...
+                },
+            },
+            // meaning anyone, any host, can receive this event when adding our app as embedded on their website,
+            // so let's be cautious with what we add to the payload
+            '*'
+        )
+        // mf-geoadmin3 was also sending the same feature in a different unstructured/string format "layerId#featureId"
+        // but this comment here https://github.com/geoadmin/mf-geoadmin3/blob/6a7b99a2cc9980eec27b394ee709305a239549f1/src/components/tooltip/TooltipDirective.js#L661-L668
```
ltshb, line 63: I proposed to move that code to an api file and out of the EmbedView component.
Some user uses the normal view in iframe and I'm not if we should also add this to the normal map view. But simply for code organization would be better to just have a function call here, or use a composable at best.

src/router/index.js
```diff
@@ -15,9 +15,59 @@ import { parseQuery, stringifyQuery } from '@/utils/url-router'
...
+]
+
+if (store.getters.hasDevSiteWarning) {
+    // adding a special route to the embed test view
+    routes.push({
+        path: '/test-embed',
+        name: 'TestEmbedView',
+        component: TestEmbedView,
+    })
+}
```
ltshb, line 69: I'm not a big fan of this test view, especially also because it doesn't really do a realistic test as iframe will always be used outside of the same domain, this could result to CORS issue not being found by this. I think codepen are better for testing such think. 
Moreover this view don't work with test link.

src/views/TestEmbedView.vue
```diff
@@ -0,0 +1,58 @@
+<script setup>
+import { onMounted, ref } from 'vue'
+
+import log from '@/utils/logging'
+
+const embed = ref(null)
+const lastEventTimestamp = ref(null)
+
+const layerIDsForIFrame = ['ch.bav.haltestellen-oev']
+const iFrameUrl = `${window.location.protocol}//${window.location.host}/#/embed?layers=${layerIDsForIFrame.join(',')}`
```
ltshb, line 10: This don't work with test link

src/views/EmbedView.vue
ltshb: Early returns at the top of the function are ok and good in such example as above or like here, but I would not do returns in the middle of a function like this

```javascript
function foo(param) {
  if (!isGood(param)) {
    return
  }
  // do something
  if (...) {
    return
  }
  // do something else
   
  // final return
  return
}
```
This would be a very poor and hard code to read as some returns would be hidden in the middle of the function.

src/router/index.js
pakb: I thought it would be helpful to check that it was working without having to write some codepen or other helper, but it isn't as good as expected.
When self-hosting the iframe (same domain) there are some weird things happening in the background.
I was able to use this test view locally all fine, but this doesn't work when deployed on our DEV staging...

I'll remove it and copy the legacy codepen to make it point to the DEV staging (and update it subsequently as soon as we have that deployed on the test viewer)

src/views/TestEmbedView.vue
pakb: view will be removed as it doesn't work on deployed environment (only localhost works)

src/views/EmbedView.vue
schtibe: @ltshb Agree!

schtibe (approved):
Apart from the little detail about the optional chaining operator, which is maybe irrelevant, this looks good to me!

ltshb (commented):
@pakb why did you not implemented it as well in the map view ? iframe don't always uses the embed view and some user might use this feature with the mapview !

pakb:
> @pakb why did you not implemented it as well in the map view ? iframe don't always uses the embed view and some user might use this feature with the mapview !

good point, see https://github.com/geoadmin/web-mapviewer/pull/863

PR #849, opened 2024-05-21
PB-518: Fix timeslider flickering and WMS time enabled layer update

ltshb:
Now when clicking on play on the time slider the layer is updated and not
removed/added which made a flickering effect.

Also solve the WMS time enabled layer which were not updated upon timestamp changes (via time config or time slider).

NOTE: The time slider flickering effect for 3D mode is not solved with this and cannot be easily solved as apparently Cesium doesn't support to update a layer, or at least I could not find a way.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-518-slider-smooth/index.html)

### DIFF
--- a/src/utils/layerUtils.js
+++ b/src/utils/layerUtils.js
@@ -46,14 +46,22 @@ export function getTimestampFromConfig(config, previewYear, isTimeSliderActive) 
 /**
  * @param {GeoAdminWMTSLayer | ExternalWMTSLayer} wmtsLayerConfig
  * @param {CoordinateSystem} projection
- * @param {Number} previewYear
- * @param {Boolean} isTimeSliderActive
+ * @param {Boolean} [options.addTimestamp=false] Add the timestamp from the time config or the
+ *   timeslider to the ur. When false the timestamp is set to `{Time}` and need to processed later
+ *   on. Default is `false`
+ * @param {Number | null} [options.previewYear=null] Default is `null`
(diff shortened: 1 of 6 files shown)


src/utils/layerUtils.js
```diff
@@ -46,14 +46,22 @@ export function getTimestampFromConfig(config, previewYear, isTimeSliderActive)
...
+ * @param {Boolean} [options.addTimestamp=false] Add the timestamp from the time config or the
+ *   timeslider to the ur. When false the timestamp is set to `{Time}` and need to processed later
+ *   on. Default is `false`
+ * @param {Number | null} [options.previewYear=null] Default is `null`
+ * @param {Boolean | null} [options.isTimeSliderActive=null] Default is `null`
  * @returns {String | null}
  */
-export function getWmtsXyzUrl(wmtsLayerConfig, projection, previewYear, isTimeSliderActive) {
+export function getWmtsXyzUrl(wmtsLayerConfig, projection, options = {}) {
+    const { addTimestamp = false, previewYear = null, isTimeSliderActive = null } = options ?? {}
```
ltkum, line 57: `isTimeSliderActive` default value is `false`. I would keep consistency here and put `false` as the default value.

ltkum (approved):
One small thing that I would change, but to be fair, it's not that big of a deal

Looks good to me :+1:

PR #854, opened 2024-05-22
PB-530: Added print date to the print API spec

ltshb:
Use the date formatting based on the locale (en => MM/DD/YYYY, H:MM PM/AM, fr-CH => DD.MM.YYYY, HH:MM

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-530-print-date/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -366,13 +366,14 @@ async function transformOlMapToPrintParams(olMap, config) {
                 },
             })
         }
-
+        const now = i18n.global.d(new Date(), 'datetime', i18n.global.locale)
         const spec = {
             attributes: {
                 map: encodedMap,
                 copyright: attributionsOneLine,
                 url: shortLink,
                 qrimage: qrCodeUrl,
+                printDate: now,
             },
             format: 'pdf',
             layout: layout.name,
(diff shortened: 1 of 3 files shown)


pakb (commented):
I'm not sure we shouldn't be sending a proper ISO-8601 datestring instead, while testing and looking at what was sent and then received I see some timezone issue.

I started a job with a date sent that was `"printDate": "5/23/2024, 8:20 AM"` but the file I received in return had `Printed on 23.05.2024 06:20` (so the GMT+2 delta)

Have you tested sending a full ISO string ?

ltshb:
> I'm not sure we shouldn't be sending a proper ISO-8601 datestring instead, while testing and looking at what was sent and then received I see some timezone issue.
> 
> I started a job with a date sent that was `"printDate": "5/23/2024, 8:20 AM"` but the file I received in return had `Printed on 23.05.2024 06:20` (so the GMT+2 delta)
> 
> Have you tested sending a full ISO string ?

Thats because the backend has not been built and deployed yet see https://github.com/geoadmin/service-print3/pull/37

pakb (approved):
Then all good, maybe we should wait for the PR on the backend side to be merged/deployed, then test the result here, and merge when we are happy with what we see

ltshb:
Tested with the new backend deployment and looks good

PR #856, opened 2024-05-23
PB-496: better selection of feature IDs for external layers

pakb:
use of the first property of the feature if no feature ID is given, so that we may have some text to show the user to differentiate features in the feature selection tooltip

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-496_better_external_feature_id_selection/index.html)

### DIFF
--- a/src/api/features/features.api.js
+++ b/src/api/features/features.api.js
@@ -400,10 +400,25 @@ async function identifyOnExternalWmsLayer(config) {
                     coordinates: [...coordinate],
                 }
             }
+            let featureId = feature.id ?? feature.properties?.id
+            if (!featureId && feature.properties) {
+                const propNames = Object.keys(feature.properties)
+                // if no feature ID defined, we take the first property that contains either "id", "title" or "name"
+                // and if nothing is found we use the first property as an "ID"
+                featureId =
+                    propNames.find((name) => {
+                        const lowerCaseName = name.toLowerCase()
+                        return (
+                            lowerCaseName.indexOf('id') !== -1 ||
+                            lowerCaseName.indexOf('title') !== -1 ||
+                            lowerCaseName.indexOf('name') !== -1
+                        )
+                    }) ?? feature.properties[Object.keys(feature.properties)[0]]
+            }
             return new LayerFeature({
                 layer,
-                id: feature.id,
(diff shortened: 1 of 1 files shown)


src/api/features/features.api.js
```diff
@@ -387,10 +387,15 @@ async function identifyOnExternalWmsLayer(config) {
                     coordinates: [...coordinate],
                 }
             }
+            let featureId = feature.id ?? feature.properties?.id
+            // if no feature ID defined, we take the first property and use it as an "ID"
+            if (!featureId && feature.properties) {
+                featureId = feature.properties[Object.keys(feature.properties)[0]]
+            }
```
ltshb, line 394: Here I would take the following as id and nothing (`-`) if none of those are available:
- id
- identifier
- name
- title

The search of those properties should be done case insensitive (id, Id, ID, ...)

pakb: There are cases (like our backend... :see_no_evil: ) which doesn't give ANY property that contains ID, but the first one in the "list of properties" is a decent identifier.

So I can start by looking if anything looks like an ID (contains some `.id` or is `id` case insensitive) and if really nothing was found default to the first prop.

ltshb: ok with the fallback to the first prop if no id, identifier, name or title is found.

src/api/features/features.api.js
```diff
@@ -400,10 +400,25 @@ async function identifyOnExternalWmsLayer(config) {
...
+                // if no feature ID defined, we take the first property that contains either "id", "title" or "name"
+                // and if nothing is found we use the first property as an "ID"
+                featureId =
+                    propNames.find((name) => {
+                        const lowerCaseName = name.toLowerCase()
+                        return (
+                            lowerCaseName.indexOf('id') !== -1 ||
+                            lowerCaseName.indexOf('title') !== -1 ||
+                            lowerCaseName.indexOf('name') !== -1
+                        )
```
ltshb, line 415: This might lead to inconsistent output where you might have the title or name as ID even if an `id` property is available, because this code rely on the properties to be sorted which they are not.

pakb:
replaced by #935

PR #858, opened 2024-05-23
PB-496: Fix external layer error and time slider issue

ltshb:
At startup we never know the projection of the external layers, the projection
is taken from the capabilities that needs to be loaded first. Therefore don't
test projection upon creation and don't do any fallback.

Also fixed an issue when a layer has only one timestamp, the time slider was
displayed but the user could not change the timestamp.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-496-external-layer-projection/index.html)

### DIFF
--- a/src/api/layers/ExternalLayer.class.js
+++ b/src/api/layers/ExternalLayer.class.js
@@ -1,7 +1,5 @@
 import AbstractLayer, { LayerAttribution } from '@/api/layers/AbstractLayer.class'
 import { InvalidLayerDataError } from '@/api/layers/InvalidLayerData.error'
-import { WGS84 } from '@/utils/coordinates/coordinateSystems'
-import log from '@/utils/logging'
 
 /**
  * Information required to create a GetFeatureInfo request to this external WM(T)S server. This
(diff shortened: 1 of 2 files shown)


src/store/modules/layers.store.js
```diff
@@ -170,7 +170,7 @@ const getters = {
     visibleLayersWithTimeConfig: (state) =>
         // Here we cannot take the getter visibleLayers as it also contain the preview and system layers
         state.activeLayers.filter(
-            (layer) => layer.visible && layer.timeConfig?.timeEntries?.length
+            (layer) => layer.visible && layer.timeConfig?.timeEntries?.length > 1
```
pakb, line 173: `(layer) => layer.visible && layer.hasMultipleTimestamps` would also work fine :wink:

src/api/layers/ExternalLayer.class.js
pakb, line 1: might want to re-organize imports, the linter is complaining about unused imports

src/store/modules/layers.store.js
ltshb: good idea

pakb (approved):

PR #860, opened 2024-05-23
PB-505: Added error when external WMTS layer don't support WGS84

ltshb:
For this I needed to be able to add/remove errors per layer as this error is
only for 3D.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-505-projection-error/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -150,6 +150,22 @@ const getters = {
         state.activeLayers.filter((layer) => layer.id === layerId),
 
     /**
+     * Retrieves layer(s) by ID.
+     *
+     * Search in active layer and in preview layer
+     *
+     * @param {string} layerId ID of the layer(s) to retrieve
+     * @returns {[AbstractLayer]} All active layers matching the ID
+     */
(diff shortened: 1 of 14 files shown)


src/store/modules/layers.store.js
```diff
@@ -149,6 +149,22 @@ const getters = {
...
 
+    /**
+     * Retrieves layer(s) by ID.
+     *
+     * Search in active layer and in preview layer
+     *
+     * @param {string} layerId ID of the layer(s) to retrieve
+     * @returns {[AbstractLayer]} All active layers matching the ID
+     */
+    getLayersById: (state) => (layerId) => {
```
pakb, line 160: might be good to rename that to `getActiveLayerById`, as it stands it could make you think it will lookup the whole layers config, but in reality will only seek matches in the active layers (preview layer is kind of an active layer too...)

ltshb: @pakb that is not a good idea, the catalogue uses `getActiveLayersById` to add/remove layer from the active layer list, here we definitely don't want to have the `previewLayer`

pakb (approved):

PR #861, opened 2024-05-23
Script to generate reports on translation use

ltkum:
RFC 
The script is used to create three files.

report.json is a json file which contains all translation keys found, state their occurrences across the code, has their translation when there is one available and state if the key is a variable called in the code (in which case, it would be difficult to find all keys that could match this one)

report.txt is a human readable version of the same report

non_explicitly_used_keys.txt is a report which lists all translations for which there was no explicit translation found (i.e : all translations currently specified, but not in the previous reports). We cannot guarantee that they are not used.

We could still upgrade the script by checking for ternary operators which gives explicit values, but this will suffice for a first draft

usage : npm run check:translations

If this is sufficient, it's usable (even if we can't find every usage)
I'm still going to check for ternary operators (we have 2 keys which use those), or cases where there is a null / false check before assigning a value to translate, they are currently counted as "non explicit".

@hansmannj : once you've run it, feel free to tell me what you'd like to change to make files that are easier to read.

[Test link](https://sys-map.dev.bgdi.ch/preview/translation-checking-script/index.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,3 +41,6 @@ scripts/check-layer-providers-results/*invalid_providers_wms.yaml
 scripts/check-layer-providers-results/*invalid_providers_wmts.yaml
 scripts/check-layer-providers-results/*invalid_providers_content.yaml
 scripts/check-layer-providers-results/*valid_providers.json
+
+# reports for translations
+translation_reports/*
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
         "build:int": "npm run build -- --mode integration",
         "build:prod": "npm run build -- --mode production",
         "update:translations": "node scripts/generate-i18n-files.js",
+	"check:translations": "node scripts/check-translations.js",
         "check:external": "npx vite-node scripts/check-external-layers-providers.js",
         "delete:reports": "rimraf tests/results/ || true",
         "delete:reports:unit": "rimraf tests/results/unit/ || true"
--- a/scripts/check-translations.js
+++ b/scripts/check-translations.js
@@ -1,0 +1,234 @@
+import fs from 'fs'
+import path from 'path'
+
+// regex constants
+const re_explicit = /[$.]t\([`'].+[`']\)/g
+const re_non_explicit = /[.$]t\((?!['`]).+\)/g
+
+// path constants
+const reportDir = 'translation_reports'
+const reportFile = 'report'
+const jsonFilePath = `${reportDir}/${reportFile}.json`
+const txtFilePath = `${reportDir}/${reportFile}.txt`
+const noExplicitUseTxtFilePath = `${reportDir}/translations_not_explicitly_used.txt`
+
+// language constants
+const langs = ['de', 'fr', 'it', 'en', 'rm']
+const translation_files_generic_path = 'src/modules/i18n/locales/lang.json'
+
+// dict used to store the current translation
+const translations = {}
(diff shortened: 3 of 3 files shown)


hansmannj:
Thanks @ltkum , I also added @pakb and @ltshb as reviewers here, since I don't want to be the bottleneck here.
It looks good to me, but I did not yet have time to test it.
Would be good though to have the extracted files soon, so I can send them to our colleagues who support us with checking the translations. I'll try to reserve some time for that on Wednesday at the latest (in case I don't find a slot today..).

Thanks a lot for creating the extraction scripts!

ltshb:
@ltkum I got this error when trying to run the script

```
npm run check:translations              

> check:translations
> node scripts/check-translations.js

node:fs:744
  handleErrorFromBinding(ctx);
  ^

Error: EISDIR: illegal operation on a directory, read
    at Object.readSync (node:fs:744:3)
    at tryReadSync (node:fs:444:20)
    at Object.readFileSync (node:fs:490:19)
    at file:///home/ltshb/repositories/web-mapviewer/scripts/check-translations.js:81:36
    at Array.forEach (<anonymous>)
    at searchDirectory (file:///home/ltshb/repositories/web-mapviewer/scripts/check-translations.js:74:11)
    at file:///home/ltshb/repositories/web-mapviewer/scripts/check-translations.js:146:13
    at Array.forEach (<anonymous>)
    at searchDirectory (file:///home/ltshb/repositories/web-mapviewer/scripts/check-translations.js:74:11)
    at file:///home/ltshb/repositories/web-mapviewer/scripts/check-translations.js:146:13 {
  errno: -21,
  syscall: 'read',
  code: 'EISDIR'
}

Node.js v18.20.3
```

.gitignore
```diff
@@ -41,3 +41,6 @@ scripts/check-layer-providers-results/*invalid_providers_wms.yaml
 scripts/check-layer-providers-results/*invalid_providers_wmts.yaml
 scripts/check-layer-providers-results/*invalid_providers_content.yaml
 scripts/check-layer-providers-results/*valid_providers.json
+
+# reports for translations
+translation_reports/*
```
ltshb, line 46: I would put the output directory next to the script location

package.json
```diff
@@ -28,6 +28,7 @@
         "build:int": "npm run build -- --mode integration",
         "build:prod": "npm run build -- --mode production",
         "update:translations": "node scripts/generate-i18n-files.js",
+	"check:translations": "node scripts/check-translations.js",
```
ltshb, line 31: Use spaces instead of tabs here

scripts/check-translations.js
```diff
@@ -0,0 +1,257 @@
+import fs from 'fs'
+import path from 'path'
+
+// regex constants
+const re_explicit = /[$.]t\([`'].+[`']\)/g
+const re_non_explicit = /[.$]t\((?!['`]).+\)/g
```
ltshb, line 6: Those regex will not work for all use case, translation can only have parameters !
I think it would be best to use i18n tooling for such use case https://kazupon.github.io/vue-i18n/guide/tooling.html#vue-i18n-extract. It cannot not be resolved by regex, but need a proper code parser.

ltshb (changes requested):

ltkum:
This is not a priority : we'll revisit this after deploy.

PR #862, opened 2024-05-27
PB-472 : profile in the context of the new floating tooltip

pakb:
Changes how the profile handles itself when the floating tooltip is active, closing the profile now only unselect the profile feature (and do not close the selected features anymore)

While the feature info is at the bottom, the profile is now shown as an "off-shoot" or second tab of the selected features, taking the whole width of the screen. Adding a "Back" button at the top of the infobox panel to go back to feature details.

Rework of the profile width management so that it doesn't overflow its parent. Also adding a little spacer between each profile information (going Composition API in the process)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-472_better_profile_infobx/index.html)


src/modules/infobox/InfoboxModule.vue
```diff
@@ -58,14 +65,52 @@ function onPrint() {
...
+        store.dispatch('clearAllSelectedFeatures', dispatcher)
+        store.dispatch('clearClick', dispatcher)
+    } else if (showElevationProfile.value) {
+        // if feature details are shown in the floating tooltip we don't want to close the detail
+        // when clicking on the X button, but only close the profile
+        onHideProfile()
+    }
+}
+function onHideProfile() {
+    store.dispatch('setProfileFeature', { feature: null, ...dispatcher })
```
ltkum, line 78: I don't really like this way of doing things. If someone closes by mistake the profile, and open it again, they will make unnecessary calls to the backend. Now I understands the `ShowGeometryProfileButton` component sets the profile visibility by dispatching a `setProfileFeature` too, so it would require some extra modifications which might not be what you'd like to do in this PR, and would require extra work.


If there is a clean and quick way to do this, it could be great. Otherwise, it's okay as it is. (I don't think a lot of people are going to close and reopen their profile a lot. Maybe twice maximum per person ^^)

ltkum (commented):
Hey, from using it a bit : 

When choosing to have the feature info at the bottom with the profile, it no longer shows the feature information. I do not know if it is by design, but it did confuse me when using it.

src/modules/infobox/InfoboxModule.vue
pakb: Regarding call to backends : IMHO out of scope of this PR, and if our caching works properly calling twice the same request will hit our backend only once.

Regarding UI :
That is what was wanted for mobile use (which is the purpose of the infobox), do not have the profile AND the feature detail at the same time because it takes to much space

pakb:
> Something that has nothing to do with your PR, but maybe it would be good to fix here: the icons in in the "window bar" of the tooltip don't have a hover tooltip description (only the print icon does).

We could have a ticket to fix those little things, but I'd really like a review of the current work in this PR if possible :wink:

schtibe:
> > Something that has nothing to do with your PR, but maybe it would be good to fix here: the icons in in the "window bar" of the tooltip don't have a hover tooltip description (only the print icon does).
> 
> We could have a ticket to fix those little things, but I'd really like a review of the current work in this PR if possible 😉

Totally fine!

src/modules/infobox/InfoboxModule.vue
```diff
@@ -58,14 +65,52 @@ function onPrint() {
...
 </script>
 
 <template>
-    <div v-if="showContainer" class="infobox card rounded-0" data-cy="infobox" @contextmenu.stop>
-        <div class="infobox-header card-header d-flex justify-content-end" data-cy="infobox-header">
+    <div
+        v-if="showContainer"
+        class="infobox card rounded-0 clear-no-ios-long-press"
+        data-cy="infobox"
+        @contextmenu.stop
```
schtibe, line 87: Is this needed? It's just a div, where does the event come from?

schtibe (approved):
Looking good to me

src/modules/infobox/InfoboxModule.vue
pakb: that was already there, my linter just made one line for each entry.

I think it's there to not have the context menu when right clicking on the whole Infobox, don't know if it is really useful :face_with_diagonal_mouth:

PR #865, opened 2024-05-27
PB-534: transparency slider works in real time

ltkum:
Issue: The transparency slider did not change the opacity in real time when users were moving the slider, only applying the changes when it was released

Fix: When clicking on the slider, it sets an interval which updates the opacity as long as we're handling the slider. This interval is set to 100ms, which allows to see the changes happen often enough to work, without burdening the backend with too much commits. We also reduce the number of commits by never committing the same value as the one stored.

If this puts too much strain on the backend, we can wait until post prod to find a more elegant solution.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-534-real-time-opacity-change/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -14,6 +14,7 @@ import ErrorButton from '@/utils/components/ErrorButton.vue'
 import TextTruncate from '@/utils/components/TextTruncate.vue'
 import ThirdPartyDisclaimer from '@/utils/components/ThirdPartyDisclaimer.vue'
 import { useTippyTooltip } from '@/utils/composables/useTippyTooltip'
+import debounce from '@/utils/debounce'
 import log from '@/utils/logging'
 
 const dispatcher = { dispatcher: 'MenuActiveLayersListItem.vue' }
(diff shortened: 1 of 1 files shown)


pakb (commented):
I think it would make sense to debounce it just a little bit, so that it doesn't fire a store action each little changes

ltkum:
@pakb : I've already put a 100ms interval to avoid firing it too fast, I could make it slower.
I experimented with also setting a difference threshold when moving the slider (not when releasing it) and it led to sometimes the opacity changing after releasing it.
I could increase the time for the interval. Do you have a value that is sufficient for "real time" in your opinion ? I was Thinking maybe 200ms.

pakb:
I was more referring on using what is in `src/utils/debounce.js` (see `FeatureStyleEdit.vue` for an example of use) instead of using `setInterval`

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -92,12 +95,28 @@ function onToggleLayerVisibility() {
...
+function onTransparencyChange() {
+    dispatchOpacity(1.0 - transparencySlider.value.value)
+}
+
+const debounceTransparencyChange = debounce(onTransparencyChange, 50)
+
+function onTransparencyCommit() {
+    log.info(
+        '[Menu Active Layers List Item component]: Committing last transparency reached and clearing the interval'
+    )
```
pakb, line 117: this log isn't up to date anymore, as you aren't using `setInterval` anymore

pakb (approved):
small log/comment issue and good to go

PR #871, opened 2024-05-28
PB-555 : fix too many resolutions being filtered out with WMTS layers

pakb:
Fix supports for 0.1 resolution

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_pb-555_load_resolution_0dot1/index.html)


ltkum (approved):
Looks good to me.

hansmannj (approved):
Thanks for fixing it that quick :rocket: 
Looks good from what I can tell :+1:

PR #872, opened 2024-05-28
PB-543 : fix Cesium CSS spillover

pakb:
Cesium CSS was altering how our buttons were shown, adding some padding or margins there and there. As we were importing the Cesium CSS file just to place the canvas (and its Cesium generated parent divs) correctly, I've switched to declare this CSS ourselves and remove the import of the Cesium CSS file.

Also removing a couple wrapping divs in the Cesium component that are not useful, as they clutter our "real" DOM with many empty divs

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_pb-543_fix_cesium_css_spillover/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumGeoJSONLayer.vue
+++ b/src/modules/map/components/cesium/CesiumGeoJSONLayer.vue
@@ -1,7 +1,5 @@
 <template>
-    <div>
-        <slot />
-    </div>
+    <slot />
 </template>
 
 <script>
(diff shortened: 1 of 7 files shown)


schtibe (approved):
Looking good! 👍 I wonder if we now could possibly revert #830, but maybe it's not even that important.

pakb:
> Looking good! 👍 I wonder if we now could possibly revert #830, but maybe it's not even that important.

I think we could, but that's not a big deal to keep it there too

PR #873, opened 2024-05-29
PB-497: Add GPX on 3D map

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-497-gpx-on-3d/index.html) (can be tested with loading gpx file)

- Load GPX to Cesium with Cesium standard way
- Add style for point to match with the old geoadmin (red circle billboard)
- Handle transparency update

image

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-497-gpx-on-3d/index.html)

### DIFF
--- a/src/modules/map/components/cesium/CesiumGPXLayer.vue
+++ b/src/modules/map/components/cesium/CesiumGPXLayer.vue
@@ -1,10 +1,4 @@
-<template>
-    <div>
-        <slot />
-    </div>
-</template>
-
-<script>
+<script setup>
 import {
     BillboardGraphics,
     Cartesian3,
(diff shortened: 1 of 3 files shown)


src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,159 @@
...
+const isPresentOnMap = ref(false)
+const gpxDataSource = ref(new GpxDataSource())
+
+const opacity = ref(props.gpxLayerConfig.opacity)
+const gpxData = ref(props.gpxLayerConfig.gpxData)
+
+const getViewer = inject('getViewer')
+watch(gpxData, () => {
+    addLayer()
+})
```
pakb, line 33: If the layer was already added, I think it would make to also remove it before adding it again.
I don't think what `addLayer` does removes the "old" data

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,159 @@
...
+    const redCircleBillboard = new BillboardGraphics({
+        image: redCircleImage,
+        width: billboardSize,
+        height: billboardSize,
+        eyeOffset: new Cartesian3(0, 0, -100), // Make sure the billboard is always seen
+        heightReference: HeightReference.CLAMP_TO_TERRAIN, // Make the billboard always appear on top of the terrain
+    })
+    const redColorMaterial = new ColorMaterialProperty(Color.RED.withAlpha(opacity.value))
+
+    const entities = gpxDataSource.value._entityCollection.values
```
pakb, line 119: why using the private `_entityCollection.values`?
there's a public equivalent with `.entities`

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,159 @@
...
+        context.beginPath()
+        context.arc(radius, radius, radius, 0, 2 * Math.PI, false)
+
+        const color = `rgba(255, 0, 0, ${opacity.value})`
+        context.fillStyle = color
+        context.fill()
+
+        // Return the data URL of the canvas drawing
+        return canvas.toDataURL()
+    }
```
pakb, line 103: you might want to have a look at this https://github.com/geoadmin/web-mapviewer/blob/37ce862f42c3359b80ee50bb00e50066bcc864df/src/modules/map/components/cesium/CesiumGeolocationFeedback.vue#L84-L111

where a circle entity is created using only Cesium stuff (no canvas).

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,159 @@
...
+            entity.polygon.outlineColor = Color.BLACK
+        }
+    }
+    getViewer().scene.requestRender()
+}
+</script>
+<template>
+    <div>
+        <slot />
+    </div>
```
pakb, line 158: no need to wrap in a div, simply give the `<slot />` as template, and no HTML will be generated (no empty div) when the component is loaded

src/modules/map/components/cesium/CesiumInternalLayer.vue
```diff
@@ -42,6 +42,7 @@
                 :z-index="zIndex"
             />
         </div>
+        <slot />
```
pakb, line 45: It's already declared bellow, I don't know why you would want to have it twice?

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,159 @@
...
+            entity.polyline.width = 1.5
+        }
+
+        if (cesiumDefined(entity.polygon)) {
+            entity.show = true
+            entity.polygon.material = redColorMaterial
+            entity.polygon.outline = true
+            entity.polygon.outlineColor = Color.BLACK
+        }
+    }
```
pakb, line 151: might be better to style the GPX by using the `.load` function options, see https://cesium.com/learn/ion-sdk/ref-doc/GpxDataSource.html#.load

pakb (commented):
Good work, only remark I have is that you are deviating a little to much from the "standard Cesium way" regarding styling GPX features

src/modules/map/components/cesium/CesiumGPXLayer.vue
ismailsunni: Cesium load the point as a billboard. Should I change it to circle/ellipse in this case?

src/modules/map/components/cesium/CesiumGPXLayer.vue
ismailsunni: I still need to update the style after the gpx loaded with this `updateStyle` function (in current case, the opacity). Or do you mean something else? I couldn't find another way to update the opacity without iterating the GPX entities.

src/modules/map/components/cesium/CesiumGPXLayer.vue
ismailsunni: We will use this approach per my discussion with @pakb

src/modules/map/components/cesium/CesiumGPXLayer.vue
ismailsunni: We will use this approach per my discussion with @pakb

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,155 @@
...
+        width: billboardSize,
+        height: billboardSize,
+        eyeOffset: new Cartesian3(0, 0, -100), // Make sure the billboard is always seen
+        heightReference: HeightReference.CLAMP_TO_TERRAIN, // Make the billboard always appear on top of the terrain
+    })
+    const redColorMaterial = new ColorMaterialProperty(Color.RED.withAlpha(opacity.value))
+
+    const entities = gpxDataSource.entities.values
+
+    entities.array.forEach((entity) => {
```
pakb, line 120: I don't think you are meant to access `.array` here but just call `entites.forEach`

pakb (commented):
image
I'm getting this error when trying to change a GPX's opacity, and it is then unchanged, maybe a error while looping (see bellow)

src/modules/map/components/cesium/CesiumGPXLayer.vue
ismailsunni: yes, it's from the autocomplete. Thanks for the eyes :)

src/modules/map/components/cesium/CesiumGPXLayer.vue
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        if (cesiumDefined(entity.polygon)) {
+            entity.show = true
+            entity.polygon.material = redColorMaterial
+            entity.polygon.outline = true
+            entity.polygon.outlineColor = Color.BLACK
+        }
+    })
+    getViewer().scene.requestRender()
```
pakb, line 149: I don't know why, but if I change the opacity of a GPX layer it disappear from the map, and then comes back again if I change the camera position (if I force another render?)

You might want to test that a bit further, if you want a GPX to test, [here's some available file](https://www.crossingswitzerland.com/map)

A wrap with a `nextTick` maybe?

ismailsunni: ah yes, with a bigger GPX file, I can see the issue you mentioned. I will investigate it.

ismailsunni: hi @pakb 
Unfortunately, `nextTick` doesn't work. If I forced render it to quickly, it's not working also. I came up with [hacky/dirty solution](https://github.com/geoadmin/web-mapviewer/pull/873/commits/403fffae881c163f9f25ec831cb5d0de52bed419) to do force rendering after some time (1500ms). It's working although there is a delay, but even in [sandcastle](https://sandcastle.cesium.com/#c=fVZtb9s2EP4rRL5UBmJKUeLUdZNsRdZ1G9K1aJpi2Dy0lEhZRGmSIKkk7pD/viMpWVLsJEAA6+6543OvZKmkdeiWsztm0DmS7A5dMsubNf4SZMnyoAzfl0o6wiUzy4PJ66VMU3RNJC2JdYKhglheIstco5cyKK8Uoejdx78QJY4s5S0xaKXvb4yAQ5YHtXPaLtL0dJ5lpzN6OmVsPp+ezOZH0/m8yqezjB7nOauqYlbhigtmG2twqdZps6JplpNZ8fLrbP5qdpyXLD+en5y8rObFUfYqP2bH5Wmek+JVhuHAn6g8L42ylsuVvePuBzMCaOdZfvI1yY6mWT71vyceuzx4jYD5J6YFKRkCdI1czRApXUMEuvl0hZxCG9WYEJin5YNtQ/sF4rwGHViO0vhuqMIC0pLEREwwOJdJ1cjScSVRspQoZCtCl3KC/vOiWBvcaywmlCb9ty+HBwL3v5Vae5KetucoyAYK1jv5AfrPar/tpWHEQbSSIvAfXDilRAHhVcogpUnJ3QaV0AdGieAUVB3kHFFVNmsmHS6Do7eC+S/oH8pvY8+gDo1LQaz9k6xZ6IbYYNNWCWUImejcFYpuMNGaSXpZc0GTFrdL3VPuaApSsC3JVnjlZc8xjUYt16ERduze+QkAXKD8ISoXLdkuriHNof3zXLnUjXvE9Xcve45rNBpzDUbYbXTMqyFyxVqGI8SaSwBkexTkHhRHuwrrmPYmONujvCWiYVu7Z3IR4INcXDOHGr2bC8RufaIFh2NlbODRgdCfbz3iqgX06ThE/Ty18zNKK7DUxFj2KwyiS3aj6MgFejeaPi6VqhARAsHZ3HFmge521vxQIdvOrvfQTxnu8PEQi2Gg3pKyHgx/QGy2lBHiFUraFUJZBYuXthisldgIEEwGaOTnEhZ52RgDsEsllJ/JRxZQX8cMJwJWKQDwirkvnlDSu0Hd3vqjEZxIWF0MS3WXTHpIbLn45xta6U1IQvAZ9kdZ+85D3NkucY95woLsOA52ZRCNyAzjwYbRwyeVK8OYfFpdQJgj7Q6xyWj7vwBa7ZC/iFdBWPwUiAKNbT+EenYt0ybEd7XPhw8s5qQg5fduLceSvLCoq0Vv+1S19iTpfav6aJRmxm2SLqF9cR56Wmn6dDcVXIhCEUNjO/Um6DEgNk3fVY8U2GfpjdA16eaq55KmWzYPUfowWANvtBZ+8mFEINhBYUbzSbnVxJV1mHwfL4q/+lXoXT50jxPnc7gKV9fgKovXdWntc6vVgi2LqxWQ/T71l0AKkrBSvYZLWD6/fX5/BepvMTw8vsyGE6qV5X7YF4gUVonGscEgOaUX6CjT9wOZYJXbEWpYfhDYjtz32MqoRtIFMquCJCf5Ier+MzwfTm2hDGVmagjljV2g2chRKOYC3dV8S/Bhf2zhuhxGuCZmxeXU8FXthm6fsA91+8en93x7Wf079HfHqat9pNnQ1Tf4ua1dzQgdXTVQFx/qweHBWajjRefuZ77WysCFY0SCcerYGibdMZsWTfmdQRdEQw89S4emZ/CAQZye73kJo/CQAU3VCHHNf0AAF2cp4HdM/cMPyvbh1j9ANx5WH11cRSHG+CyFz/2W22fRSP8/), it still takes time to render. 

if I set `requestRenderMode` to false, obviusly it works, but I am sure we don't want to set it to false.

What do you think? Perhaps you know something else to wait until entities are updated?

ismailsunni:
@pakb I have tested with removing and readding the gpx layer for chaning the opacity. It's still fragile. I have removed it for now.

PR #874, opened 2024-05-29
PB-521 : add geolocation feedback on Cesium

pakb:
There's still issues when keeping the geolocation active and swapping between 2D and 3D, but we might be able to "live with it" and have a ticket to fix that after go-live (ticket [here](https://jira.swisstopo.ch/browse/PB-559))

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-521_3d_geoloc_feedback/index.html)


ltkum (approved):
Apart from the issue already mentionned, it seems to be working fine, but I noticed that on my desktop, when I activate the geolocation, it asks twice for it before processing the query. (I'm not checking the 'remember my decision' 

This is not really that big of an issue IMO.

PR #875, opened 2024-05-29
PB-538 : split profile too long for backend into chunks

pakb:
in the same way we split the profile regarding LV95 bound, we now also cut it in portions/chunks of 3000 points, in order not to hit the hard limit of 5k point set in our backend

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-538_split_profile_request_too_long/index.html)


src/api/profile/profile.api.js
```diff
@@ -16,16 +15,60 @@ export class ProfileError {
...
+ *
+ * @type {number}
+ */
+const MAX_CHUNK_LENGTH = 3000
+
+/**
+ * @param {CoordinatesChunk | null} [chunks]
+ * @returns {null | CoordinatesChunk[]}
+ */
+function splitIfTooManyPoints(chunks = null) {
```
ltkum, line 40: It seems to be working well, but I would add a small unit test to ensure this stays so.

ltkum (approved):
I tested it with the very big kml file we used to test in PB-378, and it works fine. I would add a small unit test on `splitIfTooManyPoints` and then you're good to go.

PR #876, opened 2024-05-29
PB-228 : add aggregate layer support to 3D

pakb:
removing the visible flag being used, we already remove the component altogether when the visibility flag is false, and it hinders aggregate from being displayed (sub-layers have their visibility flag set to false in the config for some reasons...) Also removing the subdomain for WMS, as we are fine landing always on the main URL endpoint (and can be problematic for external layer support)

Note : a good aggregate layer candidate is the RBD layer loaded by default in the displayed map (will be a WMTS for a while and switch to a WMS when closer to the ground)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-228_aggregate_in_3d/index.html)


ltkum (approved):
After testing for a bit, it seems it works as expected.

PR #881, opened 2024-06-06
PB-548 : fix camera position present in URL at startup issue

pakb:
Removing 2D position URL params when 3D is active to diminish issues when we load the app directly in 3D. When zoom and/or position are present, the 3D viewer would react to changes in these values. This was creating some race condition issue at startup when a camera position was also described in the URL.

Camera initialization has also been separated from the logic to move the camera after init, so that we have a clearer logical decision (with logging)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_pb-548_camera_after_reload/index.html)

### DIFF
--- a/src/store/plugins/sync-camera-lonlatzoom.js
+++ b/src/store/plugins/sync-camera-lonlatzoom.js
@@ -7,6 +7,7 @@ import {
 } from '@/modules/map/components/cesium/utils/cameraUtils'
 import { normalizeAngle } from '@/store/modules/position.store'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
+import log from '@/utils/logging'
 
 /**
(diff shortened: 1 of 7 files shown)


src/store/plugins/sync-camera-lonlatzoom.js
```diff
@@ -45,32 +46,27 @@ export default function syncCameraLonLatZoom(store) {
...
-                    pitch: CesiumMath.toDegrees(-CesiumMath.PI_OVER_TWO),
-                    roll: 0,
-                },
-                dispatcher: self,
-            })
-        }
+    })
+    store.subscribeAction({
+        after: (action, state) => {
+            if (state.position.camera === null) {
```
ltshb, line 52: Maybe put a comment here why we use action subscription instead of mutation as we usually do mutation watcher

ltshb (approved):

PR #882, opened 2024-06-06
PB-521: Enabled geolocation high accuracy

ltshb:
While testing geolocation it often results to really poor accuracy, sometimes
up to several kilometers !

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-521-goelocation-high-precision/index.html)

### DIFF
--- a/src/store/plugins/geolocation-management.plugin.js
+++ b/src/store/plugins/geolocation-management.plugin.js
@@ -9,6 +9,8 @@ import log from '@/utils/logging'
 
 const dispatcher = { dispatcher: 'geolocation-management.plugin' }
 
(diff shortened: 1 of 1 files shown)


pakb (approved):
Looking forward using it next time I go outdoor :smile:

ltshb:
I tested it today while mountain biking and here is my test impressions:
- After enabling, I got within a few seconds an initial position about ~1km off
- In less than 10-15 seconds I got a higher precision within ~50m
- In less than 30s I got a high precision less ~10m
- I was on a small trail in the forest while testing with clear wheater
- later on I used watch position while biking at 40km/h down a road outside the forest and the geolocation work very well and was precise (although a bit slower that I was biking)

PR #886, opened 2024-06-10
PB-589 : limit WMS request to LV95 extent

pakb:
Layer that are telling, with a label at bottom, when was their last update weren't displayed in a manner that permit this label to be shown (it was out of scope of the map viewport with how the request was built)

This fixes two things :
 - Limits the request to our own WMS to LV95 extent
 - Tells OpenLayers to not request extra information around (gutter) but keep a ratio of 1:1 when requesting a WMS (this way the update label can be inside the viewport)

Good candidate to test : `ch.bazl.luftfahrthindernis`
[Test link with layer](https://sys-map.dev.bgdi.ch/preview/bug_pb-589_no_last_update_wms_label/index.html?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-farbe&layers=ch.swisstopo.zeitreihen,ch.bfs.gebaeude_wohnungs_register,ch.bav.haltestellen-oev,ch.swisstopo.swisstlm3d-wanderwege,ch.astra.wanderland-sperrungen_umleitungen,ch.bazl.luftfahrthindernis&layers_opacity=1,1,1,0.8,0.8,1&layers_visibility=false,false,false,false,false,true&layers_timestamp=18641231,,,,,)
[Test link](https://sys-map.dev.bgdi.ch/preview/bug_pb-589_no_last_update_wms_label/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
+++ b/src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
@@ -12,6 +12,7 @@ import GeoAdminWMSLayer from '@/api/layers/GeoAdminWMSLayer.class'
 import { ALL_YEARS_WMS_TIMESTAMP } from '@/api/layers/LayerTimeConfigEntry.class'
 import { WMS_TILE_SIZE } from '@/config'
 import useAddLayerToMap from '@/modules/map/components/openlayers/utils/useAddLayerToMap.composable'
+import { LV95 } from '@/utils/coordinates/coordinateSystems.js'
 import { flattenExtent } from '@/utils/coordinates/coordinateUtils'
 import CustomCoordinateSystem from '@/utils/coordinates/CustomCoordinateSystem.class'
 import { getTimestampFromConfig } from '@/utils/layerUtils'
@@ -97,6 +98,9 @@ if (gutter.value !== -1) {
 // That means that data will not be requested if the map viewport is outside the extent.
 if (wmsLayerConfig.value.extent) {
     layer.setExtent(flattenExtent(wmsLayerConfig.value.extent))
+} else if (wmsLayerConfig.value instanceof GeoAdminWMSLayer) {
+    // do not request stuff outside our technical extent with our own layers.
+    layer.setExtent(LV95.getBoundsAs(projection.value).flatten)
 }
 
 // grabbing the map from the main OpenLayersMap component and use the composable that adds this layer to the map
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -125,6 +129,11 @@ function createSourceForProjection() {
             url: url.value,
             projection: projection.value.epsg,
             params: wmsUrlParams.value,
+            // Limiting image request to exactly the size of the map viewport.
+            // We have a couple layers that state when they have lastly been updated at the bottom
+            // of the WMS iamge, and without this ratio prop this label is out of the map viewport.
```
ltkum, line 134: A little typo in the comment

ltshb, line 134: typo `iamge` => `image`

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -97,6 +98,9 @@ if (gutter.value !== -1) {
 // That means that data will not be requested if the map viewport is outside the extent.
 if (wmsLayerConfig.value.extent) {
     layer.setExtent(flattenExtent(wmsLayerConfig.value.extent))
+} else if (wmsLayerConfig.value instanceof GeoAdminWMSLayer) {
+    // do not request stuff outside our technical extent with our own layers.
+    layer.setExtent(LV95.getBoundsAs(projection.value).flatten)
 }
```
ltshb, line 104: Is this not dangerous, what if in future we add the `extent` to the GeoAdminWMSLayer, then this hack will not used anymore, maybe putting this `if` first is more future proof.

```javascript
if (wmsLayerConfig.value instanceof GeoAdminWMSLayer) {
    // do not request stuff outside our technical extent with our own layers.
    layer.setExtent(LV95.getBoundsAs(projection.value).flatten)
} else if (wmsLayerConfig.value.extent) {
    layer.setExtent(flattenExtent(wmsLayerConfig.value.extent))
}
```

ltkum (commented):
In the viewer when zooming, the box is a bit under the slightly transparent footer
image
If there is an easy way to move the box a few pixels up, that would be great, but I can live with the box being where it is

In 3d, those labels appear in the middle of every tile.
image
We can fix this issue in a second PR (as the cause is most certainly not the same),but this would be better if the behavior between 2d and 3d was the same

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
pakb: Wouldn't that be good that if the layer has some specific configured extent we use it instead of the "hack"?

pakb:
> In 3d, those labels appear in the middle of every tile. 

OK, that opened up another huge can of worms :see_no_evil: 
I found out that many (MANY!) layers have a `config3d` property in the JSON layer config, and we are to swap to the `config3d` layer ID...

That will requires a bigger change and I will do that separately

ltshb (approved):

PR #888, opened 2024-06-10
PB-587: altering feature info behavior on mobile

ltkum:
Issue : On mobile, small screen sizes could create issues where the head bar was partially inaccessible. In landscape mode it could appear under the header, which made it impossible to use, and in vertical mode, part of it was behind the right side menu.

Fix :
  In landscape mode, we set the height of the tooltip to be the minimum between 350 pixels (previous value) and 60% of the viewport height, which ensure the head bar of the tooltip is never under the header.
  In vertical mode, if the width of the screen is smaller than 400 pixels, we enforce the feature info position to be set as 'bottompanel' and we remove the button to set it as a floating tooltip. This is called both when setting the info position and on resize.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-587-tooltip-behavior-in-mobile-mode/index.html)

### DIFF
--- a/src/modules/infobox/InfoboxModule.vue
+++ b/src/modules/infobox/InfoboxModule.vue
@@ -4,6 +4,7 @@ import { computed, nextTick, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
+import { MAX_WIDTH_SHOW_FLOATING_TOOLTIP } from '@/config'
 import FeatureEdit from '@/modules/infobox/components/FeatureEdit.vue'
 import FeatureElevationProfile from '@/modules/infobox/components/FeatureElevationProfile.vue'
 import FeatureList from '@/modules/infobox/components/FeatureList.vue'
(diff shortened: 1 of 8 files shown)


src/store/modules/ui.store.js
```diff
@@ -321,13 +334,19 @@ export default {
...
                     `invalid feature Info Position given as parameter. ${upCasePos} is not a valid key`
                 )
                 return
             }
+            // when the viewport width is too small, the layout of the floating infobox will be
+            // partially under the menu, making it hard to use. In those conditions, the option to
+            // set it as a floating tooltip is disabled.
+            if (upCasePos !== 'NONE' && state.width < MAX_WIDTH_SHOW_FLOATING_TOOLTIP) {
+                upCasePos = 'BOTTOMPANEL'
+            }
```
pakb, line 349: Please use values directly from the enum "object" `FeatureInfoPositions`

src/modules/infobox/InfoboxModule.vue
```diff
@@ -40,7 +41,9 @@ const showContainer = computed(() => {
             (showElevationProfile.value && showFeatureInfoInTooltip.value))
     )
 })
-const showTooltipToggle = computed(() => showFeatureInfoInBottomPanel.value)
+const showTooltipToggle = computed(
+    () => showFeatureInfoInBottomPanel.value && width.value >= MAX_WIDTH_SHOW_FLOATING_TOOLTIP
+)
```
pakb, line 46: That's weird but if I resize my viewport to bigger than the threshold and then back to mobile, the floating toogle button stays there (as if this wasn't re-computed)

ltkum: I can't seem to manage to reproduce it and it behaves as expected locally (and on sys dev). We can take a quick look tomorrow to see if we can reproduce this on your device and check what's the problem.

pakb (approved):

PR #893, opened 2024-06-11
PB-449: Add a minimize button to give feedback and report problem

ltshb:
The first idea was to use movable but this cause trouble with the content of
the window not being at a fix size like for the tooltip but dependent on the
window content.

As simpler alternative I added the minimize button which let the user minimize
the window which is for the use case good enough.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-449-move-window/index.html)

### DIFF
--- a/src/modules/menu/components/settings/ReportProblemButton.vue
+++ b/src/modules/menu/components/settings/ReportProblemButton.vue
@@ -150,6 +150,9 @@ async function generateShortLink() {
 }
 
 function openForm() {
+    if (!props.showAsLink) {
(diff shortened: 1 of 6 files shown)


pakb (approved):
not something you should fix in the scope of this PR, but if you open both "modals" they stack over one another (feedback + report), might be better to close the other one

ltshb:
> not something you should fix in the scope of this PR, but if you open both "modals" they stack over one another (feedback + report), might be better to close the other one

Yes but this is only an issue on dev staging on prod and test map we don't have both button only one or the other 😉

PR #894, opened 2024-06-11
Fix tooltip movable window pointer and icon issues

ltshb:
The pointer was not set to `move` on the movable window header. Also clicking
on the window header buttons (close, print, ...) triggered also a move event
which is unwanted.

Made also the use movable composable usable if the component is not mounted
and fixed default params.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-movable-window/index.html)

### DIFF
--- a/src/utils/composables/useMovableElement.composable.js
+++ b/src/utils/composables/useMovableElement.composable.js
@@ -1,4 +1,4 @@
-import { ref, toValue } from 'vue'
+import { onMounted, ref, toValue } from 'vue'
 
 const defaultPadding = 4 // px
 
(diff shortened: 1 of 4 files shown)


src/utils/composables/useMovableElement.composable.js
```diff
@@ -9,16 +9,16 @@ const defaultPadding = 4 // px
  * It is possible to specify custom offset with screen border, in order to avoid some part of the
  * app (such as the header) to be accessible to the movable element
  *
- * @param element
+ * @param element Reference to an DOM element
```
pakb, line 12: small typo here `Reference to a DOM element` (DOM might have been added later :wink: )

pakb (approved):

PR #895, opened 2024-06-11
Add title to infobox

ltshb:
The infobox only had the profile title but not on object or in drawing.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-infobox-title/index.html)


pakb (commented):
There's a little improvement that could still be made, when drawing something that shows a profile the title "Edit object" appears in the profile infobox, even though the "edit" part is on the map
image

pakb (approved):

PR #896, opened 2024-06-11
PB-594: Use singular form of vegetation.

ismailsunni:
- DE: Vegetation
- EN: Vegetation
- FR: Végétation
- IT: Vegetazione
- RM: Vegietaziun

I am not a language expert, but that's what I get from google translate. Let me know if something is wrong.


ltshb (approved):

ltshb (changes requested):
Translation needs to be changed first in the googlesheet. I'll do it and run the update translation script and update this PR

ltshb (approved):

PR #897, opened 2024-06-11
PB-621: Fixed default lang with `-CH` suffix

ltshb:
The `-CH` local suffix is not supported by our API backend. The default
language was set to the default browser language which could be de-CH or fr-CH.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-621-default-lang/index.html)

### DIFF
--- a/src/modules/i18n/index.js
+++ b/src/modules/i18n/index.js
@@ -19,15 +19,13 @@ export function langToLocal(lang) {
 }
 
 export function localToLang(local) {
-    return local.replace('-CH', '')
+    return local.split('-')[0]
+}
 
 // detecting navigator's locale as the default language
 // (if it is a language served by this app)
-let matchedLanguage = null
-if (navigator.languages) {
-    // we keep the first match we found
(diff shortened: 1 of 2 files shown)


src/modules/i18n/index.js
```diff
@@ -18,6 +18,9 @@ export function langToLocal(lang) {
     return ['de', 'fr', 'it', 'rm'].includes(lang) ? `${lang}-CH` : lang
 }
 
+export function localToLang(local) {
+    return local.replace('-CH', '')
+
```
pakb, line 23: I would even go further, and change things for all local (`en-US` or `en-UK` is not covered by our use case, neither is `de-DE` or `fr-FR`)

ltshb: Done

pakb (approved):
I've never experienced the case/bug on my device so I can't really test if it fixes it, but from code point of view, looking good

ltshb:
> I've never experienced the case/bug on my device so I can't really test if it fixes it, but from code point of view, looking good

To reproduce the error you need to set the default/preferred language of chrome to de-CH or fr-CH 😉  then start the application without parameters.

PR #901, opened 2024-06-11
PB-586 improve tooltip css

ltkum:
Various issues with the tooltips CSS:

sometimes, features titles were too big
sometimes, the space between the attribute name and the attribute value was too small and hard to read.
sometimes, the alignment between an attribute's name and its value was a bit off, as they were centered on a cell (when the value or the name took more than one line to display)

Fix : We adapt the CSS for those elements, and add a tippy on top of the feature title when it's truncated.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-586-improve-tooltip-css/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureDetail.vue
+++ b/src/modules/infobox/components/FeatureDetail.vue
@@ -134,4 +134,7 @@ function getIframeHosts(value) {
 :global(.htmlpopup-content) {
     padding: 7px;
 }
+:global(td) {
+    vertical-align: top;
+}
 </style>
--- a/src/modules/infobox/components/FeatureDetail.vue
+++ b/src/modules/infobox/components/FeatureDetail.vue
@@ -134,11 +134,11 @@ function getIframeHosts(value) {
 :global(.htmlpopup-content) {
     padding: 7px;
 }
-:global(td) {
+td {
     vertical-align: top;
 }
 
-:global(td.cell-left) {
+td.cell-left {
     padding-right: 10px;
 }
 </style>
--- a/src/modules/infobox/components/FeatureDetail.vue
+++ b/src/modules/infobox/components/FeatureDetail.vue
@@ -137,4 +137,8 @@ function getIframeHosts(value) {
 :global(td) {
     vertical-align: top;
 }
+
+:global(td.cell-left) {
+    padding-right: 10px;
+}
 </style>
--- a/src/modules/infobox/components/FeatureListCategoryItem.vue
+++ b/src/modules/infobox/components/FeatureListCategoryItem.vue
@@ -40,7 +40,6 @@ const store = useStore()
 const isHighlightedFeature = computed(
     () => store.state.features.highlightedFeatureId === item.value.id
 )
-const showFeatureInfoInBottomPanel = computed(() => store.getters.showFeatureInfoInBottomPanel)
 function highlightFeature(feature) {
     store.dispatch('setHighlightedFeatureId', {
         highlightedFeatureId: feature?.id,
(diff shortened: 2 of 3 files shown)


src/modules/infobox/components/FeatureListCategoryItem.vue
```diff
@@ -74,19 +79,43 @@ function showContentAndScrollIntoView(event) {
...
+            placement: 'top',
+            delay: [500],
+            // Show tippy on long touch for mobile device
+            touch: ['hold', 500], // 500ms delay,
+        })
+    }
+})
+onUnmounted(() => {
+    tippyTitle?.destroy()
+})
```
ltshb, line 102: The component TextTruncate already handle this, so better to reuse it

ltshb (commented):
There is an issue where the zoom to extent button is put on the next line when the text is truncated 
image

src/modules/infobox/components/FeatureListCategoryItem.vue
ltshb: See https://github.com/geoadmin/web-mapviewer/blob/876902fbd0f8a206a8320f74f5e1bce6e23e5ead/src/utils/components/TextTruncate.vue#L1

ltkum: Hoo, Pascal told me about it but I thought he meant the `text-truncate` bootstrap component, not the vue component. That'll simplify things.

ltshb:
The `arrow` seems not to be aligned with the text 
image

src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -134,4 +134,11 @@ function getIframeHosts(value) {
 :global(.htmlpopup-content) {
     padding: 7px;
 }
+:global(td) {
+    vertical-align: top;
+}
+
+:global(td.cell-left) {
+    padding-right: 10px;
+}
```
ltshb, line 143: Not sure it is a good idea to use `global` on a `td` element, does this changes for the whole application ? @pakb what do you think ?

src/modules/infobox/components/FeatureListCategoryItem.vue
```diff
@@ -78,15 +80,27 @@ function showContentAndScrollIntoView(event) {
...
-        <strong>{{ name }}</strong>
+        <FontAwesomeIcon :icon="`caret-${showContent ? 'down' : 'right'}`" class="mx-2 column" />
+        <TextTruncate
+            :text="name"
+            class="column-truncate"
+            :class="{
+                'infobox-active': showFeatureInfoInBottomPanel,
+            }"
+        >
+            <span class="font-weight-bold">{{ name }}</span>
```
ltshb, line 102: Use bootstrap for boldness

src/modules/infobox/components/FeatureListCategoryItem.vue
```diff
@@ -123,4 +139,23 @@ function showContentAndScrollIntoView(event) {
...
+        // infobox take the whole width of the screen. We truncate somewhere different
+        // depending on how wide the screen is.
+        width: 80vw;
+    }
+}
+
+.font-weight-bold {
+    // this is a boostrap class, we need to use the bootstrap one
+    font-weight: bold;
+}
```
ltshb, line 160: use bootstrap class for bold text

src/modules/infobox/components/FeatureDetail.vue
ltkum: the alignment seems to work without the global. I thought we needed it to apply the style to the html popup, but it seems we can avoid using it, thus avoid side effects.

I'm removing the 'global'

and yes, the :global would have changed for the whole application

ltshb:
@ltkum I modify the solution to use bootstrap, can you double check/test that everything works, I tested myself but a second test might be good,  if you don't see any other issue you're good to merge

ltshb (approved):

PR #904, opened 2024-06-12
PB-542 : take into account "config3d" layers config property

pakb:
and improve how we parse WMS layer config, taking into account possible "wmsLayers" property

This means that the BG layer set in the URL stays the same while going 3D, and CesiumMap component looks up for 3D equivalent for any WMS/WMTS (background included) if they are defined.

This should fix the issue we had with `ch.bazl.luftfahrthindernis` in 3D, where the "Last updated" label was on each tile (3D can't load a single tile WMS, so there's a specific 3D config for this one, among others)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug_pb-542_proper_3d_layer_selection/index.html)

### DIFF
--- a/src/api/layers/GeoAdminLayer.class.js
+++ b/src/api/layers/GeoAdminLayer.class.js
@@ -18,8 +18,10 @@ export default class GeoAdminLayer extends AbstractLayer {
     /**
      * @param {String} layerData.name Name of this layer in the current lang
      * @param {LayerTypes} layerData.type See {@link LayerTypes} geoAdminLayerData.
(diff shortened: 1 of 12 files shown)


ltshb (approved):
Looks good, however it seems that the `last update` label for `ch.bazl.luftfahrthindernis` on 2D is still not working as on the old, but I guess @pakb  this is another issue not related to this PR ?

pakb:
> Looks good, however it seems that the `last update` label for `ch.bazl.luftfahrthindernis` on 2D is still not working as on the old, but I guess @pakb this is another issue not related to this PR ?

Yeah that is fixed in this PR #886

PR #905, opened 2024-06-12
PB-412: Update print to use native print inside a modal for Infobox, Map-Popover and Legend

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-412-infobox-print-improvement/index.html)


pakb (commented):
It's not working properly when the content is an iFrame (like OeV)

You might want to test out having a `nextTick` callback before triggering the browser print prompt

LukasJoss:
Is there a sensible way to test print prompts in cypress? @ltshb

src/utils/components/PrintButton.vue
```diff
@@ -0,0 +1,47 @@
...
+    showModal.value = true
+}
+function onHideParentModal(hide) {
+    emits('hideParentModal', hide)
+}
+</script>
+
+<template>
+    <button
+        id="PrintButton"
```
ltshb, line 31: Using an id here is not a good idea, usually you should have unique ID in the whole DOM of the application, so you might have several print button at the same time which will break the id uniqueness rule. I suggest to create a class, e.g. `print-button` and use the class selector in `useTippyTooltip('.print-button[data-tippy-content]')` (you don't need to define the class in CSS only in the element `class`).

src/utils/components/PrintButton.vue
```diff
@@ -0,0 +1,47 @@
...
+        class="btn btn-light btn-sm d-flex align-items-center"
+        :data-tippy-content="'print'"
```
ltshb, line 33: here you can simply use static attribute `data-tippy-content="print"`

ltshb (approved):
Great thanks

ltshb:
@LukasJoss I may have found a solution to the selection issue on the printed PDF, it seems that you can select the hidden part of the view behind the modal. To avoid this we could print from a hidden iframe (See https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_media_queries/Printing#print_an_external_page_without_opening_it), so instead of putting the content into a modal we could put it in a hidden iframe, it worth a quick try.

ltshb:
@LukasJoss  @pakb I find an easy solution to the print selections of feature popup see my commit https://github.com/geoadmin/web-mapviewer/pull/905/commits/e9c3643d77cef0ff1c1522c32d127c917d58f285, could you also test it and if it is good then we can rebase and merge this PR.

ltshb:
@pakb @LukasJoss I also found a way to print elevation profile, see last commit https://github.com/geoadmin/web-mapviewer/pull/905/commits/9f38a542a12ca01b790601d8c80527fa95ab8bd0

LukasJoss:
Thanks @ltshb 
Looks good to me.
The  one thing I found is that when printing a profile, the size of the profile is not always the, maybe it does not have the time to properly resize sometimes?
[Maps of Switzerland - Swiss Confederation - map.geo.admin.ch.pdf](https://github.com/user-attachments/files/15881940/Maps.of.Switzerland.-.Swiss.Confederation.-.map.geo.admin.ch.pdf)

ltshb:
@pakb would you like to have a quick look here or should I merge ?

pakb (approved):
outside of the two small issues that I mentioned to you through slack (which we can live with for a go-live), looking good!

the two issues are

- profile/infobox getting smaller (height) after a print of the profile (I didn't test feature detail print though)
- sometimes iframe aren't loaded when the print prompt comes in, maybe a bigger timeout could help fix that... but that's already better than the old viewer so all good to go-live with that

PR #911, opened 2024-06-13
PB-546: Put print overlay always on top

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-546-print-overlay/index.html)

PR #910, opened 2024-06-13
Updated the help link to the new help from the goe.admin.ch CMS

ltshb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-help-links/index.html)

PR #914, opened 2024-06-13
Revert "PB-546: Put print overlay always on top"

ltshb:
Reverts geoadmin/web-mapviewer#911 

It contained also the PR #910 

[Test link](https://sys-map.dev.bgdi.ch/preview/revert-911-bug-pb-546-print-overlay/index.html)


pakb (approved):
this is the one :see_no_evil:

PR #916, opened 2024-06-13
PB-546: Print overlay not centered on high resolution device (e.g. Retina or HiDPI device)

ltshb:
Somehow, probably due to an openlayer update, the print overlay was not anymore
centered on the map on HiDPI and Retina devices (e.g. Iphone, Dell XPS 15, HP ZBook G9).

After some test it turns out that the issue arised only when the DEVICE_PIXEL_RATIO
was not 1. Removing this ratio from the computed print area solved the issue.
This ratio in the print area has been introduced in 2014 in the old mf-geoadmin3
project (see https://github.com/geoadmin/mf-geoadmin3/pull/1359)
in order to solve the same issue observed here on HiDPI and Retina
device. This issue has been probably solved internally in openlayers now.

To test use Iphone and/or our HP zbook G9 without external display.
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-546-print-overlay-center/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
+++ b/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
@@ -14,7 +14,6 @@ export default function usePrintAreaRenderer(map) {
     const POINTS_PER_INCH = 72 // PostScript points 1/72"
     const MM_PER_INCHES = 25.4
     const UNITS_RATIO = 39.37 // inches per meter
-    let printRectangle = []
 
     const isActive = computed(() => store.state.print.printSectionShown)
     const printLayoutSize = computed(() => store.getters.printLayoutSize)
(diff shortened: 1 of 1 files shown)


ltshb:
The origin of the DEVICE_PIXEL_RATIO is from geoadmin/mf-geoadmin3 from this PR https://github.com/geoadmin/mf-geoadmin3/pull/1359 and issue https://github.com/geoadmin/mf-geoadmin3/issues/1332. It also been used by c2c here https://github.com/camptocamp/ngeo/pull/1634.
I don't understand yet but it seems that removing this ratio solve the issue on Iphone 11 and on our HP Zbook with high resolution screen. 
From what I remember on Iphone 11 the print area used to work with this ratio, so the non working current state might be due to an openlayer update. I quickly looked into openlayers to see if I see any recent open/close issues/pr but did not find anything relevant.

pakb (approved):

PR #920, opened 2024-06-17
PB-633: Fix map panning being too fast

ltshb:
When panning the map moved faster than the cursor.

This was due to the fact that default openlayer map interactions (like map panning)
were added twice.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-633-panning/index.html)


pakb (approved):
did we add the "move interaction" twice with the previous code?
I'm a bit surprised what you changed fixes the issue

ltshb:
> did we add the "move interaction" twice with the previous code? I'm a bit surprised what you changed fixes the issue

Yes look at the code we took the default interactions, that are apparently automatically added by ol, extend it with the new two interactions and then loop over all interactions to do an `add` 😉

PR #922, opened 2024-06-17
PB-637: Keep time layer selected year when changing the language

ltshb:
When changing the language the timed layer where reset to their default time
parameter.

The issue is that only their current year where updated but not their current
timestamp.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-637-time-layer-config-lang/index.html)


LukasJoss (approved):
lgtm

PR #929, opened 2024-06-18
PB-660: Fix elevation profile resize

ltshb:
When resizing the window from small to big the elevation profile was not re-render
properly.

See https://www.chartjs.org/docs/master/configuration/responsive.html#important-note

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-660-profile-size/index.html)


LukasJoss (approved):
lgtm

PR #930, opened 2024-06-18
PB-631 : proxify KML images when not admin.ch / bgdi.ch - #patch

pakb:
[Test link](https://sys-map.int.bgdi.ch/preview/bug_pb-631_cors_on_nested_urls/index.html)

### DIFF
--- a/src/utils/kmlUtils.js
+++ b/src/utils/kmlUtils.js
@@ -13,6 +13,7 @@ import EditableFeature, { EditableFeatureTypes } from '@/api/features/EditableFe
 import { extractOlFeatureCoordinates } from '@/api/features/features.api'
 import { DEFAULT_TITLE_OFFSET } from '@/api/icon.api'
 import { DrawingIcon } from '@/api/icon.api'
+import { API_SERVICE_PROXY_BASE_URL, API_SERVICES_BASE_URL } from '@/config.js'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import {
     allStylingSizes,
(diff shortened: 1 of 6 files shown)


src/utils/kmlUtils.js
```diff
@@ -488,4 +489,65 @@ export function parseKml(kmlLayer, projection, iconSets) {
...
+    })
+
+    // Replace all hrefs to prevent errors if image doesn't have
+    // CORS headers. Exception for *.geo.admin.ch, *.bgdi.ch and google
+    // markers icons (only https)
+    // to keep the OL magic for anchor origin.
+    // Test regex here: http://regex101.com/r/tF3vM0/9
+    // List of Google icons: http://www.lass.it/Web/viewer.aspx?id=4
+    transformedKml = transformedKml.replace(
+        /<href>http(?!(s:\/\/maps\.(google|gstatic)\.com[a-zA-Z\d.\-/_]*\.png|s?:\/\/[a-z\d.-]*(bgdi|geo.admin)\.ch))/g,
```
ltkum, line 518: In the old viewer, in the first `[a-zA-Z\d.\-/_]` , both the `.` and the `/` are also escaped.
in the second one `[a-z\d.-]`, the `.` and the `-` are escaped. 
There is no need to escape them in js, but I would still change the documentation line regarding the regex test permalink to reflect the current regex

ltkum (commented):
From what I could see, between the [legacy viewer](https://map.geo.admin.ch/?lang=de&topic=ech&bgLayer=ch.swisstopo.pixelkarte-grau&E=2659794.74&N=1191072.21&zoom=0&layers=KML%7C%7Chttps:%2F%2Fwww.uvek-gis.admin.ch%2FBAZL%2FKarte_AeMC_AME%2FAeromedical%20Examiner%20(AME).kml,KML%7C%7Chttps:%2F%2Fwww.uvek-gis.admin.ch%2FBAZL%2FKarte_AeMC_AME%2FAeromedical%20Center%20(AeMC).kml)
and [your work](https://sys-map.dev.bgdi.ch/preview/bug_pb-631_cors_on_nested_urls/index.html#/map?lang=de&center=2659794.74,1191072.21&z=1&bgLayer=ch.swisstopo.pixelkarte-grau&topic=ech&layers=KML%7Chttps://www.uvek-gis.admin.ch/BAZL/Karte_AeMC_AME/Aeromedical+Examiner+(AME).kml;KML%7Chttps://www.uvek-gis.admin.ch/BAZL/Karte_AeMC_AME/Aeromedical+Center+(AeMC).kml&featureInfo=default), 
only looking at the CORS and proxy queries : 
it mostly seems to work, but the icons for the AeMC layer (cropped-star) are not showing. We can see in the logs they hit a CORS error, but are still proxy-fied correctly within the kml (once again, if we refer to the log)

The function seems to work as expected and transform what needs to be transformed, but there might be an underlying problem which causes the layer to call for resources before being updated.

I'm a bit out of my depth here, maybe @ltshb has better insight than me.

ltshb:
@pakb  I suggest to close this, as #1008 seems to solve the kml icon issues

PR #932, opened 2024-06-18
PB-646: Fix logic to set value in style from literal.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-646-symbolizaion-geojson/index.html)

[Test link with snow layer](https://sys-s.dev.bgdi.ch/6y26qfj3ehyu)
Sample:
image

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-646-symbolizaion-geojson/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/styleFromLiterals.js
+++ b/src/modules/map/components/openlayers/utils/styleFromLiterals.js
@@ -274,8 +274,8 @@ OlStyleForPropertyValue.prototype.setOlRotation_ = function (
 }
 
 OlStyleForPropertyValue.prototype.getOlStyle_ = function (feature, resolution, properties) {
-    // A value can be 0
-    const value = properties[this.key] ? properties[this.key] : this.defaultVal
+    // Use default value if key is not found in properties
+    const value = properties[this.key] !== undefined ? properties[this.key] : this.defaultVal
     const geomType = getGeomTypeFromGeometry(feature.getGeometry())
 
     let olStyles = null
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/utils/styleFromLiterals.js
```diff
@@ -274,8 +274,8 @@ OlStyleForPropertyValue.prototype.setOlRotation_ = function (
 }
 
 OlStyleForPropertyValue.prototype.getOlStyle_ = function (feature, resolution, properties) {
-    // A value can be 0
-    const value = properties[this.key] ? properties[this.key] : this.defaultVal
+    // Use default value if key is not found in properties
+    const value = properties[this.key] !== undefined ? properties[this.key] : this.defaultVal
```
ltshb, line 278: What if the value is `null` do we want to have `null` as value or the default value int his case ?

pakb, line 278: maybe something like
```
const value = properties[this.key] ?? this.defaultVal
```
could achieve what you wanted, taking what Brice was concerned about into account

ismailsunni: I am open for both cases. I just check the old geoadmin, it only set to `defaultVal` if the value is undefined. Should I go with the old approach or should I handle the null ?

See: https://github.com/geoadmin/mf-geoadmin3/blob/develop/src/components/StylesFromLiteralsService.js#L292

ismailsunni: Currently, I update it like @pakb suggests

ltshb (approved):

PR #935, opened 2024-06-18
PB-593 : use feature label/title if defined

pakb:
instead of showing the technical ID to the user


ltkum (approved):

ltshb (commented):
Why doing this only on the vector layer ? We should do this also for other layers for example the `RBD: building status` layer also have a label (street name) that is used on old viewer while not in the new viewer.

src/utils/identifyOnVectorLayer.js
```diff
@@ -114,7 +114,7 @@ export function identifyGeoJSONFeatureAt(geoJsonLayer, coordinate, projection, r
             return new LayerFeature({
                 layer: geoJsonLayer,
                 id: feature.id,
-                name: feature.properties.station_name || feature.id,
+                name: feature.properties.station_name ?? feature.properties.label ?? feature.id,
```
ltshb, line 117: Add also comment about layer id using station_name

src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
```diff
@@ -122,7 +122,11 @@ export default function useMapInteractions(map) {
                                     new LayerFeature({
                                         layer: vectorLayer,
                                         id: olFeature.getId(),
-                                        name: olFeature.getId(),
+                                        name:
+                                            olFeature.get('label') ??
+                                            // exception for MeteoSchweiz GeoJSONs, we use the station name instead of the ID
```
ltshb, line 127: Add also comment about layer id using station_name

pakb:
replaces #856

PR #937, opened 2024-06-18
PB-645: zoom to extent on feature preselection

ltkum:
Issue : When pre selecting features, the zoom to extend would not work as expected.

Fix :
 - We process layers and their parameters before processing the zoom level, to avoid the possibility that the default value overrides the expected behavior when there is no zoom level specified.
 - We set the maxZoomLevel parameter of the `zoomToExtent` function to the Global constant `STANDARD_ZOOM_LEVEL_1_25000_MAP`, which correspond to zoom level 8, the expected zoom level.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-645-zoom-to-extent-on-feature-preselection/index.html)

### DIFF
--- a/src/router/storeSync/LayerParamConfig.class.js
+++ b/src/router/storeSync/LayerParamConfig.class.js
@@ -13,7 +13,6 @@ import {
     transformLayerIntoUrlString,
 } from '@/router/storeSync/layersParamParser'
 import { flattenExtent } from '@/utils/coordinates/coordinateUtils.js'
-import { STANDARD_ZOOM_LEVEL_1_25000_MAP } from '@/utils/coordinates/SwissCoordinateSystem.class'
 import { getExtentOfGeometries } from '@/utils/geoJsonUtils'
 import log from '@/utils/logging'
 
@@ -207,7 +206,7 @@ async function getAndDispatchFeatures(to, featuresPromise, store) {
             if (!query.z) {
                 await store.dispatch('zoomToExtent', {
                     extent: extent,
-                    maxZoom: STANDARD_ZOOM_LEVEL_1_25000_MAP,
+                    maxZoom: store.state.position.projection.get1_25000ZoomLevel(),
                     dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
                 })
             } else {
--- a/src/router/storeSync/LayerParamConfig.class.js
+++ b/src/router/storeSync/LayerParamConfig.class.js
@@ -13,6 +13,7 @@ import {
     transformLayerIntoUrlString,
 } from '@/router/storeSync/layersParamParser'
 import { flattenExtent } from '@/utils/coordinates/coordinateUtils.js'
+import { STANDARD_ZOOM_LEVEL_1_25000_MAP } from '@/utils/coordinates/SwissCoordinateSystem.class'
 import { getExtentOfGeometries } from '@/utils/geoJsonUtils'
 import log from '@/utils/logging'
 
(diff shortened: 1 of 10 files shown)


src/router/storeSync/storeSync.config.js
```diff
@@ -42,6 +42,7 @@ const storeSyncConfig = [
     // to the default projection EPSG number
     new PositionParamConfig(),
     new CameraParamConfig(),
+    new LayerParamConfig(),
```
ltshb, line 45: Here we should add a comment why the layer param needs to be before z param, a comment related to the `zoomToExtent` in LayerParamConfig. But consider also trying first using `from` instead of `to`

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -170,7 +171,7 @@ async function getAndDispatchFeatures(to, featuresPromise, store) {
             if (!query.z) {
                 await store.dispatch('zoomToExtent', {
                     extent: extent,
-                    maxZoom: 8,
+                    maxZoom: STANDARD_ZOOM_LEVEL_1_25000_MAP,
```
ltshb, line 174: I'm wondering if it would not make more sense to use here `from.query.z` instead of `to.query.z` ? This could avoid layer parameter order dependencies.

pakb, line 174: The use of this variable is valid only (and ONLY) if the current projection system is a world-wide projection (Mercator).
This is not a LV95 zoom!

ltkum: the maxZoom parameter, when given to `zoomToExtent` goes through `transformStandardZoomLevelToCustom`, which was not the case when the parameter was first introduced. (see https://github.com/geoadmin/web-mapviewer/blob/0f3a4dbb104ae68a14f2aa4881104be47507c173/src/store/modules/position.store.js#L269-L273 )
it assumes the maxZoom level you give to the function is not a LV95 zoom.

If I give the `8` from before (which is the LV95 zoom level I would like) to the function, it transforms it to zoom level 2 because it thinks it's a world wide projection level 8 zoom and that it needs to be converted to LV95

pakb: Then I'd say the issue is that we reprocess the given `maxZoom`, as if it was a Mercator zoom.

The logic should maybe be changed to :
 - if `maxZoom` is defined, use it as it is
 - if not set, use `STANDARD_ZOOM_LEVEL_1_25000_MAP`
   - Are we in a custom coordinate system? If yes, transform `STANDARD_ZOOM_LEVEL_1_25000_MAP` to match the current coordinate system

This way `maxZoom` can be set to 8, which is correct in the context of LV95.

It might be good to have new variable on each projection, called something like "default detail zoom level` or "zoom level 1:25000", in order to reduce the amount of magic numbers in our code

ltkum: I like this solution.

tests/cypress/tests-e2e/embed.cy.js
```diff
@@ -34,7 +34,7 @@ describe('Testing the embed view', () => {
             .should(
                 'have.attr',
                 'href',
-                `#/map?lang=en&center=2660000,1190000&z=1&bgLayer=test.background.layer2&topic=ech&layers=`
+                `#/map?lang=en&center=2660000,1190000z=1&bgLayer=test.background.layer2&topic=ech&layers=`
```
ltshb, line 37: You probably don't want this changes, this make your test failing

src/router/storeSync/LayerParamConfig.class.js
ltkum: we do not have access to 'from' here without changing every storesync `setValueInStore` function. Luckily, the 'to.query' is not modified at this state. The issue was only the maxZoom being modified by the zoomToExtent function

tests/cypress/tests-e2e/embed.cy.js
ltkum: good catch

pakb (approved):
While you will have to rebase, could you read through your changes and make sure you don't add (or remove) unnecessary line breaks? That pollutes the git log for not much.

Also you could 'ventilate' a bit class declarations (meaning one empty line between each members) 😉

PR #939, opened 2024-06-19
PB-590: Quick fix for timeslider preview year for layer without current year entry

ltshb:
In the timeslider when a layer did not have any matching timestamp from the preview
year, the layer was still displayed. Also the feature selection only work on
the previously configured timestamp (before entering the timeslider).

The first issue has been resolved by filtering visible layers that don't
match the preview year.

The second issue has been solved by directly setting the preview year in the
layer config and not only when unmounting.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-590-time-slider-no-data-quick/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/TimeSlider.vue
+++ b/src/modules/map/components/toolbox/TimeSlider.vue
@@ -248,6 +248,7 @@ function setPreviewYearToLayers() {
     activeLayers.value.forEach((layer, index) => {
         const year = previewYear.value
         if (
+            layer.visible &&
             layer.hasMultipleTimestamps &&
             layer.timeConfig &&
             layer.timeConfig.getTimeEntryForYear(year)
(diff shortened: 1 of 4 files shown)


LukasJoss:
@ltshb Now the feature information will also be loaded while dragging the time slider and while running the animation, i don't know if that produces a significant enough amount of requests that that should be changed post go live or if it is fine like that.

ltshb:
@LukasJoss what do you mean exactly by feature informations ? When the timeslider is moving only the layer are loading with this PR and that is what it is intended.

pakb (approved):

LukasJoss:
> @LukasJoss what do you mean exactly by feature informations ? When the timeslider is moving only the layer are loading with this PR and that is what it is intended.

@ltshb Never-mind, I rechecked and it seems to be good,

PR #944, opened 2024-06-20
PB-669 : separate html popup request from getFeature

pakb:
identify request already provides us with all what we need (feature data), the only thing missing is the HTML popup.

This means getFeature request is now only called when the app is loaded with pre-selected features.

This reduces the number of calls to our backend for an identify on the map (should be a bit faster)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-669_less_get_feature_requests/index.html)

### DIFF
--- a/src/api/features/features.api.js
+++ b/src/api/features/features.api.js
@@ -158,23 +158,19 @@ export async function identifyOnGeomAdminLayer({
             },
         }
     )
-    // firing a getFeature (async/parallel) on each identified feature
-    const featureRequests = []
+    // firing a getHtmlPopup (async/parallel) on each identified feature
+    const features = []
     if (identifyResponse.data?.results?.length > 0) {
-        // for each feature that has been identified, we will now load their metadata and tooltip content
-        identifyResponse.data.results.forEach((feature) => {
-            featureRequests.push(
-                getFeature(layer, feature.id, projection, {
+        for (const feature of identifyResponse.data.results) {
+            const featureData = await getFeatureHtmlPopup(layer, feature.id, lang)
+            features.push(
+                parseGeomAdminFeature(layer, feature, featureData, projection, {
                     lang,
-                    coordinates: coordinate.join(','),
-                    mapExtent: mapExtent.join(','),
-                    imageDisplay,
                 })
             )
-        })
+        }
     }
-    // waiting on the result of all parallel getFeature requests
-    return await Promise.all(featureRequests)
+    return features
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/api/features/features.api.js
```diff
@@ -529,108 +525,161 @@ export const identify = (config) => {
...
-            params.coord = coordinates
-        }
-        if (imageDisplay) {
-            params.imageDisplay = imageDisplay
-        }
-        if (mapExtent) {
-            params.mapExtent = mapExtent
+        if (!outputProjection) {
+            reject('An output projection is required')
         }
```
ltshb, line 617: You will break some layer if you don't add those parameter to the GetFeature, see https://github.com/geoadmin/web-mapviewer/pull/933

ltkum (approved):
this looks good to me. The one point I wasn't too sure about was whether the checks in `getFeatureHtmlPopup` were really necessary, but they won't cause any issues, and it's good to be a little bit on the safer side :)

src/api/features/features.api.js
```diff
@@ -529,108 +525,161 @@ export const identify = (config) => {
...
+ * @param {String | Number} featureId The feature ID in the BGDI
  * @param {CoordinateSystem} outputProjection Projection in which the coordinates (and possible
  *   extent) of the features should be expressed
- * @param {String} lang The language for the HTML popup
+ * @param {String} lang The language for the HTML popup. Default is `en`.
  * @returns {Promise<LayerFeature>}
  */
-const getFeature = (layer, featureID, outputProjection, options = {}) => {
-    const { lang = 'en', coordinates = null, imageDisplay = null, mapExtent = null } = options
+const getFeature = (layer, featureId, outputProjection, lang = 'en') => {
```
ltshb, line 607: In which case the `getFeature` is used, I mean in which case we need to get the htmlpopup and the feature info at the same time ?

ltshb (changes requested):

src/api/features/features.api.js
ltshb: Without `coord`, `mapExtent` and `imageDisplay` the https://api3.geo.admin.ch/services/sdiservices.html#feature-resource don't return the correct data for certain layer that was the issue with the `Moorlandschaften AuLaV` layer so you need to pass them (maybe not all of them, just saw that the old viewer used all of them)

src/api/features/features.api.js
pakb: at app startup when a pre-selected feature is present. Only use case I see

src/api/features/features.api.js
pakb: that's weird, these params (especially `mapExtent` which seems the trigger to have the correct output) are not documented on this endpoint
https://api3.geo.admin.ch/services/sdiservices.html#htmlpopup-resource

pakb: https://api3.geo.admin.ch/services/sdiservices.html#feature-resource doesn't mentioned them either

pakb: I had a quick chat with Marcel, and the layer you mentioned is a bit of an edge case, as it has two different model to show the users. It will select the model depending on the resolution of the map, hence the need for all these params (especially `mapExtent` and `imageDisplay`)

I'll add them back to the HTML popup request with a bit of documentation

src/api/features/features.api.js
pakb: I hesitated to move the loading of the HTML popup directly in the FeatureDetail component, and only deal with metadata in this file...
But then we have go-live in 2 (work) days and didn't want to spend too much time here...

pakb: there's a second usage of this function in the feature store, when a features are updated after a lang switch

ltshb (approved):
👍🏼

PR #950, opened 2024-06-24
Improve geolocation error management by removing the alert() - #patch

ltshb:
Now the geolocation error is not anymore in a window.alert() but in an error
overlay on the top right corner that can be close.

Now it looks like this 
image

[Test link](https://sys-map.int.bgdi.ch/preview/bug-geoloc/index.html)

### DIFF
--- a/src/modules/map/components/toolbox/GeolocButton.vue
+++ b/src/modules/map/components/toolbox/GeolocButton.vue
@@ -28,7 +28,7 @@ function toggleGeolocation() {
 </script>
 
 <template>
-    <!-- Here belowe we need to set the tippy to an external div instead of directely to the button,
+    <!-- Here below we need to set the tippy to an external div instead of directly to the button,
      otherwise the tippy won't work when the button is disabled -->
     <div class="geoloc-button-div" :data-tippy-content="tippyContent">
         <button
--- a/src/modules/map/components/toolbox/GeolocButton.vue
+++ b/src/modules/map/components/toolbox/GeolocButton.vue
@@ -8,10 +8,19 @@ const dispatcher = { dispatcher: 'GeolocButton.vue' }
 
 const store = useStore()
 
-useTippyTooltip('.geoloc-button[data-tippy-content]', { placement: 'left' })
+useTippyTooltip('.geoloc-button-div[data-tippy-content]', { placement: 'left' })
 
 const isActive = computed(() => store.state.geolocation.active)
 const isDenied = computed(() => store.state.geolocation.denied)
+const tippyContent = computed(() => {
+    if (isDenied.value) {
+        return 'geoloc_permission_denied'
+    }
+    if (isActive.value) {
+        return 'geoloc_stop_tracking'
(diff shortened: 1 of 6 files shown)


src/modules/map/components/toolbox/GeolocButton.vue
```diff
@@ -8,29 +8,42 @@ const dispatcher = { dispatcher: 'GeolocButton.vue' }
...
-        :class="{ active: isActive, disabled: isDenied }"
-        :data-tippy-content="isActive ? 'geoloc_stop_tracking' : 'geoloc_start_tracking'"
-        data-cy="geolocation-button"
-        @click="toggleGeolocation"
-    >
-        <svg xmlns="http://www.w3.org/2000/svg" y="0" x="0">
-            <ellipse class="geoloc-button-inner-circle" />
-        </svg>
-    </button>
+    <!-- Here belowe we need to set the tippy to an external div instead of directely to the button,
```
ismailsunni, line 31: below

ismailsunni: directly

src/utils/components/ErrorWindow.vue
```diff
@@ -0,0 +1,111 @@
...
+    hide: {
+        type: Boolean,
+        default: false,
+    },
+    /** Add a minimize button in header that will hide/show the body */
+    hasMinimize: {
+        type: Boolean,
+        default: true,
+    },
+    /** Set to true if this is and */
```
ismailsunni, line 24: I don't understand the comment here. Perhaps it should be removed?

ltkum (approved):
Looks good to me. I'm not a fan of 'empty block of codes' (the big comment in the geolocation management plugin), but it's not a problem at all, and it predates your work anyway.
I'm approving this, so you're not blocked by the small addition I suggested, which does not change the code anyway.

ismailsunni (approved):
I know the web mapviewer is supposed to be accessed from Switzerland. But I just want to mention that if I enable the geolocation from Indonesia, it behave quite strange. It tries to center the map view to somewhere north west and the map is somehow flickering.

Obviously, not a priority :)

ltshb:
> I know the web mapviewer is supposed to be accessed from Switzerland. But I just want to mention that if I enable the geolocation from Indonesia, it behave quite strange. It tries to center the map view to somewhere north west and the map is somehow flickering.
> 
> Obviously, not a priority :)

Thats a known issue with an existing ticket that will be done after go live

tests/cypress/tests-e2e/geolocation.cy.js
```diff
@@ -11,6 +11,10 @@ function getGeolocationButtonAndClickIt() {
     cy.get(geolocationButtonSelector).should('be.visible').click()
 }
 
+// TODO Those tests below are not working as expected, as the cypress-browser-permissions is not
```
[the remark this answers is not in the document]
ltshb: Good idea

PR #951, opened 2024-06-24
PB-420: improve legend css

ltkum:
Issue : when displaying a legend, the styling made it sometimes difficult to read which attributes had which value, as the columns were quite separate from each other.
Fix : When not on mobile, and not in portrait mode, we reduce the width of the table, making the columns closer to each other and making it easier to read.


src/modules/menu/components/LayerDescriptionPopup.vue
```diff
@@ -145,8 +152,14 @@ $spacing: 8px;
     }
     table {
         font-size: 100%;
-        width: 100%;
+        width: 50%;
```
ltshb, line 155: Would be better to use `width: fit-content` and add a right margin to the `td` element
```
table {
  ...
  width: fit-content;
  td {
    padding-right: 2rem;
  }
}
```
Somehow `margin-right` don't work, but padding does.

This way the column width is dependent on the content and not the window size and if some column require the whole window width they can use it without text wrapping.

ltkum: apparently, margin does not work on table internal elements. I did not think about the padding, as 'fit-content' did not work initially. It seems to work with the padding and the fit content, so I can go with your solution.

ltshb (approved):

PR #952, opened 2024-06-24
PB-698: update links - #patch

ltkum:
Issue : some localised links were outdated in the translation files.
Fix : we update the links

[Test link](https://sys-map.int.bgdi.ch/preview/bugfix-pb-698-update-links/index.html)


src/modules/i18n/locales/de.json
```diff
@@ -76,7 +76,7 @@
     "connect": "Verbinden",
     "contact_us": "Kontakt",
     "contextpopup_lv03_url": "https://www.swisstopo.admin.ch/de/landestriangulation-lv03",
-    "contextpopup_lv95_url": "https://www.swisstopo.admin.ch/de/landesvermessung-lv95",
```
ltshb, line 79: The previous link works and the new one is not pointing to LV95 description, why changing those links ?

ltshb (changes requested):

src/modules/i18n/locales/de.json
ltkum: This is a suggestion from @mariokeusen . We are redirecting to the description of the swiss coordinate system as a whole, and the page describes the lv95 system quite well.

ltshb: Then ok for me

ltshb (approved):

PR #954, opened 2024-06-25
Display also the version on prod staging - #patch

ltshb:
This helps to know when the deployment is done as it provide a easy visual feedback.

[Test link](https://sys-map.int.bgdi.ch/preview/feat-version/index.html)


ltkum (approved):
Maybe someone will be mad that this is visible, but it'll be useful :)

PR #955, opened 2024-06-25
PB-186: pre processing tool for URL parameters - #minor

ltkum:
Issue : some parameters needed to be handled in a special way before we load the map viewer, and this pre-processing happened within the legacy router. This worked fine, but this was not the place to handle it.

Fix : we now have an additional router, called only once between the legacy view and the map view, which handles the pre-processing of parameters.

[Test link](https://sys-map.int.bgdi.ch/preview/feat-pb-186-url-param-pre-process-router/index.html)


src/router/parametersPreProcess.routerPlugin.js
```diff
@@ -0,0 +1,86 @@
...
+ */
+const preProcessParam = (param_name, param_value, query) => {
+    switch (param_name) {
+        case '3d':
+            // if the 3d flag is set, we need to ensure that we are in the webmercator projection
+            if (
+                ((typeof param_value === 'boolean' && param_value) ||
+                    param_value === 'true' ||
+                    param_value === null) &&
+                (!isNumber(query.sr) || Number(query.sr) !== WEBMERCATOR.epsgNumber)
```
ltshb, line 19: I would keep like before, to keep it simple, and always overwrite the sr parameter even if it has already the correct value

src/router/parametersPreProcess.routerPlugin.js
```diff
@@ -0,0 +1,86 @@
...
+            }
+            break
+        default:
+            break
+    }
+}
+
+const handleParams = async (query) => {
+    // if so, we transfer all old param (stored before vue-router's /#) and transfer them to the MapView
+    // we will also transform legacy zoom level here (see comment below)
```
ltshb, line 34: Clean up wrong comment

src/router/parametersPreProcess.routerPlugin.js
```diff
@@ -0,0 +1,86 @@
...
+    const newQuery = structuredClone(query)
```
ltshb, line 35: I'm not sure it is a very good idea to use such new web api released only 2 years ago, this might not work for a lot of peoples, @pakb what do you think ? It might make more sense to use the lodash deep clone function

ltshb (commented):
To be honest I don't think thats the correct approach to use a preprocessing router for this use case, what if a user add manually the 3d query param in the URL ! in that case you pre-processor won't be triggered !

I think it would be wiser to do this in the handling of the query parameter 3d, which should also set the projection in the store if needed which will then trigger the sr parameter changes.

ltkum:
the failing test works locally, I'll investigate to see if it's flaky, and why if it's the case.

ltshb (approved):

PR #956, opened 2024-06-25
PB-705 : fix feature search on aggregate layers

pakb:
Using RegBL (RBD) as basis. Removing feature detail from search result, as it can be unformatted content in some cases (with RegBL it's the same as the title without formatting, so no benefit of showing it)


ltkum (approved):
looks like it's working to me.

PR #957, opened 2024-06-25
PB-387: Fix geolocation outside switzerland - #patch

LukasJoss:
- Fix endless recentering loop by restricting center to lv95 bound
- Throw error if users geolocation is outside of lv95 bound

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-387-fix-active-geolocation-outside-switzerland/index.html)


LukasJoss:
importToolMaps.cy seems very flaky, but I don't see how this is related to this PR

src/App.vue
```diff
@@ -54,12 +54,14 @@ function refreshPageTitle() {
         @pointerdown="withOutline = false"
     >
         <router-view />
-        <ErrorWindow
-            v-if="errorText"
-            title="error"
-            @close="store.dispatch('setErrorText', { errorText: null, ...dispatcher })"
-            ><div>{{ i18n.t(errorText) }}</div></ErrorWindow
-        >
+        <div>
```
ltkum, line 57: Do we need to add a `<div>` here ?

ltkum (commented):
Looks like good work. I would add e2e tests to check that we can't go out of bounds through the URL, both in 2D and 3D,

src/modules/i18n/locales/de.json
```diff
@@ -76,7 +76,7 @@
     "connect": "Verbinden",
     "contact_us": "Kontakt",
     "contextpopup_lv03_url": "https://www.swisstopo.admin.ch/de/landestriangulation-lv03",
-    "contextpopup_lv95_url": "https://www.swisstopo.admin.ch/de/landesvermessung-lv95",
+    "contextpopup_lv95_url": "https://www.swisstopo.admin.ch/de/schweizer-koordinatensystem",
```
ltshb, line 79: Those are Not correct url ans shouöd Not be in this PR

ltshb (changes requested):

src/modules/i18n/locales/de.json
LukasJoss: @ltkum You seem to have changed that entry in the spreadsheet, should the value in table be changed back?

src/App.vue
LukasJoss: The error window throws an error and cannot reappear afterwards if not though I don't know why exactly it needs the div.

src/store/modules/position.store.js
```diff
@@ -217,14 +217,22 @@ const actions = {
...
-                dispatcher,
-            })
+            if (state.projection.epsg != LV95.epsg || LV95.isInBounds(center[0], center[1])) {
+                commit('setCenter', {
+                    x: center[0],
+                    y: center[1],
+                    dispatcher,
+                })
+            } else {
+                log.debug('center received is out of bounds, ignoring')
```
ltshb, line 227: Better use a warning message here

src/store/modules/position.store.js
```diff
@@ -217,14 +217,22 @@ const actions = {
...
+            }
         } else {
-            const { x, y } = center
-            commit('setCenter', { x, y, dispatcher })
+            if (state.projection.epsg != LV95.epsg || LV95.isInBounds(center.x, center.y)) {
+                const { x, y } = center
+                commit('setCenter', { x, y, dispatcher })
+            } else {
+                log.debug('center received is out of bounds, ignoring')
```
ltshb, line 234: Dito use warning message

src/store/modules/position.store.js
ltshb: Or even better use error message.

ltshb (approved):
Good just change the logs and add a comment about why you added the div

src/utils/components/ErrorWindow.vue
```diff
@@ -35,38 +35,36 @@ const emit = defineEmits(['close'])
 </script>
 
 <template>
-    <teleport to="#main-component">
```
LukasJoss, line 38: @ltshb it looks like the additional div was necessary due to the 'in-place' teleport of this component, is there any reason to keep this teleport around or can it be deleted since this component is not used anywhere else?

PR #958, opened 2024-06-25
PB-707 : show tooltip on iframe - #patch

pakb:
The old viewer had always shown tooltip when showing the app in embed mode. With our app we changed that a bit, and that breaks a lot of our customer's integrations (through iframe).

Revert that so that the first feature selection in an iFrame shows a tooltip, only the pre-selection of feature should be impacted by the featureInfo URL param (or showTooltip in legacy)

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-707-tooltip-iframe/index.html)


ltshb (approved):
👍🏼  We can re-introduce this behavior later on by using the `featureInfo=never` parameter insteaf of `none`

PR #959, opened 2024-06-25
PB-709 : also change 3D specific config timestamp - #patch

pakb:
was forgotten when switching to this new approach that time config needs to be synced too (only the time config of the "parent" aka 2d config was changed through the UI)

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-709-3d-config-timestamp/index.html)


ltshb (approved):

ltkum (approved):
I tested it, it seems to work, code seems good

PR #961, opened 2024-06-26
PB-713: revert to use the same logic as in the old geoadmin. - #patch

ismailsunni:
[Test link](https://sys-map.int.bgdi.ch/preview/fix-713-missing-treasure-hunt-icons/index.html)


pakb (approved):
👍

PR #962, opened 2024-06-26
PB-716 : guess WGS84 coordinate order in the searchbar - #patch

pakb:
without world-wide coverage, it doesn't really make sense for our users that we do not do the same guessing as the old viewer was doing. So adding it back.

This will need to be discussed/re-thinked when we will be adding world-wide coverage, as an inverted WGS84 is still valid in this context (pointing to something outside CH...)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-716-inverted-wgs84-in-searchbar/index.html)


ismailsunni (approved):
LGTM

PR #963, opened 2024-06-26
PB-717 : add current year to time slider years - #patch

pakb:
some layers, such as ch.swisstopo.pixelkarte-farbe-pk25.noscale, uses the current year as a valid timestamp. Removing the "minus one" to the youngest year of the time slider to be able to show it.

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-717-current-year-in-timeslider/index.html)

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -313,7 +313,7 @@ export const OLDEST_YEAR = 1844
  *
  * @type {Number}
  */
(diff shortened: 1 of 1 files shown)


ltclm (approved):
thank you :muscle:

PR #965, opened 2024-06-26
PB-170: Add device orientation to geolocation button - #minor

LukasJoss:
Add vision cone to geolocation if information is available through geolocation api or magnetometer to indicate the users orientation.
If a magnetometer is available, clicking the geolocation a second time will rotate the map to align it to north

[Test link](https://sys-map.int.bgdi.ch/preview/feat-pb-170-device-orientation/index.html)


ltshb (commented):
What is this PR ? What should I review ? Could you please update the PR title to provide information on what is asked to review here

ltshb (commented):
I don't fully get the auto rotation and based on the PR description it seems that it don't work, at least on my mobile phone touching geolocation again would disable the geolocation and not re-orient to the north if not in the north.

LukasJoss:
> I don't fully get the auto rotation and based on the PR description it seems that it don't work, at least on my mobile phone touching geolocation again would disable the geolocation and not re-orient to the north if not in the north.

Does your device have absolute orientation (magnetic sensors)? If not this is expected behavior with the current code as pressing the geolocation button again should just turn off the geolocation without absolute orientation

ltshb:
is replaced by #1022

PR #966, opened 2024-06-26
PB-720: Fix default language detection - #patch

ltshb:
We always used the first supported language of the application that the browser
also supported instead of choosing the prefer browser langage that app supports

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-720-dft-lang/index.html)


pakb (approved):
Not that easy to set for Firefox users, as it's in the second sub-menu to set the preferred lang (the first one, to set display and notification) but works well when correctly set

PR #967, opened 2024-06-26
PB-726: Added a print progress bar instead of a spinner - #patch

ltshb:
The progress bar is degressive and it's completion time is set based on the
number of layer to print

[Test link](https://sys-map.int.bgdi.ch/preview/feat-print-progress/index.html)

### DIFF
--- a/src/modules/menu/components/print/MenuPrintSection.vue
+++ b/src/modules/menu/components/print/MenuPrintSection.vue
@@ -8,6 +8,7 @@ import {
     usePrint,
 } from '@/modules/map/components/openlayers/utils/usePrint.composable'
 import MenuSection from '@/modules/menu/components/menu/MenuSection.vue'
+import ProgressBar from '@/utils/components/ProgressBar.vue'
 import log from '@/utils/logging'
 import { formatThousand } from '@/utils/numberUtils'
 
@@ -27,6 +28,8 @@ const store = useStore()
 const availablePrintLayouts = computed(() => store.state.print.layouts)
 const selectedLayout = computed(() => store.state.print.selectedLayout)
 const scales = computed(() => selectedLayout.value?.scales || [])
+// approximate print duration := 8s per layer (+1 is for the background layer and to avoid 0 duration)
+const printDuration = computed(() => 8 * (store.getters.visibleLayers.length + 1))
 
 const selectedLayoutName = computed({
     get() {
@@ -198,6 +201,12 @@ defineExpose({
                 <div class="valid-feedback">{{ i18n.t('operation_successful') }}</div>
             </div>
             <div class="full-width justify-content-center">
+                <ProgressBar
+                    v-if="printStatus === PrintStatus.PRINTING"
+                    :duration="printDuration"
+                    bar-class="bg-danger"
+                    class="mb-2"
+                />
                 <button
                     v-if="printStatus === PrintStatus.PRINTING"
                     type="button"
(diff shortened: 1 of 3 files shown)


hansmannj (approved):
Nice idea! :1st_place_medal: 

Might also make sense, once the printing itself is through and the progess bar is at 100% and disappears, to have a small message there?
When testing on the train with a slow connection to print an A3, the progress bar nicely shows that printing is done.
But downloading an A3 takes quite long and there is no more progress bar of course for that.
But a small message like "downloading PDF" or something might make the user feel more comfortable?

ltshb:
@hansmann for the download time is difficult and behave differently on mobile device and desktop device. On desktop the download is left to the browser and you have the browser download file progress, while on mobile the file is open in a new window.
An issue fix could be to slightly change the success message from "Operation Successful!" to "Print Successful! Downloading..." but then even if the download is done the message will stay so not sure it helps.

pakb:
I don't really like that we do a magic number calculation to decide how to progress the bar, that might backfire in case of future outages, or unforeseen delay (bar being stuck at 99%, like in everyday life on Windows...)

There's no way of giving some steps/progress/feedback from the polling request of the backend?

ltshb:
> I don't really like that we do a magic number calculation to decide how to progress the bar, that might backfire in case of future outages, or unforeseen delay (bar being stuck at 99%, like in everyday life on Windows...)
> 
> There's no way of giving some steps/progress/feedback from the polling request of the backend?

@pakb Unfortunately not that I'm aware off. However your issue is mitigated with https://github.com/geoadmin/infra-kubernetes/pull/568 were the backend timeout after 3min and with my logic here having a degressive progress bar, it will never stays long at 99%, in the worst case for a single layer print it might stay stuck ~1min. We could either decrease the backend timeout to 2min in my opinion ?

hansmannj:
> @hansmann for the download time is difficult and behave differently on mobile device and desktop device. On desktop the download is left to the browser and you have the browser download file progress, while on mobile the file is open in a new window. An issue fix could be to slightly change the success message from "Operation Successful!" to "Print Successful! Downloading..." but then even if the download is done the message will stay so not sure it helps.

Exactly, I don't mean to have  a progress bar for the download, but a message, because the download time is unpredictable

hansmannj:
> > I don't really like that we do a magic number calculation to decide how to progress the bar, that might backfire in case of future outages, or unforeseen delay (bar being stuck at 99%, like in everyday life on Windows...)
> > There's no way of giving some steps/progress/feedback from the polling request of the backend?
> 
> @pakb Unfortunately not that I'm aware off. However your issue is mitigated with [geoadmin/infra-kubernetes#568](https://github.com/geoadmin/infra-kubernetes/pull/568) were the backend timeout after 3min and with my logic here having a degressive progress bar, it will never stays long at 99%, in the worst case for a single layer print it might stay stuck ~1min. We could either decrease the backend timeout to 2min in my opinion ?

Timeout of 2 min might do, just need to test with A3 and a ton of layers. 2 Minutes might be fully used for those cases...

boecklic (approved):
imho in general a good idea to show to users that print takes time. It's not ideal to set if fixed - arbitrary - value for when it's done, but we don't have a better option right now.

ltshb:
@pakb do you agree to use this ?

pakb:
> @pakb do you agree to use this ?

I don't disagree, I just feel it's a bad solution to a bad problem 😉 (you have a consensus, I think 🤔 )

PR #968, opened 2024-06-27
PB-742: Use permalink on report problem email - #patch

ismailsunni:
[Test link](https://sys-map.int.bgdi.ch/preview/fix-742-use-permalink/index.html)

### DIFF
--- a/src/api/feedback.api.js
+++ b/src/api/feedback.api.js
@@ -1,7 +1,6 @@
 import axios from 'axios'
 
 import { getKmlFromUrl } from '@/api/files.api'
-import { createShortLink } from '@/api/shortlink.api'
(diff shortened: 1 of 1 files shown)


src/api/feedback.api.js
```diff
@@ -50,12 +39,14 @@ export default async function sendFeedback(subject, text, options) {
             kmlData = kml
         }
 
+        const permalink = sanitizeUrl(window.location.href)
+
```
pakb, line 43: I think helpdesk might want to know if the user had some crosshair or geolocation, so no need to "sanitize" it (they want to know the state of the viewer for the feedback they received)

ismailsunni: @pakb ok, I remove the last commit so it doesn't contain sanitazion.

pakb (approved):

PR #969, opened 2024-06-27
PB-718 : adding cardinal letter support for most WGS84 search input - #patch

pakb:
now supports most input (inverted or not) with cardinal letters, such as :

- 47° 43.813' N 8° 37.738' E
- N 46° 59.000 E 006° 44.000
- E7.297165 N46.791657 

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-718-wgs84-search-bar/index.html)

### DIFF
--- a/src/utils/coordinates/coordinateExtractors.js
+++ b/src/utils/coordinates/coordinateExtractors.js
@@ -3,12 +3,24 @@ import proj4 from 'proj4'
 import { LV95, WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { reprojectUnknownSrsCoordsToWGS84 } from '@/utils/coordinates/coordinateUtils'
 import { toPoint as mgrsToWGS84 } from '@/utils/militaryGridProjection'
+import { isNumber } from '@/utils/numberUtils.js'
 
 // 47.5 7.5
 const REGEX_WEB_MERCATOR = /^\s*([\d]{1,3}[.\d]+)\s*[ ,/]+\s*([\d]{1,3}[.\d]+)\s*$/i
+// 47.5N 7.5E
+const REGEX_WEB_MERCATOR_WITH_CARDINALS =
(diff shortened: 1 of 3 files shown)


src/utils/coordinates/coordinateExtractors.js
```diff
@@ -3,12 +3,24 @@ import proj4 from 'proj4'
 import { LV95, WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { reprojectUnknownSrsCoordsToWGS84 } from '@/utils/coordinates/coordinateUtils'
 import { toPoint as mgrsToWGS84 } from '@/utils/militaryGridProjection'
+import { isNumber } from '@/utils/numberUtils.js'
 
 // 47.5 7.5
 const REGEX_WEB_MERCATOR = /^\s*([\d]{1,3}[.\d]+)\s*[ ,/]+\s*([\d]{1,3}[.\d]+)\s*$/i
+// 47.5N 7.5E
+const REGEX_WEB_MERCATOR_WITH_CARDINALS =
+    /^\s*([\d]{1,3}[.\d]+)\s*([NSEW]?)\s*[ ,/]+\s*([\d]{1,3}[.\d]+)\s*([NSEW]?)\s*$/i
```
ltshb, line 12: Should be 
```
/^\s*(\d{1,3}(\.\d+)?)\s*([NSEW]?)\s*[,/]?\s*(\d{1,3}(\.\d+)?))\s*([NSEW]?)\s*$/i
```
But are you sure that web mercator cardinal are not more than 999 ? Because your regex specify 3 digit, so the regex above would now match 999.99999999999 N but not 1000 N

ltshb (commented):
I see that most of the regex are not correct they would match number like this `11.11.11.11` due to `[.\d]+` !

ltshb:
> looks like the MGRS has some issue with the new regexs

Solved was just a `\s+` that I needed to change to `\s*` I though that in MGRS you needed two numbers, but it can be one.

ltshb (approved):

PR #970, opened 2024-06-27
PB-661 : fix Android (and more broadly) mobile feature selection - #patch

pakb:
using OL built-in event instead of doing it ourselves (and failing at that, from the looks of it)

Please test on your mobile devices if you see anything weird (especially if you have an Android phone)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-661-android-feature-selection/index.html)


src/modules/map/components/LocationPopup.vue
```diff
@@ -72,17 +72,9 @@ watch(showEmbedSharing, () => {
...
-        if (showEmbedSharing.value) {
-            if (oldQuery['lang'] != newQuery['lang']) {
-                updateShareLink()
-            }
-        }
-    }
-)
+watch(() => route.query, updateShareLink, {
+    deep: true,
+})
```
ltshb, line 77: Note here we should only update the sharelink if the share tab is selected to avoid creating shortlink for nothing !

ltshb (changes requested):
On Iphone 11 the long touch to get the position popup don't work anymore, also an issue with the shortlink this comment in code.

ltshb (approved):

PR #971, opened 2024-06-27
PB-734 : search all layers with ID before fallback to technical name - #patch

pakb:
so that 3D config don't get shown in the catalogue (they have the same technical name as the one that should be shown)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-734-topic-layers-config-mismatch/index.html)


rebert (approved):
Thanks for your fast fix !

PR #976, opened 2024-07-03
PB-771 : only react to click on the map (not in the tooltip)

pakb:


pakb:
> @pakb which problem are you exactly trying to solve here ?

Click on a feature (ÖV for instance) and try opening the second feature detail

ltshb (approved):
OK I got it from the ticket description, would be good to put it as well in the git commit message and PR description.

PR #978, opened 2024-07-03
PB-723: Keep toolbox buttons in full screen mode - #patch

ltshb:
[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-723-full-screen/index.html)


pakb (approved):
Tests need some fixing, but looking good!

PR #980, opened 2024-07-04
PB-747: Fix potential startup race condition - #patch

ltshb:
During the startup if the app.isReady was set too early, the initial URL parsing
was not done which would have left the application blank.

We observed such behavior on the prod deployment which was only randomly reproducable.
I suspect that this was the issue but cannot tell for sure.

I also improved logging of the startup procedure.

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-747-logs/index.html)

### DIFF
--- a/src/router/storeSync/storeSync.routerPlugin.js
+++ b/src/router/storeSync/storeSync.routerPlugin.js
@@ -183,6 +183,13 @@ function urlQueryWatcher(store, to, from) {
     return undefined
 }
 
+function initialUrlQueryWatcher(to, store, router) {
+    const newRoute = urlQueryWatcher(store, to)
+    if (newRoute) {
+        router.push(newRoute)
+    }
+}
+
 /**
  * Plugin that syncs what is in the URL with what is in the store (and vice-versa). It also reacts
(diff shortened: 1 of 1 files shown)


src/router/storeSync/storeSync.routerPlugin.js
```diff
@@ -232,22 +251,31 @@ const storeSyncRouterPlugin = (router, store) => {
...
-                        '[Router store plugin] app is ready, trigger initial URL query watcher'
-                    )
-                    const newRoute = urlQueryWatcher(store, to)
-                    if (newRoute) {
-                        router.push(newRoute)
-                    }
+                    // If the app was not yet ready after entering the map view, we need to
+                    // trigger the initial urlQuery watcher otherwise we have a blank application.
+                    log.info(`[store sync router] App is ready, trigger initial URL query watcher`)
+                    initialUrlQueryWatcher(to, store, router)
```
ltshb, line 263: Most of the time the initial URL query watcher happen here, if you comment this line you have exactly the same behavior observed in ticket PB-747

pakb (approved):
👍

PR #982, opened 2024-07-04
Add app version in settings menu on phone layout - #patch

pakb:
it isn't possible to read the version anywhere on phone, making it a bit hard to know if we are testing on the correct version

[Test link](https://sys-map.int.bgdi.ch/preview/fix-add-version-on-phone/index.html)

### DIFF
--- a/src/modules/menu/components/settings/MenuSettings.vue
+++ b/src/modules/menu/components/settings/MenuSettings.vue
@@ -1,6 +1,13 @@
 <script setup>
+import { computed } from 'vue'
+import { useStore } from 'vuex'
+
 import LangSwitchToolbar from '@/modules/i18n/components/LangSwitchToolbar.vue'
+import MapFooterAppVersion from '@/modules/map/components/footer/MapFooterAppVersion.vue'
 import LinksToolbar from '@/modules/menu/components/settings/LinksToolbar.vue'
+
+const store = useStore()
+const isPhoneMode = computed(() => store.getters.isPhoneMode)
 </script>
 
 <template>
@@ -11,5 +18,6 @@ import LinksToolbar from '@/modules/menu/components/settings/LinksToolbar.vue'
         <div id="menu-links">
             <LinksToolbar />
         </div>
+        <MapFooterAppVersion v-if="isPhoneMode" />
     </div>
 </template>
--- a/src/modules/menu/components/settings/MenuSettings.vue
+++ b/src/modules/menu/components/settings/MenuSettings.vue
@@ -3,21 +3,30 @@ import { computed } from 'vue'
 import { useStore } from 'vuex'
 
 import LangSwitchToolbar from '@/modules/i18n/components/LangSwitchToolbar.vue'
-import MapFooterAppVersion from '@/modules/map/components/footer/MapFooterAppVersion.vue'
 import LinksToolbar from '@/modules/menu/components/settings/LinksToolbar.vue'
+import AppVersion from '@/utils/components/AppVersion.vue'
 
 const store = useStore()
 const isPhoneMode = computed(() => store.getters.isPhoneMode)
 </script>
 
 <template>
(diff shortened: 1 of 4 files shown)


src/modules/menu/components/settings/MenuSettings.vue
```diff
@@ -1,6 +1,13 @@
 <script setup>
+import { computed } from 'vue'
+import { useStore } from 'vuex'
+
 import LangSwitchToolbar from '@/modules/i18n/components/LangSwitchToolbar.vue'
+import MapFooterAppVersion from '@/modules/map/components/footer/MapFooterAppVersion.vue'
```
ltshb, line 6: Not a good idea to do cross module import !

src/modules/menu/components/settings/MenuSettings.vue
```diff
@@ -11,5 +18,6 @@ import LinksToolbar from '@/modules/menu/components/settings/LinksToolbar.vue'
         <div id="menu-links">
             <LinksToolbar />
         </div>
+        <MapFooterAppVersion v-if="isPhoneMode" />
```
ltshb, line 21: I would not place the version on mobile PROD here, it is too predominant ! I know this is helpful for developers but not for the users. You can maybe try to keep it in the footer on mobile in the center very small and in white that we almost don't see it.

ltshb (changes requested):

src/modules/menu/components/settings/MenuSettings.vue
pakb: I moved it to shared components

src/modules/menu/components/settings/MenuSettings.vue
pakb: I made it way smaller, and not take more space than before (position absolute)

src/modules/menu/components/settings/MenuSettings.vue
```diff
@@ -1,15 +1,32 @@
 <script setup>
+import { computed } from 'vue'
+import { useStore } from 'vuex'
+
 import LangSwitchToolbar from '@/modules/i18n/components/LangSwitchToolbar.vue'
 import LinksToolbar from '@/modules/menu/components/settings/LinksToolbar.vue'
+import MapFooterAppVersion from '@/utils/components/AppVersion.vue'
```
ltkum, line 7: I would import `AppVersion` here, unless there is something I'm completely missing.

src/views/MapView.vue
```diff
@@ -17,6 +16,7 @@ import TimeSliderButton from '@/modules/map/components/toolbox/TimeSliderButton.
 import MapModule from '@/modules/map/MapModule.vue'
 import MenuModule from '@/modules/menu/MenuModule.vue'
 import { UIModes } from '@/store/modules/ui.store'
+import MapFooterAppVersion from '@/utils/components/AppVersion.vue'
```
ltkum, line 19: same as above.

ltkum (commented):
I would just change the imports, otherwise looks good to me.

ltkum (approved):

ltshb (commented):
Perso I really don't like it, it is too predominant on the mobile 
image
We should refine this before going to PROD and see with the other of the team what they think of this. On desktop it is fine as you can easily almost hide it, but here in this menu it doesn't fit in my opinion.
Also I now the usefulness to have it to debug, we don't debug that often on mobile.
If you want to have it here some other thought on how to do it
1. Make it invisible (white color) and text selectable, so you could select it to read it
2. Create a new button/link called "About", replacing the "More informations" button. This button/link on Desktop would open a window with more information like; version, application copyright, link to github repository, link to swisstopo website, ...). On mobile this could be a section in the settings menu similar to the embed section in the share menu. This would of course need some refinement in the team.
4. keep it in the footer as on desktop make it it almost invisible next to the attribution

pakb:
replaced by #1028

PR #984, opened 2024-07-04
PB-747: Enabled logs on PROD to try to understand the blank screen issue - #patch

ltshb:
[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-747-logs-2/index.html)


pakb (approved):
so that means the last fix wasn't enough 😞

ltshb:
Unfortunately not.

PR #987, opened 2024-07-04
PB-704 : iframe preview update - #patch

pakb:
Adds new change event triggered to iframe parent. It will be triggered when route query changes, as it is not possible with VueJS to get the current SRC attribute of an iframe (it always returns the value that was set with :src, without updating it...)

This way we can receive the new state of the iframe and update the code snippet accordingly.

Removing shortlink generation for embed code snippet (like old viewer), so that param can then be updated by the user copying it (and less calls to our shortlink backend)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-704-iframe-preview-update/index.html)

### DIFF
--- a/src/api/iframeFeatureEvent.api.js
+++ b/src/api/iframeFeatureEvent.api.js
@@ -1,6 +1,27 @@
-import log from '@/utils/logging.js'
+import log from '@/utils/logging'
 
 const targetWindow = parent ?? window.parent ?? window.opener ?? window.top
+
+/**
+ * All events fired by the iFrame postMessage implementation.
+ *
+ * @enum
+ */
+export const IFRAME_EVENTS = {
+    /**
+     * Event raised whenever the app changes its state (the URL changed).
+     *
+     * Payload of this event : a String with the new URL of the viewer
+     */
+    CHANGE: 'gaChange',
+    /**
+     * Event raised when a feature has been selected. Will fire as many events as there are feature
+     * selected on the map (won't bundle all features into one event)
+     *
+     * Payload of this event : a JSON containing the layerId and featureId of the selected feature
+     */
+    FEATURE_SELECTION: 'gaFeatureSelection',
+}
 
 /**
  * Sends information to the iFrame's parent about features, through the use of the postMessage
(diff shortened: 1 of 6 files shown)


ltkum (changes requested):
Quick test review before going through the code : 
The embed map seems to be taking into account only the changes to the viewer made inside the embedded link. 
For example, if I activate a layer, or zoom, then want to share an embed link, the link shown will be the one for the default state, and once in the preview I will also see the default state. Once I make change within the preview, those will be applied to the embed links.

src/api/iframeFeatureEvent.api.js
```diff
@@ -43,3 +45,29 @@ export function sendFeatureInformationToIFrameParent(features) {
...
+export function sendChangeEventToParent() {
+    if (!targetWindow) {
+        log.debug(
+            'Embed view loaded as root document of a browser tab, cannot communicate with opener/parent'
+        )
+        return
+    }
+    targetWindow.postMessage(
+        {
+            type: 'gaChange',
```
ltshb, line 66: We should start to document those API features that we implement in a wiki page !

src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -120,6 +117,18 @@ function toggleEmbedSharing() {
...
     showPreviewModal.value = !showPreviewModal.value
+    if (showPreviewModal.value) {
+        window.addEventListener('message', onPreviewChange)
+    } else {
+        window.removeEventListener('message', onPreviewChange)
+    }
+}
+
+function onPreviewChange(e) {
+    if (e?.data?.type === 'gaChange') {
```
ltshb, line 128: Might be a better idea to import the `gaChange` string from the iframe api file here

ltshb: In the api file we should have also each message type as variable and documented same as their payload.

ltshb (approved):
Looks good but documentation should be improved.

src/api/iframeFeatureEvent.api.js
pakb: I've created a (self-assigned) ticket to do just that after my vacation
https://jira.swisstopo.ch/browse/PB-813

ltkum (approved):
seems good to me :)

PR #988, opened 2024-07-04
PB-729 : fix recent earthquake tooltip - #patch

pakb:
We were missing a `coord` param for the HTML popup endpoint (undocumented) so that the HTML endpoint can select features for us instead of giving feature IDs (not optimal, but that's the way it is for now)

Also not requesting geometries (in identify requests) when the layer isn't highligtable, it was returning the extent of the whole layer as geom. Using instead the click coordinate as a geometry (adapting the code to be able to handle no geom)

[Test link with recent earthquakes](https://sys-map.int.bgdi.ch/preview/fix-pb-729-recent-earthquake-tooltip/index.html?layers=ch.bafu.gefahren-aktuelle_erdbeben)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-729-recent-earthquake-tooltip/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/OpenLayersPopover.vue
+++ b/src/modules/map/components/openlayers/OpenLayersPopover.vue
@@ -11,7 +11,7 @@ import MapPopover, { MapPopoverMode } from '@/modules/map/components/MapPopover.
 const props = defineProps({
     coordinates: {
         type: Array,
-        required: true,
+        default: null,
     },
     authorizePrint: {
         type: Boolean,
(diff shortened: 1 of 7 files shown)


src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -127,19 +127,19 @@ export function createLayerObject(parsedLayer, currentLayer, store, featuresRequ
...
-                    featuresRequests.push(
-                        getFeature(
-                            store.getters.getLayerConfigById(parsedLayer.id),
-                            featureId,
-                            store.state.position.projection,
-                            {
+                    const layer = store.getters.getLayerConfigById(parsedLayer.id)
+                    if (layer.isHighlightable) {
+                        // only highlightable feature will output something, for the others a click coordinate is required
+                        // (and we don't have it if we are here, as we are dealing with pre-selected feature in the URL at app startup)
```
ltkum, line 133: So, I know `layer` is already used outside of the scope but
I would move this out of the forEach, and have at line 124 somthing like :
`124 : const layerConfig = store.getters.getLayerConfigById(parsedLayer.id)`
`125: if (features !== undefined && layerConfig.isHighlightable) { ... `
It doesn't change that much, I must concede that, but I do find it a bit cleaner

ltkum (commented):
A few details I would change but that are not breaking anything. Good job :+1:

src/modules/map/components/openlayers/OpenLayersPopover.vue
```diff
@@ -49,6 +49,9 @@ onUnmounted(() => {
 })
 
 function getPixelForCoordinateFromMap() {
+    if (!coordinates.value) {
+        return
```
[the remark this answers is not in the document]
pakb: I renamed the function to better explain what it does, having a `get...` was suggesting it was supposed to return something, but that's not the case

src/router/storeSync/LayerParamConfig.class.js
pakb: `layer` was already computed above, so I removed the `.getters.` call in the for loop and used the already present variable instead

ltkum: that's even better then

ltkum (approved):

PR #991, opened 2024-07-05
PB-775 : show compared layer name when slider is hovered/used - #patch

pakb:
[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-775-compared-layer-name/index.html)


ltkum (approved):
Code seems good and it works well. 
At first, I had trouble seeing the label (especially when dezoomed). Maybe making the text **bold** could draw the eye to it a little bit, but it's not really a blocking issue. You can merge whenever you're satisfied with it :)

PR #992, opened 2024-07-05
PB-750 : adapt CSS to enable browser print of the map - #patch

pakb:
with only the layer attributions and eventual feature tooltip visible (no menu, no footer)

Had to adapt a bit the OpenLayerMouseTracker component so that it has a root HTML element (on which we can apply a class from the component using it)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-750-browser-printing/index.html)

### DIFF
--- a/src/scss/webmapviewer-bootstrap-theme.scss
+++ b/src/scss/webmapviewer-bootstrap-theme.scss
@@ -65,3 +65,9 @@ $nav-tabs-link-active-bg: $primary;
 .w-13 {
     width: 13%;
 }
+
+@media print {
+    .no-print {
+        display: none !important;
+    }
+}
--- a/src/views/MapView.vue
+++ b/src/views/MapView.vue
@@ -46,11 +46,12 @@ onMounted(() => {
 </script>
 
 <template>
-    <div id="map-view" class="no-print">
-        <LoadingBar v-show="showLoadingBar" />
+    <div id="map-view">
+        <LoadingBar v-show="showLoadingBar" class="no-print" />
         <MapModule>
-            <MenuModule />
+            <MenuModule class="no-print" />
             <MapToolbox
+                class="no-print"
                 geoloc-button
                 :full-screen-button="!isDrawingMode"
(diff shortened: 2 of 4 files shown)


pakb:
I will try to add the scale line too

ltkum (commented):
When trying to print, it offers to print only the left quarter / half of the map. We might have to fiddle a bit more with the CSS to have a print layout which is correct, but we are heading in the right direction (on firefox by the way, in case this worked on other browsers)

for the following print attempt, it was from the default state of the application.

image

src/scss/webmapviewer-bootstrap-theme.scss
```diff
@@ -65,3 +65,9 @@ $nav-tabs-link-active-bg: $primary;
 .w-13 {
     width: 13%;
 }
+
+@media print {
+    .no-print {
+        display: none !important;
+    }
+}
```
ltshb, line 73: This is already present in main.scss which has been added for the tooltip print.

src/views/MapView.vue
```diff
@@ -46,11 +46,12 @@ onMounted(() => {
 </script>
 
 <template>
-    <div id="map-view" class="no-print">
-        <LoadingBar v-show="showLoadingBar" />
+    <div id="map-view">
+        <LoadingBar v-show="showLoadingBar" class="no-print" />
```
ltshb, line 50: You are breaking the tooltip print here, now with your change when you print the tooltip and you select the text in the tooltip you have part of the view that is printed invisible but selectable.
image

ltshb: See https://github.com/geoadmin/web-mapviewer/pull/905/commits/e1875b1bdc72221295ce53f07d066d0feba08300 from #905

ltshb (changes requested):

pakb:
Requires a ton of work to achieve a map print without breaking the tooltip print. I won't be pursuing this more, and focus on more pressing issues

PR #996, opened 2024-07-08
PB-786: Fix geolocation error management - #patch

ltshb:
While using the geolocation you have time to times the `Unknown error` window
that appears. This is due to the fact that the geolocation can return an error,
but  while using the app on the field I noticed that it happens sometimes just
once and then everything seems to work fine. In this case the error windows is
quite annoying as you have to close it and if it happened during the initial
get current position you had to disable and re-enable the geolocation.

So now I improved the error management and do some automatic retries until 3 times
if we are the get current position phase, while in the watcher phase the error
is ignored unless it happens more that 3 times in a row.

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-786-geolocation/index.html)


ltkum (approved):
Looks good to me code wise. We'll have to see on use if it works well.

PR #998, opened 2024-07-10
PB-798: Leading/ending white spaces character is not supported in KML feature name

ltshb:
The feature name in KML doesn't support leading/ending white space. Openlayer
don't support it (also tried to used <![CDATA[]]> but don't work). On google
earth we cannot add multiline feature titles ! and when using leading space
it somehow cuts the end of the title so, not working neither.

The issue that we add is that leading white spaces were allowed and worked
while drawing, but were trimmed by openlayer when loading the drawing, so user
complained that their title did not stayed at the same place as the put them
while drawing.

To avoid this we trim the text when entering it, so the text always appears trimmed
in the view even in drawing. This way if the user add leading blank line the text
stays at the top of the icon.


ltkum (approved):
tested it and it works, good job

PR #999, opened 2024-07-10
PB-784: Use smaller map resolution for grid to avoid coordinate label…

ismailsunni:
… cut off.


ltshb (approved):
I've noticed a difference in default background layer details, but this difference is present between the PROD and INT staging and has nothing to do with this PR.

PR #1002, opened 2024-07-10
PB-761: Don't override text scale if it is zero - #patch

LukasJoss:
[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-761-default-text-size-if-text-size-equal-zero/index.html)


ltshb (approved):

ltshb:
An PR in openlayer has been already open to solve some scale issue on icon, we might add this to this PR as well, @pakb can update your PR in ol to also add this https://github.com/openlayers/openlayers/pull/15964

PR #1003, opened 2024-07-12
PB-792: Change rounding and decimal places threshold - #patch

LukasJoss:
Fixed bug which showed an area of  0 m2 if the area was just above 1000 m2
[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-792-measuring-area-rounding-issue/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureAreaInfo.vue
+++ b/src/modules/infobox/components/FeatureAreaInfo.vue
@@ -6,7 +6,6 @@ import { useStore } from 'vuex'
 import { useTippyTooltip } from '@/utils/composables/useTippyTooltip'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { reprojectGeoJsonData, transformIntoTurfEquivalent } from '@/utils/geoJsonUtils'
-import { round } from '@/utils/numberUtils'
 
 const props = defineProps({
     geometry: {
@@ -40,10 +39,10 @@ const humanReadableArea = computed(() => {
 
         const value = calculatedArea < unitThreshold ? calculatedArea : calculatedArea / divider
         result += parseFloat(value.toPrecision(precision))
-        if (calculatedArea >= unitThreshold) {
-            result += ' km'
-        } else {
+        if (calculatedArea < unitThreshold) {
             result += ' m'
+        } else {
+            result += ' km'
         }
     }
     return result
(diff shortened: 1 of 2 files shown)


ltkum (approved):

src/modules/infobox/components/FeatureAreaInfo.vue
```diff
@@ -34,7 +34,7 @@ const humanReadableArea = computed(() => {
     const calculatedArea = area(transformIntoTurfEquivalent(geometryWgs84.value))
     let result = ''
     if (calculatedArea) {
-        result += roundValueIfGreaterThan(calculatedArea, 1000, 1000000)
+        result += roundValueIfGreaterThan(calculatedArea, 10000, 1000000)
         if (calculatedArea > 10000) {
             result += ' km'
         } else {
```
ltshb, line 40: For better code readability and to be bullet proof, the constant should be defined as constant instead of hardcoding. If I understand correctly the threshold should be also used to define the unit ?

Should this not be something like this ?

```javascript
const humanReadableArea = computed(() => {
    const calculatedArea = area(transformIntoTurfEquivalent(geometryWgs84.value))
    let result = ''
    if (calculatedArea) {
        const divider = 1000000
        const threshold = 10000
        result = roundValueIfGreaterThan(calculatedArea, threshold, divider)
        if (calculatedArea > threshold) {
            result += ' km'
        } else {        
            result += ' m'
        }
    }
    return result
})

File: src/modules/infobox/components/FeatureAreaInfo.vue [hunk 2]
```diff
@@ -46,7 +46,7 @@ const humanReadableArea = computed(() => {
 
 function roundValueIfGreaterThan(value, threshold, divider) {
     if (value > threshold) {
-        return `${round(value / divider, 2)}`
+        return `${round(value / divider, 3)}`
     }
```
ltshb, line 50: Here the value of the number of decimal depends on the divider to avoid having again a 0 value ! Because if a user take a bigger divider you might end up again to a 0 value. Because this function is only used here I suggest to make the rounded decimal in case of threshold reached being a parameter and pass the 3 above in the humanReadableArea function so you have all calcul dependencies at the same place, which will make the code easier to understand and avoid such issue in future.

ltshb (commented):
You should also add some unit tests for the humanReadableArea function with different numbers also number close to the hard coded values.

PR #1004, opened 2024-07-12
PB-811: Fix go to position if geolocation is active at startup - #patch

ltshb:
If the geolocation was activated at startup using the goelocation query parameter
the application would not re-center to the position. This would break some legacy
links that use this parameter like the swisstopo "you are here" T-shirt.

This was also an issue when entering the parameter in the new query format.

We had several issue here, first the legacy parameter was totally ignored because
URLSearchParams parsed the geolocation to an empty string which was not supported
by our boolean query param.

Another issue was that per default tracking was off unless you explicitely click
on the geolocation button.

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-811-geolocalisation/index.html)


ltkum (approved):
lgtm

PR #1006, opened 2024-07-15
PB-791: Do not hide middle footer when in fullscreen - #patch

LukasJoss:
Make the middle map footer containing the infobox no longer disappear to fix following issues in fullscreen:

- When opening fullscreen while feature info is set to bottom, selecting a feature would not show an infobox
- When pressing the arrow down button on the infobox popover, the infobox would disappear
- Profiles could not be displayed, as they are embedded in the middle footer

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-791-altimeter-profile-fullscreen/index.html)

### DIFF
--- a/src/modules/map/components/footer/MapFooter.vue
+++ b/src/modules/map/components/footer/MapFooter.vue
@@ -21,12 +21,11 @@ onUpdated(() => {
 </script>
 
 <template>
-    <div
-        class="map-footer"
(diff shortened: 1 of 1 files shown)


ltkum (approved):
looks like it works to me :+1:

ltshb (approved):

PR #1007, opened 2024-07-16
PB-801: Allow all possible quotes style in coordinate extractor

ltshb:
There are many quotes style that a user might use to enter a coordinate in
the search bar, a user used the following coordinates that did not worked

47°09′43″ N 7°40′33″E

So added all possible quotes styles to the regex.


LukasJoss (approved):

ltkum (approved):
I don't think you've forgotten any way to make quotes, and it works for all of them

PR #1008, opened 2024-07-17
PB-803, PB-631, PB-175: Use proxy for all non geoadmin KML icon urls - #patch

ltshb:
Because openlayer only provide a synchrone hook function to modify the icon
url we do the proxyfy regardless if the remote server support or not CORS. However
we do run the check asynchronously and if it failed we add a warning window
on top of the screen.

This PR solve issues; PB-803, PB-631 (icon scale is too small has to be increase to see icons), PB-816

[Test link with CORS issue](https://sys-map.int.bgdi.ch/preview/bug-pb--631-cors-kml-icons/index.html#/map?lang=en&center=2604108.64,1189241.41&z=6.839&bgLayer=ch.swisstopo.pixelkarte-grau&topic=ech&layers=KML%7Chttps://www.uvek-gis.admin.ch/BAZL/Karte_AeMC_AME/Aeromedical+Examiner+(AME).kml;KML%7Chttps://www.uvek-gis.admin.ch/BAZL/Karte_AeMC_AME/Aeromedical+Center+(AeMC).kml&featureInfo=bottomPanel)
[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb--631-cors-kml-icons/index.html)

### DIFF
--- a/src/utils/components/WarningWindow.vue
+++ b/src/utils/components/WarningWindow.vue
@@ -37,9 +37,9 @@ const emit = defineEmits(['close'])
 <template>
     <div
         v-show="!hide"
-        class="simple-window card bg-danger text-white fw-bold"
+        class="simple-window card bg-warning fw-bold"
         :class="{ 'dev-disclaimer-present': hasDevSiteWarning }"
-        data-cy="error-window"
+        data-cy="warning-window"
     >
         <div
             class="card-header d-flex align-items-center justify-content-sm-end"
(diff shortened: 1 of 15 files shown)


src/utils/components/WarningWindow.vue
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Hide the modal with backdrop, can be used to temporarily hide the modal without loosing its
+     * content
+     */
+    hide: {
+        type: Boolean,
+        default: false,
+    },
+    /** Add a minimize button in header that will hide/show the body */
+    hasMinimize: {
```
LukasJoss, line 20: this variable seems to do nothing

LukasJoss (approved):
Is there a time limit how long we will support icons without cors? If yes this should maybe be added to the error message.

src/utils/components/WarningWindow.vue
ltshb: good point I'll remove this this is a copy from ErrorWindow

PR #1011, opened 2024-07-24
PB-790: Update mapfishprint to include fix for empty text on annotation.

ismailsunni:


package.json
```diff
@@ -39,7 +39,7 @@
         "@fortawesome/free-solid-svg-icons": "^6.5.2",
         "@fortawesome/vue-fontawesome": "^3.0.8",
         "@geoblocks/cesium-compass": "^0.5.0",
-        "@geoblocks/mapfishprint": "^0.2.14",
+        "@geoblocks/mapfishprint": "github:geoblocks/mapfishprint#932253b0e8ccc0c1d911cab8e02c5b2fc7a061af",
```
ltshb, line 42: Would be much better to wait and have a proper release tag here to go to prod (merging this PR will push it directly to prod !)

ltshb (changes requested):

ismailsunni:
@ltshb yes, that's what I mean with my comment in the PR description. But after I read it again, it doesn't say like that.

And yes, it shouldn't be merged until we have a new version released.

I tried to change it back to a draft PR, but it seems it's not possible.

ismailsunni:
Superseded by #1021

PR #1015, opened 2024-07-30
PB-818: Ensuring scales are respected on print - #minor

ltkum:
Issue : When we were sending print requests, the scales of the symbols and texts in kmls were off. For icons, they were too small, and the texts were too big.

Fix : We corrected the scales for icons and labels. Icons used a DPI adaptation of * 90 / 254 which was inherited from the old mapviewer. After some empirical testing, it looks like on the new mapviewer needs a *150 / 254 DPI adaptation Labels were not using the 'scale' parameter for the fonts at all, and they now also use a 150 / 254 DPI adaptation.

It is important to note that the '254' given here is a computed value, as some layers will have smaller resolutions and their DPI will be used in those situations

While testing, try to test with different screens, to ensure the result is consistent at all times.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-818-font-and-image-size-dont-match-in-print/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -83,18 +83,11 @@ class GeoAdminCustomizer extends BaseCustomizer {
         symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
         symbolizer.haloRadius = adjustWidth(symbolizer.haloRadius, this.printResolution)
         symbolizer.conflictResolution = false
-        symbolizer.fontSize =
-            Math.ceil(2 * adjustWidth(parseInt(symbolizer.fontSize), this.printResolution)) + 'px'
-
         // we try to adapt the font size and offsets to have roughly the same
         // scales on print than on the viewer.
         try {
-            symbolizer.labelYOffset = symbolizer.labelYOffset
-                ? adjustWidth(symbolizer.labelYOffset, this.printResolution)
-                : 0
-            symbolizer.labelXOffset = symbolizer.labelXOffset
-                ? adjustWidth(symbolizer.labelXOffset, this.printResolution)
-                : 0
+            symbolizer.labelYOffset = adjustWidth(symbolizer.labelYOffset, this.printResolution)
+            symbolizer.labelXOffset = adjustWidth(symbolizer.labelXOffset, this.printResolution)
             symbolizer.fontSize = `${adjustWidth(
                 parseInt(symbolizer.fontSize) * text.getScale(),
                 this.printResolution
@@ -146,8 +139,8 @@ class GeoAdminCustomizer extends BaseCustomizer {
         if (size) {
             symbolizer.graphicWidth = adjustWidth(size[0] * scale, this.printResolution)
         }
-        symbolizer.graphicXOffset = symbolizer.graphicXOffset ? symbolizer.graphicXOffset : 0
-        symbolizer.graphicYOffset = symbolizer.graphicYOffset ? symbolizer.graphicYOffset : 0
+        symbolizer.graphicXOffset = symbolizer.graphicXOffset ?? 0
+        symbolizer.graphicYOffset = symbolizer.graphicYOffset ?? 0
 
         if (symbolizer.fillOpacity === 0.0 && symbolizer.fillColor === '#ff0000') {
             // Handling the case where we need to print a circle in the end of measurement lines
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -86,9 +86,6 @@ class GeoAdminCustomizer extends BaseCustomizer {
         symbolizer.fontSize =
             Math.ceil(2 * adjustWidth(parseInt(symbolizer.fontSize), this.printResolution)) + 'px'
 
-        // Ideally this should be done in the geoblocks/mapfishprint
-        // but it's quite complex to handle all the cases
-
         // we try to adapt the font size and offsets to have roughly the same
         // scales on print than on the viewer.
         try {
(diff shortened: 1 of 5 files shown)


src/utils/styleUtils.js
```diff
@@ -161,7 +161,8 @@ export function adjustWidth(width, dpi) {
     if (!width) {
         return
     }
-    // 90 is choosen to compensate the difference DPI between the WMTS or WMS and the print DPI used.
-    // The number is from the old geoadmin (see the link above)
-    return (width * 90) / dpi
+    // 90 Was the old mapviewer magic number set to compensate for the DPI difference between the
+    // mapviewer and the service print. Currently, 150 seems to be giving better results at keeping
+    // the same scale overall.
+    return (width * 150) / dpi
```
pakb, line 167: I would move this "magic number" into a new dedicated `config/print.config.js` file (you'll see that config has been split into multiple files when you'll rebase on develop)

src/api/print.api.js
```diff
@@ -121,12 +138,23 @@ class GeoAdminCustomizer extends BaseCustomizer {
...
+
+            symbolizer.graphicYOffset = symbolizer.graphicYOffset
+                ? (symbolizer.graphicYOffset = adjustWidth(-size[1], this.printResolution))
+                : 0
         }
         if (size) {
             symbolizer.graphicWidth = adjustWidth(size[0] * scale, this.printResolution)
         }
+        symbolizer.graphicXOffset = symbolizer.graphicXOffset ? symbolizer.graphicXOffset : 0
+        symbolizer.graphicYOffset = symbolizer.graphicYOffset ? symbolizer.graphicYOffset : 0
```
pakb, line 157: can be simplified to something like `symbolizer.graphicYOffset ?? 0`

src/api/print.api.js
```diff
@@ -92,6 +92,23 @@ class GeoAdminCustomizer extends BaseCustomizer {
         symbolizer.pointRadius = adjustWidth(symbolizer.pointRadius, this.printResolution)
         symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
         symbolizer.haloRadius = adjustWidth(symbolizer.haloRadius, this.printResolution)
+
+        // we try to adapt the font size and offsets to have roughly the same
+        // scales on print than on the viewer.
+        try {
+            symbolizer.labelYOffset = symbolizer.labelYOffset
+                ? adjustWidth(symbolizer.labelYOffset, this.printResolution)
```
pakb, line 100: might be a better idea to make `adjustWidth` to return `0` if the input is falsy.
I don't really like multiple ternary operator in a row, that makes the code a bit harder to read/understand

src/api/print.api.js
ltkum: I was convinced somehow that the ?? operator was only for 'null', but it seems you're right and it works as well for undefined.

src/config/print.config.js
```diff
@@ -0,0 +1,5 @@
+// 90 Was the old mapviewer magic number set to compensate for the DPI difference between the
+// mapviewer and the service print. Currently, 144 seems to be giving better results at keeping
+// the same scale overall.
+// We're going to check thoroughly if this works as intended, or if this is a monitor dpi setting element.
+export const PRINT_MAGIC_NUMBER = 144
```
pakb, line 5: I would maybe not call it "magic number" but something that self-explains what it does.
Maybe `PRINT_DPI_COMPENSATION`?

src/config/print.config.js
```diff
@@ -0,0 +1,5 @@
+// 90 Was the old mapviewer magic number set to compensate for the DPI difference between the
+// mapviewer and the service print. Currently, 144 seems to be giving better results at keeping
+// the same scale overall.
+// We're going to check thoroughly if this works as intended, or if this is a monitor dpi setting element.
```
pakb, line 4: We should check before publishing, so this line of comment is a bit irrelevant don't you think?
A comment should help understand why this is there, not what we will do with it (otherwise write a TODO)

src/utils/styleUtils.js
```diff
@@ -158,10 +159,9 @@ export const highlightPointStyle = new Style({
 // Change a width according to the change of DPI (from the old geoadmin)
 // Originally introduced here https://github.com/geoadmin/mf-geoadmin3/pull/3280
 export function adjustWidth(width, dpi) {
-    if (!width) {
-        return
+    if (!width || isNaN(width)) {
```
pakb, line 162: you should maybe also check for `dpi` in the same fashion, as if you'd let it go trough with a value of `0` you'll have a divided by zero error bellow

tests/cypress/tests-e2e/print.cy.js
```diff
@@ -426,7 +426,12 @@ describe('Testing print', () => {
                 ).to.lessThan(1.5) // thinner than the drawn in the OL map.
             })
         })
-        it('should send a print request correctly to mapfishprint (icon and label)', () => {
+        /**
+         * Currently, we are working on the print, which might alter the way data is sent, or the
+         * way data is interpreted. When the work on the print API (and backend) is over, we will
+         * need to modify those tests by providing new pdf reports as fixtures.
+         */
+        it.skip('should send a print request correctly to mapfishprint (icon and label)', () => {
```
pakb, line 434: Is it a big effort to adapt these tests with your changes? if not please do, there are many little things that would be bad not to test.
Even if it is painful to maintain tests, they are our safeguard against doing regressions

ltkum: I was dumb, all I need to do is check on the attributes sent here. I can do that as soon as I'm sure the result is visibly good enough.

pakb (approved):

PR #1017, opened 2024-08-05
PB-804 : don't center on search query when sharing position - #minor

ltkum:
Issue : When sharing a position, if there was a search query set, the search query would take priority over the position sharing, removing the shared position pin and centering on the search query result instead.
Fix : When a position is shared, it sets the marker url parameter, which means we can decide that when a marker is present, we should not center on the query result.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-804-dont-follow-search-on-position-sharing/index.html)


pakb (commented):
I can't make it work as intended, have a look at this URL :

https://sys-map.int.bgdi.ch/preview/bugfix-pb-804-dont-follow-search-on-position-sharing/index.html#/map?swisssearch=2672000,1216250&center=2557000,1196750&crosshair=2557000,1196750

The app should start centered on Lake Neuchatel with a crosshair, but still goes to center the view on the search input

pakb (commented):
there's still some issues with the notation `crosshair=x,y`, see https://sys-map.int.bgdi.ch/preview/bugfix-pb-804-dont-follow-search-on-position-sharing/index.html#/map?swisssearch=2672000,1216250&center=2557000,1196750&crosshair=2557000,1196750

it should open the viewer over lake Neuchatel, but is still moving to the searched coordinates instead

ltkum:
> there's still some issues with the notation `crosshair=x,y`, see https://sys-map.int.bgdi.ch/preview/bugfix-pb-804-dont-follow-search-on-position-sharing/index.html#/map?swisssearch=2672000,1216250&center=2557000,1196750&crosshair=2557000,1196750
> 
> it should open the viewer over lake Neuchatel, but is still moving to the searched coordinates instead

Found out the issue and it should now work as expected

src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -22,6 +23,17 @@ function dispatchCrossHairFromUrlIntoStore(to, store, urlParamValue) {
...
+            const crossHairPosition = [parseFloat(parts[0]), parseFloat(parts[1])]
+            if (!(isNaN(crossHairPosition[0]) || isNaN(crossHairPosition[1]))) {
+                promisesForAllDispatch.push(
+                    store.dispatch('setCrossHair', {
+                        crossHair: CrossHairs.marker,
+                        crossHairPosition,
+                        dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
+                    })
+                )
+            }
```
ltshb, line 36: Where is the spec for the crosshair parameter ? If I understand correctly your code it is like this ?

- crosshair=MARKER,X,Y
- crosshair=X,Y

Was it always like this ? I think this is not a good idea and to be consistent with other parameters like the layers, the second case should be `crosshair=,X,Y` to use the default marker.

ltshb: Is this code really needed for which use case ?

ltshb (changes requested):

src/router/storeSync/CrossHairParamConfig.class.js
ltkum: The crosshair parameters used to have two ways of being handled : 
`crosshair=MARKER` would add the marker in the center
`crosshair=MARKER,X,Y` would add the marker in a specific position

During the tests, Pascal tried to only use the coordinates in the links, and we decided to add the `crosshair=X,Y` too, as thinking "I want the crosshair to be in such position" is quite instinctive.

I like your idea of a `crosshair=,X,Y` to be consistent with others similar parameters, but that would require adding the crosshair to the legacy parser. @pakb , what do you think ?

pakb: yep, go for the `,X,Y`, that is more consistent with what we already do on the `layers` URL param

src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -1,10 +1,30 @@
...
+ *
+ * @param {Object} to
+ * @param {Object} store
+ * @param {URLSearchParams} urlParamValue
+ * @returns
+ */
 
 function dispatchCrossHairFromUrlIntoStore(to, store, urlParamValue) {
     const promisesForAllDispatch = []
+    const error = new WarningMessage('parameter_error', {
```
ltshb, line 24: Dito use new ErrorMessage class

src/store/modules/ui.store.js
```diff
@@ -153,14 +154,13 @@ export default {
         showDisclaimer: true,
 
         /**
-         * Text to display in the error window
+         * A WarningMessage object containing parameters for error texts, so we can have error
```
ltshb, line 157: Should be ErrorMEssage

ltshb (approved):

PR #1018, opened 2024-08-05
PB-800 : simplify GPXs geometry - #patch

pakb:
so that we diminish, or totally avoid, the coastal paradox we've encounter so far with GPX that contains points close together, following a steep mountain (with the GPS error, that would mean it would jump back and forth on the hill, creating fake height distance, and thus giving a very VERY poor hiking time estimation)

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-800-profile-discrepancy/index.html)

### DIFF
--- a/src/utils/gpxUtils.js
+++ b/src/utils/gpxUtils.js
@@ -3,6 +3,7 @@ import bbox from '@turf/bbox'
 import { isEmpty as isExtentEmpty } from 'ol/extent'
 import GPX from 'ol/format/GPX'
 
+import { GPX_GEOMETRY_SIMPLIFICATION_TOLERANCE } from '@/config'
 import CoordinateSystem from '@/utils/coordinates/CoordinateSystem.class'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { gpxStyles } from '@/utils/styleUtils'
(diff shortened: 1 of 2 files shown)


src/utils/gpxUtils.js
```diff
@@ -44,7 +44,11 @@ export function parseGpx(gpxData, projection) {
         featureProjection: projection.epsg,
     })
     features.forEach((feature) => {
-        feature.setStyle(gpxStyles[feature.getGeometry().getType()])
+        const geom = feature.getGeometry()
+        // PB-800 : to avoid a coastline paradox we simplify the geometry of GPXs
+        // 12.5 meters is what was used in the old viewer, see https://github.com/geoadmin/mf-geoadmin3/blob/ce24a27b0ca8192a0f78f7b8cc07f4e231031304/src/components/GeomUtilsService.js#L207
+        feature.setGeometry(geom.simplify(12.5))
```
ltshb, line 50: Might make sense to already put this value in the config, instead of hardcoding it here. So if we need to also add the simplify on external KML later on we can reuse it.

ltshb (approved):
I didn't test but I guess you tested it?

pakb:
> I didn't test but I guess you tested it?

There are GPX to test against on the ticket, yes I did test, and the outcome is now the same as `mf-geoadmin3`

PR #1020, opened 2024-08-06
PB-866: Fix undefined profile name for GPX track and fix crash in left click on 3D - #patch

ltshb:
When displaying a GPX track, its title in the profile was undefined. This was
due to the fact that GPX on a ol feature seems to have no ID, but have a name.

Also fixed a crash on 3d view when clicking on the map (preventing any feature identify).

[Test link](https://sys-map.int.bgdi.ch/preview/bug-gpx-profile/index.html)

### DIFF
--- a/src/store/modules/features.store.js
+++ b/src/store/modules/features.store.js
@@ -157,9 +157,9 @@ export default {
          *   have a LineString or Polygon feature
          */
         profileFeature: null,
-        /** @type {ElevationProfile | null} Profile Data for the profile geometry */
+        /** @type {LayerFeature | null} Profile Data for the profile geometry */
         profileData: null,
-        /** @type {ProfileError | null} */
+        /** @type {ElevationProfile | null} */
         profileRequestError: null,
     },
     getters: {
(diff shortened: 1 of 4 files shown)


ltshb:
@pakb I've noticed that identify of vectore layer (kml or gpx) on 3D is not working, I quickly (2min) debug it but could not find the issue, I don't know if there is any ticket yet for this issue, could you double check if there is any ticket and create one if needed (as I'm not working today, If just fixed this for my personal usage, preparation for my weekend 😉 ) ? thanks.

src/store/modules/features.store.js
```diff
@@ -157,9 +157,9 @@ export default {
          *   have a LineString or Polygon feature
          */
         profileFeature: null,
-        /** @type {ElevationProfile | null} Profile Data for the profile geometry */
+        /** @type {LayerFeature | null} Profile Data for the profile geometry */
         profileData: null,
-        /** @type {ProfileError | null} */
+        /** @type {ElevationProfile | null} */
```
pakb, line 162: I don't think that's the good types for those, you maybe wanted to add `@type {LayerFeature | null}` to the `profileFeature` props above

pakb:
> I don't know if there is any ticket yet for this issue, could you double check if there is any ticket and create one if needed

Ticket created 👍 
https://jira.swisstopo.ch/browse/PB-867

pakb (approved):

PR #1022, opened 2024-08-07
PB-154: Geolocation device orientation (gyroscope) - #minor

ltshb:
This PR replaced the initial work of #965

Testing slack canvas: https://swisstopo.slack.com/docs/TG623EKS6/F07GV0DQS7L

It also fixes some bugs related to goelocation see commits.

Because of some issues with some android devices (mostly samsung devices), the device orientation is only enabled on device where we can be sure that it works well, meaning Iphone and Pixel phone.

[Test link](https://sys-map.int.bgdi.ch/preview/feat-pb-154-geolocation-orientation/index.html)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-154-geolocation-orientation/index.html)


src/store/plugins/geolocation-management.plugin.js
```diff
@@ -51,6 +52,12 @@ const handlePositionAndDispatchToStore = (position, store) => {
         position: positionProjected,
         ...dispatcher,
     })
+    if (position.coords.heading != null && position.coords.speed >= 0.5) {
+        store.dispatch('setGeolocationHeading', {
+            heading: toRadians(position.coords.heading),
+            ...dispatcher,
+        })
+    }
```
ltshb, line 60: @LukasJoss @pakb why using 0.5 here ? I understand that if the speed is 0 the heading will NaN as it cannot be computed, but why taking a 0.5 speed and not simply checking that heading is not NaN ? Why 0.5 and not any other number ? from where did this number come from ?

schtibe:
As discussed with @ltshb, Absolute Listener with Alpha Value is OK on
`Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/127.0.0.0 Mobile Safari/537.36`

pakb (approved):

PR #1023, opened 2024-08-07
Added SSL for the local development server - #minor

ltshb:
This helps to test some features like the geolocation on device mobile
which requires ssl (at least on iphone).

[Test link](https://sys-map.int.bgdi.ch/preview/feat-localhost-ssl/index.html)


pakb (approved):
meaning it will be an auto-signed certificate? Does it still work with the more sensitive browers? (Safari on mobile for instance)

ltshb:
> meaning it will be an auto-signed certificate? Does it still work with the more sensitive browers? (Safari on mobile for instance)

Yes the only drawback is that you always have to use https and bypass browser protection by saying each times you know what you do. This is usefull to test geolocation on mobile (iphone) as you are not allowed to enable geolocation on http ! But I could let the code in and commented so if we need it we can simply uncomment it ?

pakb:
> > meaning it will be an auto-signed certificate? Does it still work with the more sensitive browers? (Safari on mobile for instance)
> 
> @pakb Yes the only drawback is that you always have to use https and bypass browser protection by saying each times you know what you do. This is usefull to test geolocation on mobile (iphone) as you are not allowed to enable geolocation on http ! But I could let the code in and commented so if we need it we can simply uncomment it ?

Is the code still served as unsecured in parallel or only HTTPS?
If only HTTPS, I would propose to add a new `npm run dev:https` or `npm run dev:ssl` for the use case of GeolocationAPI testing (and other APIs that requires HTTPS)

PR #1025, opened 2024-08-08
PB-871 : add debug UI and URL params to override URLs used by data integration team - #minor

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-871-debug-tools-for-url/index.html)


ltshb (commented):
I like the idea of the debug modal to set the url backend.

However having the override check every where in code were we need the backend doesn't seem a good approach.

I would consider having a new approach, now the backend urls are hardcoded as constant in a javascript file, we should have a library that returns the backend urls for each backend, so instead of having `API_BASE_URL` we would have a function `getApiBaseUrl()` or even better having a singleton class `Backends` from which we could as for any backend base url. Then this class would be responsible to override any url if needed. In future we could also consider that the backend urls are not anymore hardcoded in the config but would be asked from the backend at map.geo.admin/api/backends

ltclm:
great, thank you @pakb 
one small question:
when setting the api_url to `http://localhost:9176` as here:

https://sys-map.int.bgdi.ch/preview/feat-pb-871-debug-tools-for-url/index.html#/map?lang=de&center=2609348.52,1236799.55&z=4.585&bgLayer=ch.swisstopo.pixelkarte-farbe&topic=ech&swisssearch=SWISSIMAGE+Zeitreise&layers=ch.swisstopo.swissimage-product@year=2017;ch.swisstopo.swissimage-product.metadata@year=2017@features=206028&featureInfo=default&timeSlider=2017&api_url=http://localhost:9176/

the identify request is done with the localhost url but the htmlpopup is still done by sys-api3.int.bgdi.ch.

rebert (approved):
This helps us a lot for the DI. Beside of the feedback from @ltclm about the popup, it seems to work fine. I'm looking forward to get it merged

src/utils/layerUtils.js
```diff
@@ -76,7 +78,7 @@ export function getWmtsXyzUrl(wmtsLayerConfig, projection, options = {}) {
         const layerId = wmtsLayerConfig.isExternal
             ? wmtsLayerConfig.id
             : wmtsLayerConfig.technicalName
-        return `${wmtsLayerConfig.baseUrl}1.0.0/${layerId}/default/${timestamp}/${projection.epsgNumber}/{z}/{x}/{y}.${wmtsLayerConfig.format}`
+        return `${baseUrlOverride ?? wmtsLayerConfig.baseUrl}1.0.0/${layerId}/default/${timestamp}/${projection.epsgNumber}/{z}/{x}/{y}.${wmtsLayerConfig.format}`
```
ltshb, line 81: Why using baseUrlOverride here ? it should be `getWmtsBaseUrl`, a reason to not export the baseUrlOverride to avoid such bug.

ltshb (commented):
I like your idea to split the config into several files 👍🏼

src/utils/layerUtils.js
pakb: I might have forgotten to change this one, should be good now

src/modules/map/components/openlayers/OpenLayersWMSLayer.vue
```diff
@@ -44,7 +46,7 @@ const wmsVersion = computed(() => wmsLayerConfig.value.wmsVersion || '1.3.0')
 const format = computed(() => wmsLayerConfig.value.format || 'png')
 const gutter = computed(() => wmsLayerConfig.value.gutter || -1)
 const opacity = computed(() => parentLayerOpacity.value ?? wmsLayerConfig.value.opacity)
-const url = computed(() => wmsLayerConfig.value.baseUrl)
+const url = computed(() => getBaseUrlOverride('wms') ?? wmsLayerConfig.value.baseUrl)
```
ltshb, line 49: Why not calling directly getWmsBaseUrl here ?

ltshb (approved):
Just change the one from my last comment and you good to merge

PR #1026, opened 2024-08-09
PB-830 : add specific (S)CSS for Diemo popup - #patch

pakb:
so that electric car charging location can be styled like they were on mf-geoadmin3

[Test link with diemo layer](https://sys-map.int.bgdi.ch/preview/fix-pb-830-diemo-popup/index.html?layers=ch.bfe.ladestellen-elektromobilitaet)

[Test link](https://sys-map.int.bgdi.ch/preview/fix-pb-830-diemo-popup/index.html)


ltkum (approved):
Looks good to me, and with the correct colors :)

PR #1031, opened 2024-08-19
PB 773 clickable droplet search - #minor

ismailsunni:
New screenshot:
image

[Test link](https://sys-map.int.bgdi.ch/preview/pb-773-clickable-droplet-search/index.html)


ltshb (changes requested):
We need to differentiate 2 different use case here:
1. Search result is a place (coordinate or address)
2. Search result is an object (enable the `RBD: building status` layer and type something like `bern`)

In the first use case I would reuse the location popup as the user might be interested at any coordinate or might be want to share his address/place to someone using for example whatsapp.

In the second use case I see that we already have a different behavior from the old viewer, where on the old viewer it selected the object and showed its content, it did not add a droplet. In that case we should have the same behavior as on the old viewer simply select the object (similar as if the user clicked the object).

Moreover this PR is breaking the infobox behavior to toggle between infobox and popup.

ismailsunni:
So, in short:
First use case: show the location pop up
Second use case: remove the droplet and select the object (then automatically show the info box for the object)

Regarding the toggle behaviour, there is a ticket to remove the behaviour, and use can show both location popup and infobox (PB-859).

ismailsunni:
hi @ltshb 
I have updated the approach. Now, when you click the pin, it will show the pop up location (as shown in the screenshot above).

I will implement the second use case in another PR, since it's quite different in term of area of code, it's ok.

Thanks

ltshb (approved):

PR #1039, opened 2024-08-27
PB-773: Auto select the feature from search result.

ismailsunni:
Peek 2024-08-27 10-11 - autoselect feature

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-773-auto-select-feature-from-search/index.html)

### DIFF
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -85,6 +85,7 @@ export function sanitizeTitle(title = '') {
  * @extends LayerSearchResult
  * @extends LocationSearchResult
  * @typedef {Object} LayerFeatureSearchResult
+ * @param {GeoAdminLayer} layer The layer of this feature.
  */
 
 /**
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -258,6 +258,7 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
                     ...locationContent,
                     resultType: SearchResultTypes.FEATURE,
                     title,
+                    layer,
                 }
             }) ?? []
         )
--- a/src/modules/map/components/openlayers/utils/useVectorLayer.composable.js
+++ b/src/modules/map/components/openlayers/utils/useVectorLayer.composable.js
@@ -13,19 +13,21 @@ import { randomIntBetween } from '@/utils/numberUtils'
  * Caveat: it will not update the style!
  *
  * @param {Map} map
- * @param {Readonly<Ref<Number>>} zIndex
  * @param {Readonly<Ref<Feature[]>>} features
- * @param {Function} styleFunction
- * @param {Function} onFeatureSelectCallback
+ * @param {Readonly<Ref<Number>>} [config.zIndex] The Z index of the layer
+ * @param {Function} [config.styleFunction] A function that defines the style for the vector layer.
+ * @param {Function} [config.onFeatureSelectCallback] A callback function that is called when a
+ *   feature is selected.
+ * @param {boolean} [config.deselectAfterSelect] If true, the selected feature will be deselected
+ *   after the select callback is called
  */
-export default function useVectorLayer(
-    map,
-    features,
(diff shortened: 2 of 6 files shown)


ltshb:
@ismailsunni could you update the PR to also have the feature to select droplet from your last PR.

ismailsunni:
@pakb done, also updated with develop

ltshb (commented):
There is an issue with the droplet from #1031, you can first select the droplet (when it is a location, address or coordinate) but if you move the map you cannot click it again

ismailsunni:
@ltshb I fix the issue, please check and see it's working for you. Thanks!

src/store/modules/search.store.js
```diff
@@ -152,6 +153,44 @@ const actions = {
...
+                    dispatch('setZoom', {
+                        zoom: entry.zoom,
+                        dispatcher: dipsatcherSelectResultEntry,
+                    })
+                }
+
+                // Automatically select the feature
+                try {
+                    getFeature(
+                        getters.getActiveLayersById(entry.layerId)[0],
```
pakb, line 174: might be a good idea to add the layer object to the search result instead. It might be available when the search request is fired in `search.api.js`, so adding it to the typedef `LayerSearchResult` is preferable

src/modules/map/components/openlayers/utils/useVectorLayer.composable.js
```diff
@@ -23,7 +23,8 @@ export default function useVectorLayer(
     features,
     zIndex = -1,
     styleFunction = highlightFeatureStyle,
-    onFeatureSelectCallback = () => {}
+    onFeatureSelectCallback = () => {},
+    deselectAfterSelect = false // If true, the selected feature will be deselected after the callback is called
```
pakb, line 27: could you change the way params are given and use a config object approach?
this starts to have too many arguments to be able to know which is what

src/api/search.api.js
```diff
@@ -258,6 +258,7 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
                     ...locationContent,
                     resultType: SearchResultTypes.FEATURE,
                     title,
+                    layer,
```
pakb, line 261: don't forget to document the JSDoc for the `LayerFeatureSearchResult` typedef with this new addition 😉

ismailsunni: ah yes, of course!

src/store/modules/search.store.js
```diff
@@ -152,6 +153,39 @@ const actions = {
...
+                    dispatch('setCenter', {
+                        center: entry.coordinate,
+                        dispatcher: dipsatcherSelectResultEntry,
+                    })
+                    dispatch('setZoom', {
+                        zoom: entry.zoom,
+                        dispatcher: dipsatcherSelectResultEntry,
+                    })
+                }
+
```
ltkum, line 170: this is the exact same code as the first part of the location code, which makes sense. Perhaps we could add a small 'zoomToEntry' function here, to avoid code repetition.

ltkum (commented):
I would prefer if we avoided the small repetition, but it's not that big of a deal. Otherwise it seems to work well, thanks. I think we also need to be able to select the locations, but unless I am mistaken, this is in another PR.

ismailsunni:
@ltkum what do you mean about location? If we select a location, what should we select? Because there is no object to select. Also based on this comment from Brice https://github.com/geoadmin/web-mapviewer/pull/1031#pullrequestreview-2245304064

ltkum:
> @ltkum what do you mean about location? If we select a location, what should we select? Because there is no object to select. Also based on this comment from Brice [#1031 (review)](https://github.com/geoadmin/web-mapviewer/pull/1031#pullrequestreview-2245304064)

The ticket mention that before, it was possible to click the marker that appeared on a location after searching, but as I said, this part was handled in another ticket :+1:

PR #1044, opened 2024-08-28
PB-843 : use LV03 to LV95 reframe services

pakb:
instead of reprojecting through proj4, which led to poor precision

[Test link with legacy E and N param](https://sys-map.dev.bgdi.ch/preview/fix-pb-843-lv03-reframe/index.html?E=559224&N=101520&crosshair=marker&z=13)
[Same link on PROD](https://map.geo.admin.ch/?E=559224&N=101520&crosshair=marker&z=13)

The difference is about a meter, no more. But for some users, that was a big deal

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-843-lv03-reframe/index.html)

### DIFF
--- a/src/store/modules/search.store.js
+++ b/src/store/modules/search.store.js
@@ -1,3 +1,5 @@
+import proj4 from 'proj4'
+
 import search, { SearchResultTypes } from '@/api/search.api'
 import { isWhat3WordsString, retrieveWhat3WordsLocation } from '@/api/what3words.api'
 import coordinateFromString from '@/utils/coordinates/coordinateExtractors'
@@ -38,9 +40,9 @@ const actions = {
         if (query.length >= 2) {
             const currentProjection = rootState.position.projection
             // checking first if this corresponds to a set of coordinates (or a what3words)
-            const coordinates = coordinateFromString(query, currentProjection)
+            const extractedCoordinate = await coordinateFromString(query)
             let what3wordLocation = null
-            if (!coordinates && isWhat3WordsString(query)) {
+            if (!extractedCoordinate && isWhat3WordsString(query)) {
                 try {
                     what3wordLocation = await retrieveWhat3WordsLocation(query, currentProjection)
                 } catch (error) {
@@ -51,7 +53,15 @@ const actions = {
                 }
             }
 
-            if (coordinates) {
+            if (extractedCoordinate) {
+                let coordinates = [...extractedCoordinate.coordinate]
+                if (extractedCoordinate.coordinateSystem !== currentProjection) {
+                    coordinates = proj4(
+                        extractedCoordinate.coordinateSystem.epsg,
+                        currentProjection.epsg,
+                        coordinates
+                    ).map(currentProjection.roundCoordinateValue)
+                }
                 const dispatcherCoordinate = `${dispatcher}/search.store/setSearchQuery/coordinate`
                 dispatch('setCenter', {
                     center: coordinates,
--- a/src/store/modules/search.store.js
+++ b/src/store/modules/search.store.js
@@ -1,11 +1,10 @@
-import proj4 from 'proj4'
-
 import reframe from '@/api/lv03Reframe.api'
 import search, { SearchResultTypes } from '@/api/search.api'
 import { isWhat3WordsString, retrieveWhat3WordsLocation } from '@/api/what3words.api'
 import coordinateFromString from '@/utils/coordinates/coordinateExtractors'
 import { STANDARD_ZOOM_LEVEL_1_25000_MAP } from '@/utils/coordinates/CoordinateSystem.class'
-import { LV03, LV95 } from '@/utils/coordinates/coordinateSystems'
+import { LV03 } from '@/utils/coordinates/coordinateSystems'
+import { reprojectAndRound } from '@/utils/coordinates/coordinateUtils'
 import CustomCoordinateSystem from '@/utils/coordinates/CustomCoordinateSystem.class'
 import log from '@/utils/logging'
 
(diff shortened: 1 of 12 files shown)


src/utils/coordinates/coordinateExtractors.js
```diff
@@ -240,33 +322,50 @@ const executeAndReturn = (regex, text, outputProjection, extractor = numericalEx
...
+        }
+    }
+    const lv03result = await extractLV03Coordinates(text)
+    if (lv03result) {
+        return {
+            // coordinates have been reframed to LV95
+            coordinateSystem: LV95,
+            coordinate: lv03result,
+        }
+    }
```
ltshb, line 353: I would do the extract without reframed and after the extract do the reframe, it would make the code easier to read understand, having a `lv03result` with LV95 projection sounds wrong.

src/utils/coordinates/coordinateExtractors.js
```diff
@@ -240,33 +322,50 @@ const executeAndReturn = (regex, text, outputProjection, extractor = numericalEx
...
-
-        { regex: REGEX_METRIC_COORDINATES, extractor: numericalExtractor },
-        { regex: REGEX_MILITARY_GRID, extractor: mgrsExtractor },
-    ]
-        .map((config) => {
-            // going through each config and extracting the result (can be undefined if not a match)
-            return executeAndReturn(config.regex, text, toProjection, config.extractor)
-        })
-        .find((result) => Array.isArray(result)) // returning the first value that is a coordinate array (will return undefined if nothing is found)
+    const wgs84result = extractWGS84Coordinates(text)
```
ltshb, line 332: Pay attention to camelcase nomenclature, should be `wgs84Result` same for other results below

src/utils/coordinates/coordinateExtractors.js
ltshb:
```javascript
const lv03Result = extractLV03Coordinates(text)
if (lv03Result) {
  const lv95Result = await reframe(..)
  return {...}
}
```

src/utils/coordinates/coordinateExtractors.js
```diff
@@ -75,28 +73,119 @@ const REGEX_WGS_84_WITH_SECONDS_PREFIXED = new RegExp(
...
+            })
+        } else if (LV03.isInBounds(coordinates[1], coordinates[0])) {
+            return reframe({
+                inputCoordinates: [coordinates[1], coordinates[0]],
+                inputProjection: LV03,
+            })
+        }
+    }
+    return undefined
+}
```
ltshb, line 138: See other comment I would not do the reframe here and keep the function as the other projection, doing only one thing, extracting. This would make the code easier to understand/read.

ltshb: The return value should be in correct order x,y like for lv95 this check can be done here to normalize the coordinate axis order.

src/store/modules/search.store.js
```diff
@@ -51,7 +55,27 @@ const actions = {
...
+                            inputCoordinates: coordinates,
+                        })
+                        extractedCoordinate.coordinateSystem = LV95
+                    }
+
+                    coordinates = proj4(
+                        extractedCoordinate.coordinateSystem.epsg,
+                        currentProjection.epsg,
+                        coordinates
+                    ).map(currentProjection.roundCoordinateValue)
```
ltshb, line 77: Not sure that overwriting the `extractedCoordinate.coordinateSystem = LV95` is a good idea as well as passing the reframe lv03 to lv95 to proj4, make the code a bit hard to understand and error prone. Would be better to have an else statement for the proj4 projection and not overwrite the extractedCoordinate.

src/store/modules/search.store.js
```diff
@@ -51,7 +55,29 @@ const actions = {
...
+                    // So we pass through a LV95 reframe (done by a backend service that knows all deformations between the two)
+                    // and then go to the wanted coordinate system
+                    if (extractedCoordinate.coordinateSystem === LV03) {
+                        coordinates = await reframe({
+                            inputProjection: LV03,
+                            inputCoordinates: coordinates,
+                        })
+                        coordinates = proj4(LV95.epsg, currentProjection.epsg, coordinates).map(
+                            currentProjection.roundCoordinateValue
+                        )
```
ltshb, line 72: ? I don't get it why using reframe and proj4 ?

pakb: because if we are using something else than LV95, reframe only returns LV95 coordinates (say when we will be using mercator as the main projection)

ltshb: If I understood correctly the goal of using reframe was for user that used lv03 to lv95 and not lv03 to other projection ? Am I correct. Because converting from lv03 to lv95 to wgs84 or any other will anyway add some imprecision ?
Would it be not simpler and explicite to have something
```javascript
if (extractedCoordinate.coordinateSystem === LV03 && currentProjection.epsg == LV95) {
   coordinates = await reframe(...)
} else {
   coordinates = proj4(...)
}
```

Or at least in you current solution only call proj4 if the current projection is not LV95.

pakb: the reframe backend service only deals with coordinates in LV03 or LV95 (output too) so proj4 is still required if we are aiming for anything else.
What requires special care (the dedicated backend) is the transformation between LV03 and LV95, as it's not a linear change

pakb: calling proj4 on the same in and out will just returns the coordinates, that won't use any calculation (that's why I let it go through without checking)

ltshb (approved):

PR #1045, opened 2024-08-29
PB 898 Fix rumantsch footer link to geo.admin.ch - #patch

schtibe:
There's no RM version of the page, so we fall back to DE for that link

[Test link](https://sys-map.int.bgdi.ch/preview/bug-pb-898-fix-rumantsch-footer-link/index.html)

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-898-fix-rumantsch-footer-link/index.html)

### DIFF
--- a/src/modules/i18n/locales/rm.json
+++ b/src/modules/i18n/locales/rm.json
@@ -155,7 +155,7 @@
     "drop_me_here": "Trair vi la datoteca KML nà qua",
     "east": "ost",
     "ech": "Catalog da geodatas",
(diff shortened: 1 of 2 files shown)


pakb (commented):
I'm wondering if we shouldn't move that link into our translation file instead of managing i18n stuff for specific language here

schtibe:
I had the same thought. Then again I wondered if this is ever going to change, and if not, if it's spamming the translation file. I don't know, I trust your experience!

pakb (approved):
👍

PR #1046, opened 2024-08-30
PB-871 : only dispatch if something has changed

pakb:
it was dispatching any time any URL param changed, leading to the part of the code reloading the topic/layers config being called way too often

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-871-too-many-refresh/index.html)


ltshb (approved):

pakb:
There's still a bug in 3D I'd like to fix within this PR directly.

PR #1047, opened 2024-08-30
PB-856 : add a style selector for KML in layer settings

pakb:
this enables users to opt-out of the default style (Google Earth like) we apply by default to external KML, and let them go back to the GeoAdmin (all red) style that was previously (on mf-geoadmin3) applied across the board.

adding this choice to the layer URL extra attributes

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-856-add-kml-style-selector/index.html)

### DIFF
--- a/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -6,12 +6,15 @@
 
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import { computed, onMounted, ref, toRefs } from 'vue'
+import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
 import AbstractLayer from '@/api/layers/AbstractLayer.class'
 import GPXLayer from '@/api/layers/GPXLayer.class.js'
 import KMLLayer from '@/api/layers/KMLLayer.class.js'
+import { allKmlStyles } from '@/api/layers/KmlStyles.enum'
 import MenuActiveLayersListItemTimeSelector from '@/modules/menu/components/activeLayers/MenuActiveLayersListItemTimeSelector.vue'
+import DropdownButton, { DropdownItem } from '@/utils/components/DropdownButton.vue'
 import ErrorButton from '@/utils/components/ErrorButton.vue'
 import TextTruncate from '@/utils/components/TextTruncate.vue'
 import ThirdPartyDisclaimer from '@/utils/components/ThirdPartyDisclaimer.vue'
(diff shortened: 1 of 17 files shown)


src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -94,6 +97,16 @@ onMounted(() => {
...
 
+watch(currentKmlStyle, (newStyle) => {
+    const updatedLayer = layer.value.clone()
+    updatedLayer.style = newStyle
+    store.dispatch('updateLayer', {
+        index: index.value,
+        layer: updatedLayer,
+        ...dispatcher,
+    })
+})
```
ltshb, line 108: You don't need to do a clone here, you could simply do

```javascript
store.dispatch('updateLayer', {
  index: index.value,
  layer: {
     style: newStyle,
  },
  ...dispatcher,
})
```

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -213,67 +226,75 @@ function duplicateLayer() {
...
+            </div>
+            <div v-if="isLayerKml" v-show="showLayerDetail" class="menu-layer-item-details mt-n2">
+                <label class="menu-layer-transparency-title text-nowrap me-2">
+                    {{ t('vector_feedback_select_style') }}
+                </label>
+                <select v-model="currentKmlStyle" class="form-select form-select-sm">
+                    <option v-for="style in allKmlStyles" :key="style" :value="style">
+                        {{ t(`kml_style_${style.toLowerCase()}`) }}
+                    </option>
+                </select>
```
ltshb, line 296: We should be careful with the drop down box, we currently have 3 different types:

- select type like here and a few other places
- Home made pure bootstrap like in import tool and search bar
- tippy like in drawing menu style size selectors

Perso I don't like the look and feel of the select type as the dropdown box is squared and don't match the rest of the styling. I also find the tippy not so nice, because the dropdown placement is not always at the bottom.

Can we not make the <select> drop down look like the rest of the box, rounded corner with a grey border instead of black ? Or should we implement our own select that match the style of the search bar and import map tools ?

ltshb: Otherwise some top margin to the new line in the settings would be nicer.

pakb: I've switch to the DropdownButton component, might be better for consistency yes...

ltshb (commented):
When opening the drawing menu and clicking on add icon, the icon is missing and you have the following type error 
image

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
ltshb: Well now in the menu we have 3 types, the DropdownButton component is only used in drawing currently.
Might suggestion for this PR would be to also change the downdown in the print menu so we have in the menu only two styles the tippy dropdown and the import tool home made advanced dropdown with button group.
Then in a separate PR we could simply change one component `DropdownButton` to match the same style as the sear bar and import map tools and we have one single style over the whole app 😄

pakb: https://jira.swisstopo.ch/browse/PB-1048
I've created a task to do that later on, I don't want to spend more time on this right now (quite enough on my table)

ismailsunni:
I tried to load `Rossier.kml` from the JIRA ticket, but it doesn't allow me anymore. It works in production
image

ismailsunni (approved):
LGTM, the issue I found aboe is fixed also

PR #1052, opened 2024-09-02
PB-912 Fix language of background layer attribution

schtibe:
(the reason I am doing this was because I fixed something small when nobody else was around, this is like the extension of that)
Some notes on the fix: the backgroundLayer was copied inside the store. When one changes the language, the attributions would therefore not be updated. One solution would be to listen to language changes and trigger the update (see first commit), but IMO this would diffuse the responsibility of the data further. So I changed the store a bit so the currentLayer comes through a getter, streamlining the data flow.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-912-swisstopo-source/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -33,11 +33,11 @@ const cloneActiveLayerConfig = (getters, layer) => {
 
 const state = {
     /**
-     * Current background layer
+     * Current background layer ID
      *
-     * @type AbstractLayer
+     * @type string
      */
-    currentBackgroundLayer: null,
+    currentBackgroundLayerId: null,
     /**
      * Currently active layers (that have been selected by the user from the search bar or the layer
      * tree)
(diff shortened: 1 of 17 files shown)


src/store/modules/layers.store.js
```diff
@@ -220,22 +228,16 @@ const actions = {
...
      */
-    setBackground({ commit, getters }, { bgLayer, dispatcher }) {
-        const layerIdOrObject = bgLayer
-        let futureBackground
-        if (typeof layerIdOrObject === 'string') {
-            futureBackground = getters.getLayerConfigById(layerIdOrObject)
-        } else if (layerIdOrObject instanceof AbstractLayer) {
-            futureBackground = getters.getLayerConfigById(layerIdOrObject.id)
+    setBackground({ commit, getters }, { bgLayerId }) {
+        if (bgLayerId == null) {
```
pakb, line 235: triple `===` please 😉

pakb (approved):

src/store/modules/layers.store.js
schtibe: Yes sir! 😉
  Can't we enforce this with eslint?

schtibe: Yes sir 😉
Can't we enforce this with eslint?

pakb: we could maybe... https://eslint.org/docs/latest/rules/eqeqeq
 there are rarely cases where we want to double equals, but it would make sense to then disable the rule

schtibe: As discussed on Slack, I will maybe check this with another branch, as 1) adding the rule would probably require quite many adjustments throughout the code base and 2) the rule should actually already be enabled by default

PR #1053, opened 2024-09-02
PB-859: Use the same approach to compute polygon area.

ismailsunni:
image

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-869-different-area-calculation/index.html)

### DIFF
--- a/src/modules/infobox/components/FeatureAreaInfo.vue
+++ b/src/modules/infobox/components/FeatureAreaInfo.vue
@@ -1,11 +1,11 @@
 <script setup>
-import area from '@turf/area'
+import { Geodesic, PolygonArea } from 'geographiclib-geodesic'
 import { computed, toRefs } from 'vue'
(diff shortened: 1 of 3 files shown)


ltshb (approved):
Code looks good but the PR and code should be better documented, eg. why using the GeographicLib ? From where come this library etc...

ismailsunni:
@ltshb I don't know either why or when we decide to use GeographicLib in the first time. In this PR I just simply make the area calculated the same way.

pakb (commented):
Might be good to reference that you are copying here what you found out [here](https://github.com/geoadmin/web-mapviewer/blob/c2dc9cc7068b4ed878234582a59e066c039e9785/src/utils/geodesicManager.js#L155-L172)

or even better, have both location use the same piece of code

pakb (approved):

PR #1055, opened 2024-09-02
PB-878: Filter out KML outside print extent

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-878-filter-out-kml/index.html)

Tested with this file: 

[super-big.zip](https://github.com/user-attachments/files/16864204/super-big.zip)

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-878-filter-out-kml/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -15,6 +15,7 @@ import {
     getWmsBaseUrl,
 } from '@/config/baseUrl.config'
 import i18n from '@/modules/i18n'
+import store from '@/store'
 import log from '@/utils/logging'
 import { adjustWidth } from '@/utils/styleUtils'
 
@@ -25,11 +26,12 @@ const MAX_PRINT_SPEC_SIZE = 1 * 1024 * 1024 // 1MB in bytes (should be in sync w
 
 class GeoAdminCustomizer extends BaseCustomizer {
     /** @param {string[]} layerIDsToExclude List of layer names to exclude from the print */
-    constructor(layerIDsToExclude, printResolution) {
-        super()
+    constructor(printExtent, layerIDsToExclude, printResolution) {
+        super(printExtent)
         this.layerIDsToExclude = layerIDsToExclude
         this.printResolution = printResolution
         this.layerFilter = this.layerFilter.bind(this)
+        this.geometryFilter = this.geometryFilter.bind(this)
         this.line = this.line.bind(this)
         this.text = this.text.bind(this)
         this.point = this.point.bind(this)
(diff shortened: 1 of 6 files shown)


src/api/print.api.js
```diff
@@ -25,11 +26,12 @@ const MAX_PRINT_SPEC_SIZE = 1 * 1024 * 1024 // 1MB in bytes (should be in sync w
 
 class GeoAdminCustomizer extends BaseCustomizer {
     /** @param {string[]} layerIDsToExclude List of layer names to exclude from the print */
-    constructor(layerIDsToExclude, printResolution) {
-        super()
+    constructor(printExtent, layerIDsToExclude, printResolution) {
```
pakb, line 29: could you add the two missing params in the JSDoc above (with types and all)?

src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
```diff
@@ -118,6 +118,21 @@ export default function usePrintAreaRenderer(map) {
...
+
+        store.commit('setPrintExtent', {
+            printExtent: [
+                topLeftCoordinate[0],
+                rightBottomCoordinate[1],
+                rightBottomCoordinate[0],
+                topLeftCoordinate[1],
+            ],
+            dispatcher,
+        })
```
pakb, line 135: I don't know if this would help, but there are a couple extent utils in the file `src/utils/coordinateUtils.js` that might be of use here.
There's a function to flatten an extent that is expressed as `[ [x0, y0], [x1, y1] ]` and the opposite to make it a double coordinate array

If you want to use them, you could move these two function in the newly created `src/utils/extentUtils.js` file instead

ismailsunni: It's different I guess, since this code compute the coordinate-based extent from the pixel-based extent. The order is in different direction (e.g. pixel is from top-left while the coordinate is from left-bottom).

If it's ok, I will keep it in this location because it's very specific. I have added a comment to explain the order.

tests/cypress/tests-e2e/print.cy.js
```diff
@@ -294,8 +304,101 @@ describe('Testing print', () => {
...
+            const localKmlFile = `tests/cypress/fixtures/import-tool/${localKmlFileName}`
+
+            // Test local import
+            cy.log('Switch to local import')
+            cy.get('[data-cy="import-file-local-btn"]:visible').click()
+            cy.get('[data-cy="import-file-local-content"]').should('be.visible')
+
+            // Attach a local KML file
+            cy.log('Test add a local KML file')
+            cy.readFile(localKmlFile, null).as('kmlFixture')
```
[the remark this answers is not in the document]
ismailsunni: ok, noted. I did it actually just for testing if it might fix the CI. But it seems no. I have reverted it back to use `fixture`

tests/cypress/tests-e2e/print.cy.js
```diff
@@ -294,8 +304,101 @@ describe('Testing print', () => {
                 expect(layers[0]['matrices'][0]['matrixSize']).to.deep.eq([1, 1])
             })
         })
-        it('should send a print request correctly to mapfishprint (with KML layer)', () => {
-            startPrintWithKml('import-tool/external-kml-file.kml')
+        it.skip('should send a print request correctly to mapfishprint (with KML layer)', () => {
```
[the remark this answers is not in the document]
ismailsunni: I was only used for testing if it's the one which cause the CI error. But it happens randomly.

PR #1058, opened 2024-09-05
PB-66: Add new babs icons to mapviewer

LukasJoss:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-66-add-new-babs-icons-to-mapviewer/index.html)


pakb (commented):
couple questions about the whole thing :

- why have this description URL? We won't be using it, as i18n keys are then also given in the list of icons endpoint.
- I'm wondering if we should "tag" the sets with a `lang` ISO code value, so that we may have a lang selector alongside the choice of "Civil icons", instead of having 3 times the civil icons in the list
- When showing description on icons that aren't available in the current app lang, I would show them all (prefixed with ISO code)
- I would maybe add the description of the icon to the description of the marker, would make sense to know what kind of icon it is also outside of the drawing module

LukasJoss:
> couple questions about the whole thing :
> 
>     * why have this description URL? We won't be using it, as i18n keys are then also given in the list of icons endpoint.
> 
Should the backend for the set instead have a boolean, e.g. has_description?
>     * I'm wondering if we should "tag" the sets with a `lang` ISO code value, so that we may have a lang selector alongside the choice of "Civil icons", instead of having 3 times the civil icons in the list
> 
I assume this should also be provided by the back-end?
>     * When showing description on icons that aren't available in the current app lang, I would show them all (prefixed with ISO code)
> 
What do you mean with "icons that aren't available in the current app lang"? All icons should be available in all languages
>     * I would maybe add the description of the icon to the description of the marker, would make sense to know what kind of icon it is also outside of the drawing module

hansmannj:
There's currently strange behavior using the test link (which might be ourdated?):
image

hansmannj:
Not sure if that is expected behavior: The mouseover text for single icons only appears, if the icon grid is expanded. 
When only the first row is slightly visible, hovering over the icons does not fire the tooltip with the translations.

pakb (approved):

PR #1059, opened 2024-09-06
PB-303: use icon size from backend

LukasJoss:


src/api/icon.api.js
```diff
@@ -4,15 +4,7 @@ import { getViewerDedicatedServicesBaseUrl } from '@/config/baseUrl.config'
...
- * NOTE: The size should match the received size for the scale from the backend. It is needed to
- * avoid race condition when exporting/saving KML. Openlayer requires the size to compute the
- * scale.
- *
- * TODO: take the default size from the backend icon API
- */
+/** Default Icon parameters for the URL. */
 export const DEFAULT_ICON_URL_PARAMS = {
     scale: 1,
     size: [48, 48],
```
ltshb, line 10: I would remove the default icon size from this structure and not export it. This default size is only used by `DEFAULT_MARKER_TITLE_OFFSET` in this file and should not be used outside of this file (not exported)

PR #1060, opened 2024-09-09
PB-900: Add no shared icon and text

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-900-local-kml-cannot-share/index.html)


src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -295,7 +299,23 @@ watch(
...
+                    class="d-flex flex-row gap-2 justify-content-center mt-2"
+                    data-cy="warn-share-local-file-container"
+                >
+                    <FontAwesomeIcon
+                        class="data-disclaimer-tooltip text-warning align-self-center"
+                        icon="circle-exclamation"
+                        data-cy="menu-external-disclaimer-icon-cloud"
+                    />
+                    <div
+                        class="small text-wrap text-center text-warning"
```
pakb, line 313: I would keep the text the standard color, only have the icon in `text-warning` (we do not have other places where we color text yellow yet, I don't find that it looks particularly good contrast-wise)

src/store/modules/layers.store.js
```diff
@@ -213,6 +213,42 @@ const getters = {
...
+    /**
+     * Returns true if the layer comes from a third party (external layer or KML layer) which has
+     * been imported from a local file.
+     *
+     * KML layer are treated as external when they are generated by another user (no adminId).
+     *
+     * @param {string} layerId Layer ID of the layer to check for data disclaimer
+     * @returns {Boolean}
+     */
+    hasDataDisclaimerLocal: (state) => (layerId) => {
```
pakb, line 226: I would maybe rename this function. Maybe simply something like
`isLocalFile`

pakb: Might also make sense to receive directly the layer object so you don't have to do the lookup here (both usage of this function are aware of the full object)

src/store/modules/layers.store.js
```diff
@@ -213,6 +213,42 @@ const getters = {
...
+     */
+    hasAnyExternalDataLocal: (state) => () => {
+        return state.activeLayers.some((layer) => {
+            const isBaseUrlValidUrl = /^\w+:\/\//.test(layer?.baseUrl)
+            return (
+                !isBaseUrlValidUrl &&
+                (layer.isExternal || (layer?.type === LayerTypes.KML && !layer?.adminId))
+            )
+        })
+    },
```
pakb, line 251: I would also rename this one accordingly.
You should re-use the function above. You can get access to getters by adding them to the signature of the overall function
```Javascript
hasAnyLocaFile: (state, getters) => {
    return state.activeLayers.some((layer) => getters.isLocalFile(layer))
})
```

src/utils/components/ThirdPartyDisclaimer.vue
```diff
@@ -39,6 +39,10 @@ export default {
             type: String,
             default: '',
         },
+        isExternalDataLocal: {
```
pakb, line 42: `isLocalFile` would be better IMHO

tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -213,9 +222,10 @@ describe('The Import File Tool', () => {
         cy.log('Test the external layer disclaimer in the footer')
         cy.closeMenuIfMobile()
         cy.get('[data-cy="layer-copyright-example.com"]')
+            .eq(0)
```
pakb, line 225: why using this `eq(0)` here?

tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -213,9 +222,10 @@ describe('The Import File Tool', () => {
...
             .should('be.visible')
             .should('have.class', 'text-primary')
-        cy.get('[data-cy="layer-copyright-example.com"]').realHover()
+        cy.get('[data-cy="layer-copyright-example.com"]').eq(0).realHover()
```
pakb, line 228: same question

src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -193,12 +193,19 @@ function duplicateLayer() {
...
                 :complete-disclaimer-on-click="true"
                 :source-name="attributionName"
+                :is-external-data-local="hasDataDisclaimerLocal"
             >
                 <FontAwesomeIcon
-                    ref="tooltipAnchor"
-                    class="data-disclaimer-tooltip text-primary p-2"
-                    icon="user"
-                    data-cy="menu-external-disclaimer-icon"
+                    v-if="hasDataDisclaimer && !hasDataDisclaimerLocal"
```
pakb, line 199: no need to test for `hasDataDisclaimer` as the parent is only added when it's `true`

you could then test for the validity of `hasDataDisclaimerLocal` and add a `v-else` to the other icon

tests/cypress/tests-e2e/shareShortLink.cy.js
```diff
@@ -225,4 +225,64 @@ describe('Testing the share menu', () => {
             })
         }
     )
+    context(
```
pakb, line 228: You are doing way to much testing in the `beforeEach` hook, I think it makes more sense to transform the whole thing into a single test case

tests/cypress/tests-e2e/importToolFile.cy.js
sommerfe: i removed it

tests/cypress/tests-e2e/importToolFile.cy.js
sommerfe: also removed

src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -295,7 +299,23 @@ watch(
...
+                        icon="circle-exclamation"
+                        data-cy="menu-external-disclaimer-icon-cloud"
+                    />
+                    <div
+                        class="small text-wrap text-center"
+                        data-cy="warn-share-local-file"
+                        v-html="t('warn_share_local_file')"
+                    ></div>
+                </div>
+                <!-- eslint-enable vue/no-v-html -->
```
pakb, line 318: Might be better to move this eslint disabling rule to be set just after the line where `v-html` is used, here it is then disabled for part of the code where all rules should be applied.

src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -6,6 +6,7 @@
                 v-if="source.hasDataDisclaimer"
                 :source-name="source.name"
                 :complete-disclaimer-on-click="!source.url"
+                :is-external-data-local="source.isExternalDataLocal"
```
pakb, line 9: you should also rename your prop declaration to `:is-local-file` (in ThirdPartyDisclaimer the props is now called `isLocalFile`)

src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -62,6 +63,7 @@ export default {
                                 name: attribution.name,
                                 url: attribution.url,
                                 hasDataDisclaimer: this.hasDataDisclaimer(layer.id),
+                                isExternalDataLocal: this.isLocalFile(layer),
```
pakb, line 66: then it would also make sense to match the new name of the props in ThirdPartyDisclaimer and have the source property named the same

src/utils/components/ThirdPartyDisclaimer.vue
```diff
@@ -50,7 +54,9 @@ export default {
             currentLang: (state) => state.i18n.lang,
         }),
         tooltipContent() {
-            return this.$t('external_data_tooltip')
+            return this.isExternalDataLocal
```
pakb, line 57: I don't think a `this.isExternalDataLocal` still exists, as you've renamed the prop to `isLocalFile`

src/utils/components/ThirdPartyDisclaimer.vue
```diff
@@ -61,14 +67,19 @@ export default {
     mounted() {
         if (this.showTippy) {
             this.tippyInstance = tippy(this.$refs.tippyAnchor, {
-                theme: 'primary',
+                theme: this.isExternalDataLocal ? 'secondary' : 'primary',
```
pakb, line 70: same

src/utils/components/ThirdPartyDisclaimer.vue
```diff
@@ -61,14 +67,19 @@ export default {
...
                 content: this.tooltipContent,
                 arrow: true,
                 placement: 'top',
                 touch: false,
                 delay: 250,
                 onCreate: (instance) => {
-                    instance.popper.setAttribute('data-cy', `tippy-third-part-disclaimer`)
+                    instance.popper.setAttribute(
+                        'data-cy',
+                        this.isExternalDataLocal
```
pakb, line 79: same

pakb (commented):
don't forget to hit the "re-request review" button on GitHub, so that I know I should have a look at the code 😉 
(or add me the first time to the reviewers)

src/modules/menu/components/share/MenuShareEmbed.vue
```diff
@@ -295,7 +299,22 @@ watch(
                 </div>
                 <!-- eslint-disable vue/no-v-html-->
                 <div class="small text-wrap text-center" v-html="t('share_disclaimer')"></div>
-                <!-- eslint-enable vue/no-v-html-->
```
pakb, line 298: the re-enabling of the rule should still be added, it otherwise disable this rule for the remainder of the file

pakb (approved):

PR #1063, opened 2024-09-12
PB-970: Add support for simple KMZ file

ltshb:
Only support the kml file inside the KMZ.


pakb (approved):
All good code wise.
Only thing I was wondering is : should we start documenting which exception will be raised/thrown by our file loading utils? There starts to be a lot of different reasons the `handleFileContent` or other similar functions could fail by raising errors.

PR #1066, opened 2024-09-13
PB-1000 : add support for local GeoTIFF import

pakb:
online GeoTIFF requires some work done on `service-proxy`, as it currently answers with `413 Payload Too Large` (obviously...)
I've tested with a Dropbox link (hence the use of the proxy).

[Test link with a GeoTIFF already added from STAC](https://sys-map.dev.bgdi.ch/preview/feat-pb-1000-geotiff-support/index.html#/map?lang=en&center=2556732.67,1101442.52&z=6.976&bgLayer=ch.swisstopo.pixelkarte-farbe&topic=ech&layers=GEOTIFF%7Chttps://data.geo.admin.ch/ch.swisstopo.swissimage-dop10/swissimage-dop10_2017_2556-1101/swissimage-dop10_2017_2556-1101_0.1_2056.tif)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1000-geotiff-support/index.html)


src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -39,31 +41,57 @@ export function isGpx(fileContent) {
...
         const kmz = await unzipKmz(content, source)
-        textContent = kmz.kml
+        parsedContent = kmz.kml
         linkFiles = kmz.files
+    } else if (source.endsWith('.tif')) {
+        log.debug(`File content might be a COGTIFF, attempting a parse as such`)
+        try {
+            parsedContent = await fromArrayBuffer(content)
+        } catch (err) {
+            log.debug('parsing as COGTIFF failed, moving on with other formats', err)
```
ltshb, line 61: Here it should be an error and we should throw an error in order to provide a user feedback that the tif file is invalid

src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -39,31 +41,57 @@ export function isGpx(fileContent) {
...
+    let parsedContent
     let linkFiles
     if (isZipContent(content)) {
         log.debug(`File content is a zipfile, assume it is a KMZ archive`)
         const kmz = await unzipKmz(content, source)
-        textContent = kmz.kml
+        parsedContent = kmz.kml
         linkFiles = kmz.files
+    } else if (source.endsWith('.tif')) {
+        log.debug(`File content might be a COGTIFF, attempting a parse as such`)
```
ltshb, line 57: I'm not sure it is a good idea to test for the extension, we don't do this for GPX and KML.
Maybe we should rethink this logic here and use the same as for zip file by checking the magic number header of tiff file (`49 49 2A 00` or `4D 4D 00 2A` depending on endianness)

ltshb: What about for people using `.tiff` extension or saving the file without extension on a server ?

src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -39,31 +41,57 @@ export function isGpx(fileContent) {
...
+        parsedContent = new TextDecoder('utf-8').decode(content)
     }
-    if (isKml(textContent)) {
+    if (parsedContent instanceof GeoTIFF) {
+        layer = new GeoTIFFLayer({
+            fileSource: source,
+            visible: true,
+            opacity: 1.0,
+            // For local files : OpenLayers needs a Blob, and not an already parsed GeoTIFF instance
+            data: originalFile,
```
ltshb, line 73: Here the original file will be null if imported using the import tool file if I'm correct ?

ltshb (commented):
Do you have any tiff file to test or any data.geo.admin.ch url to test as well, as far as I know we should be able to use data from service-stac so I wonder what did not worked.

src/modules/menu/components/advancedTools/ImportFile/utils.js
pakb: that's correct, it will be available through the `fileSource` prop

src/modules/menu/components/advancedTools/ImportFile/utils.js
pakb: how do you find out which byes comes first in file? is there a website for that?

ltshb: I used for zip file detection the same approach using the header magic number without being concerned about endianness which might be a mistake, don't know

pakb:
Some GeoTIFF can be found here : https://data.geo.admin.ch/browser/#/collections/ch.swisstopo.swissimage-dop10

src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -39,31 +41,59 @@ export function isGpx(fileContent) {
...
+                store.dispatch('zoomToExtent', {
+                    extent: normalizeExtent(geoTIFFExtent),
+                    ...dispatcher,
+                })
+            }
+            // we are done here, so to not run the parsed content part below, we stop the function here
+            return layer
+        } catch (err) {
+            log.debug('parsing as GeoTIFF failed', err)
+            throw new Error(`Could not parse GeoTIFF from ${source}`)
```
ltshb, line 81: You should also update the `invalid_kml_gpx_file_error` key to include GeoTiff file as invalid, this translation key will also be used for invalid geotiff if you throw a generic Error here, if you want a more specific error message then you should create a new error class, but I think updating the key above (key name and translation) should be sufficient.

ltshb (commented):
@pakb The data.geo.admin.ch link that you provide was a link to the stac browser pointing to whole collections, not a link to a STAC asset file, that's why you could not import it. Using the following link https://data.geo.admin.ch/ch.swisstopo.swissimage-dop10/swissimage-dop10_2017_2517-1139/swissimage-dop10_2017_2517-1139_0.1_2056.tif I had an issue, the import tool say `successfull` but I had a crash in the console and the file was not loaded
image

image

src/utils/utils.js
```diff
@@ -278,3 +278,22 @@ export function isZipContent(content) {
...
+ * @param {ArrayBuffer} content
+ * @returns {boolean} Return true if the content is a TIFF content
+ */
+export function isTiffContent(content) {
+    // Check the first 4 bytes for the TIFF file signature
+    // there are two signatures that can be found here https://en.wikipedia.org/wiki/TIFF
+    const tiffSignatures = [
+        [0x49, 0x49, 0x2a, 0x00],
+        [0x4d, 0x4d, 0x00, 0x2a],
+    ]
```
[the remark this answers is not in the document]
pakb: There are ways to check if the underlying system is big or little endian, but AFAIK most modern infra (even mobile) use little endian.
So let's stick with only testing little endian then

src/modules/i18n/locales/de.json
```diff
@@ -172,7 +172,7 @@
     "drawing_attached": "Zeichnung als Anhang hinzugefügt",
     "drawing_too_large": "Ihre Zeichnung ist zu gross, entfernen Sie einige Details.",
     "drop_invalid_url": "URL ist ungültig.",
-    "drop_me_here": "KML Datei hierhin ziehen",
+    "drop_me_here": "Datei hier ablegen (KML, GPX, GeoTIFF)",
```
ltshb, line 175: There is also the placeholder to update 
image

I've just noticed that with the KMZ support I forgot to update those translation and the placeholder, could you add as well KMZ when you already add GeoTIFF ?

src/modules/menu/components/advancedTools/ImportFile/utils.js
```diff
@@ -39,31 +41,59 @@ export function isGpx(fileContent) {
...
+            store.dispatch('addLayer', { layer, ...dispatcher })
+            // looking into the GeoTIFF file to get the extent
+            const geoTIFFImage = await parsedContent.getImage()
+            const geoTIFFExtent = geoTIFFImage.getBoundingBox()
+            if (geoTIFFExtent) {
+                store.dispatch('zoomToExtent', {
+                    extent: normalizeExtent(geoTIFFExtent),
+                    ...dispatcher,
+                })
+            }
```
ltshb, line 76: What if extent is not within switzerland, should we not do the same as for KML and GPX here ?

ltshb (commented):
You also need to add .tiff and .tif to `acceptedFileTypes`. Also updte hte `no_file` placeholder with GeoTiff (would be nice to add as well KMZ that I forgot 😉

src/modules/menu/components/advancedTools/ImportFile/utils.js
pakb: good point, yes we should do the same logic.
I'll try to gather all extent use case together

pakb:
here a couple GeoTIFF out of bounds can be found https://download.osgeo.org/geotiff/samples/zi_imaging/

src/modules/i18n/locales/de.json
ltshb, line 175: Still missing KMZ

ltshb (approved):
Just one KMZ missing in translation 😉 
Thanks a lot

src/modules/i18n/locales/de.json
pakb: oops 🙈 added

PR #1067, opened 2024-09-16
PB-984: add error feedback for all URL parameters

ltkum:
Issue : when modifying the URL directly, users received no feedback about what they did wrong when making mistakes, especially on a productive environment where most logs are deactivated.
Fix : we now give a feedback for every parameter that is wrongfully filled. We also added some extra cases for the layers, where we want to give a feedback on each individual layer, and for the time slider when everything is in order, but there are no layers supporting it.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-984-error-feedback-for-all-url-parameters/index.html)

### DIFF
--- a/src/api/errorQueues.api.js
+++ b/src/api/errorQueues.api.js
@@ -15,9 +15,9 @@ export function getStandardErrorMessage(query, urlParamName) {
  * @param {Boolean} is_valid Is the value valid or not
  * @returns
  */
-export function getStandardValidationResponse(query, is_valid, urlParamName) {
+export function getStandardValidationResponse(query, isValid, urlParamName) {
     return {
-        valid: is_valid,
-        errors: is_valid ? null : getStandardErrorMessage(query, urlParamName),
+        valid: isValid,
+        errors: isValid ? null : getStandardErrorMessage(query, urlParamName),
     }
 }
--- a/src/api/errorQueues.api.js
+++ b/src/api/errorQueues.api.js
@@ -1,7 +1,10 @@
 import ErrorMessage from '@/utils/ErrorMessage.class'
 
-export function getStandardErrorMessage(query) {
-    return new ErrorMessage('url_parameter_error', { param: this.urlParamName, value: query })
+export function getStandardErrorMessage(query, urlParamName) {
+    return new ErrorMessage('url_parameter_error', {
+        param: urlParamName,
+        value: query,
+    })
 }
 
 /**
@@ -12,9 +15,11 @@ export function getStandardErrorMessage(query) {
  * @param {Boolean} is_valid Is the value valid or not
  * @returns
  */
-export function getStandardValidationResponse(query, is_valid) {
+export function getStandardValidationResponse(query, is_valid, urlParamName) {
+    console.warn('inGETSTANDARD')
+    console.warn(is_valid)
     return {
         valid: is_valid,
-        errors: is_valid ? null : this.getStandardErrorMessage(query),
+        errors: is_valid ? null : getStandardErrorMessage(query, urlParamName),
     }
 }
--- a/src/api/errorQueues.api.js
+++ b/src/api/errorQueues.api.js
@@ -16,8 +16,6 @@ export function getStandardErrorMessage(query, urlParamName) {
  * @returns
  */
 export function getStandardValidationResponse(query, is_valid, urlParamName) {
-    console.warn('inGETSTANDARD')
-    console.warn(is_valid)
     return {
         valid: is_valid,
         errors: is_valid ? null : getStandardErrorMessage(query, urlParamName),
--- a/src/api/errorQueues.api.js
+++ b/src/api/errorQueues.api.js
@@ -1,0 +1,20 @@
+import ErrorMessage from '@/utils/ErrorMessage.class'
+
+export function getStandardErrorMessage(query) {
+    return new ErrorMessage('url_parameter_error', { param: this.urlParamName, value: query })
+}
+
+/**
+ * Return the standard feedback for most parameters given in the URL: if the query is validated, it
+ * can proceed and be set in the store.
+ *
+ * @param {any} query The value of the URL parameter given
+ * @param {Boolean} is_valid Is the value valid or not
+ * @returns
+ */
+export function getStandardValidationResponse(query, is_valid) {
+    return {
+        valid: is_valid,
+        errors: is_valid ? null : this.getStandardErrorMessage(query),
+    }
+}
--- a/src/api/errorQueues.api.js
+++ b/src/api/errorQueues.api.js
@@ -12,7 +12,7 @@ export function getStandardErrorMessage(query, urlParamName) {
  * can proceed and be set in the store.
  *
  * @param {any} query The value of the URL parameter given
- * @param {Boolean} is_valid Is the value valid or not
+ * @param {Boolean} isValid Is the value valid or not
  * @returns
  */
 export function getStandardValidationResponse(query, isValid, urlParamName) {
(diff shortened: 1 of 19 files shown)


src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -116,6 +109,7 @@ export default class CrossHairParamConfig extends AbstractParamConfig {
             keepInUrlWhenDefault: false,
             valueType: String,
             defaultValue: null,
+            acceptedValues: acceptedValues, // TODO : implement this to have a common behavior amongst all params
```
ltshb, line 112: No TODO in code

src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -99,6 +77,21 @@ function generateCrossHairUrlParamFromStoreValues(store) {
     return null
 }
 
+function acceptedValues(store, query) {
```
ltshb, line 80: The name of the functions implies that it would return a list and not a boolean

ltshb: it seems to me that it is more a validate function, so `validate` seems better.

src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -99,6 +77,21 @@ function generateCrossHairUrlParamFromStoreValues(store) {
...
+    if (query) {
+        const parts = query.split(',')
+        let crossHair = parts[0]
+        let crossHairPosition = [parseFloat(parts[1]), parseFloat(parts[2])]
+
+        return (
+            (typeof query === 'string' || query instanceof String) &&
```
ltshb, line 87: Testing for string after the split don't make sense, moreover the input should come from the vue router which is either a string or null/undefined, meaning that the if at line 81 would test for existence and non empty string.

src/router/storeSync/CrossHairParamConfig.class.js
```diff
@@ -99,6 +77,21 @@ function generateCrossHairUrlParamFromStoreValues(store) {
...
+            (crossHair || crossHairPosition) &&
```
ltshb, line 88: Here crossHairPosition will always be true (unless parseFloat(undefined) crash

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -244,6 +245,44 @@ function generateLayerUrlParamFromStoreValues(store) {
...
+                if (!layer.visible || layer.opacity) {
+                    value += `,${layer.visible ? '' : 'f'},${layer.opacity ?? 1}`
+                }
+                let error_type = layer.isExternal
+                    ? 'url_external_layer_no_scheme_error'
+                    : 'url_layer_error'
+                store.dispatch('addError', {
+                    error: new ErrorMessage(error_type, { layer: value }),
+                    dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
+                })
```
ltshb, line 280: The behavior of this function should be consistent, here it dispatch the error and always return true while in other case it don't do the dispatch ! That is not easy to understand and follow and is error prone. This `acceptedValues` which should be more `validate` should have a consistent behavior overall all validations. For example it could return always an object

```
{
  valid: true | false,
  errors: undefined | [ErrorMessage]
}
```

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -244,6 +245,44 @@ function generateLayerUrlParamFromStoreValues(store) {
...
+                    value = `${layer.type}|`
+                }
+                value += layer.id
+                Object.entries(layer.customAttributes).forEach(
+                    ([attribute_name, attribute_value]) =>
+                        (value += `@${attribute_name}=${attribute_value}`)
+                )
+                if (!layer.visible || layer.opacity) {
+                    value += `,${layer.visible ? '' : 'f'},${layer.opacity ?? 1}`
+                }
```
ltshb, line 273: It doesn't make sense to rewrite the whole layer parameter here and it might not match the one from url due to default values and due to order of custom attributes, simply use the layer.id, keep it simple stupide.

src/router/storeSync/layersParamParser.js
```diff
@@ -50,12 +50,14 @@ export function decodeUrlLayerId(urlLayerId) {
...
         const decodedLayerId = layerId ? decodeExternalLayerParam(layerId) : decodedBaseUrl
         return {
+            isExternal: true,
             id: decodedLayerId,
             type: layerType,
             baseUrl: decodedBaseUrl,
         }
     }
     return {
+        isExternal: false,
```
ltshb, line 60: We need to take care of the definition of `isExternal` ! In the layer object (e.g. WMTSLayer, KMLLayer, ...) isExternal has a slightly different meaning which will lead to errors in future or even now ! Here you take every layer that has a type aka starting with `LAYER_TYPE|` as external while in the layer abstract KML layers generated by the application will be an internal even thought it starts with `KML|` ! I would not add this flag here just for the validation purpose. For the validation simply check for the types and with your filter for layer config, you must have a `baseUrl` to check for so no need for this `isExternal`

src/router/storeSync/abstractParamConfig.class.js
```diff
@@ -35,6 +38,7 @@ export default class AbstractParamConfig {
         keepInUrlWhenDefault = true,
         valueType = String,
         defaultValue = null,
+        acceptedValues = null,
```
ltshb, line 41: See comment about naming.

ltshb (changes requested):
I did not find the ticket from the sprint ? Should we not concentrate on the sprint tickets based on our team discussions?

src/router/storeSync/CrossHairParamConfig.class.js
ltkum: I must admit it started as a list, and then thought it would be better for it to be a boolean directly, but forgot to change the name. This is a good suggestion.

src/api/errorQueues.api.js
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Return the standard feedback for most parameters given in the URL: if the query is validated, it
+ * can proceed and be set in the store.
+ *
+ * @param {any} query The value of the URL parameter given
+ * @param {Boolean} is_valid Is the value valid or not
+ * @returns
+ */
+export function getStandardValidationResponse(query, is_valid, urlParamName) {
```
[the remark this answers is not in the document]
ltkum: 01j8jjs19xcmdxyktf6rhbmfpx

src/api/errorQueues.api.js
```diff
@@ -0,0 +1,23 @@
...
+        value: query,
+    })
+}
+
+/**
+ * Return the standard feedback for most parameters given in the URL: if the query is validated, it
+ * can proceed and be set in the store.
+ *
+ * @param {any} query The value of the URL parameter given
+ * @param {Boolean} is_valid Is the value valid or not
```
pakb, line 15: That also needs an update

src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -254,6 +256,42 @@ function generateLayerUrlParamFromStoreValues(store) {
...
+                )
+            }
+        })
+    const valid = faulty_layers.length < parsed.length
+    if (!valid) {
+        return getStandardValidationResponse(query, valid, this.urlParamName)
+    }
+    return {
+        valid,
+        errors: faulty_layers.length === 0 ? null : faulty_layers,
```
pakb, line 291: 🐍🫷

pakb (approved):
some more snake case to remove and we are good to go

PR #1068, opened 2024-09-17
PB-1006: Fixed CORS issue with admin.ch hosts

ltshb:
Some admin.ch hosts like uvek-gis.admin.ch don't support yet CORS, therefore
they still need to go through proxy.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1006-kml-admin.ch/index.html)


hansmannj (approved):
:+1: Thanks!

PR #1071, opened 2024-09-18
PB-901: Modify name of KML file

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-901-kml-change-name-file/index.html)


sommerfe:
Apparently i made a mistake during the rebase so most of my changes are gone, so i have to insert the right code

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -67,6 +76,18 @@ function onCloseClearConfirmation(confirmed) {
         store.dispatch('setDrawingMode', { mode: null, ...dispatcher })
     }
 }
+
+async function onSaveName() {
+    isActiveLayerNameEditMode.value = !isActiveLayerNameEditMode.value
+    if (activeKmlLayerName.value === activeKmlLayer.value.name) return
+    store.dispatch('updateLayer', {
+        index: activeKmlLayerIndex.value,
+        layer: { ...activeKmlLayer.value, name: activeKmlLayerName.value },
```
pakb, line 85: I'm not sure a spread will be sufficient, as there is nested objects (such as the KML metadata).
If you want a short handed way of dealing with a targeted update, you can use the `updateLayers` action in this way
```Javascript
store.dispatch('updateLayers', {
    layers: [
        {
            id: activeKmlLayer.value.id,
            name: activeKmlLayerName.value,
        }
    ]
})
```

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -87,6 +108,41 @@ function onDeleteLastPoint() {
                 class="card text-center drawing-toolbox-content shadow-lg rounded-bottom rounded-top-0 rounded-start-0"
                 :class="{ 'rounded-bottom-0': isPhoneMode }"
             >
+                <div
+                    class="d-flex justify-content-center align-items-center gap-2 p-4"
+                    v-if="activeKmlLayer"
```
pakb, line 113: If you could find a way to not hide it at the opening of the drawing module, I would prefer that.
It is a bit weird that the UI "grows" after the first save occurs.

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -87,6 +108,41 @@ function onDeleteLastPoint() {
...
+                    <label for="activeKmlLayerName" class="mr-3">
+                        {{ i18n.t('file_name') }}
+                    </label>
+                    <div class="input-group">
+                        <input
+                            id="activeKmlLayerName"
+                            v-model="activeKmlLayerName"
+                            type="string"
+                            class="form-control"
+                            :disabled="!isActiveLayerNameEditMode"
```
pakb, line 124: Can you change that to something like https://getbootstrap.com/docs/5.3/forms/form-control/#readonly-plain-text ?

The edit button should maybe be separated from the text then.

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -87,6 +108,41 @@ function onDeleteLastPoint() {
...
+                            :placeholder="`${i18n.t('draw_layer_label')}`"
+                        />
+                        <button
+                            v-if="!isActiveLayerNameEditMode"
+                            class="btn btn-outline-secondary"
+                            type="button"
+                            @click="isActiveLayerNameEditMode = !isActiveLayerNameEditMode"
+                        >
+                            {{ i18n.t('edit_button') }}
```
pakb, line 133: I'm fine not having a text, but use the icon https://fontawesome.com/icons/pen?f=classic&s=solid

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -87,6 +108,41 @@ function onDeleteLastPoint() {
...
+                        </button>
+                        <button
+                            v-else
+                            class="btn btn-outline-secondary"
+                            type="button"
+                            @click="onSaveName"
+                        >
+                            {{ i18n.t('save_button') }}
```
pakb, line 141: you can use `validate` key here, I'll add the necessary i18n entries later

src/store/modules/layers.store.js
```diff
@@ -395,6 +400,7 @@ const actions = {
      * @param {string} dispatcher Action dispatcher name
      */
     updateLayer({ commit, getters }, { index, layer, dispatcher }) {
+        console.log('updateLayer', index, layer.name)
```
pakb, line 403: You can check your console in the browser and make sure you've activated the "Verbose" level, this information is already present (each mutation and actions are logged, with they payload etc...)

pakb (commented):
don't keep `console.log` when pushing code.
If you feel this log could help us later down the line, you can use the logging utils from `src/utils/logging` (there's a `debug()` or `error()` function that can be imported from this file)

src/modules/drawing/components/DrawingToolbox.vue
sommerfe: I could disable it when there is no active kml layer yet is that ok ?

src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -67,6 +75,20 @@ function onCloseClearConfirmation(confirmed) {
...
     }
 }
+
+async function onSaveName() {
+    if (activeKmlLayerName.value === activeKmlLayer.value.name) return
+    store.dispatch('updateLayer', {
+        index: activeKmlLayerIndex.value,
+        layer: {
+            id: activeKmlLayer.value.id,
+            name: activeKmlLayerName.value,
```
pakb, line 85: Wouldn't it make more sense to sanitize it here, instead of only when exporting it?

src/store/modules/layers.store.js
```diff
@@ -144,6 +144,18 @@ const getters = {
             (layer) => layer.visible && layer.type === LayerTypes.KML && !layer.isExternal
         ) ?? null,
 
+    /**
+     * Get index of current KML layer selected for drawing.
+     *
+     * That is the KML layer that will be used when the drawing mode is opened.
+     *
+     * When no KML layer is in active layers then -1 is returned.
```
pakb, line 152: That is why you've added this getter, but it can work for other type of layers too. You might want to reflect that in the comment and have something more general as a description (and maybe end by saying this is mainly used to get the index of the active KML layer)

tests/cypress/tests-e2e/drawing.cy.js
pakb, line 1: For future changes in this tooltip's size, I think it would be good to store (or even better to import in the JS from the SCSS!) the size of the tooltip, and have clicks done relative to this size

pakb (commented):
From the user's perspective it doesn't really make sense that they can't start by filling the name of what they want to draw and then start drawing.

As you've already used a `ref` to store the name of the drawing (of the KML), I'd really want that we can edit this before the KML even exists. In the `onSaveName`, we could check that the KML exists before dispatching the change, and watch the activeKmlLayer for initial creation to set the name correctly.

I think this would really be better for user experience, I'm betting on multiple question/helpdesk interactions with question like "Why can't I change the drawing name"

src/store/modules/layers.store.js
```diff
@@ -78,6 +78,12 @@ const state = {
      * @type AbstractLayer[]
      */
     systemLayers: [],
+    /**
+     * Name of the active KML layer for drawing.
+     *
+     * @type {String | null}
+     */
+    activeKmlLayerName: null,
```
pakb, line 86: Not sure you need to store that, it's modified only in the context of the DrawingModule (or more precisely in DrawingToolbox) and I don't think it needs to be "accessed" or changed in any other way than reading the KML metadata later on.

IMHO you could remove this from the store

src/modules/drawing/components/DrawingExporter.vue
```diff
@@ -37,7 +38,7 @@ function exportDrawing() {
         type = 'application/gpx+xml;charset=UTF-8'
     } else {
         fileName = generateFilename('.kml')
-        content = generateKmlString(projection.value, features)
+        content = generateKmlString(projection.value, features, activeKmlLayerName.value)
```
pakb, line 41: would be better to access the name of the `KMLLayer` stored at `getters.activeKmlLayer` and not store that separately (as said above)

src/store/modules/layers.store.js
sommerfe: The problem is when there is no active kml layer yet but you already want to save the name the name has to be stored somewhere so it won't work to remove this variable from the store

pakb: Cant' you keep track of it in the component only?
If it's not possible to have it any other place than the store, I'd move it to `drawing.store.js` to clarify what is the purpose of this (and maybe name it, `drawingName` or something like that)

sommerfe: I think it depends on the desired behaviour
for example what should be the behaviour when you change the name, click on validate, then return to the main menu without having drawn anything. Should then be the saved name be there when you return to the drawing interface or not?

sommerfe:
When you change the name and then click on "Retour / Terminer Dessin" you are returned to the main menu but there is still the loading indication at the top so there might be an issue

pakb (approved):
Last part of the job (and not the easiest) : fix cypress testing 😉

pakb (commented):
Can you clean up a bit the git log? There's still my `WIP` commit message that should not be merged to `develop` as they are

You can squash stuff, and remove my commits, that's your work anyway

PR #1073, opened 2024-09-20
PB-1017: un-proxy external graphics in print specs

ltkum:
Issue : Service print can't use service-proxy like the mapviewer, so when external graphics are proxied in the mapviewer to go around CORS issues, those same graphics can't be rendered in print jobs

Fix : we un-proxy the 'externalGraphics' attribute of styles in the print spec, so the original URL is sent to service print.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-1017-unproxify-print-jobs/index.html)


ismailsunni (approved):
LGTM, I also tested with the sample KML given in slack

PR #1075, opened 2024-09-23
PB-960: handle long press events error on some mobiles

ltkum:
Issue : on some phones, a long press event would also fire a single press event afterwards, causing the application to open both a location popup and an infobox.
Cause : setting the longClickTriggered boolean was not registered while in the longClickTimeout function, most likely due to a scoping issue.

Fix : we set the longClickTriggered boolean in the 'onMapRightClick' event instead. We also added a small parameter in the function to tell if we need to set up the flag or not, as desktop users won't need this.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-960-handle-longpress-event-conflict/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
+++ b/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
@@ -184,6 +184,7 @@ export default function useMapInteractions(map) {
 
     function onMapRightClick(event) {
         clearTimeout(longClickTimeout)
+        longClickTriggered = event.updateLongClickTriggered ?? false
         store.dispatch('click', {
             clickInfo: new ClickInfo({
                 coordinate: event.coordinate,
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
```diff
@@ -182,8 +182,9 @@ export default function useMapInteractions(map) {
         })
     }
 
-    function onMapRightClick(event) {
+    function onMapRightClick(event, updateLongClickTriggered = false) {
         clearTimeout(longClickTimeout)
+        longClickTriggered = updateLongClickTriggered
```
ismailsunni, line 187: Why not set the value to `false` here? Because I don't see the `onMapRightClick` function is called with different value for `updateLongClickTriggered`.

ltkum: Well, it was meant to be set to true in the `longClickTimeout` call but I somehow forgot to add the argument, thanks for pointing this out.

ismailsunni (approved):

PR #1076, opened 2024-09-25
PB-742: Do not decodeURI twice.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-742-malformed-link/index.html)

### DIFF
--- a/src/router/legacyPermalinkManagement.routerPlugin.js
+++ b/src/router/legacyPermalinkManagement.routerPlugin.js
@@ -49,7 +49,6 @@ const handleLegacyParam = (
 ) => {
     const { projection } = store.state.position
     let newValue
-    let alreadyDecoded = false
     let key = param
     switch (param) {
         case 'zoom':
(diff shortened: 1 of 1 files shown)


src/router/legacyPermalinkManagement.routerPlugin.js
```diff
@@ -158,9 +160,9 @@ const handleLegacyParam = (
         // When receiving a query, the application will encode the URI components
         // We decode those so that the new query won't encode encoded character
         // for example, we avoid having " " becoming %2520 in the URI
-        newQuery[key] = decodeURIComponent(newValue)
+        newQuery[key] = alreadyDecoded ? newValue : decodeURIComponent(newValue)
```
ltkum, line 163: ,I would replace the 'alreadyDecoded' boolean by a `param==='layers'` check here, and integrate the line 114 explanation comment in the comment above. 
Unless there are other legacy parameters handling which decode the URI, of course, in which case we'll have to assign true to `alreadyDecoded` in those case instead.

ltkum (commented):
I would make a small change to avoid extra variables, but the logic is otherwise good :)

ltkum (approved):

PR #1078, opened 2024-09-26
PB-1003: no more long mouse left clicks

ltkum:
Issue : pressing the left mouse button for a long period of time would cause the location popup to appear

 Cause : in openlayers, the singleClick event represents both mouse left clicks and mobiles touchdown events

 Fix : We check the pointer type of the original event, ensuring we only handle the cases where this was triggered on a mobile.

[Test link](https://sys-map.dev.bgdi.ch/preview/bugfix-pb-1003-stop-long-left-click-interaction/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
+++ b/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
@@ -212,7 +212,8 @@ export default function useMapInteractions(map) {
             // triggering a long click on the same spot after 500ms, so that mobile cas have access to the
             // LocationPopup by touching the same-ish spot for 500ms
             longClickTimeout = setTimeout(() => {
-                if (!mapHasMoved) {
+                // we need to ensure long mouse clicks don't trigger this.
+                if (!mapHasMoved && event.originalEvent.pointerType === 'touch') {
                     // we are outside of OL event handling, on the HTML element, so we do not receive map pixel and coordinate automatically
                     const pixel = map.getEventPixel(event)
                     const coordinate = map.getCoordinateFromPixel(pixel)
--- a/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
+++ b/src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
@@ -20,6 +20,11 @@ import log from '@/utils/logging'
 const dispatcher = {
     dispatcher: 'useMapInteractions.composable',
 }
+const longPressEvents = [
+    'touch',
+    'pen',
+    '', // This is necessary for IoS support
+]
 
 export default function useMapInteractions(map) {
     const store = useStore()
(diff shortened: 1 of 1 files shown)


src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
```diff
@@ -212,7 +212,8 @@ export default function useMapInteractions(map) {
             // triggering a long click on the same spot after 500ms, so that mobile cas have access to the
             // LocationPopup by touching the same-ish spot for 500ms
             longClickTimeout = setTimeout(() => {
-                if (!mapHasMoved) {
+                // we need to ensure long mouse clicks don't trigger this.
+                if (!mapHasMoved && event.originalEvent.pointerType === 'touch') {
```
pakb, line 216: sadly this blocks iOs touch too... 
in the doc [here](https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType) it says that browser not supporting this feature will fill the `pointerType` with `""`, we could maybe try to react to this value too (maybe that will add back iOs...)
So something like
```Javascript
if (!mapHasMoved && ['' /* iOs support */, 'touch'].includes(event.originalEvent.pointerType)) {
    // ...
}
```

pakb: or do the negative test
```Javascript
if (!mapHasMoved && event.originalEvent.pointerType !== 'mouse') {
    // ...
}
```

ltkum: the negative seems simpler and more explicit in my opinion. But yeah, I did not think that IoS would be causing problems in that manner

pakb (commented):
seems like iOS wants us to think they are a mouse... It's not working either.

another approach will be to change the logic here, and instead of relying only on `pointerdown` and `pointerup` (which are usually more geared toward a mouse) we could also listen to `touchstart` and use the timer there.

I think iOS (at least) hooks itself correctly with `touchstart`...

src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
```diff
@@ -20,6 +20,8 @@ import log from '@/utils/logging'
 const dispatcher = {
     dispatcher: 'useMapInteractions.composable',
 }
+// these are the events that can trigger a long press event. This is necessary for IoS support
+const longPressEvents = ['touch', '', 'pen']
```
pakb, line 24: Maybe split your comment in multiple lines, something like
```Javascript
// these are the events that can trigger a long press event.
const longPressEvents = [
    'touch',
     'pen',
    '', // This is necessary for iOS support
]
```
As it's not clear currently if all are necessary for iOS support, or which one of them is

pakb (approved):

src/modules/map/components/openlayers/utils/useMapInteractions.composable.js
ltkum: yeah, the comment predates the time I had a good name for the variable ^^.

PR #1079, opened 2024-09-26
PB-985: Make Layer Info Movable

ismailsunni:
Update screencast:

Peek 2024-10-08 15-16 - layer description movable

I'm not super expert in CSS, so any suggestions are welcome :)

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-985-floating-layer-info/index.html)

### DIFF
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -59,7 +59,12 @@ onMounted(async () => {
 </script>
 
 <template>
-    <ModalWithBackdrop :title="title" :allow-print="true" @close="emit('close', layerId)">
+    <ModalWithBackdrop
+        :title="title"
+        :allow-print="true"
+        :movable="true"
+        @close="emit('close', layerId)"
+    >
         <div class="layer-description" data-cy="layer-description-popup">
             <h4 v-if="!isExternal && !htmlContent" class="mb-0">
                 <font-awesome-icon spin :icon="['fa', 'spinner']" />
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -64,6 +64,7 @@ onMounted(async () => {
         allow-print
         movable
         allow-minimize
+        compact
         :use-black-backdrop="false"
         @close="emit('close', layerId)"
     >
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -63,6 +63,7 @@ onMounted(async () => {
         :title="title"
         allow-print
         movable
+        allow-minimize
         :use-black-backdrop="false"
         @close="emit('close', layerId)"
     >
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -59,12 +59,7 @@ onMounted(async () => {
 </script>
 
 <template>
-    <ModalWithBackdrop
-        :title="title"
-        :allow-print="true"
-        :movable="true"
-        @close="emit('close', layerId)"
-    >
+    <ModalWithBackdrop :title="title" allow-print movable @close="emit('close', layerId)">
         <div class="layer-description" data-cy="layer-description-popup">
             <h4 v-if="!isExternal && !htmlContent" class="mb-0">
                 <font-awesome-icon spin :icon="['fa', 'spinner']" />
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -59,7 +59,13 @@ onMounted(async () => {
 </script>
 
 <template>
-    <SimpleWindow :title="title" movable allow-print @close="emit('close', layerId)">
+    <SimpleWindow
+        :title="title"
+        movable
+        allow-print
+        initial-position="top-left"
+        @close="emit('close', layerId)"
+    >
         <div class="layer-description" data-cy="layer-description-popup">
             <h4 v-if="!isExternal && !htmlContent" class="mb-0">
                 <font-awesome-icon spin :icon="['fa', 'spinner']" />
--- a/src/modules/menu/components/LayerDescriptionPopup.vue
+++ b/src/modules/menu/components/LayerDescriptionPopup.vue
@@ -5,7 +5,7 @@ import { useStore } from 'vuex'
 
 import AbstractLayer from '@/api/layers/AbstractLayer.class'
 import { getLayerDescription } from '@/api/layers/layers.api'
-import ModalWithBackdrop from '@/utils/components/ModalWithBackdrop.vue'
+import SimpleWindow from '@/utils/components/SimpleWindow.vue'
 
 const props = defineProps({
     layer: {
@@ -59,15 +59,7 @@ onMounted(async () => {
 </script>
 
 <template>
-    <ModalWithBackdrop
-        :title="title"
-        allow-print
-        movable
-        allow-minimize
-        compact
-        :use-black-backdrop="false"
-        @close="emit('close', layerId)"
-    >
+    <SimpleWindow :title="title" movable @close="emit('close', layerId)">
         <div class="layer-description" data-cy="layer-description-popup">
             <h4 v-if="!isExternal && !htmlContent" class="mb-0">
                 <font-awesome-icon spin :icon="['fa', 'spinner']" />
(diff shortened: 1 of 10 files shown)


ismailsunni:
Need to fix the test first, it seems the approach is wrong :)

ltkum (changes requested):
In the ticket, there was also the idea of adding a 'minimize' button if this did not take too much effort. In the `mapPopoverComponent`, this is handled with this button : https://github.com/geoadmin/web-mapviewer/blob/693e590984ed0dc3c71db15179bc2f2434c92c52/src/modules/map/components/MapPopover.vue#L156-L163, which toggles a boolean telling us if we should show the content or not. If it's not too much effort, this could be a good idea to add it now.

src/utils/components/ModalWithBackdrop.vue
```diff
@@ -119,11 +120,48 @@ export default {
...
+        this.modalClass = {
+            'start-50': true,
+            'top-50 translate-middle': !this.top,
+            'translate-middle-x on-top-with-padding': this.top,
+        }
+        if (this.movable) {
+            this.$nextTick(() => {
+                this.setInitialPosition()
+            })
         }
```
ltkum, line 165: In the `mapPopopver` module, we handle the movable aspect of the component with a `useMovableElement` function, like so : https://github.com/geoadmin/web-mapviewer/blob/693e590984ed0dc3c71db15179bc2f2434c92c52/src/modules/map/components/MapPopover.vue#L83-L113

I think this would be good to keep only one way to move popups in the viewer.

src/modules/menu/components/LayerDescriptionPopup.vue
```diff
@@ -59,7 +59,12 @@ onMounted(async () => {
 </script>
 
 <template>
-    <ModalWithBackdrop :title="title" :allow-print="true" @close="emit('close', layerId)">
+    <ModalWithBackdrop
+        :title="title"
+        :allow-print="true"
+        :movable="true"
```
pakb, line 65: Small hint, if you want to set a Boolean prop to true, you can just add its name and "voilà"
```
<ModalWithBackdrop
    :title="title"
    allow-print
    movable
    @close="emit('close', layerId)"
>
```

pakb (commented):
As stated by Martin, it would make more sense to share the "movable" part of the `MapPopover.vue` here (through some composable or component hierarchy) instead of recreating it.

pakb:
I would also shrink the width of the container to fit its content, but not to go further anymore.
You can also get rid of the black backdrop as soon as this is movable (that's the end goal, have the legend visible, but still be able to manipulate the map)

pakb (commented):
If possible, I'd lower the width of the movable layer info to somewhat similar width as the feature tooltip

ismailsunni:
Hi @ltkum @pakb 

I have updated my works, now it contains:
- The usage of useMovableComponent (add new prop to handle initial position)
- Add button to minimize
- Reduce the width, now it's more or less similar as the info box
- Remove the black backdrop (add some new props)

I have updated the screencast in the first comment

Feel free to test and suggestions are welcomed.

src/utils/components/ModalWithBackdrop.vue
```diff
@@ -226,6 +221,10 @@ export default {
             // But for desktop we let the size be dynamic with max to 90% of the view
             max-width: 80vw;
             max-height: 90svh;
+
+            &.modal-popup-compact {
+                width: 20vw;
```
pakb, line 226: could you use the variable `$overlay-width` here? (from `variables.module.scss`)
This way it will be exactly the same size as the feature tooltip.

With relative width, it tends to break apart on mobile if you don't deal with smaller screen width separately

src/utils/composables/useMovableElement.composable.js
```diff
@@ -132,9 +135,22 @@ export function useMovableElement(element, options = {}) {
...
+            initialPositionClasses.forEach((className) => {
+                htmlElement.classList.remove(className)
+            })
+            htmlElementStyle.top = `${rect.top}px`
+            htmlElementStyle.left = `${rect.left}px`
+        } else {
+            htmlElementStyle.top = `${top}px`
+            htmlElementStyle.left = `${left}px`
+        }
+        firstMovement = false
```
pakb, line 153: Shouldn't you also be adding them back after the move ended?

ismailsunni: No, because after we move the modal, the location of the modal will be handled by the `useMovableElement`. If not, the modal will jump to another location (e.g. if there is a `translate` or `start-50` initially)

src/utils/components/ModalWithBackdrop.vue
ismailsunni: ah ok, thanks!

src/utils/composables/useMovableElement.composable.js
pakb: Ok meaning you can give your modal some initial placement with for instance `end-0` (float right) and set it here in this `initialPositionClasses` so that it will be removed and placement be entirely managed by this composable 👍

ismailsunni:
hi @pakb 
could you show me tje issue in a smaller screen? I have tried different device size, and it seems all works well. The only difference is the width of the layer description (now it sets to `$overlay-width`)

src/utils/composables/useMovableElement.composable.js
ismailsunni: yes, exactly

ltshb (changes requested):
It breaks the normal modal windows like the third party disclaimer and the import file window. In this case the black drop is removed and the close function when clicking outside of the window don't work anymore.
For the import file window, it might also good to make it movable by adding the `movable` option to it. For the disclaimer the modal with backdrop should be kept.
But to be honest I would not add the movable to the `ModalWithBackdrop`, to me a modal window should not be movable.

Can you not replace the `ModalWithBackrop` by the `MapPopover` component for the layer description ? Or even better use the `SimpleWindow` component like for the report a problem and add the movable to it. This component is used by the report a problem window which is not movable but would be surely nice to have it movable as well.

ismailsunni:
@ltshb I have used SimpleWindow for the LayerDescription, add movable to it, and make the report window movable.

ltshb (commented):
Very nice and very simple code 😉 
Just one more small improvement is the initial window position. For the feedback window having it on the top right of the map make sense as the feedback button is on the top right and it leave most of the map free so the user can add his feedback while consulting the map. But for the layer description it doesn't make sense and is disturbing on big screen, the user click on the info button which is on the left of the screen and the window appears on the opposite side of the screen.

I propose to add a `initial-position` property to the `SimpleWindow` with the following two values:
- `top-right`
- `top-left`
- `center`
And use `center` as default which makes more sense for default value. `bottom-right`, `bottom-left`, ... can be later on implemented if needed. We can also use `top-right` as default and let `center` non-implemented. Of course then feedback would be `top-right` and layer description `top-left`.
It might also make sense for UI consistency to add the Import a File (import external KML file) window as well movable with the `SimpleWindow` with `top-left` as initial value.

pakb (commented):
Nice! I link the use of the composable, pretty neat!

I upvote Brice's last points, and then we are good to go

ismailsunni:
Updated screenshots:

Import file:
image
Layer description and feedback dialog
image

Smaller screen:

image
image
image

src/utils/composables/useMovableElement.composable.js
```diff
@@ -123,18 +131,36 @@ export function useMovableElement(element, options = {}) {
...
+        const htmlElement = toValue(element)
+        const htmlElementStyle = htmlElement.style
+
+        // In case the original element has CSS class that affects its position, we remove them first if the user is dragging the element
+        if (
+            firstMovement &&
+            initialPositionClasses.length > 0 &&
+            movementSource === MovementSource.MOUSE_DRAG
+        ) {
+            console.log('first movement')
```
ltshb, line 152: No `console.log()`

src/utils/composables/useMovableElement.composable.js
```diff
@@ -147,6 +173,7 @@ export function useMovableElement(element, options = {}) {
      * here.
      */
     function constrainsElementWithinViewport() {
+        console.log('constrainsElementWithinViewport')
```
ltshb, line 176: No console.log

ltshb (commented):
Very nice.

For the import file, I wonder if it would not be better to have it also as `top-left` without `wide` ? But if removing the wide, we probably need some min-width, otherwise the windows is very small. Something like this 
image
Of course the min-width should be probably dynamic based on the screen size, in the printscreen I used 400px which don't work on mobile phone.

ismailsunni:
@ltshb 

Normal browser size
image

Smaller browser size
image

The mobile will be the same as my previous screenshot.

ltshb (approved):
Very nice

PR #1084, opened 2024-10-01
PB-951: do not store swisssearch param in url

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-951-not-store-swisssearch/index.html)

### DIFF
--- a/src/modules/menu/components/search/SearchBar.vue
+++ b/src/modules/menu/components/search/SearchBar.vue
@@ -9,7 +9,7 @@ const dispatcher = { dispatcher: 'SearchBar' }
 
 const store = useStore()
 
-const isPristine = ref(true)
+const isPristine = ref(true) // if search bar is not yet modified by the user
 const showResults = ref(false)
 const searchInput = ref(null)
 const searchValue = ref('')
@@ -20,16 +20,21 @@ const searchQuery = computed(() => store.state.search.query)
 const hasResults = computed(() => store.state.search.results.length > 0)
 const isPhoneMode = computed(() => store.getters.isPhoneMode)
 
-watch(hasResults, (newValue) => {
-    // if an entry has been selected from the list, do not show the list again
-    // because the list has been hidden by onEntrySelected. Also if the search bar is pristine (not
-    // yet modified by the user) we don't want to show the result (e.g. at startup with a swisssearch
-    // query param)
-    if (!selectedEntry.value && !isPristine.value) {
-        log.debug(`Search has result changed to ${newValue}, change the show result to ${newValue}`)
-        showResults.value = newValue
-    }
-})
+watch(
+    hasResults,
+    (newValue) => {
+        // if an entry has been selected from the list, do not show the list again
+        // because the list has been hidden by onEntrySelected.
+        if (!selectedEntry.value) {
+            log.debug(
+                `Search has result changed to ${newValue}, change the show result to ${newValue}`
+            )
+            showResults.value = newValue
+        }
+    },
+    // we need to run the watcher immediately to make sure the result list is displayed on the first load
+    { immediate: true }
+)
 
 watch(showResults, (newValue) => {
     if (newValue && isPhoneMode.value && store.state.ui.showMenu) {
--- a/src/router/storeSync/SearchParamConfig.class.js
+++ b/src/router/storeSync/SearchParamConfig.class.js
@@ -2,6 +2,7 @@ import AbstractParamConfig, {
     STORE_DISPATCHER_ROUTER_PLUGIN,
 } from '@/router/storeSync/abstractParamConfig.class'
 
+const URL_PARAM_NAME = 'swisssearch'
 /**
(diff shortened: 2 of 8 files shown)


src/modules/menu/components/search/SearchBar.vue
```diff
@@ -20,16 +20,21 @@ const searchQuery = computed(() => store.state.search.query)
...
-        showResults.value = newValue
-    }
-})
+watch(
+    hasResults,
+    (newValue) => {
+        // if an entry has been selected from the list, do not show the list again
+        // because the list has been hidden by onEntrySelected.
+        if (!selectedEntry.value) {
+            // if (!selectedEntry.value) {
```
ltshb, line 29: Clean up

src/modules/menu/components/search/SearchBar.vue
```diff
@@ -20,16 +20,21 @@ const searchQuery = computed(() => store.state.search.query)
...
-    if (!selectedEntry.value && !isPristine.value) {
-        log.debug(`Search has result changed to ${newValue}, change the show result to ${newValue}`)
-        showResults.value = newValue
-    }
-})
+watch(
+    hasResults,
+    (newValue) => {
+        // if an entry has been selected from the list, do not show the list again
+        // because the list has been hidden by onEntrySelected.
```
ltshb, line 27: Why changing the old comment ?

src/modules/menu/components/search/SearchBar.vue
```diff
@@ -20,16 +20,21 @@ const searchQuery = computed(() => store.state.search.query)
...
+        if (!selectedEntry.value) {
+            // if (!selectedEntry.value) {
+            log.debug(
+                `Search has result changed to ${newValue}, change the show result to ${newValue}`
+            )
+            showResults.value = newValue
+        }
+    },
+    { immediate: true }
```
ltshb, line 36: Why do you need the immediate ? would be good to comment it

src/modules/menu/components/search/SearchBar.vue
sommerfe: because i removed the && !isPristine.value from the if statement so now if it is pristine the results are shown
this is to show the results when the url has swisssearch in it

src/modules/menu/components/search/SearchBar.vue
sommerfe: i added a comment

src/router/storeSync/abstractParamConfig.class.js
```diff
@@ -201,4 +205,25 @@ export default class AbstractParamConfig {
...
+            const promiseAfterSetValuesInStore = this.afterSetValuesInStore()
+            if (promiseAfterSetValuesInStore) {
+                promiseAfterSetValuesInStore.then(() => {
+                    resolve()
+                })
+            } else {
+                resolve()
+            }
+        })
+    }
```
pakb, line 228: you could maybe simplify the complexity by turning this into an `async` function

```
async afterPopulateStore() {
    if (!this.afterSetValuesInStore) {
        return
    }
    return await this.afterSetValuesInStore()
}
```
(might need some debug, I just tried transforming this on the fly without running it)

src/router/storeSync/SearchParamConfig.class.js
```diff
@@ -18,13 +19,37 @@ function dispatchSearchFromUrl(to, store, urlParamValue) {
...
 }
 
+/**
+ * This will remove the query param from the URL It is necessary to do this in vanilla JS because
+ * the router does not provide a way to remove a query without reloading the page which then removes
+ * the value from the store.
+ *
+ * @param {Object} key The key to remove from the URL
+ */
+function removeQueryParamFromHref(key) {
```
pakb, line 29: What if we start the app with a hash? such as `http://localhost:8080/#/map?swissearch=test`
Will it work too?

sommerfe: yes it still works

src/router/storeSync/abstractParamConfig.class.js
```diff
@@ -27,16 +27,19 @@ export default class AbstractParamConfig {
...
      * @param {Boolean | Number | String | null} defaultValue
+     * @param {Function} afterSetValuesInStore A function that will be called after the store values
+     *   have been set
      */
     constructor({
         urlParamName,
         mutationsToWatch,
         setValuesInStore,
-        extractValueFromStore,
+        extractValueFromStore = (_) => '',
```
pakb, line 37: might be better to use the default value, meaning `null` as return value here

pakb (approved):

src/router/storeSync/abstractParamConfig.class.js
pakb: I meant something like `extractValueFromStore: () => null,`

sommerfe: i wonder if it is necessary, i think just making the function undefined should work also

PR #1086, opened 2024-10-02
PB-722 : Cesium rework

pakb:
Rewriting the whole thing into Composition API, and fixing / improving many things in the process

I tried to apply the same logic as what we did with all OpenLayers component, naming them as their OL equivalents.
The only thing I found hard to do the same was camera and click/interaction management. As we have to wait for some condition to be met before initializing (Mercator) adding these two as composable was not ideal, as they also had to have this logic. So instead I've created components that are added when the viewer is created.

I took the opportunity to remove our shallow copy of `ol-cesium` that was there to translate GeoJSON style into Cesium equivalent, and wrote something tailor made for us. I tested with Meteoschweiz layers, don't hesitate to try all the GeoJSON you know!

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-722-cesium-rework/index.html)


pakb:
replaced by many little smaller PRs

PR #1088, opened 2024-10-04
Enforced triple eq in condition statements

schtibe:
Enable the eslint rule [eqeqeq](https://eslint.org/docs/v8.x/rules/eqeqeq) which enforces the use of triple equals (`===` and `!=`).

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-eqeqeq-rule-in-eslint/index.html)


pakb (commented):
Some test fail, maybe some values were set to `'undefined'` (string type) and now are filtered out 🤔

pakb (approved):

PR #1090, opened 2024-10-08
PB-984 : only accept array of warnings/errors when dispatching

pakb:
It was confusing to be able to send a single error/warning while having to use the name in plural when dispatching. Only accepting array of both now (and fixing some places where it was wrongly dispatched before)

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-984-couple-fixes-with-warnings/index.html)


ltkum (approved):
I don't know why your test fails here, as it seems it's not linked to the code. What I see looks good. Try to check if the tests are flaky or if there is a side effect somewhere, but it doesn't look to me like it is linked to one another.

pakb:
Forgot one place (was a bit harder to detect with search tools)

PR #1091, opened 2024-10-08
PB-1023: enable search inside kml

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1023-enable-search-inside-kml/index.html)

### DIFF
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -1,3 +1,4 @@
+import bbox from '@turf/bbox'
 import center from '@turf/center'
 import { points } from '@turf/helpers'
 import axios from 'axios'
@@ -8,10 +9,11 @@ import { getServiceSearchBaseUrl } from '@/config/baseUrl.config'
 import i18n from '@/modules/i18n'
 import CoordinateSystem from '@/utils/coordinates/CoordinateSystem.class'
 import { LV95, WGS84 } from '@/utils/coordinates/coordinateSystems'
+import { normalizeExtent } from '@/utils/coordinates/coordinateUtils'
 import CustomCoordinateSystem from '@/utils/coordinates/CustomCoordinateSystem.class'
 import LV95CoordinateSystem from '@/utils/coordinates/LV95CoordinateSystem.class'
-import { getGpxExtent, parseGpx } from '@/utils/gpxUtils'
-import { getKmlExtent, parseKml } from '@/utils/kmlUtils'
+import { parseGpx } from '@/utils/gpxUtils'
+import { parseKml } from '@/utils/kmlUtils'
 import log from '@/utils/logging'
 
 import LayerTypes from './layers/LayerTypes.enum'
@@ -324,10 +326,8 @@ async function searchKmlLayerFeatures(outputProjection, queryString, layer) {
         )
         if (!includedFeatures.length) return []
 
-        const extent = getKmlExtent(layer.kmlData)
-
         return includedFeatures.map((feature) =>
-            createSearchResultFromLayer(layer, feature, extent, outputProjection)
+            createSearchResultFromLayer(layer, feature, outputProjection)
         )
     } catch (error) {
         log.error(
@@ -357,10 +357,8 @@ async function searchGpxLayerFeatures(outputProjection, queryString, layer) {
         )
         if (!includedFeatures.length) return []
 
-        const extent = getGpxExtent(layer.gpxData)
-
         return includedFeatures.map((feature) =>
-            createSearchResultFromLayer(layer, feature, extent, outputProjection)
+            createSearchResultFromLayer(layer, feature, outputProjection)
         )
     } catch (error) {
         log.error(
@@ -380,13 +378,15 @@ async function searchGpxLayerFeatures(outputProjection, queryString, layer) {
  * @param {CoordinateSystem} outputProjection
  * @returns {SearchResult}
  */
-function createSearchResultFromLayer(layer, feature, extent, outputProjection) {
+function createSearchResultFromLayer(layer, feature, outputProjection) {
     const featureName = feature.values_.name || layer.name || ''
     const coordinates = extractOlFeatureCoordinates(feature)
     const zoom = outputProjection.get1_25000ZoomLevel()
-    // TODO is this the correct way we should handle this? It works fine for polygons but for lines it might be a bit off
-    const point = points(coordinates)
-    const centerPoint = center(point)
+
+    const coordinatePoints = points(coordinates)
+    const centerPoint = center(coordinatePoints)
+    const normalExtent = normalizeExtent(bbox(coordinatePoints))
+
     const featureId = feature.id_ ?? feature.ol_uid
     const layerContent = {
         resultType: SearchResultTypes.LAYER,
@@ -404,7 +404,7 @@ function createSearchResultFromLayer(layer, feature, extent, outputProjection) {
         description: feature.values_.description ?? '',
         featureId: featureId,
         coordinate: centerPoint.geometry.coordinates,
-        extent,
+        extent: normalExtent,
         zoom,
     }
     return {
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -286,22 +286,22 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
 }
 
 /**
- * Searches for the query string in the layer inside the provided search fields
+ * Searches for the query string in the feature inside the provided search fields
  *
- * @param layer
+ * @param feature
  * @param {String} queryString
  * @param {String} searchFields
  * @returns {Boolean}
  */
-function isQueryInLayer(layer, queryString, searchFields) {
+function isQueryInFeature(feature, queryString, searchFields) {
     const queryStringClean = queryString
         .trim()
         .toLowerCase()
         .normalize('NFD')
         .replace(/[\u0300-\u036f]/g, '')
     return searchFields.some((field) => {
-        const value = layer[field]?.toString()
-        return value && value.trim().toLowerCase().includes(queryStringClean)
+        const value = feature.values_[field]?.toString()
+        return !!value && value.trim().toLowerCase().includes(queryStringClean)
     })
 }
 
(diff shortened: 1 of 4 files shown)


src/api/search.api.js
```diff
@@ -272,6 +279,137 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * @param feature
+ * @param {ol/extent|null} extent
+ * @param {CoordinateSystem} outputProjection
+ * @returns {SearchResult}
+ */
+function createSearchResultFromLayer(layer, feature, extent, outputProjection) {
+    const featureName = feature.values_.name || layer.name || ''
+    const coordinates = extractOlFeatureCoordinates(feature)
+    const zoom = outputProjection.get1_25000ZoomLevel()
+    // TODO is this the correct way we should handle this? It works fine for polygons but for lines it might be a bit off
```
sommerfe, line 381: Is this the correct way of handling finding the point on which the search result should zoom to ?

pakb: I would say : if there's an "valid" extent (meaning it has different bottomLeft/topRight coordinates) you could run a `zoomToExtent` action on that. It is otherwise a good choice to go with the current projection\s 1:25'000 zoom level 👍

src/store/modules/search.store.js
```diff
@@ -174,18 +181,37 @@ const actions = {
...
+                    } else {
+                        // For imported KML and GPX files
+                        let features = []
+                        if (entry.layer.type === LayerTypes.KML) {
+                            features = parseKml(entry.layer, rootState.position.projection, [])
+                        }
+                        if (entry.layer.type === LayerTypes.GPX) {
+                            features = parseGpx(entry.layer, rootState.position.projection, [])
+                        }
+                        // TODO see if there is a better way of finding the feature
```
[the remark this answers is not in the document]
pakb: as you are in the store module, and we try to avoid at all cost to link the map to the store : no I don't think so.
Or you would have to do that somewhere else where you have access to the map and the store (meaning in a component), but I bet it makes more sense to have that alongside all the other search types.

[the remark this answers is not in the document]
sommerfe: Ok thanks then i will leave it like this

src/api/search.api.js
```diff
@@ -272,6 +281,135 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * @param {CoordinateSystem} outputProjection
+ * @param {String} queryString
+ * @param layer
+ * @returns {SearchResult[]}
+ */
+async function searchKmlLayerFeatures(outputProjection, queryString, layer) {
+    try {
+        const searchFields = ['name', 'description', 'id']
+        const features = parseKml(layer, outputProjection, [])
+        if (!features || !features.length) return []
```
ltshb, line 316: See coding guidelines https://github.com/geoadmin/doc-guidelines/blob/master/JAVASCRIPT.md#4-single-statement

src/api/search.api.js
```diff
@@ -272,6 +281,135 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+
+        const includedFeatures = features.filter((feature) =>
+            isQueryInFeature(feature, queryString, searchFields)
+        )
+        if (!includedFeatures.length) return []
```
ltshb, line 321: https://github.com/geoadmin/doc-guidelines/blob/master/JAVASCRIPT.md#4-single-statement

src/api/search.api.js
```diff
@@ -272,6 +281,135 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * @param {CoordinateSystem} outputProjection
+ * @param {String} queryString
+ * @param layer
+ * @returns {SearchResult[]}
+ */
+async function searchGpxLayerFeatures(outputProjection, queryString, layer) {
+    try {
+        const searchFields = ['name', 'description', 'id']
+        const features = parseGpx(layer.gpxData, outputProjection, [])
+        if (!features || !features.length) return []
```
ltshb, line 347: https://github.com/geoadmin/doc-guidelines/blob/master/JAVASCRIPT.md#4-single-statement

src/api/search.api.js
```diff
@@ -272,6 +281,135 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+
+        const includedFeatures = features.filter((feature) =>
+            isQueryInFeature(feature, queryString, searchFields)
+        )
+        if (!includedFeatures.length) return []
```
ltshb, line 352: https://github.com/geoadmin/doc-guidelines/blob/master/JAVASCRIPT.md#4-single-statement

src/api/search.api.js
```diff
@@ -1,13 +1,22 @@
...
 import CoordinateSystem from '@/utils/coordinates/CoordinateSystem.class'
 import { LV95, WGS84 } from '@/utils/coordinates/coordinateSystems'
+import { normalizeExtent } from '@/utils/coordinates/coordinateUtils'
 import CustomCoordinateSystem from '@/utils/coordinates/CustomCoordinateSystem.class'
 import LV95CoordinateSystem from '@/utils/coordinates/LV95CoordinateSystem.class'
+import { parseGpx } from '@/utils/gpxUtils'
+import { parseKml } from '@/utils/kmlUtils'
 import log from '@/utils/logging'
 
+import LayerTypes from './layers/LayerTypes.enum'
```
pakb, line 18: can you change that to a non-relative import?

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * @param {String} queryString
+ * @param {String} searchFields
+ * @returns {Boolean}
+ */
+function isQueryInFeature(feature, queryString, searchFields) {
+    const queryStringClean = queryString
+        .trim()
+        .toLowerCase()
+        .normalize('NFD')
+        .replace(/[\u0300-\u036f]/g, '')
```
pakb, line 297: what are these two specific chars? (this needs a little comment explaining what you are doing 😉 )

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
     }
 }
 
+/**
+ * Searches for the query string in the feature inside the provided search fields
+ *
+ * @param feature
```
pakb, line 287: can you give this param a type?

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+        return !!value && value.trim().toLowerCase().includes(queryStringClean)
+    })
+}
+
+/**
+ * Searches for the query string in the KML layer
+ *
+ * @param {CoordinateSystem} outputProjection
+ * @param {String} queryString
+ * @param layer
```
pakb, line 309: this also need some type

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+        return []
+    }
+}
+
+/**
+ * Searches for the query string in the GPX layer
+ *
+ * @param {CoordinateSystem} outputProjection
+ * @param {String} queryString
+ * @param layer
```
pakb, line 344: same

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * Searches for the query string in the KML layer
+ *
+ * @param {CoordinateSystem} outputProjection
+ * @param {String} queryString
+ * @param layer
+ * @returns {SearchResult[]}
+ */
+async function searchKmlLayerFeatures(outputProjection, queryString, layer) {
+    try {
+        const searchFields = ['name', 'description', 'id']
```
pakb, line 314: as these are the same for GPX, you could declare them as a const at the beginning of the file (and this way share the const)

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+        return includedFeatures.map((feature) =>
+            createSearchResultFromLayer(layer, feature, outputProjection)
+        )
+    } catch (error) {
+        log.error(
+            `Failed to search layer features for layer ${layer.id}, fallback to empty result`,
+            error
+        )
+        return []
+    }
```
pakb, line 369: everything under this point is the same for KML, please factorize so that we do not copy paste code parts.

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+}
+
+/**
+ * Creates the SearchResult for a layer
+ *
+ * @param layer
+ * @param feature
```
pakb, line 376: need some types

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+ * @param {ol/extent|null} extent
+ * @param {CoordinateSystem} outputProjection
+ * @returns {SearchResult}
+ */
+function createSearchResultFromLayer(layer, feature, outputProjection) {
+    const featureName = feature.values_.name || layer.name || ''
```
pakb, line 382: you should prefer the use of `??` for variable declaration instead of `||` (which could lead to funny thing with boolean type, `??` is specifically made for "assign if value before is invalid")

src/api/search.api.js
```diff
@@ -272,6 +281,141 @@ async function searchLayerFeatures(outputProjection, queryString, layer, lang, c
...
+    const coordinates = extractOlFeatureCoordinates(feature)
+    const zoom = outputProjection.get1_25000ZoomLevel()
+
+    const coordinatePoints = points(coordinates)
+    const centerPoint = center(coordinatePoints)
+    const normalExtent = normalizeExtent(bbox(coordinatePoints))
+
+    const featureId = feature.id_ ?? feature.ol_uid
```
pakb, line 390: I think you could also try to run `feature.get('id')` (it might be cleaner than using private `feature.id_`). But that presuppose `feature` is of type `ol/Feature`

src/api/search.api.js
```diff
@@ -321,6 +465,22 @@ export default async function search(config) {
...
+        )
+    }
+
+    if (layersToSearch.some((layer) => layer.type === LayerTypes.GPX)) {
+        allRequests.push(
+            ...layersToSearch
+                .filter((layer) => layer.type === LayerTypes.GPX)
+                .map((layer) => searchGpxLayerFeatures(outputProjection, queryString, layer))
+        )
+    }
```
pakb, line 482: As both `searchKmlLayerFeatures` and `searchGpxLayerFeatures` are doing a similar job (only the feature parsing differs) you could maybe try having only one single function that can deal with both types, and then join these two IFs into one

src/api/search.api.js
sommerfe: if i use ?? but the `feature.values_.name = ''` then it will take the empty string, but in this case i want that it does not take it but instead uses layer.name so i intentionally use `||`
i can add a comment to make it clear

pakb (approved):

PR #1092, opened 2024-10-08
PB-984: avoid duplicate errors

ltkum:
Issue : sometimes, errors and warning might get added multiple times, and as they are new object instances, the set had issue ensuring unicity

Fix : we filter the errors and warnings received, ensuring we only add messages with a different message AND different parameters.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-984-ensure-unicity/index.html)

### DIFF
--- a/src/store/modules/ui.store.js
+++ b/src/store/modules/ui.store.js
@@ -389,12 +389,21 @@ export default {
         setShowDisclaimer({ commit }, { showDisclaimer, dispatcher }) {
             commit('setShowDisclaimer', { showDisclaimer, dispatcher })
         },
-        addErrors({ commit }, { errors, dispatcher }) {
+        addErrors({ commit, state }, { errors, dispatcher }) {
             if (
                 errors instanceof Array &&
                 errors.filter((error) => error instanceof ErrorMessage).length === errors.length
             ) {
-                commit('addErrors', { errors, dispatcher })
+                errors = errors.filter(
+                    (error) =>
+                        // we only add the errors that are not existing within the store
+                        [...state.errors].filter((otherError) => {
+                            error.isEquals(otherError)
+                        }).length === 0
+                )
+                if (errors.length > 0) {
+                    commit('addErrors', { errors, dispatcher })
+                }
             } else {
                 throw new Error(
                     `Error ${errors} dispatched by ${dispatcher} is not of type ErrorMessage, or not an Array of ErrorMessages`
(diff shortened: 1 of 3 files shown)


ltkum:
@sommerfe : since you're the one that found the issue, can you quickly check if you manage to reproduce it ? I might need to make a more thorough check on the params, since those are an object too.

sommerfe:
When I import a KML file i still get a warning that i have to close 4 times to make it disappear, then a different one is visible that i can close with one click so maybe there is still something missing

src/store/modules/ui.store.js
```diff
@@ -412,13 +421,22 @@ export default {
...
                     warnings.length
             ) {
-                commit('addWarnings', { warnings, dispatcher })
+                warnings = warnings.filter(
+                    (warning) =>
+                        // we only add the warnings that are not existing within the store
+                        [...state.warnings].filter((otherWarning) => {
+                            warning.isEquals(otherWarning)
+                        }).length === 0
+                )
```
sommerfe, line 436: i'm not sure that the double filter works, i think this might be better for the warnings and the errors

```suggestion
        warnings = warnings.filter(
            (warning) =>
                // we only add the warnings that are not existing within the store
                !state.warnings.some((otherWarning) => warning.isEquals(otherWarning))
        )
```

ltkum: The `.some` method will encounter the same issue as the filter, but it will be better and easier to read. I'll have to implement the check for the `params` in warnings and errors to try to get rid of this. Thanks for the feedback :)

sommerfe:
Now i don't get any warnings or errors when importing a KML file, is this correct or should there still be some warnings ?

ltkum:
> Now i don't get any warnings or errors when importing a KML file, is this correct or should there still be some warnings ?

If you import a KML through the menu, there should be no errors / warnings. (sorry for the late answer, I was sick on friday)

src/store/modules/ui.store.js
```diff
@@ -412,13 +418,22 @@ export default {
...
             ) {
-                commit('addWarnings', { warnings, dispatcher })
+                warnings = warnings.filter(
+                    (warning) =>
+                        // we only add the warnings that are not existing within the store
+                        ![...state.warnings].some((otherWarning) => {
+                            warning.isEquals(otherWarning)
+                        })
+                )
+                log.error(warnings)
```
ismailsunni, line 433: perhaps `log.warn`?

ismailsunni (approved):
LGTM, more complete than mine :D

src/store/modules/ui.store.js
ltkum: this looks like a forgotten log while developing. I'll remove it entirely. Thanks :)

PR #1093, opened 2024-10-11
PB-1028 : add a parser for each type of file we want to support

pakb:
Instead of having a big chunk of code that does the heavy lifting (loosely copied or similar in places like the load-kml and load-gpx plugin, plus the drag and drop handler) I gathered all this logic in one place, giving it the necessary options/config to handle all situation and be written only once.

There is now only one way of dealing with a KML/KMZ/GPX file from external/local source, the dedicated parser.

List of other things done while working on this :
 - changing updateLayer action to receive a layer ID instead of an index (was never used because of the hassle to get this index). I replaced uses of updateLayers when it was used for a single layer.
 - move all extent utils in... extentUtils.js
 - FileInput can now receive params for i18n placeholders in the error message
 - Fixing a missing rename with the local file disclaimer, it was not the correct tooltip shown for local files because of that
 - Adding a zoom to extent button in the layer list, if the layer provides an extent (meaning KML/GPX and soon COG)

[Test link with multiple files from Dropbox](https://sys-map.dev.bgdi.ch/preview/feat-pb-1028-import-file-rework/index.html?layers=KML%7Chttps://www.dropbox.com/scl/fi/n0wivr7h7cqgqc75lamk5/sample.kml?rlkey=1ez6goev55zc3tdo7jrk0hwmi%26st=b67gpice%26dl=0;GPX%7Chttps://www.dropbox.com/scl/fi/i6l23gya5fcy9lpsyeyyk/sample.gpx?rlkey=1c4n5y3jeqz3wza116mx3ex6t%26st=n83i7v88%26dl=0;KML%7Chttps://www.dropbox.com/scl/fi/l7tijlo5bnc1vyoowhyvk/sample.kmz?rlkey=37xynvq5g0kggdo1knn5smfol%26st=mbxhpbxq%26dl=0)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1028-import-file-rework/index.html)


src/api/files.api.js
```diff
@@ -349,10 +350,10 @@ export async function getFileFromUrl(url, options = {}) {
     const { timeout = null, responseType = null } = options
```
schtibe, line 350: It seems that getFileFromUrl is only called from the classes in ImportFile/parser. Since they all inherit FileParser, could it go there?

src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js
```diff
@@ -0,0 +1,240 @@
...
+     * @returns {Promise<ArrayBuffer>}
+     */
+    async getContentThroughServiceProxy(fileUrl) {
+        if (!this.serviceProxyConfiguration) {
+            throw new Error('No service-proxy configuration for this parser')
+        }
+        // service proxy will stream the response, so we will only receive Content-Type application/binary
+        // we need to check the content of what is returned
+        return await getContentThroughServiceProxy(fileUrl)
+    }
```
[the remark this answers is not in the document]
pakb: was a left over from previous displacement of code, I removed it

src/api/files.api.js
```diff
@@ -349,10 +350,10 @@ export async function getFileFromUrl(url, options = {}) {
...
     if (/^https?:\/\/localhost/.test(url) || isInternalUrl(url)) {
         // don't go through proxy if it is on localhost or the internal server
-        return axios.get(url, { timeout, responseType })
+        return await axios.get(url, { timeout, responseType })
```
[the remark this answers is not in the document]
pakb: my IDE was detecting that not using await was returning the wrong type of Promise, and after carefully reading the code using these await clarify what will be returned

src/api/files.api.js
pakb: We could, it was previously used in multiple unrelated places.
I might want to leave it here, as it is doesn't really have the file parser "touch", and fits quite well in the "File API" file

src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js
```diff
@@ -0,0 +1,227 @@
...
+            return this.fileContentTypes.includes(headResponse.headers.get('content-type'))
+        } catch (error) {
+            if (allowServiceProxy) {
+                log.debug(
+                    `[FileParser][${this.constructor.name}] could not access resource through HEAD request, trying through service-proxy`,
+                    fileUrl
+                )
+                return this.validateContentThroughServiceProxy(
+                    await getContentThroughServiceProxy(fileUrl)
+                )
```
ltshb, line 158: Should not also do a HEAD here on the proxy instead of a GET ? In the docu of the function you talk about only doing a HEAD but that not 100%, how can we avoid doing two GET over the proxy ?

src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js
```diff
@@ -0,0 +1,227 @@
...
+            )
+            throw new Error('Invalid parsing configuration')
+        }
+        if (this.isLocalFile(fileSource)) {
+            if (await this.isLocalFileParsingPossible(fileSource)) {
+                return this.parseLocalFile(fileSource, currentProjection)
+            }
+        } else if (await this.isOnlineFileParsingPossible(fileSource, options)) {
+            return this.parseUrl(fileSource, currentProjection, options)
+        }
```
ltshb, line 223: Here you will do two HEAD requests ?

ltshb (commented):
There is some issue with duplicate requests and also with the load bar.
- I imported the file https://bergschaaf.de/gps/2022/20220227_Fruemseltal.kmz via the import tool
  - I had 2 HEAD requests to the URL
  - I had 2 GET request over the proxy for the URL
  - After the import the load bar (red bar on top) was still active

src/modules/menu/components/advancedTools/ImportFile/parser/KMLParser.class.js
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @param {ArrayBuffer | String} fileContent
+ * @returns {boolean}
+ */
+export function isKml(fileContent) {
+    let stringValue = fileContent
+    if (fileContent instanceof ArrayBuffer) {
+        stringValue = new TextDecoder('utf-8').decode(fileContent)
+    }
+    return /^\s*(<\?xml\b[^>]*\?>)?\s*(<!--(.*?)-->\s*)*<(kml:)?kml\b[^>]*>[\s\S.]*<\/(kml:)?kml\s*>/g.test(
```
schtibe, line 22: I guess you could spare the regex hassle with simply this:
```
const parser = new DOMParser()
parser.parser.parseFromString("...", "application/xml")

Comment (ltshb) [reply src/modules/menu/components/advancedTools/ImportFile/parser/KMLParser.class.js:22]: @schtibe `<kml:kml>` is a valid kml and we do have users using it.

File: src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js (hunk 5 above)
Comment (pakb) [reply src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js:236]: that should now be fixed

File: src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js [hunk 7]
```diff
@@ -0,0 +1,227 @@
...
+     *   service-proxy). When given, no other request will be made on the file source, but this
+     *   content will be used instead.
+     * @returns {Promise<Boolean>}
+     */
+    async isOnlineFileParsingPossible(fileUrl, options = {}) {
+        const { allowServiceProxy = false, mimeType = null, loadedContent = null } = options
+
+        if (mimeType) {
+            return this.fileContentTypes.includes(mimeType)
+        }
```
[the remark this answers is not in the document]
pakb: It will now check the content too (but if MIME type is a match, it will use it)

src/modules/menu/components/advancedTools/ImportFile/parser/FileParser.class.js
pakb: Would a HEAD on the proxy work?
As I've switched to also detecting file content (instead of only relying on MIME types) I don't think it could add a ton to do an extra HEAD request. Your take?

ltshb: HEAD on proxy should work, but we need to try...

ltshb: I tested it HEAD on proxy is working as expected, returning the same headers as the source with CORS headers

ltkum (approved):
This looks good to me, and I did not catch anything that seemed problematic. Good job.

PR #1094, opened 2024-10-14
PB-1041: Fix Conflict Layer Name/ID 

ismailsunni:
I also ended up refactoring on how we retrieve a layer. LayerID is not enough since it's possible to have two layers with same layerID but different baseURL.

I am thinking also about same layerID, same baseURL, but different type of layer (e.t. WMTS vs WMS). But perhaps it's even more corner case, that can be addressed later.

Peek 2024-10-14 14-40 - layer name conflict

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1041-conflict-layer-name/index.html)

### DIFF
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -397,9 +397,9 @@ const actions = {
         { index = null, layerId = null, isExternal = null, baseUrl = null, dispatcher }
     ) {
         if (layerId) {
-            commit('removeLayersById', { layerId, dispatcher })
-        } else if (index !== null) {
             commit('removeLayersById', { layerId, isExternal, baseUrl, dispatcher })
+        } else if (index !== null) {
+            commit('removeLayerByIndex', { index, dispatcher })
         } else {
             log.error(
                 `Failed to remove layer: invalid parameter: ${index}, ${layerId}, ${dispatcher}`
--- a/src/store/modules/layers.store.js
+++ b/src/store/modules/layers.store.js
@@ -232,14 +232,14 @@ const getters = {
         (state, getters) =>
         (layerId, isExternal = null, baseUrl = null) => {
             const layer = getters.getActiveLayersById(layerId, isExternal, baseUrl)[0]
-            console.log(
-                'hasDataDisclaimer',
-                layerId,
-                isExternal,
-                baseUrl,
-                layer,
-                layer?.isExternal || (layer?.type === LayerTypes.KML && !layer?.adminId)
-            )
+            // console.log(
+            //     'hasDataDisclaimer',
+            //     layerId,
+            //     isExternal,
+            //     baseUrl,
+            //     layer,
+            //     layer?.isExternal || (layer?.type === LayerTypes.KML && !layer?.adminId)
+            // )
             return layer?.isExternal || (layer?.type === LayerTypes.KML && !layer?.adminId)
         },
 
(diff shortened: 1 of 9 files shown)


src/store/modules/layers.store.js
```diff
@@ -5,8 +5,14 @@ import { getGpxExtent } from '@/utils/gpxUtils.js'
...
 import log from '@/utils/logging'
 
-const getActiveLayersById = (state, layerId) =>
-    state.activeLayers.filter((layer) => layer.id === layerId)
+const getActiveLayersById = (state, layerId, isExternal = null, baseUrl = null) => {
+    return state.activeLayers.filter((layer) => {
+        const matchesLayerId = layer.id === layerId
+        const matchesIsExternal = isExternal === null || layer.isExternal === isExternal
+        const matchesBaseUrl = baseUrl === null || layer.baseUrl === baseUrl
+        return matchesLayerId && matchesIsExternal && matchesBaseUrl
```
ltkum, line 13: this bit of code is repeated three times across the store. It would be cleaner to turn it into a function like so : 
``` 
function matchTwoLayers(layerId, isExternal=null, baseUrl=null, layerToMatch) {
const matchesLayerId = layer.id === layerId
        const matchesIsExternal = isExternal === null || layer.isExternal === isExternal
        const matchesBaseUrl = baseUrl === null || layer.baseUrl === baseUrl
        return matchesLayerId && matchesIsExternal && matchesBaseUrl
}
```

then you call it here (line 9), in getActiveLayersById (line 178) and in removeLayerById (line 818) by checking for a false value.That way, if we need to change the way we filter layers by their various identifiers, we can do it by changing the code only in one place.

src/store/modules/layers.store.js
```diff
@@ -165,10 +171,20 @@ const getters = {
...
      * @param {string} layerId ID of the layer(s) to retrieve
+     * @param {string} isExternal If the layer must be external, not, or both
+     * @param {string} baseUrl Base URL of the layer(s) to retrieve
      * @returns {[AbstractLayer]} All active layers matching the ID
      */
-    getActiveLayersById: (state) => (layerId) =>
-        state.activeLayers.filter((layer) => layer.id === layerId),
+    getActiveLayersById:
+        (state) =>
+        (layerId, isExternal = null, baseUrl = null) => {
```
ltkum, line 180: see comment above

src/store/modules/layers.store.js
```diff
@@ -809,14 +834,21 @@ const mutations = {
...
+                (layer2Update) => {
+                    log.debug(`update layer`, layer2Update, layer)
+                    Object.assign(layer2Update, layer)
+                }
+            )
         })
     },
-    removeLayersById(state, { layerId }) {
-        state.activeLayers = state.activeLayers.filter((layer) => layer.id !== layerId)
+    removeLayersById(state, { layerId, isExternal = null, baseUrl = null }) {
```
ltkum, line 845: see comment above

src/store/modules/layers.store.js
```diff
@@ -165,10 +171,20 @@ const getters = {
      * Retrieves active layer(s) by ID
      *
      * @param {string} layerId ID of the layer(s) to retrieve
+     * @param {string} isExternal If the layer must be external, not, or both
+     * @param {string} baseUrl Base URL of the layer(s) to retrieve
      * @returns {[AbstractLayer]} All active layers matching the ID
```
ltkum, line 176: it could be a good idea to update the comment, as the method has changed a bit, even if the goal is clear.

ltkum (commented):
It seems to be working, and apart from a small change to make the code easier to maintain in the future, I have not seen anything problematic within the code. I have seen that the `getLayersById` method (line 197) has not been changed, and was curious if it could cause a potential issue in the future. To be on the safe side, I would also make it check against the 'isExternal' and 'baseUrl' parameters (set to null by default), so that every check meant to identify a layer by its `id` has the same sets of parameters to check.

ltkum (approved):
looks good to me, thanks for the work :)

PR #1095, opened 2024-10-14
PB-875: CTRL-Left click selection features for vector (KML/GPX/GeoJSON) and multiple elements

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-875-ctrl-right-click-select-multiple-features/index.html)

### DIFF
--- a/src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
+++ b/src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
@@ -1,8 +1,16 @@
 import { booleanIntersects } from '@turf/boolean-intersects'
-import { lineString, point, polygon } from '@turf/helpers'
+import { circle } from '@turf/circle'
+import {
+    geometryCollection,
+    lineString,
+    multiLineString,
+    multiPoint,
+    multiPolygon,
+    point,
+    polygon,
+} from '@turf/helpers'
 import { platformModifierKeyOnly } from 'ol/events/condition'
 import GeoJSON from 'ol/format/GeoJSON'
-import { LineString, Point, Polygon } from 'ol/geom'
 import { DragBox } from 'ol/interaction'
 import { useStore } from 'vuex'
 
@@ -98,14 +106,35 @@ export function useDragBoxSelect() {
  *   is not supported.
  */
 function fromOlGeometryToTurfGeometry(olGeometry) {
-    if (olGeometry instanceof Point) {
-        return point(olGeometry.getCoordinates())
+    if (!olGeometry || typeof olGeometry.getCoordinates !== 'function') {
+        log.error('Invalid OpenLayers geometry provided.', olGeometry)
+        return null
     }
-    if (olGeometry instanceof Polygon) {
-        return polygon(olGeometry.getCoordinates())
+
+    // Mapping OpenLayers geometry types to Turf.js functions
+    const geometryMapping = {
+        Point: point,
+        MultiPoint: multiPoint,
+        LineString: lineString,
+        MultiLineString: multiLineString,
+        Polygon: polygon,
+        MultiPolygon: multiPolygon,
+        GeometryCollection: geometryCollection,
+        Circle: function (olGeometry) {
+            const center = olGeometry.getCenter()
+            const radius = olGeometry.getRadius()
+            return circle(center, radius)
+        },
     }
-    if (olGeometry instanceof LineString) {
-        return lineString(olGeometry.getCoordinates())
+
+    const olGeometryType = olGeometry.getType()
+
+    const turfGeometryFunction = geometryMapping[olGeometryType]
+
+    if (turfGeometryFunction) {
+        return turfGeometryFunction(olGeometry.getCoordinates())
+    } else {
+        log.error('Unsupported geometry type:', olGeometryType)
+        return null
     }
-    return null
 }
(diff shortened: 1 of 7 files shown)


src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
```diff
@@ -20,16 +28,63 @@ export function useDragBoxSelect() {
     )
     dragBoxSelect.on('boxend', () => {
         const selectExtent = dragBoxSelect.getGeometry()?.getExtent()
+        if (selectExtent?.length !== 4) return
```
ltshb, line 31: coding style guidelines

src/utils/layerUtils.js
```diff
@@ -97,3 +101,36 @@ export function indexOfMaxResolution(projection, layerMaxResolution) {
...
+/**
+ * Creates a LayerFeature object from an OpenLayers feature and a layer.
+ *
+ * @param {ol.Feature} olFeature - The OpenLayers feature to convert.
+ * @param {AbstractLayer} layer - The layer associated with the feature.
+ * @returns {LayerFeature | null} The created LayerFeature object or null if the feature has no
+ *   geometry.
+ */
+export function createLayerFeature(olFeature, layer) {
+    if (!olFeature?.getGeometry()) return null
```
ltshb, line 114: coding style

src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
```diff
@@ -20,16 +28,63 @@ export function useDragBoxSelect() {
...
+                    })
+                    return olFeatures.map((feature) => ({ feature: feature, layer }))
+                }
             })
-        }
+            .filter((result) => {
+                const geometry = fromOlGeometryToTurfGeometry(result.feature.getGeometry())
+                return geometry && dragBox && booleanIntersects(dragBox, geometry)
+            })
+            .map(({ feature, layer }) => createLayerFeature(feature, layer))
```
ltshb, line 59: I'm not sure it is a good idea to put this here, it only select GeoJSON, GPX and KML features, where is the code to select other features types ? In my opinion both code should sit next to each other to improve code organization.

src/utils/layerUtils.js
```diff
@@ -97,3 +101,36 @@ export function indexOfMaxResolution(projection, layerMaxResolution) {
...
+        name:
+            olFeature.get('label') ??
+            // exception for MeteoSchweiz GeoJSONs, we use the station name instead of the ID
+            // some of their layers are
+            // - ch.meteoschweiz.messwerte-niederschlag-10min
+            // - ch.meteoschweiz.messwerte-lufttemperatur-10min
+            olFeature.get('station_name') ??
+            // GPX track feature don't have an ID but have a name !
+            olFeature.get('name') ??
+            olFeature.getId(),
```
ltshb, line 127: This code seems to be copy paste from other place(s) ? we should maybe put it into a function and use it everywhere where it is needed

src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
sommerfe: the other features are selected in click-on-map-management.plugin.js
the features of the GeoJSON, GPX and KML files are added to the ClickInfo features which are then also used in the click-on-map-management.plugin.js so i think the code is well placed here, i can change the name of the features to vectorFeatures to make it clear that those are just the vector features

src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
```diff
@@ -20,16 +28,63 @@ export function useDragBoxSelect() {
     )
     dragBoxSelect.on('boxend', () => {
         const selectExtent = dragBoxSelect.getGeometry()?.getExtent()
+        if (selectExtent?.length !== 4) return
 
-        if (selectExtent?.length === 4) {
-            store.dispatch('click', {
-                clickInfo: new ClickInfo({ coordinate: selectExtent }),
-                ...dispatcher,
+        const dragBox = polygon(dragBoxSelect.getGeometry()?.getCoordinates())
```
pakb, line 35: https://turfjs.org/docs/api/polygon I think this could break the code if `getGeomtry()` doesn't return something relevant, you might want to check that it has something valid before sending it to this TurfJS helper

src/utils/identifyOnVectorLayer.js
```diff
@@ -5,11 +5,12 @@ import pointToLineDistance from '@turf/point-to-line-distance'
 import proj4 from 'proj4'
 import { reproject } from 'reproject'
 
-import LayerFeature from '@/api/features/LayerFeature.class'
 import { WGS84 } from '@/utils/coordinates/coordinateSystems'
 import { reprojectGeoJsonData, transformIntoTurfEquivalent } from '@/utils/geoJsonUtils'
 import log from '@/utils/logging'
 
+import { createLayerFeature } from './layerUtils'
```
pakb, line 12: please use "absolute" import (`@/utils/layerUtils`)

pakb (commented):
Looking good! I've sent you a file (on Slack) where I couldn't select some features, might be a missing geometry type in the code you added

pakb (approved):

PR #1096, opened 2024-10-15
PB-768: swisssearch now centers when it's given coordinates in the URL

ltkum:
- Issue : Swisssearch coordinates in the query were not centering the map on the coordinates.
- Cause : We sometimes end up going twice through the storesync, which meant that, since the store was setting the center default param, and it differed from the query, the center param was overwritten by the default value.
- Fix : To counter this, we set the center through the search before the center parameter itself, preventing him to try to set its default value. We also stop the center from populating the store when there is a swisssearch parameter and no crosshair

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-768-swisssearch-center-in-legacy/index.html)

### DIFF
--- a/src/router/storeSync/storeSync.config.js
+++ b/src/router/storeSync/storeSync.config.js
@@ -51,6 +51,8 @@ const storeSyncConfig = [
                 'sr'
             ),
     }),
+    //
+    new SearchParamConfig(),
     // Position must be processed after the projection param,
     // otherwise the position might be wrongly reprojected at app startup when SR is not equal
     // to the default projection EPSG number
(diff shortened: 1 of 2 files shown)


src/router/storeSync/storeSync.config.js
```diff
@@ -51,6 +51,8 @@ const storeSyncConfig = [
                 'sr'
             ),
     }),
+    //
+    new SearchParamConfig(),
```
ltshb, line 55: You should add a comment about why the position of this param is important

ltshb (approved):

PR #1098, opened 2024-10-16
PB-877: add swisssearch legacy url limit attribute

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-877-add-swisssearch-legacy-url-limit-attribute/index.html)

### DIFF
--- a/src/api/search.api.js
+++ b/src/api/search.api.js
@@ -197,13 +197,14 @@ function parseLocationResult(result, outputProjection) {
     }
 }
 
-async function searchLayers(queryString, lang, cancelToken) {
+async function searchLayers(queryString, lang, cancelToken, limit = 0) {
     try {
         const layerResponse = await generateAxiosSearchRequest(
             queryString,
             lang,
             'layers',
-            cancelToken.token
+            cancelToken.token,
+            { limit }
         )
         // checking that there is something of interest to parse
         const resultWithAttrs = layerResponse?.data.results?.filter((result) => result.attrs)
@@ -222,15 +223,17 @@ async function searchLayers(queryString, lang, cancelToken) {
  * @param queryString
  * @param lang
  * @param cancelToken
+ * @param limit
  * @returns {Promise<LocationSearchResult[]>}
  */
-async function searchLocation(outputProjection, queryString, lang, cancelToken) {
+async function searchLocation(outputProjection, queryString, lang, cancelToken, limit = 0) {
     try {
         const locationResponse = await generateAxiosSearchRequest(
             queryString,
             lang,
             'locations',
-            cancelToken.token
+            cancelToken.token,
+            { limit }
         )
         // checking that there is something of interest to parse
         const resultWithAttrs = locationResponse?.data.results?.filter((result) => result.attrs)
@@ -427,10 +430,17 @@ let cancelToken = null
  * @param {String} config.lang The lang ISO code in which the search must be conducted
  * @param {GeoAdminLayer[]} [config.layersToSearch=[]] List of searchable layers for which to fire
  *   search requests. Default is `[]`
+ * @param {number} config.limit The maximum number of results to return
  * @returns {Promise<SearchResult[]>}
  */
 export default async function search(config) {
-    const { outputProjection = null, queryString = null, lang = null, layersToSearch = [] } = config
+    const {
+        outputProjection = null,
+        queryString = null,
+        lang = null,
+        layersToSearch = [],
+        limit = 0,
+    } = config
     if (!(outputProjection instanceof CoordinateSystem)) {
         const errorMessage = `A valid output projection is required to start a search request`
         log.error(errorMessage)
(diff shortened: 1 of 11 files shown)


src/store/modules/search.store.js
```diff
@@ -147,6 +169,7 @@ const actions = {
      * @param {SearchResult} entry
      */
     selectResultEntry: ({ dispatch, getters, rootState }, { entry, dispatcher }) => {
+        console.log('selectResultEntry', entry)
```
ltshb, line 172: Don't use `console.log` but `log.debug` from the log library

ltshb (commented):
I've two comprehension question that should be at least documented on the PR description:

- Why removing the `swisssearch` url parameter when it has been entered by the user ? What is the advantage of this ? If I read the ticket associated with this PR, the only issue that we have is that when a user is entering a search text in the search bar the `swisssearch` is set, so the changes should be very straight forward by not syncing store to url param, but only url param to store.
- Why adding this strange limit support and how it works ? from where does it come from ? is it documented ? I could not find anything in ticket related to this ? Based on the PR title it is a legacy attribute so it should be in the legacy url param parser or has it been decided to keep this as official ?
Also the PR title only reflect the limit attribute but not the changes asked in ticket.

ltshb:
I've just seen that this PR contains also #1084 this answer partly my questions above, you should base this PR to #1084 this would ease review and understanding, but then pay attention when merging to merge first #1084 and then rebase this one to develop before merging.

ltshb (changes requested):
See comment in ticket, I think this needs some refinement on how to do it.

src/router/storeSync/SearchAutoSelectConfig.class.js
```diff
@@ -0,0 +1,41 @@
...
+            setValuesInStore: dispatchSearchFromUrl,
+            afterSetValuesInStore: () => removeQueryParamFromHref(URL_PARAM_NAME),
+            extractValueFromStore: (store) => store.state.search.swisssearchAutoSelect,
+            keepInUrlWhenDefault: false,
+            valueType: Boolean,
+            defaultValue: false,
+            validateUrlInput: null,
+        })
+    }
+}
```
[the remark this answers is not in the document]
sommerfe: So you mean create a new class like the AbstractParamConfig which is then used as parent for the SearchAutoSelectConfig ?
I think this would just eliminate the the redundancy of the afterSetValuesInStore which is in both cases the same, but the dispatchAutoSelect function that is used is not the same so i think there is not that much of a gain by doing it, it just makes the code more complicated in my opinion but we can talk about it if you want

src/api/search.api.js
```diff
@@ -454,10 +464,11 @@ export default async function search(config) {
 
     /** @type {Promise<SearchResult[]>[]} */
     const allRequests = [
-        searchLayers(queryString, lang, cancelToken),
-        searchLocation(outputProjection, queryString, lang, cancelToken),
+        searchLayers(queryString, lang, cancelToken, limit),
+        searchLocation(outputProjection, queryString, lang, cancelToken, limit),
     ]
 
+    // TODO limit also in the local kml and gpx files ?
```
pakb, line 471: no need to do that, you can remove the TODO comment

src/router/storeSync/SearchAutoSelectConfig.class.js
```diff
@@ -0,0 +1,41 @@
+import AbstractParamConfig, {
+    STORE_DISPATCHER_ROUTER_PLUGIN,
+} from '@/router/storeSync/abstractParamConfig.class'
+import { removeQueryParamFromHref } from '@/utils/searchParamUtils'
+
+import { URL_PARAM_NAME_SWISSSEARCH } from './SearchParamConfig.class'
```
pakb, line 6: can you use non-relative import here?

src/router/storeSync/SearchAutoSelectConfig.class.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+export default class SearchAutoSelectConfig extends AbstractParamConfig {
+    constructor() {
+        super({
+            urlParamName: URL_PARAM_NAME,
+            mutationsToWatch: ['setAutoSelect'],
+            setValuesInStore: dispatchAutoSelect,
+            afterSetValuesInStore: () => removeQueryParamFromHref(URL_PARAM_NAME),
+            extractValueFromStore: (store) => store.state.search.swisssearchAutoSelect,
```
pakb, line 34: might need to be updated to `store.state.search.autoSelect` (although it shouldn't have any consequence, as we do not sync this after start up)

pakb (commented):
[This link](https://sys-map.dev.bgdi.ch/preview/feat-pb-877-add-swisssearch-legacy-url-limit-attribute/index.html?swisssearch=Seftigenstrasse%20264&swisssearch_autoselect=true) should auto select Swisstopo office but nothing comes in the search bar, there might still be some things to fix

src/router/storeSync/SearchAutoSelectConfig.class.js
sommerfe: this needs to be there else it will keep swisssearch_autoselect=false in the url

pakb (approved):

PR #1101, opened 2024-10-21
PB-1066: Fix error window geolocation can not be closed

ismailsunni:
Original issue is caused by multtiple tries of geolocation, but keep getting failed result. It will show many `Error Window` with the same value, so when you close only 1 or 2, it seems the error window is still there.

I am fixing this by adding the number of similar error in the error window title. When you click the close button, it will remove the similar error altogether.

image

It can be discussed though. Another alternatives:

1. Put another button to close all the similar error window. So we have two button, one will close one error and one will close all similar error. Drawback: it might be confusing and polluting the UI.
2. Add another action to close all the window, for example right-click. It will need a tooltip for it
3. Remove the count in the error window altogether. We will always close all similar error window. It might be a good solution, but it failed to show that the application requesting the geolocation several times

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -78,8 +78,7 @@ function refreshPageTitle() {
             v-if="error"
             title="error"
             :count="errorCount"
-            @close="store.dispatch('removeError', { error, ...dispatcher })"
-            @close-all="store.dispatch('removeSimilarError', { error, ...dispatcher })"
+            @close="store.dispatch('removeSimilarError', { error, ...dispatcher })"
         >
             <div>
                 {{ i18n.t(error.msg, error.params) }}
--- a/src/App.vue
+++ b/src/App.vue
@@ -27,6 +27,15 @@ const error = computed(() => {
     }
     return null
 })
+const errorCount = computed(() => {
+    let count = 0
+    store.state.ui.errors.forEach((e) => {
+        if (error.value.equals(e)) {
+            count++
+        }
+    })
+    return count
+})
 const warning = computed(() => {
     if (store.state.ui.warnings.size > 0) {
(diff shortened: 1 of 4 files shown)


ltkum:
there is a PR here (https://github.com/geoadmin/web-mapviewer/pull/1092) which goal is to stop identical errors to populate the list. The issue was caused by the equality check from the `set` not being thorough enough. That being said, it was discussed to set a number on the right to tell the user how many messages he has (even if they're not the same error)

ismailsunni:
@ltkum yes, I was thinking the same approach as yours also. But I was not sure which one should we choose. I don't mind if we go with your PR's approach since it will also fix the issue. Is it ready to be reviewed?

> That being said, it was discussed to set a number on the right to tell the user how many messages he has (even if they're not the same error)

I can add this feature (since I already implemented it in this PR) after you merge your PR.

ltkum:
> @ltkum yes, I was thinking the same approach as yours also. But I was not sure which one should we choose. I don't mind if we go with your PR's approach since it will also fix the issue. Is it ready to be reviewed?
> 
> > That being said, it was discussed to set a number on the right to tell the user how many messages he has (even if they're not the same error)
> 
> I can add this feature (since I already implemented it in this PR) after you merge your PR.

It should be ready. I'll add you to the reviewers :)

ismailsunni:
hi @ltshb since we already merge @ltkum 's PR that also solves the issue, what should I add from this PR? I was thinking about adding the number of error in the error message's title (as shown in my screencast)

ismailsunni:
similar approach already done by @ltkum on #1092 with additional bug fix I made #1112 . I am closing this PR.

PR #1104, opened 2024-10-24
PB-722 : import cesium CSS globally

pakb:
that fixes the issues we had with touch device navigation in 3D

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-722-cesium-touch/index.html)


ltshb (approved):
Looks good to me

PR #1106, opened 2024-10-28
PB-1080: show both warning and error feedback

ltkum:
- Issue : When we have both an error feedback and a warning feedback, they would be displayed one atop each other rather than in a column.
- Fix : We create a small component to hold both the errors and warning feedback.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1080-show-warnings-and-errors-simultaneously/index.html)


src/App.vue
```diff
@@ -21,17 +20,10 @@ const i18n = useI18n()
...
-    return null
-})
-const warning = computed(() => {
-    if (store.state.ui.warnings.size > 0) {
-        return store.state.ui.warnings.values().next().value
-    }
-    return null
+const showFeedbackPopup = computed(() => {
+    console.error(store.state.ui.errors)
+    console.error(store.state.ui.warnings)
```
pakb, line 25: don't use `console.error` but `log.error` instead (if this makes sense to stay long term in the code, and isn't some debugging help for your PR)

src/App.vue
```diff
@@ -21,17 +20,10 @@ const i18n = useI18n()
...
+    return store.state.ui.errors.size + store.state.ui.warnings.size > 0
```
pakb, line 26: you might want to wrap the + operation with parenthesis, to be sure the greater than won't be interpreted first and end up with a `number + boolean` operation

src/utils/components/FeedbackPopup.vue
```diff
@@ -0,0 +1,74 @@
...
+    if (store.state.ui.warnings.size > 0) {
+        return store.state.ui.warnings.values().next().value
+    }
+    return null
+})
+</script>
+
+<template>
+    <div class="feedback-window">
+        <ErrorWindow
```
pakb, line 26: might be nice to have some kind of an error/warning count if more than one

src/App.vue
ltkum: good catch, I'm fairly sure this is a debugging help :)

src/utils/components/FeedbackPopup.vue
ltkum: I think @ismailsunni had something in one of his PR. If this is not on its way to be merged, I'll add this too.

ismailsunni: Joining the discussion, in my PR, I kept adding the error to the store even if they are the same. But I show the number of same error in the component title. When user close the error, I remove the same error in the store.

So, it depends on the error count. Is it the number of ALL error or the number of the SAME error?

src/App.vue
ltkum: the linter removes the parenthesis automatically, so it's safe to assume the addition has precedence over the check.

src/utils/components/FeedbackPopup.vue
ismailsunni: btw, I just had this corner case in my mind. Let's say, there is currently error message A. The user doesn't close it. Then a new error message B pops up (covering the error message A). The user doesn't close it also. When there is a new error message A, it won't appear above the error message B (because it's not added in the store). Do you think it's ok to not show the latest error message? It might give a wrong feedback to the user.

Again, this is a corner case.

ltkum: The current behavior of the messages is to pick the first of the list, not the last. The corner case would be :
User gets message A, doesn't close it
User gets message B, doesn't see it (but a little '2' will be shown)
User gets message A again, no change visible.

We are working under the assumption that people want to close all the red things everywhere ^^. If they decide to ignore error messages, it's their own responsibility ^^

src/utils/components/ErrorWindow.vue
```diff
@@ -40,7 +42,9 @@ const emit = defineEmits(['close'])
             class="card-header d-flex align-items-center justify-content-sm-end"
             data-cy="window-header"
         >
-            <span v-if="title" class="me-auto text-truncate">{{ i18n.t(title) }}</span>
+            <span v-if="title" class="me-auto text-truncate"
+                >{{ i18n.t(title) }}({{ errorCount }})</span
```
pakb, line 46: I think it would make sense to show the error count only if there are more than one error

src/utils/components/WarningWindow.vue
```diff
@@ -40,7 +42,9 @@ const emit = defineEmits(['close'])
             class="card-header d-flex align-items-center justify-content-sm-end"
             data-cy="window-header"
         >
-            <span v-if="title" class="me-auto text-truncate">{{ i18n.t(title) }}</span>
+            <span v-if="title" class="me-auto text-truncate"
+                >{{ i18n.t(title) }}({{ warningCount }})</span
```
pakb, line 46: same here for warnings

src/utils/components/ErrorWindow.vue
```diff
@@ -40,7 +44,9 @@ const emit = defineEmits(['close'])
             class="card-header d-flex align-items-center justify-content-sm-end"
             data-cy="window-header"
         >
-            <span v-if="title" class="me-auto text-truncate">{{ i18n.t(title) }}</span>
+            <span v-if="title" class="me-auto text-truncate"
+                >{{ i18n.t(title) }}{{ errorCount }}</span
```
pakb, line 48: would prefer to have something like

```
<script>
// ...
const errorCount = computed(() => store.state.ui.errors.size)
// ...
</script>

<template>
// ...
 <span v-if="title" class="me-auto text-truncate">
    {{ i18n.t(title) }}
    <span v-if="errorCount > 1">({{ errorCount }})</span>
</span>
// ...
</template>
```

Keeping templating stuff (parenthesis around, choice to show it or not) located in the template, and not use some string construction + ternary operator to achieve this

src/utils/components/WarningWindow.vue
```diff
@@ -24,6 +24,10 @@ const store = useStore()
 const showBody = ref(true)
 const hasDevSiteWarning = computed(() => store.getters.hasDevSiteWarning)
 
+const warningCount = computed(() =>
+    store.state.ui.warnings.size > 1 ? `(${store.state.ui.warnings.size})` : ''
+)
```
pakb, line 29: same as for the error, keep templating choice in the template please

src/utils/components/ErrorWindow.vue
ltkum: yeah, it would be cleaner :)

pakb:
do you have a good example that creates more than one warning/errors?

ltkum:
> do you have a good example that creates more than one warning/errors?

usually what I do is : 


> do you have a good example that creates more than one warning/errors?

https://sys-map.dev.bgdi.ch/preview/fix-pb-1080-show-warnings-and-errors-simultaneously/index.html#/map?lang=en&center=2660000,1190000&z=1&bgLayer=ch.swisstopo.pixelkarte-farbe&topic=ech&layers=KML|fakeLocalHost1.kml;KML|AnotherFakeLocalhost.kml

pakb (approved):

ismailsunni (approved):

PR #1107, opened 2024-10-30
PB-603: fix hover selected vertex and feature

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-603-balloon-drawing-drag-mouse-hover/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingTooltip.vue
+++ b/src/modules/drawing/components/DrawingTooltip.vue
@@ -86,66 +86,62 @@ function onPointerMove(event) {
     let featureDrawingMode = featureUnderCursor?.get('type').toUpperCase()
     let translationKeys
 
-    if (hoveringSelectedFeature) {
-        mapElement.classList.add(cssGrab)
-        mapElement.classList.remove(cssPointer)
+    if (drawingMode.value) {
+        if (this.currentlySketchedFeature && !pointFeatureTypes.includes(drawingMode.value)) {
+            let hoveringFirstVertex = false
+            let hoveringLastVertex = false
+            // The last two coordinates seem to be some OL internal points we don't need.
+            let coordinates = getVertexCoordinates(this.currentlySketchedFeature).slice(0, -2)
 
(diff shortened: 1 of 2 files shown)


src/modules/drawing/components/DrawingTooltip.vue
```diff
@@ -86,28 +86,29 @@ function onPointerMove(event) {
     let featureDrawingMode = featureUnderCursor?.get('type').toUpperCase()
     let translationKeys
 
-    if (hoveringSelectedFeature) {
-        mapElement.classList.add(cssGrab)
-        mapElement.classList.remove(cssPointer)
-
+    if (hoveringSelectedFeature && !drawingMode.value) {
```
pakb, line 89: Could you try changing a bit the order of if/else if/if ? I feel like having `if (drawingMode.value)` first, then `else if(hoveringSelectedFeature)` and finally `else if (hoveringSelectableFeature)`, would clarify a bit more what's happening. (but there's a little `else` block to take care of, don't know if that's possible)

src/modules/drawing/components/DrawingTooltip.vue
```diff
@@ -86,28 +86,29 @@ function onPointerMove(event) {
...
         let hoveringVertex = getVertexCoordinates(featureUnderCursor).some((coordinate) => {
             let pixel = olMap.getPixelFromCoordinate(coordinate)
             return pointWithinTolerance(pixel, event.pixel, DRAWING_HIT_TOLERANCE)
         })
+        translationKeys = `select_feature_${featureDrawingMode}`
 
-        // Display a help tooltip when modifying
-        if (hoveringVertex || pointFeatureTypes.includes(featureDrawingMode)) {
```
pakb, line 99: What was the problem of detecting that it was a point type here?

sommerfe: The problem is that even if you don't hover the vertex itself but rather the balloon then the tippy of the vertex shows

pakb (approved):

PR #1108, opened 2024-11-01
PB-1116 : move drawing name to store

pakb:
there was some bugs by keeping it only in the toolbox. For instance deleting the drawing and adding new features would not give a correct drawing name.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1116-drawing-name-fix/index.html)

### DIFF
--- a/src/modules/drawing/components/DrawingToolbox.vue
+++ b/src/modules/drawing/components/DrawingToolbox.vue
@@ -7,14 +7,14 @@ import { useStore } from 'vuex'
 
 import { EditableFeatureTypes } from '@/api/features/EditableFeature.class'
 import DrawingExporter from '@/modules/drawing/components/DrawingExporter.vue'
+import DrawingHeader from '@/modules/drawing/components/DrawingHeader.vue'
 import DrawingToolboxButton from '@/modules/drawing/components/DrawingToolboxButton.vue'
 import SharePopup from '@/modules/drawing/components/SharePopup.vue'
 import { DrawingState } from '@/modules/drawing/lib/export-utils'
 import useSaveKmlOnChange from '@/modules/drawing/useKmlDataManagement.composable'
 import ModalWithBackdrop from '@/utils/components/ModalWithBackdrop.vue'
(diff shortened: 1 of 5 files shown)


src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -78,15 +79,11 @@ onMounted(() => {
...
-watch(drawingName, () => {
-    sanitizeDrawingName()
-    debounceSaveDrawing({
-        // Lowering the risk of the user changing the name and leaving the drawing module without the name being saved.
-        // It is terrible design that this can occur, but moving the drawing name management other places raises the complexity tenfold
-        // so that is a trade-off we can live with.
-        debounceTime: 500,
-        drawingName: drawingName.value.trim(),
-    })
+watch(drawingName, (newName) => {
```
schtibe, line 82: I was once told that one should avoid to have too many `watch`. The reasoning was some performance issues, I don't know if that's true, but maybe it's also because of unintended side effects or something?

Anyway, personally I think this here could be solved like this:
```js
const editDrawing = computed({
  get() {
    return store.state.drawing.name
  },
  set(value) {
    debounceSaveDrawingName(value)
  }
})

```

I guess one argument could be that like this you have it in one location and don't have to copy values around?

schtibe (commented):
Just some small comment about code organization. But I'm not sure if it works 100% correctly:

- I open the drawing drawer (hehe)
- Place a marker on the map
- Change the name of the drawing
- Go back (Back / finish drawing)
- The Loading indicator at the top keeps playing KITT from the Knight Rider

Now going one step further

- Go to the Drawing again
- Delete it 
- Again place something on the map
- Go back / Finish drawing 
- Now the layer has an error indicator with "KML/GPX file is empty"

src/modules/drawing/components/DrawingToolbox.vue
schtibe: Very nice!

schtibe (approved):
👍

PR #1113, opened 2024-11-04
PB-1046: Update warning message.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1046-update-warning-message/index.html)


hansmannj (approved):
Thanks! :rocket:

IMHO it's ok to not reference the icon set name in the warning. If the user asks helpdeks, we will have a link to the KML anyways and can check the icon set in use there.

One question: Will this warning only be thrown when trying to edit a KML with legacy icon sets, or already when opening it in the "non-editing" mode?

In the latter case we might need to update the text of the warning message unfortunately again. 

Eg to a more generic version:
"Warning, your drawing contains icons, that are no longer available. If you replace ..."

ismailsunni:
it's when I open the draw and measure menu, as below:
image

to open, use this [link](https://sys-map.dev.bgdi.ch/preview/pb-1046-update-warning-message/index.html#/map?lang=en&center=2660000,1190000&z=1&bgLayer=ch.swisstopo.pixelkarte-farbe&topic=ech&layers=KML%7Chttps://sys-public.dev.bgdi.ch/api/kml/files/Kzc2hHunQ_GfAzhdS2nf4Q&featureInfo=default).

I believe it's already in edit mode, but no feature selected.

Do you think we need to update the warning message, or not @hansmannj ?

PR #1114, opened 2024-11-05
PB-980: use single OpenLayersWMSLayer for WMS group to reduce the number of GetMap request.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-980-single-request-getmap/index.html)


ltkum (commented):
This work well for reducing the number of getMap queries, however, I could see a difference between the layer as it would be shown on prod, and the result with your PR :
With your PR : 
image

In prod :
image

We can see the opacity is not the same over some part of the map (Seefeld for example). It is possible that we lose some information when using only one request.

If you could investigate for a bit about the cause of this difference, this would be great. Otherwise, it seems to work fine, good job :)

ismailsunni:
hi @ltkum
thanks for pointing this one. After an investigation, it seems, it's a bug in the prod. If you check the opacity, it should be 0.75 as shown in this PR instance. In the prod instance, it's not.

If you check again, in prod, you can't update the transparency. It always show the same transparency. Meanwhile, in this PR instance, when you update the transparency, the layer transparency is also changing.

So, I conclude that this PR also fixes that transparency bug.

src/modules/map/components/openlayers/OpenLayersInternalLayer.vue
```diff
@@ -89,13 +90,23 @@ function shouldAggregateSubLayerBeVisible(subLayer) {
...
         <div v-if="layerConfig.type === LayerTypes.GROUP">
-            <OpenLayersInternalLayer
-                v-for="(layer, index) in layerConfig.layers"
-                :key="`${layer.id}-${index}`"
-                :layer-config="layer"
-                :parent-layer-opacity="layerConfig.opacity"
-                :z-index="zIndex + index"
-            />
+            <!-- If the layer is a group but it is a WMS layer, we use OpenLayersWMSLayer directly so the WMS request is handled on the layer group level, reducing the number of request -->
+            <div v-if="layerConfig instanceof ExternalWMSLayer">
```
pakb, line 94: that won't catch anything, as if above we check for `layerConfig.type === LayerTypes.GROUP`, this will always be an instance of `ExternalGroupOfLayers`

ismailsunni: @pakb I compared it with `ExternalWMSLayer` not `ExternalGroupOfLayers`, which are different. Or, am I missing something here?

ismailsunni:
per my discussion with @pakb , we will try to remove `ExternalGroupOfLayers` class and update `ExternalWMSLayer` to have a children

src/api/layers/ExternalWMSLayer.class.js
```diff
@@ -47,6 +47,9 @@ export default class ExternalWMSLayer extends ExternalLayer {
      *   Default is `true`
      * @param {String} externalWmsData.baseUrl Base URL to build WMS requests (no endpoint / URL
      *   param defined)
+     * @param {ExternalWMSLayer[]} externalLayerData.layers Description of the layers being part of
+     *   this group (they will all be displayed at the same time, in contrast to an aggregate
```
pakb, line 51: maybe rename this comment from `this group` to `this WMS layer` or something like that

src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -245,21 +243,27 @@ export default class WMSCapabilitiesParser {
...
                 legends,
                 isLoading: false,
                 availableProjections,
+                hasTooltip: queryable,
                 getFeatureInfoCapability: this.getFeatureInfoCapability(ignoreError),
                 currentYear,
+                customAttributes: params,
+                dimensions: dimensions,
+                timeConfig: this._getTimeConfig(layerId, dimensions),
             })
```
pakb, line 267: this is exactly the same call to the constructor (plus the `layers` obviously) that outside the `if` block, might be better to add `layers` to the one outside and remove this return altogether

tests/cypress/tests-e2e/importToolMaps.cy.js
```diff
@@ -272,16 +272,8 @@ describe('The Import Maps Tool', () => {
...
-        // once as single layer
-        cy.checkOlLayer([
-            bgLayer,
-            `${itemId}-1`,
-            `${itemId}-2`,
-            `${itemId}-3`,
-            `${itemId}-1`,
-            singleLayerId,
-        ])
+        // NOTE here below itemId-1 should be present once as a single layer only
```
pakb, line 275: we can remove this comment, it doesn't bring much (was relevant when a layer was twice in the ULR, but now...)

pakb (approved):

PR #1119, opened 2024-11-07
PB-1130: fix selecting search result update search query

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1130-select-search-result/index.html)

### DIFF
--- a/src/store/modules/search.store.js
+++ b/src/store/modules/search.store.js
@@ -235,6 +235,7 @@ const actions = {
 
                 break
         }
+        commit('setSearchQuery', { query: entry.sanitizedTitle, dispatcher })
     },
 }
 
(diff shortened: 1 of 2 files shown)


ismailsunni (approved):
LGTM

pakb (changes requested):
No, don't do that!
the goal was to keep what the user searched before, and could go back to the results without having to clear the search bar and type again

sommerfe:
> No, don't do that! the goal was to keep what the user searched before, and could go back to the results without having to clear the search bar and type again
> 
> That's something I've intentionaly fixed/added while working on the some Search related PR in the last couple weeks

oh ok good to know then i will close the PR and issue

PR #1121, opened 2024-11-08
PB-580: Add warning when drawing could get lost otherwise

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-580-add-warning-leaving-drawing-mode/index.html)

### DIFF
--- a/src/modules/i18n/locales/en.json
+++ b/src/modules/i18n/locales/en.json
@@ -173,6 +173,7 @@
     "draw_type_marker": "Line / surface",
     "drawing_attached": "Drawing added as attachment",
     "drawing_empty_cannot_edit_name": "Please add something to the drawing before editing its name.",
+    "drawing_not_shared_admin_warning": "Warning, you have not copied/saved the link enabling you to edit your drawing at a later date. You risk not being able to edit your drawing if you reload or close the page.",
     "drawing_too_large": "Your drawing is too large, remove some features",
     "drop_invalid_url": "URL is not valid.",
     "drop_me_here": "Drop file here (KML, KMZ, GPX, COG)",
--- a/src/modules/drawing/components/DrawingHeader.vue
+++ b/src/modules/drawing/components/DrawingHeader.vue
@@ -1,15 +1,24 @@
 <script setup>
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
-import { computed, ref } from 'vue'
+import { computed, ref, toRefs } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
 import TextTruncate from '@/utils/components/TextTruncate.vue'
 
+const props = defineProps({
+    isClosingInToolbox: {
+        type: Boolean,
+        default: false,
+    },
+})
+const { isClosingInToolbox } = toRefs(props)
+
 const emits = defineEmits(['close'])
 const store = useStore()
 
 const isClosing = ref(false)
+const isClosingWithWarning = computed(() => isClosing.value && isClosingInToolbox.value)
 const drawingTitle = computed(() => store.state.drawing.drawingOverlay.title)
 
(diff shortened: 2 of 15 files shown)


ltshb (commented):
Implementation looks just, just an issue regarding the store import.

Otherwise I think the warning message should be improved. @hansmannj 
 @Luke252 a warning about non exported drawing might be a good idea, but in my opinion the warning should say why we should export it or copy the share link. We should also differentiate if the drawing has only be shared via normal link without admin link.  So if the user don't click on export and/or copy admin link the warning should say that the drawing cannot be edited anymore once the webpage is reloaded !

pakb (commented):
Implementation looks good, but I'm against this message.
It will confuse more than help IMHO.

ltshb (changes requested):
I'm a bit lost here. The PR title says to add a warning when closing the drawing (as the code seems to do) but the warning is not added ? More over based on the previous comments on this PR we should not do it, if it has been decided to do it, it MUST BE documented in the PR. Finally looking at the ticket the changes required seems to have differ but are not 100% clear to me, @hansmannj @Luke252 What has been here decided ? We SHOULD document such decision in ticket. But I'm still quite against such warning, or it should provide a much better info to the user as simply "Attention, your drawing has not been shared.", because such warning is not useful for the purpose of the ticket. I mean maybe the user don't want to share it, but simply recover later on.

src/modules/i18n/locales/en.json
```diff
@@ -172,6 +172,7 @@
     "draw_type_marker": "Line / surface",
     "drawing_attached": "Drawing added as attachment",
     "drawing_empty_cannot_edit_name": "Please add something to the drawing before editing its name.",
+    "drawing_not_shared_admin_warning": "Attention, you have not generated a link to continue to edit your drawing. You won't be able to edit your drawing after closing this window.",
```
ltshb, line 175: @Luke252 @hansmannj For all translation I would say `after reloading the website` instead of `closing this windows` (or something like this), because it is more correct, you can close the drawing and re-open it to get the admin link as long as you don't reload the page.

ltshb (commented):
Found some issues while testing:
1. No warning set for subsequent drawing, here how to reproduce
  a. Create a drawing and save the admin link
  b. Close the drawing: no warning as expected because the admin link has been copied
  c. Delete the drawing
  d. Create a new drawing without copying the new admin link
  e. Close the new drawing -> no warning ! In this case the warning should come.
2. When copying the share link from the drawing, the crosshair param is added to the url (on current dev staging is not the case)

hansmannj:
> Found some issues while testing:
> 
> 1. No warning set for subsequent drawing, here how to reproduce
>    a. Create a drawing and save the admin link
>    b. Close the drawing: no warning as expected because the admin link has been copied
>    c. Delete the drawing
>    d. Create a new drawing without copying the new admin link
>    e. Close the new drawing -> no warning ! In this case the warning should come.
> 2. When copying the share link from the drawing, the crosshair param is added to the url (on current dev staging is not the case)

Good catch! :+1:

hansmannj:
Don't we also want to catch the case, where neither share nor admin link are copied and throw a corresponding warning then?

hansmannj:
> Don't we also want to catch the case, where neither share nor admin link are copied and throw a corresponding warning then?

Ah, ok, that seem's to be covered already, sorry!

src/modules/i18n/locales/en.json
hansmannj: OK, my suggestion from above does not make much sense, since the admin link is already "thrown at the user" in this case, which makes sense.
So, yes, `after reloading the website` is fine for me!

ltshb (commented):
You still have an issue when opening the share link of the drawing it adds the cross marker

image

Other issues seems fine now

ltshb (approved):
Thanks

PR #1122, opened 2024-11-08
PB-1121: compare slider support for COGtiffs

ltkum:
Issue : COGTiffs are rendered as a webGL layer, which has a different graphical context than other OpenLayers layers. This causes the compare slider to not work on them.

Fix : when we deal with a COGtiff layer, we instead use the webgl context tools to make the slice.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1121-support-compare-slider-for-cogtiffs/index.html)

### DIFF
--- a/src/modules/map/components/CompareSlider.vue
+++ b/src/modules/map/components/CompareSlider.vue
@@ -1,27 +1,21 @@
 <script setup>
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
-import { unByKey } from 'ol/Observable'
-import { computed, inject, nextTick, onMounted, onUnmounted, ref, watch } from 'vue'
+import { getRenderPixel } from 'ol/render'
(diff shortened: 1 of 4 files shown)


pakb:
If you load (import) https://data.geo.admin.ch/ch.swisstopo.rapidmapping/data/2024-012-MISOX/RM_2024_MISOX_DOPmosaic_25cm_20240628_RGB8BIT_LV95_COG.tif while the compare slider is active, the newly created layer isn't cut.
Can you check that that way of "using" the compare also works as expected?

ltkum:
> If you load (import) https://data.geo.admin.ch/ch.swisstopo.rapidmapping/data/2024-012-MISOX/RM_2024_MISOX_DOPmosaic_25cm_20240628_RGB8BIT_LV95_COG.tif while the compare slider is active, the newly created layer isn't cut. Can you check that that way of "using" the compare also works as expected?

Sure

pakb (approved):

PR #1123, opened 2024-11-08
PB-1131: add warning when loading http resource

ltkum:
Issue: When importing an external layer through an http source, we didn't give a warning to the user.

Fix: If the resource can be loaded, and the scheme is only http, we give a warning so the user know he should use https resources.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1131-add-warning-when-loading-http-resource/index.html)


src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
```diff
@@ -76,6 +79,13 @@ async function loadFile() {
 
     try {
         await handleFileSource(fileUrl.value, false)
+        if (!fileUrl.value.match(/https:\/\//)) {
```
sommerfe, line 82: i don't think this will catch urls like http://foo.com?url=https://bar.com
maybe something like `\G/https:\/\//` could work

src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
```diff
@@ -76,6 +79,13 @@ async function loadFile() {
...
+            store.dispatch('addWarnings', {
+                warnings: [
+                    new WarningMessage('import_http_external_file_warning', { url: fileUrl.value }),
```
ismailsunni, line 85: We supply an extra param `url`, but in the translation, we don't use it. Do you think it's necessary?

src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
ltkum: ho yeah, good catch :)

src/modules/menu/components/advancedTools/ImportFile/ImportFileOnlineTab.vue
ltkum: I did the code before the translation, and forgot to remove the unused param. thanks :)

sommerfe (approved):

pakb (commented):
can you only add/edit i18n keys related to your specific work?
the next PRs will thank you for not creating a rebase hell :wink:

PR #1124, opened 2024-11-12
PB-1140: add a minimal scale for print specs

ltkum:
Issue : When sending a print spec, if the scale was too small, it would be sent to the backend in a scientific notation (for example : '1.00045e-11px'), which couldn't be interpreted by the backend.

Fix : We set a new minimal scale value to ensure we always send readable specs, which is enforced as long as the scale is greater than 0.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1140-minimal-scale-for-printing/index.html)

### DIFF
--- a/src/api/__tests__/print.api.spec.js
+++ b/src/api/__tests__/print.api.spec.js
@@ -1,7 +1,7 @@
 import { expect } from 'chai'
 import { describe, it } from 'vitest'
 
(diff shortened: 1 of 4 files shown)


ltkum:
I can't reproduce the failing tests locally. In both case, it says that 'process' is not defined, and I can find it in the package lock as a dependency for cypress. I don't know exactly why those tails consistently fail, but my guess is that it has nothing to do with the current code.

PR #1125, opened 2024-11-12
PB-849: create profile multi segement gpx

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-849-profile-multi-segment-gpx/index.html)

### DIFF
--- a/src/modules/i18n/locales/de.json
+++ b/src/modules/i18n/locales/de.json
@@ -496,6 +496,7 @@
     "profile_no_data": "kein Datum",
     "profile_poi_down": "Tiefster Punkt",
     "profile_poi_up": "Höchster Punkt",
+    "profile_segment": "Segment {segmentNumber}",
     "profile_slope_distance": "Wegstrecke",
     "profile_title": "Profil",
     "profile_too_many_points_error": "Die Profilanfrage war zu gross und konnte nicht verarbeitet werden.",
--- a/src/modules/infobox/components/FeatureElevationProfilePlot.vue
+++ b/src/modules/infobox/components/FeatureElevationProfilePlot.vue
@@ -5,6 +5,21 @@
         @mouseenter="startPositionTracking"
         @mouseleave="stopPositionTracking"
     >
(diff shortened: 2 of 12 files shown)


pakb (commented):
I think you will have to split segments, and ask for each one of them a different profile. Without that, the profile service will fill the blank with a straight line.

Here's an example : https://sys-map.dev.bgdi.ch/preview/feat-pb-849-profile-multi-segment-gpx/index.html#/map?lang=en&center=2595517.76,1205066.81&z=5.359&bgLayer=ch.swisstopo.pixelkarte-grau&topic=ech&layers=ch.astra.veloland@features=890.00&featureInfo=default

if you hover over the profile chart, you'll see at the end that it follows a straight line from the last point of the big loop to the beginning of the smaller one. It would make more sense to add some kind of pagination to the profile, with segments count or something like that
image

src/modules/i18n/locales/de.json
pakb, line 1: can you add your new i18n key (in all 5 files) just after the one called `profile_poi_up`? this way next time we run the translation script the output will be the same.

src/modules/infobox/components/FeatureElevationProfilePlot.vue
```diff
@@ -5,6 +5,18 @@
         @mouseenter="startPositionTracking"
         @mouseleave="stopPositionTracking"
     >
+        <div v-if="elevationProfile.segmentsCount > 1" class="d-flex gap-1 segment-container">
+            <button
+                v-for="(_, index) in elevationProfile.segments"
+                :key="index"
+                class="btn btn-light text-nowrap"
+                :class="{ 'btn-dark': index === elevationProfile.activeSegmentIndex }"
```
pakb, line 13: I know it's red, and usually red is more for "danger", but I think it would be better to use `btn-primary` anyway. The black button is only used once in the app, for the open/close menu button.

pakb (approved):

PR #1128, opened 2024-11-15
PB-732: fix open profile track partially outside bounds

sommerfe:


src/modules/infobox/components/FeatureElevationProfile.vue
```diff
@@ -80,7 +80,7 @@ function onCSVDownload() {
         <div v-if="!hasData">
             <LoadingBar v-if="!profileRequestError" />
             <span v-else class="text-danger" data-cy="profile-error-message">
-                {{ i18n.t(profileRequestError.messageKey) }}
+                {{ i18n.t(profileRequestError.message) }}
```
ltshb, line 83: This will probably break other error message like PofileError  just above the throw of outofbounderror. Also the error should be translated in all languages.

ltshb (changes requested):

PR #1129, opened 2024-11-18
PB-199: add new point on an existing line

ismailsunni:
Some todos:
- [x] Update style when extending the line
- [x] Make the add vertex button smaller
- [x] Put the buttons so that it's not covering the line (i.e. use azimuth?)
- [x] Check if it's work measurement line (not showing the buttons)
- [x] Replace the old feature, instead adding a new one

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-199-add-new-point/index.html)

### DIFF
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -6,6 +6,7 @@ import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
 import { IS_TESTING_WITH_CYPRESS } from '@/config/staging.config'
+import AddVertexButtonOverlay from '@/modules/drawing/components/AddVertexButtonOverlay.vue'
 import DrawingInteractions from '@/modules/drawing/components/DrawingInteractions.vue'
 import DrawingToolbox from '@/modules/drawing/components/DrawingToolbox.vue'
 import DrawingTooltip from '@/modules/drawing/components/DrawingTooltip.vue'
@@ -32,6 +33,20 @@ const featureIds = computed(() => store.state.drawing.featureIds)
 const isDrawingEmpty = computed(() => store.getters.isDrawingEmpty)
 const noFeatureInfo = computed(() => store.getters.noFeatureInfo)
 const online = computed(() => store.state.drawing.online)
+const selectedEditableFeatures = computed(() => store.state.features.selectedEditableFeatures)
+const selectedLineString = computed(() => {
+    if (selectedEditableFeatures.value && selectedEditableFeatures.value.length > 0) {
+        const selectedFeature = selectedEditableFeatures.value[0]
+        if (selectedFeature.geometry.type === 'LineString') {
+            return selectedFeature
+        } else {
+            return null
+        }
+    } else {
+        return null
+    }
+})
+
 const hasKml = computed(() => {
     if (online.value) {
         return !!activeKmlLayer.value
@@ -211,5 +226,9 @@ async function closeDrawing() {
         <DrawingToolbox @remove-last-point="removeLastPoint" @close-drawing="closeDrawing" />
         <DrawingTooltip />
         <DrawingInteractions ref="drawingInteractions" />
+        <AddVertexButtonOverlay
+            v-if="selectedLineString"
+            :line-string="selectedLineString"
+        ></AddVertexButtonOverlay>
     </div>
 </template>
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -228,7 +228,7 @@ async function closeDrawing() {
         <DrawingInteractions ref="drawingInteractions" />
         <AddVertexButtonOverlay
             v-if="selectedLineString"
-            :line-string="selectedLineString"
+            :line-string="selectedLineString.geometry"
         ></AddVertexButtonOverlay>
     </div>
 </template>
--- a/src/modules/drawing/DrawingModule.vue
+++ b/src/modules/drawing/DrawingModule.vue
@@ -44,12 +44,9 @@ const selectedLineString = computed(() => {
             selectedFeature.featureType === EditableFeatureTypes.LINEPOLYGON
         ) {
             return selectedFeature
-        } else {
-            return null
         }
-    } else {
-        return null
     }
+    return null
 })
 const showAddVertexButton = computed(() => {
     return store.state.drawing.editingMode === EditMode.MODIFY && !!selectedLineString.value
@@ -245,6 +242,6 @@ async function closeDrawing() {
         <AddVertexButtonOverlay
             v-if="showAddVertexButton"
             :line-string="selectedLineString.geometry"
-        ></AddVertexButtonOverlay>
+        />
     </div>
 </template>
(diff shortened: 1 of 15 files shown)


src/modules/drawing/DrawingModule.vue
```diff
@@ -32,6 +35,26 @@ const featureIds = computed(() => store.state.drawing.featureIds)
...
+            selectedFeature.geometry.type === 'LineString' &&
+            selectedFeature.featureType === EditableFeatureTypes.LINEPOLYGON
+        ) {
+            return selectedFeature
+        } else {
+            return null
+        }
+    } else {
+        return null
+    }
```
pakb, line 52: you can bunch that into one single `return null` at the end if no `if` was used

src/modules/drawing/DrawingModule.vue
```diff
@@ -211,5 +242,9 @@ async function closeDrawing() {
         <DrawingToolbox @remove-last-point="removeLastPoint" @close-drawing="closeDrawing" />
         <DrawingTooltip />
         <DrawingInteractions ref="drawingInteractions" />
+        <AddVertexButtonOverlay
+            v-if="showAddVertexButton"
+            :line-string="selectedLineString.geometry"
+        ></AddVertexButtonOverlay>
```
pakb, line 248: you can directly close the tag if no slot

src/modules/drawing/components/AddVertexButton.vue
```diff
@@ -0,0 +1,58 @@
+<script setup>
+import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
+import { onMounted, ref } from 'vue'
+import { useI18n } from 'vue-i18n'
+import { useStore } from 'vuex'
+
+import { EditMode } from '@/store/modules/drawing.store'
+// import { EditableFeatureTypes } from '@/api/features/EditableFeature.class'
```
pakb, line 8: maybe need some cleanup

src/modules/drawing/components/AddVertexButton.vue
```diff
@@ -0,0 +1,58 @@
...
+        default: 'modify_add_vertex',
+    },
+    // If true, the button will add a vertex in the reverse direction
+    reverse: {
+        type: Boolean,
+        default: false,
+    },
+})
+useTippyTooltip('#addVertexButton [data-tippy-content]', { placement: 'left' })
+const emit = defineEmits(['button-mounted'])
```
pakb, line 26: I would order that a bit differently, having

- props
- emits
- definition of refs
- import of dependencies (store, i18n)
- use of composables

src/modules/drawing/components/AddVertexButtonOverlay.vue
```diff
@@ -0,0 +1,107 @@
+<script setup>
+import { LineString } from 'ol/geom'
+import Overlay from 'ol/Overlay'
+import { onMounted, ref, watch } from 'vue'
+import { inject } from 'vue'
+import { onUnmounted } from 'vue'
+
+import AddVertexButton from './AddVertexButton.vue'
```
pakb, line 8: can you switch to a non relative import here?

src/modules/drawing/components/AddVertexButtonOverlay.vue
```diff
@@ -0,0 +1,107 @@
...
+        stopEvent: true,
+        offset: [-25, -25],
+    })
+    olMap.addOverlay(lastButtonOverlay.value)
+    updateButtonPositions()
+}
+
+onMounted(() => {
+    updateButtonPositions()
+    watch(() => props.lineString.coordinates, updateButtonPositions)
```
pakb, line 91: that shouldn't be in the mounted hook
you could also use a `computed` to get the coordinates, so you wouldn't need the lambda here

src/modules/drawing/components/AddVertexButtonOverlay.vue
```diff
@@ -0,0 +1,107 @@
...
+    // Get unit vector in opposite direction
+    const ux = -dx / length
+    const uy = -dy / length
+
+    // There is minus in y-direction because the y-axis is inverted in the map
+    return [ux * distance, -uy * distance]
+}
+
+const updateButtonPositions = () => {
+    const coordinates = props.lineString.coordinates
```
pakb, line 47: and if you switch to a computed you can replace the use here by a use of the composed too

src/modules/drawing/components/useModifyInteraction.composable.js
```diff
@@ -54,17 +56,78 @@ export default function useModifyInteraction(features) {
...
+        type: 'LineString', // Only works for LineString
+        minPoints: 2,
+        stopClick: true,
+        // only left-click to draw (primaryAction)
+        condition: (e) => primaryAction(e),
+        wrapX: true,
+    })
+
+    watch(
+        () => store.state.drawing.editingMode,
```
pakb, line 70: I'd prefer if you express your link to the store by binding it to a `computed` instead of using it directly

src/modules/drawing/components/useModifyInteraction.composable.js
```diff
@@ -54,17 +56,78 @@ export default function useModifyInteraction(features) {
...
+        (newValue) => {
+            if (newValue === EditMode.EXTEND && features.getArray().length > 0) {
+                const selectedFeature = features.getArray()[0]
+                if (store.state.drawing.reverseLineStringExtension) {
```
pakb, line 74: same here

src/modules/i18n/locales/rm.json
```diff
@@ -391,6 +391,7 @@
     "meteoschweiz_service_link_href": "http://www.meteoschweiz.admin.ch",
     "meteoschweiz_service_link_label": "MeteoSvizra",
     "mobile_redirect": "Versiun mobila",
+    "modify_add_vertex": "Thov nevo punkto vaś te lungǎres i alosardi linia",
```
pakb, line 394: I've updated the i18n files, and check this translation here, my tool has something else.
Can you use
`Agiuntar in nov punct per prolungar la lingia tschernida` instead?

pakb (commented):
as said on Slack, we should try to keep the geometry coordinates in order, even if we reverse edit (add something at the beginning)

src/modules/i18n/locales/rm.json
ismailsunni: @pakb I have rebased with develop, so I am not sure how else can I check if the translation is updated or not.

src/modules/drawing/DrawingModule.vue
```diff
@@ -32,6 +35,23 @@ const featureIds = computed(() => store.state.drawing.featureIds)
 const isDrawingEmpty = computed(() => store.getters.isDrawingEmpty)
 const noFeatureInfo = computed(() => store.getters.noFeatureInfo)
 const online = computed(() => store.state.drawing.online)
+const selectedEditableFeatures = computed(() => store.state.features.selectedEditableFeatures)
+const selectedLineString = computed(() => {
+    if (selectedEditableFeatures.value && selectedEditableFeatures.value.length > 0) {
+        const selectedFeature = selectedEditableFeatures.value[0]
+        if (
+            selectedFeature.geometry.type === 'LineString' &&
+            selectedFeature.featureType === EditableFeatureTypes.LINEPOLYGON
```
pakb, line 44: can you also add `EditableFeatureTypes.MEASURE` here?

pakb (approved):

PR #1135, opened 2024-11-25
PB-774: Reset to default background layer when clicking on the logo

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-774-do-not-reset-bg-layer/index.html)

### DIFF
--- a/src/modules/menu/components/header/HeaderWithSearch.vue
+++ b/src/modules/menu/components/header/HeaderWithSearch.vue
@@ -20,13 +20,14 @@ const header = ref(null)
 const store = useStore()
 const { t } = useI18n()
 
+const currentBackground = computed(() => store.state.layers.currentBackgroundLayerId)
 const currentLang = computed(() => store.state.i18n.lang)
-const currentTopicId = computed(() => store.state.topics.current)
 const currentTopic = computed(() => store.getters.currentTopic)
-const isPhoneMode = computed(() => store.getters.isPhoneMode)
+const currentTopicId = computed(() => store.state.topics.current)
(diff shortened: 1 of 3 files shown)


src/modules/menu/components/header/HeaderWithSearch.vue
```diff
@@ -48,8 +49,10 @@ function updateHeaderHeight() {
 }
 
 function resetApp() {
-    // an app reset means we keep the lang and the current topic but everything else is thrown away
-    window.location = `${window.location.origin}?lang=${currentLang.value}&topic=${currentTopicId.value}`
+    // an app reset means we keep the lang, the current topic, and its default background layer but everything else is thrown away
+    // We keep the default background layer of the current topic because the app always set to `ech` topic and its default background layer before we can even get the topic from the URL
+    const defaultBackgroundLayerId = currentTopic.value?.defaultBackgroundLayer?.id
```
pakb, line 54: I'd maybe guard against a `null` value here by having something like
```
const defaultBackgroundLayerId = currentTopic.value?.defaultBackgroundLayer?.id ?? currentBackground.value
```
What do you think?

ismailsunni: yes, I think it's better to have a guard. I was thinking that all topic has a default background layer. Added new.

pakb (commented):
Can you add a relevant Cypress test for that? So that we make sure we do not regress on this later on 😉

ismailsunni:
@pakb test added

PR #1136, opened 2024-11-26
PB-1216: Add warning on sharing when there is a local files.

ismailsunni:
image

Let me know if there is suggestion for the warning text.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1216-warning-sharing-local-file/index.html)


pakb (commented):
I think it would make more sense to have the "Share link input" and its button with this yellow-ish color, and a tooltip that shows when we hover it stating what you've added in the warning message.

Having a warning message popping up each time you open the share section is not ideal and not comfortable for users IMO

pakb:
image
Can you try keeping the text color unchanged, and change the background of the input instead? (Maybe have a border with the same color but darker).
I find it a bit hard to read what's there with yellow over white background

ismailsunni:
@pakb here is updated style for the warning.
image
I have added the e2e test also.

src/modules/menu/components/share/MenuShareInputCopyButton.vue
```diff
@@ -1,17 +1,119 @@
...
                 readonly="readonly"
                 :value="inputText"
                 data-cy="menu-share-input-copy-button"
                 @focus="$event.target.select()"
             />
             <button
-                class="btn btn-outline-group"
+                class="btn"
+                :class="{
+                    'warning-input': hasWarning,
```
pakb, line 114: you should not use self-created class here to apply the color, but use `btn-warning` or such Bootstrap tools instead (`bg-warning`, etc...).

With the current way of doing it, the hover color is out of place (and I bet we can find some cases where the color won't be consistent by doing it ourselves)

src/modules/menu/components/share/MenuShareInputCopyButton.vue
```diff
@@ -1,17 +1,119 @@
...
+    }
+})
+</script>
+
 <template>
-    <div v-if="inputText">
+    <div
+        v-if="inputText"
+        id="input-copy-button"
+        data-tippy-content="warn_share_local_file"
```
pakb, line 97: the tooltip color should match the warning from the button (add a new tooltip style if necessary)

PR #1139, opened 2024-11-26
PB-1221 : partially out of bound KML was ignored entirely

pakb:
The KML extent detection was not performing as expected with KML that had only one dimension out of LV95 bounds.

To counteract that, I've reverted the logic in the extent calculation, transforming LV95 bounds into WGS84 to do the matching instead of transforming a potentially big WGS84 extent to LV95 (lead to big math error...)

[Test link with sample file](https://sys-s.dev.bgdi.ch/5tsm1i0pza4b)
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1221-kml-extent-false-positive/index.html)

### DIFF
--- a/src/utils/extentUtils.js
+++ b/src/utils/extentUtils.js
@@ -84,26 +84,35 @@ export function getExtentIntersectionWithCurrentProjection(
     ) {
         return null
     }
-    let extentInCurrentProjection = flattenExtent(extent)
+    let currentProjectionAsExtentProjection = currentProjection.bounds.flatten
     if (extentProjection.epsg !== currentProjection.epsg) {
-        extentInCurrentProjection = projExtent(
-            extentProjection,
+        // we used to reproject the extent here, but there's problem arising if current projection is LV95 and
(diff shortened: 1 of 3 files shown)


src/utils/extentUtils.js
```diff
@@ -84,26 +84,33 @@ export function getExtentIntersectionWithCurrentProjection(
     ) {
         return null
     }
-    let extentInCurrentProjection = flattenExtent(extent)
+    let currentProjectionAsExtentProjection = currentProjection.bounds.flatten
+    let finalExtent = flattenExtent(extent)
```
sommerfe, line 88: why declare finalExtent here and not just declare it later with 
`
let finalExtent = getExtentIntersection(flattenExtent(extent), currentProjectionAsExtentProjection)
`

pakb: good point, now that it's not used in the next if block I could move the declaration further down

sommerfe (approved):

ltkum (approved):
lgtm

PR #1141, opened 2024-11-26
PB-1222 : fix mishandling of feature ID that are "number-like"

pakb:
some feature IDs ends with something like "1314.070" and the layer parsing code was transforming them into number, removing the trailing zero from the ID.

This broke many "Link to object" link generated by the backend.

Also fixing an issue with the time slider warning not being wrapped in an array, and generating a crash when its value was not set to something valid

[Test link with "Link to object" params](https://sys-map.dev.bgdi.ch/preview/fix-pb-1222-broken-link-to-object/index.html?ch.vbs.schiessanzeigen=1314.070&time=None&lang=de&topic=ech)
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1222-broken-link-to-object/index.html)


ltkum (approved):
lgtm

PR #1147, opened 2024-11-28
Add a debug tool to help us find relevant layers

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/add-debug-tool-layer-finder/index.html)


ltkum (approved):
As we discussed this morning, It could be good to add a way to say 'we want this/these aspects to be valid and I don't care about the rest', but it's already a great tool and we can upgrade it when we have time ;)

PR #1148, opened 2024-11-29
PB-1204 Cesium WMS as composable and tested with Cypress

pakb:
fixing an issue with preview layer being kept on the map after added through the catalogue component

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1204-cesium-wms-composable/index.html)


ismailsunni (approved):
LGTM

PR #1149, opened 2024-11-29
PB-1192: add test move and resize legend

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1192-add-test-open-move-legend-panel/index.html)


pakb (commented):
I'll dig the test later, but so far a couple points that could already be fixed

pakb (approved):
Let's move on, not that big of a deal if we couldn't test a drag/resize gesture. Might come later out-of-the-box in Cypress

PR #1159, opened 2024-12-03
PB-1254: fix embed preview in chrome

ltkum:
Issue: In chrome-based web browsers, the layers would be removed from the preview and the link when opening the embed preview unless the bgLayer was 'void'.

Cause: There might be a need to refactor completely the embed sharing component, but for now, it seems the issue was caused by the bgLayer parameter being handled before the layers. This would cause a situation where the router would trigger a mutation, go to a new URL with no layers, and overwrite what should have been in store. As the default background layer is the void one, we did not go through this when we used the empty background layer.

Fix: by changing the order in which the parameters are handled, we can ensure layers are handled before the background layer, and thus ensure they are kept.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1254-allow-embed-layer-sharing/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/embed.cy.js
+++ b/tests/cypress/tests-e2e/embed.cy.js
@@ -1,6 +1,13 @@
 /// <reference types="cypress" />
 
 describe('Testing the embed view', () => {
+    function checkUrlParams(urlToCheck, validationUrl) {
+        const href = new URLSearchParams(urlToCheck.replace('#map', ''))
+        const validation = new URLSearchParams(validationUrl.replace('#map', ''))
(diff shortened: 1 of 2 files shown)


tests/cypress/tests-e2e/embed.cy.js
pakb, line 1: can you change all your `invoke(...).then(...)` into `invoke(...).should(...)` ? This would add retry-ability to all your tests (and should make them a bit more resilient on the CI)

ltkum: sure

pakb (approved):

PR #1162, opened 2024-12-09
PB-1151: show a warning when a layer is partially out of bound

ltkum:
This PR changes three things : 

1. We now show once again profiles when they're partially out of bounds
2. We refactored the feature Elevation component to use the composition API
3. We added a tooltip when there is no data to explain that there is no data. 

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1151-support-partially-out-of-bounds-kml/index.html)


src/api/profile/profile.api.js
```diff
@@ -57,20 +57,47 @@ export function splitIfTooManyPoints(chunks = null) {
...
+            let coordinate = point.coordinate
+            if (outputProjection.epsg !== LV95.epsg) {
+                coordinate = proj4(LV95.epsg, outputProjection.epsg, coordinate)
+            }
+            points.push({
+                coordinate,
+                dist: startingDist + point.dist,
+                elevation: point.elevation,
+                hasElevationData: point.hasElevationData,
+            })
```
pakb, line 85: this is copy pasted and needs to be written only once.
please move things around so that we do not repeat ourselves

src/modules/infobox/components/FeatureElevationProfilePlot.vue
pakb, line 1: `$t` uses in the template should also be migrated to `i18n.t`

tests/cypress/fixtures/service-alti/profile.fixture.out.of.bounds.json
pakb, line 1: this fixture needs some data, or else being renamed to empty profile

pakb (commented):
When requesting a measure line from Germany (way north of LV95 limit) to the south of Switzerland (within LV95 extent), it still creates a multi-segmented profile, with only one of them.

image

Only the second segment is the one we want here, the first segment is totally empty, and should not be there

src/api/profile/profile.api.js
```diff
@@ -57,18 +57,33 @@ export function splitIfTooManyPoints(chunks = null) {
...
         if (outputProjection.epsg !== LV95.epsg) {
             coordinate = proj4(LV95.epsg, outputProjection.epsg, coordinate)
         }
         points.push({
             coordinate,
-            dist: startingDist + rawData.dist,
-            elevation: rawData.alts.COMB,
-            hasElevationData: rawData.alts.COMB !== null,
+            dist: startingDist + point.dist,
+            elevation: point.elevation !== undefined ? point.elevation : point.alts.COMB,
```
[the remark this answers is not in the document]
ltkum: no, because `elevation: null` is technically a valid value. What I want to test is the presence of the parameter

src/api/profile/profile.api.js
```diff
@@ -57,18 +57,33 @@ export function splitIfTooManyPoints(chunks = null) {
...
+ *   [Number, Number], elevation: null, hasElevationData: false}, ... ]}
+ * @param {Number} startingDist
+ * @param {SwissCoordinateSystem} outputProjection
+ * @returns {ElevationProfileSegment}
+ */
 function parseProfileFromBackendResponse(backendResponse, startingDist, outputProjection) {
     const points = []
-    backendResponse.forEach((rawData) => {
-        let coordinate = [rawData.easting, rawData.northing]
+    // we can receive an array of coordinates, or a dict
```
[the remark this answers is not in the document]
ltkum: I'll replace it with an `object`

tests/cypress/fixtures/service-alti/profile.fixture.out.of.bounds.json
ltkum: this is also what we receive if we get a profile when within  LV95 bounds, but out of the service-alti data extent. I can rename it "empty profile", but I wanted to be explicit with what I am trying to fetch which will give me this as data

ltkum:
We extracted the vue component refactor in the composition API to give it its own PR, we will close this as there was a confusion between what was done and what was expected :)

PR #1164, opened 2024-12-10
PB-1176: Fix cesium static assets location

ltshb:
During the build the cesium static assets were put into the root of the application.
This was an issue for our deployment strategy were we have all assets in a
versioned folder and were we first deploy the versioned folder before deploying
the entry point index.html that point to the versioned folder. This allow for
a pseudo atomic deployment.

So now all the cesium static assets have been placed into VERSION/cesium/ path.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1176-cesium-build/index.html)

### DIFF
--- a/vite.config.mts
+++ b/vite.config.mts
@@ -20,7 +20,7 @@ if (!appVersion) {
 
 const __dirname = dirname(fileURLToPath(import.meta.url))
 const cesiumSource = `${__dirname}/node_modules/cesium/Source`
-const cesiumWorkers = '../Build/Cesium/Workers'
+const cesiumStaticDir = `./${appVersion}/cesium/`
 
 /**
  * We use manual chunks to reduce the size of the final index.js file to improve startup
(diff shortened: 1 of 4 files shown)


vite.config.mts
```diff
@@ -21,6 +21,7 @@ if (!appVersion) {
 const __dirname = dirname(fileURLToPath(import.meta.url))
 const cesiumSource = `${__dirname}/node_modules/cesium/Source`
 const cesiumWorkers = '../Build/Cesium/Workers'
```
pakb, line 23: not sure it would also be a good idea to replace this relative path with something similar to the line above

iwibot: but then we need to modify further down. However I'll remove this `cesiumWorkers` constant as it is used only in one place and use the path hardcoced like for the other assets, it make the code easier to understand.

ltshb: but then we need to modify further down. However I'll remove this cesiumWorkers constant as it is used only in one place and use the path hardcoced like for the other assets, it make the code easier to understand.

PR #1165, opened 2024-12-10
PB-1176: Minify the index.html

ltshb:
This remove the unnecessary white spaces

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1176-minify/index.html)


ltshb:
This alone don't improve the situation because currently the index.html is big enough so that Cloudfront is compressing it and have a final size over network of 884 Bytes. If we minify it then it is too small to get compressed by Cloudfront and the final size over the network is 1.3K.
If we would gzip the file before deploying, together with the minify we could gain 200 Bytes.

PR #1168, opened 2024-12-11
PB-1290 : read time slider oldest/youngest year from layers config

pakb:
instead of having hard-coded values. This means we can now show the next year's data point without waiting for the 1st of January, and also show older data than the arbitrarily chosen limit of 1844.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1290-next-year-in-time-slider/index.html)


ltkum (approved):
I got a brain freeze for a bit as I mistook oldest and youngest time for a few moments, but when I came to my senses, I couldn't find any fault within the code. :)

PR #1169, opened 2024-12-11
PB-1223: add test more information feature info

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1223-add-test-more-information-feature-info/index.html)


pakb (approved):

ltkum (commented):
The work is good, but there is a small things that bothers me regarding calling the api directly. I would at least add a comment and replace this call with an intercept on a fixture when we are sure the api3 tests cover the extended popup

ltkum (approved):
looks good to me

PR #1174, opened 2024-12-17
PB-1278: Fix Icon with offset (pin) changes positon

ismailsunni:
Need to be reviewd by @ltkum as this revert his change in https://github.com/geoadmin/web-mapviewer/pull/1015

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1278-icon-offset/index.html)

Sample on 1:500 and 1:5000. Looks good
image

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1278-icon-offset/index.html)

### DIFF
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -136,7 +136,7 @@ class GeoAdminCustomizer extends BaseCustomizer {
             // don't ask why it works, but that's the best I could do.
             // Note(IS): I'm not sure if this is the best way to handle the Y offset. Need to be tested of different case.
             // Per issue 1278, it's correct. But Martin did it differently, but all test is pass.
-            symbolizer.graphicYOffset = symbolizer.graphicYOffset ? symbolizer.graphicYOffset : 0
+            symbolizer.graphicYOffset = symbolizer.graphicYOffset ?? 0
         }
         if (size) {
             symbolizer.graphicWidth = adjustWidth(size[0] * scale, this.printResolution)
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -134,7 +134,7 @@ class GeoAdminCustomizer extends BaseCustomizer {
                   )
                 : 0
             // don't ask why it works, but that's the best I could do.
-            symbolizer.graphicYOffset = Math.round(symbolizer.graphicYOffset) ?? 0
+            symbolizer.gaphicYOffset = Math.round(1000 * symbolizer.gaphicYOffset ?? 0) / 1000
         }
         if (size) {
             symbolizer.graphicWidth = adjustWidth(size[0] * scale, this.printResolution)
--- a/src/api/print.api.js
+++ b/src/api/print.api.js
@@ -134,10 +134,9 @@ class GeoAdminCustomizer extends BaseCustomizer {
                   )
                 : 0
             // don't ask why it works, but that's the best I could do.
-
-            symbolizer.graphicYOffset = symbolizer.graphicYOffset
-                ? (symbolizer.graphicYOffset = adjustWidth(-size[1], this.printResolution))
-                : 0
+            // Note(IS): I'm not sure if this is the best way to handle the Y offset. Need to be tested of different case.
+            // Per issue 1278, it's correct. But Martin did it differently, but all test is pass.
+            symbolizer.graphicYOffset = symbolizer.graphicYOffset ? symbolizer.graphicYOffset : 0
         }
         if (size) {
             symbolizer.graphicWidth = adjustWidth(size[0] * scale, this.printResolution)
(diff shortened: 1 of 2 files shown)


src/api/print.api.js
```diff
@@ -134,10 +134,9 @@ class GeoAdminCustomizer extends BaseCustomizer {
                   )
                 : 0
             // don't ask why it works, but that's the best I could do.
-
-            symbolizer.graphicYOffset = symbolizer.graphicYOffset
-                ? (symbolizer.graphicYOffset = adjustWidth(-size[1], this.printResolution))
-                : 0
+            // Note(IS): I'm not sure if this is the best way to handle the Y offset. Need to be tested of different case.
+            // Per issue 1278, it's correct. But Martin did it differently, but all test is pass.
+            symbolizer.graphicYOffset = symbolizer.graphicYOffset ? symbolizer.graphicYOffset : 0
```
pakb, line 139: could be simplified to
```javascript
symbolizer.graphicYOffset = symbolizer.graphicYOffset ?? 0
````

File: src/utils/styleUtils.js [hunk 2]
```diff
@@ -161,7 +161,7 @@ export const highlightPointStyle = new Style({
 // Change a width according to the change of DPI (from the old geoadmin)
 // Originally introduced here https://github.com/geoadmin/mf-geoadmin3/pull/3280
 export function adjustWidth(width, dpi) {
-    if (!width || isNaN(width) || width <= 0.0 || !dpi || isNaN(dpi) || dpi <= 0) {
+    if (!width || isNaN(width) || !dpi || isNaN(dpi) || dpi <= 0) {
```
pakb, line 164: maybe @ltkum can better tell if this might break something or not...
I think that was a "trade-off" fix so that we would not have to add some fixes to the mapfishprint library or backend. I'll let him tell us what was the state on that

ismailsunni: Yes, I need @ltkum to the changes in this PR.

For this code, I remove it because for `offset`, it might be a negative number.

ltkum: From what I remember and can see : 

The `width <= 0.0` line was here because, afterwards, we use a math.max to set the scale size to its minimal value if it's smaller than it (as very small numbers would break the print). Currently, you don't check anymore if it's negative, which means it will set its offset to min scale when it's negative.

what I would do : 
```
function adjustWidth(width, dpi) {
if (!width || isNaN(width) || width <= 0.0 || !dpi || isNaN(dpi) || dpi <= 0) {return 0}
if (width <= 0) {return -adjustWidth(-width, dpi}
return Math.max((width * PRINT_DPI_COMPENSATION) / dpi, MIN_PRINT_SCALE_SIZE)
}
```

that way, your offsets are not set to 0, nor to MIN_PRINT_SCALE_SIZE

src/api/print.api.js
```diff
@@ -134,10 +134,9 @@ class GeoAdminCustomizer extends BaseCustomizer {
                   )
                 : 0
             // don't ask why it works, but that's the best I could do.
-
-            symbolizer.graphicYOffset = symbolizer.graphicYOffset
-                ? (symbolizer.graphicYOffset = adjustWidth(-size[1], this.printResolution))
-                : 0
+            // Note(IS): I'm not sure if this is the best way to handle the Y offset. Need to be tested of different case.
```
ismailsunni, line 137: hi @ltkum what do you think about this one?

ismailsunni: that's correct. I have updated accordingly.

src/api/print.api.js
```diff
@@ -134,10 +134,7 @@ class GeoAdminCustomizer extends BaseCustomizer {
                   )
                 : 0
             // don't ask why it works, but that's the best I could do.
-
-            symbolizer.graphicYOffset = symbolizer.graphicYOffset
-                ? (symbolizer.graphicYOffset = adjustWidth(-size[1], this.printResolution))
-                : 0
+            symbolizer.graphicYOffset = Math.round(symbolizer.graphicYOffset) ?? 0
```
ltkum, line 137: this won't work : 
Math.round(undefined) gives 'NaN', which satisfies the null operand and so it gives NaN instead of 0.
Also I would prefer we round it a little less, here is my proposal : 
```
symbolizer.gaphicYOffset = symbolizer.gaphicYOffset ?? 0
symbolizer.gaphicYOffset = Math.round(1000 * symbolizer.gaphicYOffset) / 1000
```
that way, we have a very insignificant rounding which still satisfies the current print at all times.

(we can go with a standard ternary operator instead if you prefer this approach)

ltkum (commented):
a last small change and we should be good to go :)

src/api/print.api.js
ismailsunni: ah ok, thanks!

ltkum (approved):
Good to go :)

Thanks

PR #1177, opened 2024-12-18
PB-1213: Extend line to polygon

ismailsunni:
Since `DrawInteraction` only supports extending a `LineString` geometry, I need to create a new feature, to "extend" it. Then this new feature's geometry will replace the old one.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1213-polygon-from-line/index.html)


sommerfe (approved):

ltkum (changes requested):
There is currently an issue : if I try to extend a line other than the last line made, it will extend the last line made instead.

I'll provide a video as soon as I manage to get a capture working :)

ltkum:
https://github.com/user-attachments/assets/a476a784-c3d6-4342-8fb7-16d0dbf40d87

ismailsunni:
thanks @ltkum I have fixed it and add test to make sure it's fixed.

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -67,31 +67,48 @@ describe('Drawing module tests', () => {
...
                 .then((drawingLayer) => drawingLayer.getSource().getFeatures())
                 .should((features) => {
-                    expect(features).to.be.an('Array').lengthOf(1)
-                    const [feature] = features
-                    expect(feature.getGeometry().getType()).to.eq('Polygon')
-                    const polygonCoordinates = feature.getGeometry().getCoordinates()
-                    expect(polygonCoordinates).to.be.an('Array').lengthOf(1)
-                    expect(polygonCoordinates[0]).to.be.an('Array').lengthOf(numberOfPoints)
+                    console.log(description)
+                    console.log(features)
```
ltkum, line 91: This looks like debug logs.

If the logs are importants to use during tests, use cy.log and a small description of the content of the log.
If those are debug logs, you should remove them.

tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -67,31 +67,48 @@ describe('Drawing module tests', () => {
...
+                    features.forEach((feature) => {
+                        console.log(feature.get('description'))
+                    })
```
ltkum, line 94: this looks like this loop is only here for debug purposes. Same as above :)

ltkum (commented):
I would remove the debug logs before merging this. Otherwise this is good work, thanks.

ltkum (approved):

src/modules/drawing/components/DrawingInteractions.vue
```diff
@@ -8,28 +8,59 @@ import DrawingMarkerInteraction from '@/modules/drawing/components/DrawingMarker
...
+            log.error('Invalid feature type for extend mode')
+
+            selectedInteraction = null
+        }
+    }
+    // Make sure that the modify interaction is disabled when we are in draw / extend mode
+    if (selectedInteraction) {
+        selectInteraction.value?.setActive(false)
+    } else {
+        selectInteraction.value?.setActive(true)
```
pakb, line 61: `selectInteraction.value?.setActive(!selectedInteraction)`
could do the same, no?

src/modules/drawing/components/DrawingInteractions.vue
```diff
@@ -8,28 +8,59 @@ import DrawingMarkerInteraction from '@/modules/drawing/components/DrawingMarker
...
+import log from '@/utils/logging'
 // DOM References
 const selectInteraction = ref(null)
 const currentInteraction = ref(null)
-
 const store = useStore()
 const currentDrawingMode = computed(() => store.state.drawing.mode)
 const editMode = computed(() => store.state.drawing.editingMode)
 
+let selectedLineFeature = null
```
pakb, line 22: you should make a `ref` out of this, as it might change without recreating the whole `useExtendLineInteraction` down the line

src/modules/drawing/components/useExtendLineInteraction.composable.js
```diff
@@ -0,0 +1,25 @@
+import { EditableFeatureTypes } from '@/api/features/EditableFeature.class'
+import useDrawingModeInteraction from '@/modules/drawing/components/useDrawingModeInteraction.composable'
+import { drawLineStyle } from '@/modules/drawing/lib/style'
+
+export default function useExtendLineInteraction({
```
pakb, line 5: please describe what are the input types in a JSDoc here

src/modules/drawing/components/useDrawingModeInteraction.composable.js
```diff
@@ -26,6 +29,7 @@ export default function useDrawingModeInteraction({
     useGeodesicDrawing = false,
     snapping = false,
     drawEndCallback = null,
+    startingFeature = null,
```
pakb, line 32: might be good to also add some JSDoc on what types are expected here, it starts to get quite big...

src/modules/drawing/components/ExtendLineInteraction.vue
```diff
@@ -0,0 +1,25 @@
...
+const emits = defineEmits({
+    drawEnd(payload) {
+        return payload instanceof Feature
+    },
+})
+const { removeLastPoint } = useExtendLineInteraction({
+    drawEndCallback: (feature) => {
+        emits('drawEnd', feature)
+    },
+    startingFeature: props.startingFeature,
```
pakb, line 17: I'm not sure it's a good idea to give that without wrapping it in a `ref`. The composable down the line isn't called once more when this value changes (if you select another line of the same drawing), so reactivity might be lost here.
But after testing, it seems to work fine...

src/modules/drawing/components/useDrawingModeInteraction.composable.js
```diff
@@ -146,49 +194,54 @@ export default function useDrawingModeInteraction({
...
+            // see https://openlayers.org/en/latest/apidoc/module-ol_interaction_Draw-Draw.html#finishDrawing
+            interaction.finishDrawing()
+            store.dispatch('addDrawingFeature', { featureId: drawnFeature.getId(), ...dispatcher })
+            store.dispatch('setDrawingMode', { mode: null, ...dispatcher })
+            if (drawEndCallback) {
+                drawEndCallback(drawnFeature)
+            }
+            // Here we need to save work in next tick to have the drawingLayer source updated.
+            // Otherwise, the source might not yet be updated with the new/updated/deleted feature
+            nextTick().then(debounceSaveDrawing)
```
pakb, line 243: would a watch (deep watch if necessary) on the drawing layer also solve this issue?
Here you are waiting an arbitrary life cycle before committing to saving, but we can't be sure the next tick will be the one that has saved/stored everything you need, isn't it?

pakb (commented):
tested it, works well, good work!
A couple questions, and some request for more JSDocs on the many composable used here and we should be good to go

src/modules/drawing/components/DrawingInteractions.vue
ismailsunni: yes :)

src/modules/drawing/components/ExtendLineInteraction.vue
ismailsunni: I handled the selection in `DrawingInteractions.vue`, see [this](https://github.com/geoadmin/web-mapviewer/pull/1177/commits/7d885b6874f700975b888d9412697e03d9ade4a6#diff-6cb2ff831d9178e2c52712eb1b44e2437ba8c6b600351228eaed444ad8a13d3aR79-R91). Basically, if the feature selected is changed (from the select interaction) it will update the selectedLineFeature.

Or am I wrong?

src/modules/drawing/components/useDrawingModeInteraction.composable.js
ismailsunni: It's not my original code actually. It's already there in develop branch, see [this](https://github.com/geoadmin/web-mapviewer/blob/develop/src/modules/drawing/components/useDrawingModeInteraction.composable.js#L189-L191). I only did a small refactor, i.e. put it in a if statement. So, I am not really sure what to do.

here is the original PR that introduce the code: https://github.com/geoadmin/web-mapviewer/pull/561/files#diff-3423e4552122109b2a6f9dac47e5d135482d4822ce8a85c113ec2c983463a51cR170-R172

PR #1178, opened 2024-12-18
PB-326: check external provider get map

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-326-check-external-provider-get-map/index.html)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,6 +102,7 @@
                 "prettier-plugin-jsdoc": "^1.3.0",
                 "rimraf": "^6.0.1",
                 "sass": "1.77.6",
+                "sharp": "^0.33.5",
                 "start-server-and-test": "^2.0.9",
                 "typescript": "~5.6.3",
                 "vite": "^5.4.11",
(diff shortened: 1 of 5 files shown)


ltshb (commented):
After a quick look at the code it looks good, however I tested it and I've got the following provider as invalid although it apparently don't support CORS:
- https://vogis.cnv.at/mapserver/mapserv?map=i_naturschutz_v_wms.map
- https://www.basemap.at/wmts/1.0.0/WMTSCapabilities.xml
I haven't checked all provider with failures, but it seems that there is a bug in the script, the first goal is to make sure the provider Get capabilities support CORS, if not no other check should be done as we can't use it in the viewer.

ltshb (approved):
Looks good. 
Based on the result of the script
```
Checked 347 providers, 22 where invalids and 7 don't support CORS
```
Within this ticket, the 22 invalids provider should be analyzed and you should open new tickets as following:
- New ticket for provider with invalid CORS, @hansmannj provider with invalid CORS should be notified to be modified and if they cannot be changed then we need to remove them from our list.
- New ticket for invalid provider due to provider error, they should be notified and changed or removed from the list
- New ticket for invalid provider due to a bug in our code (I've seen that some of them return apparently proper image or get capabilities but somehow cannot be displayed, they need to be analyzed to see why, e.g. https://services.geo.sg.ch/wss/service/SG00073_WMS/guest?).

sommerfe:
Following the remarks of @ltshb i created the jira ticket PB-1324

PR #1179, opened 2024-12-20
PB-1310: 3D navigation compass was not showing

ltkum:
Issue: The compass in the 3d viewer was hidden

Cause: We checked before there was an instance of the component if we could add the scene and the clock to the compass element, meaning it would create an empty compass.

Fix: We wait for the map to be ready before adding the compass

[Test link](https://sys-map.dev.bgdi.ch/preview/hotfix-compass-not-showing/index.html)


hansmannj (approved):
Thanks :+1:
Looks good as far as I can tell!

PR #1180, opened 2025-01-02
PB-1196, PB-1197: timeslider and compare slider test

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1197-timeslider-and-compare-slider-test/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/utils.js
+++ b/tests/cypress/tests-e2e/utils.js
@@ -1,0 +1,10 @@
+/// <reference types="cypress" />
+
+export function moveTimeSlider(x) {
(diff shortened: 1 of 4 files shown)


tests/cypress/tests-e2e/utils.js
```diff
@@ -0,0 +1,10 @@
+/// <reference types="cypress" />
+
+export function moveTimeSlider(x) {
+    cy.get('[data-cy="time-slider-bar-cursor-grab"]').trigger('mousedown', { which: 1 })
```
sommerfe, line 4: I think `which` is deprecated, you should use `button` but in this case it is `button: 0`

sommerfe (approved):

pakb (approved):

PR #1182, opened 2025-01-06
PB-1297: make position of text labels configurable

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1297-make-position-of-text-labels-configurable/index.html)


ltshb (commented):
I like very much the UI :-) 

However the offset is not very optimal, the top and down offset is in my opinion way too much:
image

In the other-hand the left/right offset is way too small in comparison
image

Also it seems that the top offset has been increased in comparison to the production deployment, see below
image

ltshb (approved):
Looks good to me Thanks

PR #1184, opened 2025-01-07
PB-1195: Add test for printing WMS and WMTS layers.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1195-test-external-layer-printing/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/print.cy.js
+++ b/tests/cypress/tests-e2e/print.cy.js
@@ -13,6 +13,14 @@ describe('Testing print', () => {
 
         interceptCapabilities()
         interceptShortLink()
+
+        interceptPrintRequest()
+        interceptPrintStatus()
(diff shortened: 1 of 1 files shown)


tests/cypress/tests-e2e/print.cy.js
```diff
@@ -565,4 +568,294 @@ describe('Testing print', () => {
...
+                {
+                    fixture: '256.png',
+                }
+            ).as('externalWMTS')
+        })
+
+        it('prints external WMS correctly', () => {
+            interceptPrintRequest()
+            interceptPrintStatus()
+            interceptDownloadReport()
```
pakb, line 666: Could also be moved to the `beforeEach` hook, as there's no test not needing that in this test file

pakb (approved):

PR #1187, opened 2025-01-08
PB-1187: Add more geolocation test

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1187-geolocation-test/index.html)

### DIFF
--- a/src/store/plugins/geolocation-management.plugin.js
+++ b/src/store/plugins/geolocation-management.plugin.js
@@ -6,6 +6,7 @@ import { LV95, WGS84 } from '@/utils/coordinates/coordinateSystems'
 import ErrorMessage from '@/utils/ErrorMessage.class'
 import log from '@/utils/logging'
 import { round } from '@/utils/numberUtils'
+const { GeolocationPositionError } = window
 
 const dispatcher = { dispatcher: 'geolocation-management.plugin' }
 
@@ -91,7 +92,7 @@ const handlePositionError = (error, store, state, options = {}) => {
     const { reactivate = false } = options
     log.error('Geolocation activation failed', error)
     switch (error.code) {
-        case error.PERMISSION_DENIED:
+        case GeolocationPositionError.PERMISSION_DENIED:
             store.dispatch('setGeolocationDenied', {
                 denied: true,
                 ...dispatcher,
@@ -101,7 +102,7 @@ const handlePositionError = (error, store, state, options = {}) => {
                 ...dispatcher,
             })
             break
-        case error.TIMEOUT:
+        case GeolocationPositionError.TIMEOUT:
             store.dispatch('setGeolocation', { active: false, ...dispatcher })
             store.dispatch('addErrors', {
                 errors: [new ErrorMessage('geoloc_time_out', null)],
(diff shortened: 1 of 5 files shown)


src/store/plugins/geolocation-management.plugin.js
```diff
@@ -101,20 +102,23 @@ const handlePositionError = (error, store, state, options = {}) => {
...
             break
         default:
-            if (IS_TESTING_WITH_CYPRESS && error.code === error.POSITION_UNAVAILABLE) {
-                // edge case for e2e testing, if we are testing with Cypress and we receive a POSITION_UNAVAILABLE
-                // we don't raise an error as it's "normal" in Electron to have this error raised (this API doesn't work
-                // on Electron embedded in Cypress : no Geolocation hardware detected, etc...)
-                // the position will be returned by a mocked up function by Cypress we can ignore this error
-                // we do nothing...
+            if (
+                IS_TESTING_WITH_CYPRESS &&
```
pakb, line 114: we could get rid if this `IS_TESTING_WITH_CYPRESS` here, and apply the code bellow anytime this error code is found? (could change it to another case in the switch if it's the case)

pakb: Or if the goal is to disable the retry function, we could add some `IS_TESTING_WITH_CYPRESS` case in the code below, such as
```javascript
errorCount += IS_TESTING_WITH_CYPRESS ? 3 : 1
```

tests/cypress/tests-e2e/geolocation.cy.js
```diff
@@ -49,54 +65,104 @@ describe('Geolocation cypress', () => {
...
+            cy.goToMapView({}, true, { errorCode: GeolocationPositionError.TIMEOUT })
+            getGeolocationButtonAndClickIt()
+            testErrorMessage('geoloc_time_out')
+        })
+        it('shows an alert telling the user geolocation is not available for other reason', () => {
+            cy.goToMapView({}, true, { errorCode: GeolocationPositionError.POSITION_UNAVAILABLE })
+            getGeolocationButtonAndClickIt()
+            testErrorMessage('geoloc_unknown')
+        })
+    })
```
pakb, line 167: you can wrap up all these 3 its into a single one, that would lower our it count (that's our hard limit with CypressCloud (we have 100k tests a month, a test is a `it()` block)

src/store/plugins/geolocation-management.plugin.js
ismailsunni: I am not sure about the original intention, but it seems it's used to prevent trying more than once when doing test. I have updated as you suggested.

tests/cypress/tests-e2e/geolocation.cy.js
ismailsunni: unfortunately, I can only reduce the number of `it` from 3 to 2. I couldn't find a way to put the one testing `POSITION_UNAVAILABLE` in the same `it` block as the other 2. I have tried several tricks to do it (e.g. `cy.reload`, `restore stub`) but it still failed.

pakb: looks like it breaks a bit your testing to split them off...
Then never mind, keep them separated

ismailsunni: yep, when I put into 2 blocks, it pass locally but not in CI. I change it back to 3 blocks now.

pakb (approved):

PR #1193, opened 2025-01-13
PB-1186: Add test for landscape mobile from the portrait one and adapt accordingly.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1186-test-mobile-landscape/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/menuTray.cy.js
+++ b/tests/cypress/tests-e2e/menuTray.cy.js
@@ -213,57 +213,117 @@ function checkScrollbarVisibility(topicsScrollable, activeLayersScrollable) {
 }
 
 describe('Test menu tray ui', () => {
-    it('check correct sizing of menu and autoclosing of menu sections', function () {
-        init(30, 30)
-        checkOpenSections(['topics', 'activeLayers'])
-        measureMenu(true)
-        checkScrollbarVisibility(false, false)
+    context('Portrait mobile', () => {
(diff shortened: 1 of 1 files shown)


tests/cypress/tests-e2e/menuTray.cy.js
```diff
@@ -213,57 +213,117 @@ function checkScrollbarVisibility(topicsScrollable, activeLayersScrollable) {
...
         })
     })
-    it('no scrolling if menus are small enough', () => {
-        init(3, 2)
-        checkOpenSections(['topics', 'activeLayers'])
-        measureMenu(false)
-        checkScrollbarVisibility(false, false)
+    context('Landscape mobile', () => {
+        beforeEach(() => {
+            cy.viewport(568, 320) // Landscape mobile for the default viewport (320 x 568)
```
ltkum, line 273: if I had one thing to say about this code, it's that I don't like hard coded values within the code (I would rather have a constant files for cypress tests). However, this is how it is done in every single test with a different viewport, so I'll suggest we keep it as it is, and I'll create a ticket to gather constants (and functions) in a single file.

ismailsunni: Yes, I completely agree with you. It's better to have it as a constant, like LANDSCAPE_MOBILE, PORTRAIT_MOBILE, etc. Let's wait for a new ticket :)

PR #1194, opened 2025-01-13
PB-400: Add test for hiding the dropdown adter selecting a search result

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-400-test-search-result-is-hidden/index.html)

### DIFF
--- a/tests/cypress/tests-e2e/search/search-results.cy.js
+++ b/tests/cypress/tests-e2e/search/search-results.cy.js
@@ -467,6 +467,8 @@ describe('Test the search bar result handling', () => {
             expect(feature[1]).to.be.approximately(coordinates[1], acceptableDelta)
         })
 
+        // ----------------------------------------------------------------------
+        cy.log('Ensuring the search dialog closes once you have selected an item')
         cy.get('[data-cy="search-results-locations"] [data-cy="search-result-entry"]')
             .as('locationSearchResults')
             .first()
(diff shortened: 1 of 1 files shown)


tests/cypress/tests-e2e/search/search-results.cy.js
```diff
@@ -466,5 +466,18 @@ describe('Test the search bar result handling', () => {
...
         })
+
+        cy.get('[data-cy="search-results-locations"] [data-cy="search-result-entry"]')
+            .as('locationSearchResults')
+            .first()
+            .invoke('text')
+            .then((text) => text.trim())
+            .should('eq', '1530 Payerne')
+
+        // For https://jira.swisstopo.ch/browse/PB-400
```
ltkum, line 477: to be fair, I don't think anyone will check the Jira tickets. I would add the following at the very beginning of the whole section you added : 

```
// ----------------------------------------------------------------------
cy.log('Ensuring the search dialog closes once you've selected an item')
cy.get('[data-cy ...
```
That way, we make a separation between this test case and the previous ones.

ltkum (approved):
I would change one thing, but as this is only adding a log and removing a comment, I don't want you to be blocked by having to wait for another review.

tests/cypress/tests-e2e/search/search-results.cy.js
ismailsunni: ah yes, that's actually correct. We can still also go to the ticket by checking the commit message or PR on the line of code.

PR #1197, opened 2025-01-14
PB-1314: Report Problem drawing on map

sommerfe:


src/modules/drawing/components/useDrawingModeInteraction.composable.js
```diff
@@ -252,7 +253,10 @@ export default function useDrawingModeInteraction({
             drawnFeature.setStyle(geoadminStyleFunction)
             // see https://openlayers.org/en/latest/apidoc/module-ol_interaction_Draw-Draw.html#finishDrawing
             interaction.finishDrawing()
-            store.dispatch('setIsDrawingModified', { value: true, ...dispatcher })
+            // we do not need to share the drawing when the drawing is from the report problem tool
+            if (online.value) {
+                store.dispatch('setIsDrawingModified', { value: true, ...dispatcher })
+            }
```
pakb, line 259: from my understanding, this `isDrawingModified` is used only in a getter in `drawing.store.js` (the one called `showNotSharedDrawingWarning`.
Why not add `online` there too? We don't want to warn the user for non shared drawing when it's offline (not online), so it would make more sense to cut this feature there, instead of divert/twist this `isDrawingModified` flag for this purpose.

sommerfe: yes i could add the `online` to the `showNotSharedDrawingWarning` verification but we definitely need to verify here if online or not because if we set setIsDrawingModified to true anyways we do not get a warning in the report problem drawing but if we open the normal drawing mode and do not draw anything here and close the mode we will still get the warning because the flag `isDrawingModified` got set to true by the report problem mode

pakb (approved):

PR #1200, opened 2025-01-16
PB-1348 : revert Cesium to 1.119.0

pakb:
to fix the issue with label flickering.

Also updating all other libs in the process.
I had to delete and recreate the package-lock.json for things to work properly (related to migration to Node22 maybe?)


ltkum (approved):
Too long, didn't review :)

Do we have the information documented somewhere, that new versions of cypress cause the bug? And is the issue being worked on by the people behind cypress ? It would be a good idea to check on this from time to time to try to still update the library when possible.

pakb:
> Do we have the information documented somewhere, that new versions of cypress cause the bug? And is the issue being worked on by the people behind cypress ? It would be a good idea to check on this from time to time to try to still update the library when possible.

It looks like somebody has some clue on that, but I don't have any idea how the Cesium community works... So wait and see

PR #1201, opened 2025-01-16
PB-995 : add import KML in 3D test through the import popup

pakb:
we only tested pre-loading KMLs through the URL param in 3d/layers.cy.js, so here's a test that import a KML through the Advanced Tool menu section

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-995-add-import-kml-3d-test/index.html)


schtibe (approved):
:+1:

PR #1208, opened 2025-01-17
PB-1255: Unintended segmentation of KML

sommerfe:


ltkum (approved):
Seems to work fine, and the layers which should have chunks still have them.

PR #1209, opened 2025-01-20
PB-1236: infobox doesn't change language when open

sommerfe:


src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
```diff
@@ -26,13 +26,22 @@ const aLayerIsDragged = ref(false)
...
 
+watch(activeLayers, (newActiveLayer) => {
+    if (showLayerDescriptionIndex.value !== null) {
+        onShowLayerDescription(
+            newActiveLayer[showLayerDescriptionIndex.value],
+            showLayerDescriptionIndex.value
+        )
+    }
+})
+
```
ltkum, line 44: We should be watching a language change, not a layer change, here. This will be called if I add or remove a layer from the active layers, which is not what we want.
For example, if I add the public transports stops layer, click somewhere, and add the RBD building layer, the popup will now show the building layer which is not what we want.

We might need to do a quick check with `getIndexOfActiveLayerById` to find the layer again after the lang change, do something like 
```
if (showLayerDescription.value) {
showLayerDescription.value = activeLayers.value[getIndexOfActiveLayerById(showLayerDescription.value)]
}
```

ltkum (changes requested):
While this works for the language change, it introduces a new unwanted behavior and the logic behind it doesn't sit well with me (if we want to trigger something on a language change, we should put the watcher on the language).

src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
ltkum: We discussed about this with Pascal, and we have found a more elegant solution, but it is quite difficult to explain. He will either send you a gist (or make a commit) on this issue.

Also, I made a mistake about the active layers side effects in this case :)

ltkum (approved):
looks good to me :)

PR #1210, opened 2025-01-20
PB-1355 : don't show drawing lost warning when we open download PDFs

pakb:
using window.open to trigger a download (of the print results) was triggering the beforeUnload event in MapView.vue's code, and showing the "you'll lose your drawing" warning, while we stay on the page in the end.

Also fixing some usage of the old $t to the new Composition API equivalent

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1355-no-warning-after-print/index.html)

### DIFF
--- a/src/store/modules/ui.store.js
+++ b/src/store/modules/ui.store.js
@@ -175,6 +175,17 @@ export default {
          * @type Boolean
          */
         showDragAndDropOverlay: false,
+
+        /**
+         * Flag set to true when the app is opening a new tab.
+         *
+         * This helps us decide to show or not show a warning popup for lost changes if the user
+         * closes the tab. In some cases, we are opening a new tab ourselves (print result) and
+         * don't want this popup to show up.
+         *
+         * @type Boolean
+         */
+        isOpeningNewTab: false,
     },
     getters: {
         showLoadingBar(state) {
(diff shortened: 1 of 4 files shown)


src/store/modules/ui.store.js
```diff
@@ -175,6 +175,17 @@ export default {
...
+        /**
+         * Flag set to true when the app is opening a new tab.
+         *
+         * This helps us decide to show or not show a warning popup for lost changes if the user
+         * closes the tab. In some cases, we are opening a new tab ourselves (print result) and
+         * don't want this popup to show up.
+         *
+         * @type Boolean
+         */
+        isOpeningNewTab: false,
```
ltkum, line 188: I don't like this approach, and would prefer we play around with the closeTab event interception, but I can see why this would be simpler to do this way :) We can go with this if you're satisfied with it :)

ltkum (approved):
in the long run, I woulld prefer a cleaner solution without using the store, but for a quick fix, I can let this go :)

src/store/modules/ui.store.js
pakb: for now we will live with it, I'll create a task to sort this out for the next release

PR #1216, opened 2025-01-21
PB-1031: redirect github link to release page

ltkum:
The github link now leads to the specific release

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1031-add-github-link-to-release/index.html)

### DIFF
--- a/src/config/staging.config.js
+++ b/src/config/staging.config.js
@@ -29,7 +29,7 @@ export const APP_VERSION = __APP_VERSION__
  * @type {String}
  */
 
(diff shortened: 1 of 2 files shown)


pakb (commented):
I'd separate the AppVersion button into two buttons, one with only the GitHub icon (and that redirect to https://github.com/geoadmin/web-mapviewer) and the other one containing the version (and redirect to the version if and only if the version is stable (not containing dirty or beta keyword)

pakb (approved):

PR #1221, opened 2025-01-24
PB-1377 : fix bug with line crossing Switzerland

pakb:
when drawing a line crossing without stop LV95 bounds, the function that was splitting this line into in and out chunks wasn't behaving properly (it needed at least one coordinate in to work well).

I've switch the code splitting this up to use TurfJS function instead of a home-made one. Also switching to using the main TurfJS NPM entry, instead of importing each little pieces alone. Our bundler should do a good enough job of removing any part of the bigger package we don't use anyway.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1377-fix-split-if-out-of-bounds/index.html)


pakb:
Will be done in the same work packages as the Elevation profile module

PR #1222, opened 2025-01-24
PB-1004: Drawing polygone - Move delete point to right click

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1004-delete-point-right-click/index.html)


src/modules/drawing/components/useModifyInteraction.composable.js
```diff
@@ -34,7 +34,12 @@ export default function useModifyInteraction(features) {
     const modifyInteraction = new ModifyInteraction({
         features,
         style: editingVertexStyleFunction,
-        deleteCondition: (event) => noModifierKeys(event) && singleClick(event),
+        condition: (event) =>
+            primaryAction(event) ||
+            (event.type === 'pointerdown' &&
+                event.originalEvent.button === 2 &&
+                noModifierKeys(event)), // To delete a point with right click (contextmenu) one has to first select the point and then right click on it, therefore this select vertex condition is needed
+        deleteCondition: (event) => event.type === 'contextmenu' && noModifierKeys(event),
```
pakb, line 42: On second thought, that means we lose the ability to delete on mobile. That's not "that big of a deal" IMHO, but would it be possible to differentiate the binding for mobile and non-mobile ? (using the `isPhoneMode` getter maybe?)

If we could bind the `dblclick` event here for mobile, that would maybe give this feature back for this form factor.

pakb (approved):

src/modules/drawing/components/useModifyInteraction.composable.js
sommerfe: yes i can check if i can do this, we could also do double click for non mobile if you think this is better

pakb (approved):

PR #1225, opened 2025-01-29
PB-577: Sanitize kml content

sommerfe:


ismailsunni:
I believe this PR already covers the ticket from the issue. Unless I missed somehting, it's ready to be reviewed.

packages/mapviewer/src/config/staging.config.js
```diff
@@ -71,3 +71,10 @@ export const REPORT_PROBLEM_HOSTNAMES = [
  * @type {String[]}
  */
 export const WHITELISTED_HOSTNAMES = ['test.map.geo.admin.ch', 'map.geo.admin.ch']
+
+/**
+ * List of blocked file extensions
+ *
+ * @type {String[]}
+ */
+export const BLOCKED_EXTENSIONS = ['exe', 'bat', 'cmd', 'sh', 'msi', 'scr', 'vbs', 'dll']
```
pakb, line 80: Could go into a new `security.config.js` file, as this has not much to do with staging. `WHITELISTED_HOSTNAMES` could also be moved into this new file

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -1,59 +1,87 @@
...
-const { feature } = defineProps({
+const props = defineProps({
     feature: {
         type: SelectableFeature,
         required: true,
     },
 })
 
+const { feature } = toRefs(props)
+
```
pakb, line 23: Vue3.5 automatically adds reactivity to props, no need to `toRefs` them, previous version of the code was fine.

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -1,59 +1,87 @@
...
 const { t } = useI18n()
 
 const store = useStore()
-const hasFeatureStringData = computed(() => typeof feature?.data === 'string')
-const popupDataCanBeTrusted = computed(() => feature.popupDataCanBeTrusted)
+const hasFeatureStringData = computed(() => typeof feature.value?.data === 'string')
+const popupDataCanBeTrusted = computed(() => feature.value.popupDataCanBeTrusted)
```
pakb, line 28: you can remove the `.value` accessor to all `feature` call after you've removed the props `toRefs`

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -1,59 +1,87 @@
...
+                const ext = src.pathname.split('.').pop().toLowerCase()
+                if (BLOCKED_EXTENSIONS.includes(ext)) {
+                    node.outerHTML = blockedExternalContentString
+                    return
+                }
+            } catch (error) {
+                log.error(error)
+                node.outerHTML = blockedExternalContentString
+                return
+            }
```
pakb, line 77: the only thing you do differently in these two cases is how you select the "content" (once `const url = new URL(node.getAttribute('href'))` and the other is `const src = new URL(node.getAttribute('src'))`)
Can you find a way to have the next part of the code only written once?

And maybe explain a little bit the security concern you are dealing here in the code with a little comment before all that.

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -22,38 +22,61 @@ const { feature } = defineProps({
...
+
     DOMPurify.addHook('afterSanitizeAttributes', function (node) {
-        // set all elements owning target to target=_blank
-        if ('target' in node) {
+        // Check the A tag and add target="_blank" and rel="noopener noreferrer" to prevent tabnabbing
+        if (node.tagName === 'A') {
             node.setAttribute('target', '_blank')
-            node.setAttribute('rel', 'noopener')
+            node.setAttribute('rel', 'noopener noreferrer')
+            if (handleNode(node, 'href')) return
```
pakb, line 68: no inline if, use braces please 😉

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -22,38 +22,61 @@ const { feature } = defineProps({
...
+        }
+        // Check the IFRAME tag
+        if (node.tagName === 'IFRAME') {
+            if (handleNode(node, 'src')) return
```
pakb, line 72: this one can be merged, and also use braces please after the if

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -62,15 +85,17 @@ function getIframeHosts(value) {
...
-            return new URL(iframe.src).hostname
-        } catch (error) {
-            log.error(`Invalid iframe source "${iframe.src}" cannot get hostname`, error)
-            return iframe.src
-        }
-    })
-    return hosts.filter((host) => !WHITELISTED_HOSTNAMES.includes(host))
+    return Array.from(dom.getElementsByTagName('iframe'))
+        .map((iframe) => {
+            if (!iframe.src) return null
```
pakb, line 90: no inline if

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -6,13 +6,13 @@ import { useI18n } from 'vue-i18n'
...
 import SelectableFeature from '@/api/features/SelectableFeature.class.js'
-import { WHITELISTED_HOSTNAMES } from '@/config/staging.config'
+import { BLOCKED_EXTENSIONS, WHITELISTED_HOSTNAMES } from '@/config/security.config'
 import FeatureAreaInfo from '@/modules/infobox/components/FeatureAreaInfo.vue'
 import FeatureDetailDisclaimer from '@/modules/infobox/components/FeatureDetailDisclaimer.vue'
 import CoordinateCopySlot from '@/utils/components/CoordinateCopySlot.vue'
 import allFormats from '@/utils/coordinates/coordinateFormat'
 
-const { feature } = defineProps({
+const props = defineProps({
```
pakb, line 15: props should be deconstructed as soon as defined, see [`CONTRIBUTING.md#vue-composition-api`](https://github.com/geoadmin/web-mapviewer/blob/develop/CONTRIBUTING.md#vue-composition-api)

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -22,38 +22,63 @@ const { feature } = defineProps({
...
     DOMPurify.addHook('afterSanitizeAttributes', function (node) {
-        // set all elements owning target to target=_blank
-        if ('target' in node) {
+        // Check the A tag and add target="_blank" and rel="noopener noreferrer" to prevent tabnabbing
+        if (node.tagName === 'A') {
             node.setAttribute('target', '_blank')
-            node.setAttribute('rel', 'noopener')
+            node.setAttribute('rel', 'noopener noreferrer')
+            if (handleNode(node, 'href')){
+                return
```
pakb, line 69: is it really necessary to return something from `handleNode`?
It doesn't really matter if `true` or `false` is returned, it will eventually go to the end of this function anyway.

IMHO `handleNode` could return nothing and that would also work as intended.

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
ismailsunni: ah yes, thanks!

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
ismailsunni: agree.

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -32,28 +32,51 @@ const sanitizedFeatureDataEntries = computed(() => {
...
+                node.outerHTML = blockedExternalContentString
+                return true
+            }
+        } catch (error) {
+            log.error(error)
+            node.outerHTML = blockedExternalContentString
+            return true
+        }
+        return false
+    }
```
pakb, line 61: you can remove the `return` values here, no more needed.

can you also contextualize a bit the `log.error` by giving some error message before the error object? Will be way easier to find in the code when we see the error in the JS console later.

pakb (approved):

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -32,28 +32,47 @@ const sanitizedFeatureDataEntries = computed(() => {
...
 function sanitizeHtml(htmlText, withIframe = false) {
+    const blockedExternalContentString = t('blocked_external_content')
+
+    // Replacing possibly malicious code with `blockedExternalContentString` instead of removing it
+    function handleNode(node, attribute) {
+        try {
+            const url = new URL(node.getAttribute(attribute))
+            const ext = url.pathname.split('.').pop().toLowerCase()
+            if (BLOCKED_EXTENSIONS.includes(ext)) {
+                node.outerHTML = blockedExternalContentString            }
```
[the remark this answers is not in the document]
ismailsunni: Indeed, it should go to the next line.

PR #1226, opened 2025-01-30
PB 1052: Add 3d tooltips for buildings

ltkum:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1052-add-3d-building-tooltips/index.html)


src/modules/i18n/locales/de.json
```diff
@@ -9,6 +9,28 @@
...
+    "Mauer gross": "Mauer gross",
+    "Mauer gross gedeckt": "Mauer gross gedeckt",
+    "Offenes Gebaeude": "Offenes Gebäude",
+    "Sakraler Turm": "Sakrales Gebäude",
+    "Sakrales Gebaeude": "Sakrales Gebäude",
+    "Treibhaus": "Treibhaus",
+    "Turm": "Turm",
+    "Unterirdisches Gebaeude": "Unterirdisches Gebäude",
+    "Verbindungsbruecke": "Verbindungsbrücke",
+    "Verbindungsbrücke": "Verbindungsbrücke",
```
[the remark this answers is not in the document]
ltkum: This is the possible contents of the "building type" fields in the layer. If you go a `building.getProperty('OBJEKTART')`, it will return a german text, not a key. The layer content is not localised and is only in german.

Also, there are some little modifications between the text we receive and the text we want to show.

src/modules/map/components/cesium/CesiumPopover.vue
```diff
@@ -128,16 +128,18 @@ export default {
...
-                this.getViewer().scene,
-                Cartesian3.fromDegrees(
-                    this.wgs84Coordinates[0],
-                    this.wgs84Coordinates[1],
-                    this.coordinatesHeight,
-                    Ellipsoid.WGS84,
-                    tempCartesian3
-                )
-            )
+            const cartesianCoords = SceneTransforms.wgs84ToWindowCoordinates
```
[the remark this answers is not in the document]
ltkum: In some situations, the Scenetransforms object wouldn't have a `wgsToWindowCoordinates` function, which caused this to crash. 
I am currently running Cesium 1.119.0 locally.

I'll look into this to see why the scenetransforms could end up not having such a member, and how to handle it in a more elegant manner

src/modules/i18n/locales/de.json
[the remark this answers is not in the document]
pakb: If it's not localized, that's too bad, but we don't localize on the front end side.
Let's just show in German and be done with it. I don't want specific i18n keys for a specific layer.

packages/mapviewer/src/api/layers/GeoAdmin3DLayer.class.js
```diff
@@ -17,15 +17,16 @@ import { get3dTilesBaseUrl } from '@/config/baseUrl.config'
...
      * @param {string} layerId The ID of this layer
+     * @param {string} layerName The Name of this layer. Default to the layer Id
      * @param {String | null} urlTimestampToUse If this layers' JSON is stored in a dedicated timed
      *   folder, it can be described with this property. This will be added at the end of the URL,
      *   before the /tileset.json (or /style.json, depending on the layer type)
      * @param {boolean} use3dTileSubFolder If the JSON file stored in the /3d-tiles/ sub-folder on
      *   the S3 bucket
      */
-    constructor(layerId, urlTimestampToUse = null, use3dTileSubFolder = true) {
+    constructor(layerId, layerName = null, urlTimestampToUse = null, use3dTileSubFolder = true) {
```
pakb, line 27: Can you take the opportunity to switch to having all args for this constructor declared as a single object, and de-structure it as a first line?

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -1,27 +1,40 @@
...
 import { ClickInfo, ClickType } from '@/store/modules/map.store'
+import { createPixelExtentAround } from '@/utils/extentUtils'
 import { identifyGeoJSONFeatureAt } from '@/utils/identifyOnVectorLayer'
 
 const dispatcher = { dispatcher: 'CesiumInteractions.vue' }
 
 const getViewer = inject('getViewer')
 
+const hoveredHighlightPostProcessor = ref(null)
+const clickedHighlightPostProcessor = ref(null)
```
pakb, line 31: you might not want to store that as ref, as it will wrap them in reactivity proxies

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -31,14 +44,32 @@ const visiblePrimitiveLayers = computed(() =>
...
         viewer.screenSpaceEventHandler.setInputAction(
             onContextMenu,
             ScreenSpaceEventType.RIGHT_CLICK
         )
+        viewer.screenSpaceEventHandler.setInputAction(onMouseMove, ScreenSpaceEventType.MOUSE_MOVE)
     }
 })
+function initialize3dHighlights() {
+    hoveredHighlightPostProcessor.value = PostProcessStageLibrary.createEdgeDetectionStage()
+    hoveredHighlightPostProcessor.value.uniforms.color = Color.BLUE
```
pakb, line 64: Please use our colors defined in `src/modules/map/components/cesium/utils/highlightUtils.js`

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -31,14 +44,32 @@ const visiblePrimitiveLayers = computed(() =>
...
+    hoveredHighlightPostProcessor.value.uniforms.length = 0.01
+    hoveredHighlightPostProcessor.value.selected = []
 
+    clickedHighlightPostProcessor.value = PostProcessStageLibrary.createEdgeDetectionStage()
+    clickedHighlightPostProcessor.value.uniforms.color = Color.LIME
```
pakb, line 69: same

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -53,7 +84,46 @@ function getCoordinateAtScreenCoordinate(x, y) {
...
+    })
+    return feature
+}
+
+function handleClickHighlight(features, coordinates) {
+    clickedHighlightPostProcessor.value.selected = []
 
+    if (hoveredHighlightPostProcessor.value.selected.length > 0) {
+        features.push(
+            createBuildingFeature(hoveredHighlightPostProcessor.value.selected[0], coordinates)
```
pakb, line 119: why only the first value? The highlighter shows sometime more than one "roof" under the cursor, it would feel better to have both buildings show up in the selection

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
pakb, line 1: the post processor don't fill all edges of buildings but only the "border" so to speak, is there any options to make it color all edges of the geometry ?

pakb (commented):
As said in the previous review's comment : we don't do layer specific layer/translation on the viewer side, that's something that has to be in the data, or it isn't.
Please remove all translation keys that aren't used directly by the viewer's code, and let the German text of buildings go through

src/modules/i18n/locales/de.json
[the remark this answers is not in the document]
ltkum: fine by me. I removed the translation keys and the translation call on the content

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -53,7 +87,42 @@ function getCoordinateAtScreenCoordinate(x, y) {
...
     return coordinates
 }
+function createBuildingFeature(building, coordinates) {
+    const id = building.getProperty('EGID') ?? building.getProperty('UUID')
+    const data = {
+        EGID: building.getProperty('EGID') ?? '-',
+        building_type: building.getProperty('OBJEKTART') ?? '-',
+        building_height: building.getProperty('GESAMTHOEHE') ?? '-',
+        max_roof_height: building.getProperty('DACH_MAX') ?? '-',
+        ground_level: building.getProperty('GELAENDEPUNKT') ?? '-',
```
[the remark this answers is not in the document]
ltkum: We can, but we were requested to get those attributes :)

With that being said : 
 - We're waiting for an integration to get a standard tooltip anyway, so me might as well not make a special case here.

packages/mapviewer/src/store/modules/cesium.store.js
```diff
@@ -1,17 +1,41 @@
...
+    layerId: CESIUM_CONSTRUCTIONS_LAYER_ID,
+    layerName: '3d_constructions',
+    urlTimestampToUse: 'v1',
+    use3dTileSubFolder: false, // buildings JSON has already been migrated to the new URL nomenclature
+})
+
+// TODO :
+// move this to cesium config
+// create "local" locales translations json with layer prefixes for the keys
+// generalize generation
```
pakb, line 38: that might be your reminder todo list, can be removed no?

packages/mapviewer/src/store/modules/cesium.store.js
```diff
@@ -60,6 +84,17 @@ export default {
...
         isViewerReady: false,
+        /**
+         * An object containing the fetchable parameters for each cesium layer
+         *
+         * It will be used in the following manner : id is what property of the feature is used for
+         * the id of a feature in the layer. data represents the keys for each data (which will be
+         * translated), and which data property will be associated to it (this will never be
+         * translated)
+         *
+         * @type Object
```
pakb, line 95: type can be changed to the new `@type LayerTooltipConfig[]`

pakb (commented):
I'd like to find a way to describe that some buildings don't have some data, instead of showing nothings.

You've listed the keys we should be finding in each building am I right?
We could use (and change its name before that) the translation key `time_slider_legend_tooltip_no_data` (=> `no_data` will be a better key) to signify the user this building has no data (yet)

ltkum:
> I'd like to find a way to describe that some buildings don't have some data, instead of showing nothing.
> 
> You've listed the keys we should be finding in each building am I right? We could use (and change its name before that) the translation key `time_slider_legend_tooltip_no_data` (=> `no_data` will be a better key) to signify the user this building has no data (yet)

I would agree with you that there should be something else than a dash (`-`). however, it was a request from the topo that a dash should appear (instead of a "data not available").

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -125,18 +128,16 @@ function getIframeHosts(value) {
...
-                    class="fw-bold"
-                >
-                    {{ t(key) }}
+                <div class="d-flex flex-row justify-content-between">
+                    <div class="fw-bold">{{ t(key) }}</div>
+                    <!-- eslint-disable vue/no-v-html-->
+                    <div
+                        data-cy="feature-detail-description-content"
+                        class="text-end pr-3"
+                        v-html="t(value)"
```
pakb, line 137: why switching to a `v-html` for feature details?
that might create new security issues with external layers (possible phishing vector)

ltkum: I did not switch to `v-html`, it was already a `v-html` and I added a flexbox around the two divs to have them on a horizontal layout and added a translation around it. This is already used as a v-html for our external layers.

pakb (approved):

PR #1229, opened 2025-02-03
PB-1384 Contributing to the modularization

schtibe:
Mostly work for fixing some tests.

- Install Cypress manually
- Improve the tooltip composable, allowing for lazy referencing the DOM elements. This fixes a bug in the menu layer items where there were too many tooltips instantiated
- Added some more log output in the cypress commands
- Fixed the url transformation in the e2e tests that resulted in `feature=[object Object]` URLs

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1384-profile-npm-modules-lthee/index.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "@turf/turf": "^7.2.0",
+    "cypress": "^14.0.1",
     "vue": "^3.5.13",
     "vue-i18n": "^11.1.0"
   },
--- a/packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
+++ b/packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
@@ -6,7 +6,7 @@
 
 import { FontAwesomeIcon } from '@fortawesome/vue-fontawesome'
 import log from 'geoadmin/log'
-import { computed, onMounted, ref, toRefs } from 'vue'
+import { computed, onMounted, ref, toRefs, useTemplateRef } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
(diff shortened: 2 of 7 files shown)


packages/mapviewer/src/utils/composables/useTippyTooltip.js
```diff
@@ -110,15 +119,21 @@ export function useTippyTooltip(
     }
 
     function refreshTippyAttachment() {
+        // if the selector is a function, call it.
```
schtibe, line 122: This was a byproduct of debugging the tooltips. If we don't have this we sometimes get too many tooltip instances in the DOM because the selector was too "wide".

package.json
```diff
@@ -30,6 +30,7 @@
   },
   "dependencies": {
     "@turf/turf": "^7.2.0",
+    "cypress": "^14.0.1",
```
pakb, line 33: I'd maybe move that to `devDependencies`

packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
```diff
@@ -62,7 +62,7 @@ const emit = defineEmits(['showLayerDescriptionPopup', 'toggleLayerDetail', 'mov
 const store = useStore()
 const { t } = useI18n()
 
-useTippyTooltip('.menu-layer-item [data-tippy-content]')
+const menuLayerItem = useTemplateRef('menuLayerItem')
```
pakb, line 65: That's huge, this new `useTemplateRef` from Vue 3.5.
We should move all our old `... = ref(null)` from before to this new notation

pakb: created a ticket for that https://jira.swisstopo.ch/browse/PB-1428

pakb (approved):

packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersListItem.vue
schtibe: Yes, I didn't know that either! Makes it so much better

PR #1227, opened 2025-01-31
PB-1384 : modularize proj, number and coordinate utils

pakb:
in prep work to the elevation profile module, I started modularizing what will be used by both the viewer and the future elevation profile component.

I used the opportunity of having a smaller scoped project to switch to Typescript entirely all the utilities. Migration to ESLint9 was also done here, meaning there are some changes because of new linter rules. Some ESLint8 plugins weren't available in ESLint9 so I had to find equivalents, but they don't have exactly the same way of linting

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1384-profile-npm-modules/index.html)

PR #1232, opened 2025-02-07
Fix the eslint/prettier conflict

schtibe:
Fix eslint/prettier relationship
    
Provide a combination of eslint and prettier that sort of respects `vue/max-attributes-per-line`.
    
Configure eslint and prettier so that the attributes in HTML code are put on single lines if there are more than two. The formatting is done entirely by Prettier, since [eslint doesn't want to be a formatting tool anymore](https://eslint.org/blog/2023/10/deprecating-formatting-rules/)
    
This also disables the running of prettier from eslint, which is the [recommended way](https://github.com/vuejs/eslint-config-prettier?tab=readme-ov-file#use-separate-commands-for-linting-and-formatting)

There is some further info in this [gist](https://gist.github.com/schtibe/0044efa67b6a33802a77d16fe49aa6ec) on how I came to this conclusion.

We will also have to add `npm run format:check` to the pipeline.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-prettier-eslint-conflict/index.html)


eslint.config.mjs
pakb, line 1: `'vue/max-attributes-per-line': ['error', { singleline: 1, multiline: 1 }],` is now totally ignored, I don't mind 2 elements being inlined, but I liked the "split if more than 2" approach from https://eslint.vuejs.org/rules/max-attributes-per-line

pakb: https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#plugins
in fact this eslint-plugin-prettier removes most if not all other rules we've set so far, don't know what should be the way to go here (embrace prettier and configure it to our liking, or remove it and use multiple plugins to achieve what we want?)

schtibe: > `'vue/max-attributes-per-line': ['error', { singleline: 1, multiline: 1 }],` is now totally ignored, I don't mind 2 elements being inlined, but I liked the "split if more than 2" approach from https://eslint.vuejs.org/rules/max-attributes-per-line

Right :thinking:  My line of thinking was that it is now as it was before your modularization change (e.g. it was changed from [single to multiline](https://github.com/geoadmin/web-mapviewer/pull/1227/files#diff-c42a3d331cfec0ddd4d441314ee2f35fe2fca87ef6ca0fc6b7c49c5dcaaed3abL87) in the modularization PR now this would put it back to [single line](https://github.com/geoadmin/web-mapviewer/pull/1232/files#diff-c42a3d331cfec0ddd4d441314ee2f35fe2fca87ef6ca0fc6b7c49c5dcaaed3abR87))

schtibe: > https://github.com/prettier/eslint-config-prettier?tab=readme-ov-file#plugins in fact this eslint-plugin-prettier removes most if not all other rules we've set so far, don't know what should be the way to go here (embrace prettier and configure it to our liking, or remove it and use multiple plugins to achieve what we want?)

I guess that's because it's handing the formatting over to prettier? We've had prettier before, didn't we? 

Personally I like to keep formatting tools as close to the standard and also use the tools that are solicited by the maintainers. Of course this is mostly your product so in the end I'll follow your opinion :) 

I will put in a little bit more work, trying to understand the mechanisms more deeply. I just had a look at a bare vue setup, and stumbled on [this](https://github.com/vuejs/eslint-config-prettier?tab=readme-ov-file#use-separate-commands-for-linting-and-formatting).

packages/geoadmin/src/utils/coordinates.ts
```diff
@@ -14,9 +14,10 @@ type Single3DCoordinate = [number, number, number]
  *
  * @param coordinate The raw coordinate as array.
  * @param digits Decimal digits to round to.
- * @param [withThousandsSeparator=true] If thousands should be separated with a single quote character
+ * @param [withThousandsSeparator=true] If thousands should be separated with a single quote
+ *   character. Default is `true`
  * @param [enforceDigit=false] If set to true, we want to have that many figures after the period.
- *   Otherwise, we don't care.
+ *   Otherwise, we don't care. Default is `false`
```
schtibe, line 20: I don't know where this now comes from :thinking:

.prettierignore
```diff
@@ -1 +1,4 @@
-src/modules/drawing/lib/modifyInteraction.js
+package.json
+package-lock.json
+**/*tests/samples/**
+**/*tests/cypress/fixtures/**
```
pakb, line 4: having fixtures "pretty" isn't that bad, no?

packages/mapviewer/src/modules/infobox/components/FeatureAreaInfo.vue
```diff
@@ -62,7 +62,8 @@ const humanReadableArea = computed(() => {
             <span
                 class="align-middle"
                 data-cy="feature-area-information"
-            >{{ humanReadableArea }}<sup>2</sup></span>
+                >{{ humanReadableArea }}<sup>2</sup></span
+            >
```
pakb, line 66: don't know if I like that...
We could add a linebreak after the span declaration IMHO

packages/mapviewer/src/modules/drawing/components/DrawingToolboxButton.vue
```diff
@@ -59,9 +59,8 @@ function setDrawingMode() {
         <small
             v-if="!isPhoneMode"
             class="d-sm-block"
-        >{{
-            t(`draw_${drawingMode.toLowerCase()}`)
-        }}</small>
+            >{{ t(`draw_${drawingMode.toLowerCase()}`) }}</small
+        >
```
pakb, line 63: Same here but with the <small> tag, I usually tend to just line break after the declaration...

packages/mapviewer/src/modules/map/components/LocationPopupPosition.vue
```diff
@@ -164,7 +164,8 @@ async function updateHeight() {
                 <a
                     href="https://epsg.io/4326"
                     target="_blank"
-                >{{ WGS84Format.label }}</a>
+                    >{{ WGS84Format.label }}</a
+                >
```
pakb, line 168: yep, I really don't like that it pushes the closing `>` to the next line 😞

packages/mapviewer/src/modules/map/components/openlayers/debug/OpenLayersDeviceOrientationDebugInfo.vue
```diff
@@ -103,9 +103,7 @@ async function copyValue(event, value) {
                     v-for="parameter in parameters"
                     :key="parameter.title"
                 >
-                    <div class="text-decoration-underline fw-bold">
-                        {{ parameter.title }}:
-                    </div>
+                    <div class="text-decoration-underline fw-bold">{{ parameter.title }}:</div>
```
pakb, line 106: I'm confused, it broke many things into multiple line, and for this one, it does the opposite?

pakb (commented):
some little comments, I'm still digging into it

packages/mapviewer/scripts/check-layer-providers-results/invalid_providers.yaml
pakb, line 1: don't know if this should be included (maybe our .gitignore isn't up to date anymore with the modularization)

packages/mapviewer/public/icon.svg
pakb, line 1: maybe we shouldn't touch this file, except if we minify it in the build process. Here we are increasing its size for nothing

packages/mapviewer/index.html
pakb, line 1: maybe we shouldn't touch this file, except if we minify it in the build process. Here we are increasing its size for nothing

packages/mapviewer/src/assets/svg/swiss-flag.svg
pakb, line 1: maybe we shouldn't touch this file, except if we minify it in the build process. Here we are increasing its size for nothing

packages/mapviewer/src/modules/i18n/locales/de.json
pakb, line 1: weird that you need to change translations for this type of work, are you sure it's not a rebase problem?

.prettierrc
```diff
@@ -13,5 +13,7 @@
                 "tabWidth": 2
             }
         }
-    ]
+    ],
+    "singleAttributePerLine": true
+    // "htmlWhitespaceSensitivity": "ignore"
```
pakb, line 18: can that be cleaned up?

pakb (approved):

PR #1234, opened 2025-02-10
Install vite-plugin-vue-devtools

schtibe:
In another, fresh vue installation I noticed this plugin. I did a little research gra what it is, and it's apparently an integrated vue dev tools. On their website, the [state](https://devtools.vuejs.org/guide/vite-plugin) that this is the preferred way of using the dev tools.

What this does it it renders this little button at the bottom of the page with which one can access the dev tools:
image

I don't know if this is a good idea or not, let me hear your opinions. The dev tools in my firefox don't work half of the time. 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-vite-plugin-vue-devtools/index.html)

### DIFF
--- a/packages/mapviewer/package.json
+++ b/packages/mapviewer/package.json
@@ -116,6 +116,7 @@
         "vite-node": "catalog:",
         "vite-plugin-static-copy": "catalog:",
         "vitest": "^3.0.5",
+        "vite-plugin-vue-devtools": "catalog:",
         "write-yaml-file": "catalog:",
         "yargs": "catalog:"
     },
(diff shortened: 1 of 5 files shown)


pakb (commented):
I used to use a ton the chrome extension called Vue devtools, but it got quite power hungry and made my browser stale so I removed it. It was also not really maintained at the pace Vue was going, so if there's a new "kid in the block" that fulfill this role, I'm all for it!

Would it be possible to activate this thing on our DEV and INT staging? I don't see it pops up on the test link

packages/mapviewer/package.json
```diff
@@ -116,6 +116,7 @@
         "vite-node": "catalog:",
         "vite-plugin-static-copy": "catalog:",
         "vitest": "^3.0.5",
+        "vite-plugin-vue-devtools": "^7.7.1",
```
pakb, line 119: can you move this version definition to the catalog and use `"catalog:" here instead?

pakb (approved):

packages/mapviewer/package.json
schtibe: :+1: :heavy_check_mark:

PR #1235, opened 2025-02-11
PB-1443: re-introduce zoom buttons on mobile

ltkum:
Issue : Some people expressed that removing the zoom buttons on mobile was an issue for them, either regarding the hardware they are using or because of outside issues which stopped them from using the mapviewer correctly.

Fix : We bring the zoom buttons back.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1444-put-back-the-zoom-buttons-on-mobile/index.html)


pakb (approved):
Please fix the lint 😉

PR #1236, opened 2025-02-12
PB-1384 : split geoadmin modules into individual modules

pakb:
This was necessary so that I could move the coordinate utils and number utils, whilst having the projection definition as a module. As coordinate utils is using some of proj definition, I thought it would make more sense to create a big "coordinates" module that do all things related coordinates (proj, transformation etc...)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1384-split-modules/index.html)


schtibe (approved):
Nice!

PR #1240, opened 2025-02-13
PB-1200: Add Geolocation Test in 3D Map

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1200-geolocation-test-3d/index.html)

Can I someone check that the geolocation in 3D works fine inside Switzerland? @pakb @hansmannj or @ltkum 

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1200-geolocation-test-3d/index.html)

### DIFF
--- a/packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
+++ b/packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
@@ -39,13 +39,13 @@ function setCenterIfInBounds(store, center) {
                             pitch: -90,
                             roll: 0,
                         },
-                        dispatcher: self,
-                    })
-                }
-                    store.dispatch('setZoom', {
-                        zoom: store.state.position.projection.get1_25000ZoomLevel(),
                         ...dispatcher,
                     })
+                }
+                store.dispatch('setZoom', {
+                    zoom: store.state.position.projection.get1_25000ZoomLevel(),
+                    ...dispatcher,
+                })
             }
         }
     } else {
--- a/packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
+++ b/packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
@@ -1,6 +1,7 @@
 import { LV95, WGS84 } from '@geoadmin/coordinates'
 import log from '@geoadmin/log'
 import { round } from '@geoadmin/numbers'
+import { Math as CesiumMath } from 'cesium'
 import { isEqual } from 'lodash'
 import proj4 from 'proj4'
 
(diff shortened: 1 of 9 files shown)


hansmannj:
> [Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1200-geolocation-test-3d/index.html)
> 
> Can I someone check that the geolocation in 3D works fine inside Switzerland? @pakb @hansmannj or @ltkum
> 
> [Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1200-geolocation-test-3d/index.html)

Thanks @ismailsunni, just tested and it works fine for me :+1:

packages/mapviewer/tests/cypress/tests-e2e/3d/geolocation.cy.js
pakb, line 1: This looks a lot like a full copy-paste from the 2D geolocation tests.

I'd prefer every geolocation tests stays in the `geolocation.cy.js` file and that you find a way to share most of the logic between the two contexts (2D and 3D), especially the part that handles browser permissions, as it looks like the biggest problem here

pakb (commented):
Can you check if it's easy to zoom in when in 3D and we activate the geolocation? Maybe some kind of zoom to extent or something

ismailsunni:
@pakb what do you mean with easy to zoom? I have added a test where the zoom in or zoom out button is cliced, and check the height of the camera (to see if it's really zooming/out) while keeping the current location.

Or do you have something in mind?

pakb (commented):
as discussed yesterday, please make sure that activating geolocation for the first time while in 3D zooms the view over the user's location

ismailsunni:
Hi @pakb 
I made two decision/assumptions in this PR to implemetn the zoom to geolocation in 3D:
- The camera will be pointing down, it's complicated to keep the camera's orientation and calculating the correct z-value for the camera position
- I need to calculate the height with the cesium viewer clientWidth. But to do this, I need the viewer object, which I don't have the access from the plugin (not sure if it's possible). So, I use the store width, I assume this is the same one as the viewer's width = window's widht.

packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
```diff
@@ -27,10 +28,24 @@ function setCenterIfInBounds(store, center) {
...
+                            x: store.state.position.camera.x,
+                            y: store.state.position.camera.y,
+                            z: store.state.position.camera.z,
+
+                            heading: -CesiumMath.toRadians(store.state.position.rotation),
+                            pitch: -90,
+                            roll: 0,
+                        },
+                        dispatcher: self,
+                    })
```
pakb, line 43: is it to reset to camera to top down view?

your `dispatcher` here is a bit weird, maybe swap it for `...dispatcher`

packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
```diff
@@ -27,10 +28,24 @@ function setCenterIfInBounds(store, center) {
...
+                }
+                    store.dispatch('setZoom', {
+                        zoom: store.state.position.projection.get1_25000ZoomLevel(),
+                        ...dispatcher,
+                    })
```
pakb, line 48: not sure the linter worked as intended, you added a `tab` to those lines but these tabs are not needed 🤔

packages/mapviewer/src/store/plugins/geolocation-management.plugin.js
ismailsunni: Yes, it's to reset the angle.

pakb (approved):

PR #1241, opened 2025-02-13
PB-1436 Switch to using pnpm instead of npm

schtibe:
# PNPM

pnpm is a system that re-uses node_modules across multiple folders and projects. This results in way faster installs and less disk usage

### Usage
The main difference is that one needs to run `pnpm install` instead of `npm install`, `pnpm run` instead of `npm run` etc.

### Lock files
There are no `package-lock.json` files anymore. Instead, we now have `pnpm-lock.yaml`. Package-locks are now gitignored.

### Catalogs and Workspaces

tl;dr: Add packages to `pnpm-workspace.yaml` and in our sub-packages reference with `catalog: `.

The file `pnpm-workspaces.yaml` is where this workspaces is defined. For one, it defines the packages that are included in this mono-repo. This helps with certain command and especially with cross-linking the packages.

Secondly, there's a catalog definition. In there are the packages defined, that should be the same across all of the sub-packages. This helps preventing the problem that one package might update to a newer version while another one stays at the same, resulting in a conflict. Generally speaking, we're now adding the 3rd party packages to the workspace file and reference them in the sub-packages.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1436-pnpm/index.html)


CONTRIBUTING.md
```diff
@@ -44,42 +44,44 @@ This tag will be a bump in :
...
+| `pnpm run test:component`          | Opens up the cypress component tests                                                                                                                                       |
+| `pnpm run test:component:headless` | Run cypress component tests in headless mode                                                                                                                               |
+| `pnpm run test:component:ci`       | Run cypress component tests                                                                                                                                                |
+| `pnpm run update:translations`     | Update translation files according to our Google Spreadsheet. See [above](#tooling-for-translation-update) for required tools.                                             |
 
 All scripts commands starting a webserver or using one (`dev` and all things related to cypress) will determine the port to use by looking for the next one available starting at `8080`.
 
 ### What about `package-lock.json` file?
```
pakb, line 76: you might need to update the file name here

package.json
```diff
@@ -1,65 +1,71 @@
...
-        "test:e2e:ci": "npm run test:e2e:ci --workspace=web-mapviewer",
-        "test:e2e:headless": "npm run test:e2e:headless --workspace=web-mapviewer",
-        "test:component": "npm run test:component --workspace=web-mapviewer",
-        "test:component:ci": "npm run test:component:ci --workspace=web-mapviewer",
+        "dev": "pnpm run dev --workspace=web-mapviewer",
+        "build-all": "pnpm run -r build",
+        "build:dev": "pnpm -r run build:dev",
+        "build:int": "pnpm -r run build:int",
+        "build:prod": "pnpm -r run build:prod",
+        "preview": "pnpm run -r --filter=!web-mapviewer build --filter= && pnpm run -r --filter=web-mapviewer preview",
```
pakb, line 10: there's an empty `--filter=` here, is that wanted?

packages/geoadmin-log/package.json
```diff
@@ -19,16 +17,10 @@
...
-        "vite-plugin-dts": "^4.5.0",
-        "vitest": "^3.0.4",
-        "vue-tsc": "^2.2.0"
+        "build": "pnpm run type-check && pnpm run generate-types && vite build",
+        "build:dev": "pnpm run build -- --mode development",
+        "build:int": "pnpm run build -- --mode integration",
+        "build:prod": "pnpm run build -- --mode production"
     },
     "peerDependencies": {
         "proj4": "^2.15.0"
```
pakb, line 26: can we also move that into the pnpm catalog?

pakb (approved):
as would see our greatest Kazakh hero : very niiice!

package.json
schtibe: Nope that's a mistake

packages/geoadmin-log/package.json
schtibe: Yes indeed!

PR #1247, opened 2025-02-18
PB-1209: Fix no feature tooltip when select feature from search.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1209-select-feature-no-info-from-search/index.html)


pakb (commented):
can you add relevant tests in `search-results.cy.js` so that we make sure this is there to stays?

PR #1249, opened 2025-02-20
PB-1004 : only use right click to trigger "delete point" action

pakb:
users on touch devices could trigger a delete by double tapping on the last point they were drawing, which wasn't ideal.

As I've run out of ideas on how to fix this on mobile, I've simply removed the possibility to delete a point on mobile as we lack "gestures" to achieve it.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1004-revert-use-of-right-click/index.html)


schtibe (commented):
I think the help text needs updating: 
image

Also, right clicking on a point that isn't an edge or on the line deletes the last point. Is that intended?

PR #1250, opened 2025-02-20
Don't render invisible layer catalogue items

schtibe:
Don't render the invisible layer catalogue items. Only render them when the nested navigation is un-collapsed.
This produces quite a performance gain, as many sub-sequent components don't need rendering. This includes `<TextTruncate>`, where each of them creates a ResizeObserver on the DOM. Without this change we have ~1100 resize observers, with the change there are only ~250.

I will investigate those ~250 maybe in another attempt, some of them might be resolved when switching to the new tooltip.

This results in quite the performance gain, mostly noticable in development mode.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-less-text-truncate/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/menu/components/LayerCatalogueItem.vue
+++ b/packages/mapviewer/src/modules/menu/components/LayerCatalogueItem.vue
@@ -243,7 +243,7 @@ function containsLayer(layers, searchText) {
 
 <template>
     <div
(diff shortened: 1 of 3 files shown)


pakb (approved):
Good old `v-show` playing a trick on us, nice finding

PR #1251, opened 2025-02-20
this is to allow luca to do screenshots [DO NOT MERGE]

ltkum:


packages/mapviewer/src/config/staging.config.js
pakb, line 1: I'd propose to let use decide to "opt-out" of the "dev site warning" in the DebugToolbar, this way we could check our layout without it whenever we want

ltkum: That's a good idea. i wanted to give Luca a very quick layout so he could make screenshots, but I can adapt that.

ltkum:
we'll close this and do https://swissgeoplatform.atlassian.net/browse/PB-1551 to allow the CMS users to do their screenshots.

PR #1252, opened 2025-02-21
PB-1137 : create a full HTML confederation text (for the logo)

pakb:
We will want to re-size that part of the UI when we print, and using a flatten PNG for that part of the logo will be a problem.
The Swiss flag is already a SVG, so I thought why not change the text to also be rendered by HTML/CSS.

Importing Frutiger Light to the project for the confederation logo/text, and switching the whole code base to NotoSans (according to CD Bund https://www.bk.admin.ch/bk/de/home/dokumentation/cd-bund.html)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1137-confederation-text-as-html/index.html)


pakb:
> One thing that I noted is that the Confederation lines seem to have a bit more font-weight in the original. I checked with other `admin.ch` sites, but I doesn't seem to be consistent.

I tried to stick the most I could to the PDF document (which might be outdated BTW...), and then I compared that to what we had. Our "main text" was bolder (which isn't in the spec), maybe we could lower the font weight yeah...

> (Also the margin between the names and `In collaboration with the cantons` increased a bit. Just mentioning that in case it's not intentional.)

That is intentional, you are supposed to add a full blank line between the two blocks (which I did by adding a top margin of one line height)

PR #1253, opened 2025-02-24
PB-1359: No drawing warning outside the drawing menu

ismailsunni:


packages/mapviewer/src/modules/drawing/DrawingModule.vue
```diff
@@ -195,12 +201,26 @@ onBeforeUnmount(() => {
...
+    if (showNotSharedDrawingWarning.value) {
+        showNotSharedDrawingWarningModal.value = true
+        // This provokes the alert message to appear when trying to close the tab.
+        // During Cypress tests this causes the test to run indefinitely, so to prevent this we skip the alert.
+        if (!IS_TESTING_WITH_CYPRESS) {
+            event.returnValue = true
+            event.preventDefault()
+            return
+        }
+    }
```
pakb, line 221: are we testing the popup in Cypress?
if not, we could entirely skip the change of `showNotSharedDrawingWarningModal` by editing the if case so that
```
if (!IS_TESTING_WITH_CYPRESS && showNotShareDrawingWarning.value) {
    // ...
}
```

pakb (approved):

PR #1254, opened 2025-02-24
PB-1436 Specify packageManager in package.json

schtibe:
The build on CI fails because the pnpm command adds the `packageManager` field to package.json, making the working tree dirty. This subsequently prevents the build from succeeding.

As it seems, having this field is actually desirable, thus adding it here.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-1436-pnpm/index.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,5 +68,5 @@
             "sharp"
         ]
     },
-    "packageManager": "pnpm@10.4.1"
+    "packageManager": "pnpm"
 }
--- a/package.json
+++ b/package.json
@@ -67,5 +67,6 @@
             "cypress",
             "sharp"
         ]
-    }
+    },
+    "packageManager": "pnpm@10.4.1"
 }


pakb (approved):
let's see what the CI has to say about it 😉

schtibe:
So apparently a specific version is needed. This seems to be a `corepack` thing: when the version is specified, then the exact version will be used, or that's how I understand it at this moment.

Maybe we shouldn't be using `corepack` on the CI yet. Or maybe we should be using it locally, but I don't quite understand how that's supposed to work. Also, Corepack is flagged `experimental`, so my intuition leans more towards not using it on the CI.

schtibe:
Closing this in favour of [not using corepack](https://github.com/geoadmin/infra-terraform-bgdi/pull/768/commits/818c0bb15d021a1b244692ec2aeb9c60badccaee) until this is officially deemed more stable

PR #1255, opened 2025-02-27
PB-1456: Fix drag layer issue

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1456-duplicated-layer-when-dragging/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
+++ b/packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
@@ -43,7 +43,7 @@ onMounted(() => {
         onStart: function () {
             aLayerIsDragged.value = true
         },
-        onEnd: function (event) {
+        onUpdated: function (event) {
             aLayerIsDragged.value = false
             const { newIndex, oldIndex } = event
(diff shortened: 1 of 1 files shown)


packages/mapviewer/src/modules/menu/components/activeLayers/MenuActiveLayersList.vue
```diff
@@ -46,7 +47,27 @@ onMounted(() => {
...
+                    const nonDraggableChildren =
+                        activeLayersList.value.querySelectorAll('[draggable="false"]')
+                    if (nonDraggableChildren.length > 0) {
+                        nonDraggableChildren.forEach((child) => child.remove())
+                        log.debug('Non-draggable children removed:', nonDraggableChildren)
+                    } else {
+                        log.debug('No non-draggable children found')
+                    }
+                    onMoveLayer(reverseIndex(oldIndex), reverseIndex(newIndex))
+                })
```
pakb, line 67: works ok. nice find!
can you add a little comment to explain why we are doing this search and destroy on draggable=false?
Just make it clear that is is to fix this bug so we don't remove it in the future because we don't understand why 😉

pakb (approved):

PR #1257, opened 2025-03-10
PB-1480: Fix bug with swissearch and crosshair together

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html)

## Bug description

The issue occurs when the swisssearch value is a valid location and also a valid coordinate. An example from the JIRA issue: http://map.geo.admin.ch/?swisssearch=46.5057,6.6278&zoom=8&crosshair=marker. The query is **swisssearch=46.5057,6.6278**. It is a valid coordinate, and when we search the location, it returns a valid object:
```json
{
    "fuzzy": "true",
    "results": [
        {
            "attrs": {
                "detail": "dorfstrasse 46 5057 reitnau 4281 reitnau ch ag",
                "featureId": "3098536_0",
                "geom_quadindex": "021131022303023202112",
                "geom_st_box2d": "BOX(2646205.076 1233096.405,2646205.076 1233096.405)",
                "label": "Dorfstrasse 46 <b>5057 Reitnau</b>",
                "lat": 47.247169494628906,
                "lon": 8.048954963684082,
                "num": 46,
                "objectclass": "",
                "origin": "address",
                "rank": 7,
                "x": 1233096.375,
                "y": 2646205.0,
                "zoomlevel": 10
            },
            "id": 974163,
            "weight": 1546
        }
    ]
}
```

## Fix

It seems we don't keep the swisssearch in the URL anymore when sharing location, so the center should be set to the swisssearch location.

### Legacy URL Parser

- swisssearch=x,y: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.3163,7.6347
- swisssearch=x,y and crosshair: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.3163,7.6347&crosshair=cross

**With the buggy coordinate**
- Only swisssearch: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278
- Swisssearch and crosshair: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278&crosshair=cross

### New URL Parser

- swisssearch=x,y: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.3163,7.6347
- swisssearch=x,y and crosshair:https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.3163,7.6347&crosshair=cross
- swisssearch=x,y and crosshair with crosshair position: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.3163,7.6347&crosshair=cross,2660113,1185272

**With the buggy coordinate**
- Only swisssearch: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.5057,6.6278
- Swisssearch and crosshair: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.5057,6.6278&crosshair=cross
- swisssearch and crosshair + location: https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html#/map?swisssearch=46.5057,6.6278&crosshair=cross,2660113,1185272

### Full list of URL variations

I write all possible combinations of the URL [here](https://docs.google.com/spreadsheets/d/1mH_ZVba6OvOlWPl4REC6pfz447OLYbPwSOw_o4E6iZY/edit?usp=sharing) 

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html)

### DIFF
--- a/packages/mapviewer/tests/cypress/support/commands.js
+++ b/packages/mapviewer/tests/cypress/support/commands.js
@@ -85,7 +85,7 @@ Cypress.Commands.add(
             queryParams.lang = 'en'
         }
         if (
-            !['lat', 'lon', 'x', 'y', 'center', '3d'].some((unwantedKey) =>
+            !['lat', 'lon', 'x', 'y', 'center', '3d', 'swisssearch'].some((unwantedKey) =>
                 Object.keys(queryParams).includes(unwantedKey)
             )
         ) {
--- a/packages/mapviewer/tests/cypress/support/commands.js
+++ b/packages/mapviewer/tests/cypress/support/commands.js
@@ -81,6 +81,10 @@ Cypress.Commands.add(
             }
         }
 
+        if (!('lang' in queryParams)) {
+            queryParams.lang = 'en'
+        }
+
         if (
             !['lat', 'lon', 'x', 'y', 'center', '3d', 'swisssearch'].some((unwantedKey) =>
                 Object.keys(queryParams).includes(unwantedKey)
--- a/packages/mapviewer/tests/cypress/support/commands.js
+++ b/packages/mapviewer/tests/cypress/support/commands.js
@@ -81,9 +81,6 @@ Cypress.Commands.add(
             }
         }
 
-        // if (!('lang' in queryParams)) {
-        //     queryParams.lang = 'en'
-        // }
         if (
             !['lat', 'lon', 'x', 'y', 'center', '3d', 'swisssearch'].some((unwantedKey) =>
                 Object.keys(queryParams).includes(unwantedKey)
--- a/packages/mapviewer/tests/cypress/support/commands.js
+++ b/packages/mapviewer/tests/cypress/support/commands.js
@@ -67,7 +67,7 @@ Cypress.Commands.add(
         queryParams = {},
         withHash = true,
         geolocationMockupOptions = { latitude: 47, longitude: 7, errorCode: null },
-        fixturesAndIntercepts = {}
+        fixturesAndIntercepts = {},
     ) => {
         // Intercepts passed as parameters to "fixturesAndIntercepts" will overwrite the correspondent
         // default intercept.
@@ -81,9 +81,9 @@ Cypress.Commands.add(
             }
         }
 
-        if (!('lang' in queryParams)) {
-            queryParams.lang = 'en'
-        }
+        // if (!('lang' in queryParams)) {
+        //     queryParams.lang = 'en'
+        // }
         if (
             !['lat', 'lon', 'x', 'y', 'center', '3d', 'swisssearch'].some((unwantedKey) =>
                 Object.keys(queryParams).includes(unwantedKey)
--- a/packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
+++ b/packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
@@ -502,19 +502,49 @@ describe('Test the search bar result handling', () => {
         cy.get('@locationSearchResults').should('not.be.visible')
     })
 
+    function testQueryPositionCrosshairStore({
+        searchQuery,
+        expectedCenter,
+        expectedPinnedLocation,
+        expectedCrosshair = null,
+        expectedCrosshairPosition = null,
+    }) {
+        // check the query
+        cy.readStoreValue('state.search.query').should('eq', searchQuery)
+        // check the center of the map
+        cy.readStoreValue('state.position.center').should((positionCenter) =>
+            checkLocation(expectedCenter, positionCenter)
+        )
+        // check the location of pinnedLocation
+        cy.readStoreValue('state.map.pinnedLocation').should((pinnedLocation) =>
+            checkLocation(expectedPinnedLocation, pinnedLocation)
+        )
+        // check the crosshair
+        if (expectedCrosshair !== null) {
+            cy.readStoreValue('state.position.crossHair').should('eq', expectedCrosshair)
+        } else {
+            cy.readStoreValue('state.position.crossHair').should('be.null')
+        }
+        // check the crosshair position
+        if (expectedCrosshairPosition !== null) {
+            cy.readStoreValue('state.position.crossHairPosition').should((crossHairPosition) =>
+                checkLocation(expectedCrosshairPosition, crossHairPosition)
+            )
+        } else {
+            cy.readStoreValue('state.position.crossHairPosition').should('be.null')
+        }
+    }
+
     it.only('handle swisssearch and crosshair together correctly', () => {
         const latitude = 46.3163
         const longitude = 7.6347
-        const swisssearch = `${latitude},${longitude}`
+        const swisssearchCoordinate = `${latitude},${longitude}`
+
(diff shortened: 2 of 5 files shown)


ltkum:
There is something strange going on here : 

With your solution, if I go through the legacy parser --> [https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278&zoom=8&crosshair=marker](https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278&zoom=8&crosshair=marker)  it works fine, but if I go through the current parser, it goes to the address. I don't think the issue is with the `shouldCenter` boolean. This boolean had been introduced so that sharing locations would not be overriden by a swisssearch being present. Since then, the swisssearch parameter is no longer kept in the URL, but people can still play around with it.

Also, this only happens when the application is loaded for the first time.

From what I could see, upon starting in map mode, the application will go through `setSearchQuery` twice. The first time because of the parameter, and it will have its setCenter set to true. The second time, it goes through the else https://github.com/geoadmin/web-mapviewer/blob/915ea3c01da22f2713678d20e9b1196cb3fad097/packages/mapviewer/src/store/modules/search.store.js#L165-L184
it can still see the coordinates, but the `shouldCenter` parameter is false.

It's not called by the searchParam itself. Perhaps it's the `onMounted` function of the searchBar which creates a small conflict.

ltkum (commented):
There is something strange going on here : 

With your solution, if I go through the legacy parser --> [https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278&zoom=8&crosshair=marker](https://sys-map.dev.bgdi.ch/preview/bug-pb-1480-swissearch-crosshair/index.html?swisssearch=46.5057,6.6278&zoom=8&crosshair=marker)  it works fine, but if I go through the current parser, it goes to the address. I don't think the issue is with the `shouldCenter` boolean. This boolean had been introduced so that sharing locations would not be overriden by a swisssearch being present. Since then, the swisssearch parameter is no longer kept in the URL, but people can still play around with it.

Also, this only happens when the application is loaded for the first time.

From what I could see, upon starting in map mode, the application will go through `setSearchQuery` twice. The first time because of the parameter, and it will have its setCenter set to true. The second time, it goes through the else https://github.com/geoadmin/web-mapviewer/blob/915ea3c01da22f2713678d20e9b1196cb3fad097/packages/mapviewer/src/store/modules/search.store.js#L165-L184
it can still see the coordinates, but the `shouldCenter` parameter is false.

It's not called by the searchParam itself. Perhaps it's the `onMounted` function of the searchBar which creates a small conflict.

packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
```diff
@@ -44,9 +45,49 @@ function checkDescendantOf(selector, message) {
...
+    // check the location of pinnedLocation
+    cy.readStoreValue('state.map.pinnedLocation').should((pinnedLocation) =>
+        checkLocation(expectedPinnedLocation, pinnedLocation)
+    )
+    // check the crosshair
+    if (expectedCrosshair !== null) {
+        cy.readStoreValue('state.position.crossHair').should('eq', expectedCrosshair)
+    } else {
+        cy.readStoreValue('state.position.crossHair').should('be.null')
+    }
```
ltkum, line 78: unless I am mistaken, I think if you remove the if-else statement and only do : 
`cy.readStoreValue('state'.position.crosshair').should('eq', expectedCrosshair)`
it should work both when its value is null and when it is not.

packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
```diff
@@ -44,9 +45,49 @@ function checkDescendantOf(selector, message) {
...
+    // check the crosshair position
+    if (expectedCrosshairPosition !== null) {
+        cy.readStoreValue('state.position.crossHairPosition').should((crossHairPosition) =>
+            checkLocation(expectedCrosshairPosition, crossHairPosition)
+        )
+    } else {
+        cy.readStoreValue('state.position.crossHairPosition').should('be.null')
+    }
```
ltkum, line 86: I was thinking if it would be better to have the null verification happen in the `checkLocation` function, but I think having it here is easier to read about what happens.

packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
```diff
@@ -44,9 +45,49 @@ function checkDescendantOf(selector, message) {
...
+    expect(result[1]).to.approximately(expected[1], acceptedDelta)
+}
+
+function testQueryPositionCrosshairStore({
+    searchQuery,
+    expectedCenter,
+    expectedPinnedLocation,
+    expectedCrosshair = null,
+    expectedCrosshairPosition = null,
+}) {
```
ltkum, line 62: In your tests, you're always giving the expected Crosshair value and expected crosshair position. This is good in my opinion as we want test to be explicits as much as possible. I would remove the default values of `expectedCrosshair` and `expectedCrosshairPosition` here, to enforce future possible tests to be explicits.

packages/mapviewer/tests/cypress/support/commands.js
```diff
@@ -81,11 +81,8 @@ Cypress.Commands.add(
             }
         }
 
-        if (!('lang' in queryParams)) {
-            queryParams.lang = 'en'
-        }
```
pakb, line 86: I'd maybe keep that, we can't be certain that the CI instance we receive will have its locale as English, and some tests are testing for translated texts, so we really need to be 100% certain we are using English by default.

ismailsunni: Ah yes, I forget to put it back. There was an issue with redo the search in local instance but not in the cypress test (since it's always set to 'en').
It got me thinking, it might be not the best way, since the case where the lang is not set in the URL is not covered in our test case. Anyway, perhaps it's good for now.

packages/mapviewer/tests/cypress/tests-e2e/search/search-results.cy.js
ismailsunni: ah yes, it's correct. Thanks!

ltkum (approved):
this looks good to me, thanks for your work :)

pakb (approved):

PR #1258, opened 2025-03-10
Feat pb 1383 layers module

schtibe:


packages/geoadmin-layers/src/parsers/externalWMSCapabilitiesParser.ts
```diff
@@ -392,60 +503,67 @@ export default class WMSCapabilitiesParser {
...
                 url = attribution.OnlineResource
                 title = attribution.Title || new URL(attribution.OnlineResource).hostname
             } else {
-                title = this.Service?.Title || new URL(this.Service?.OnlineResource).hostname
+                title =
+                    this.capabilities.Service?.Title ||
+                    new URL(this.capabilities.Service?.OnlineResource).hostname
             }
         } catch (error) {
+            // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
pakb, line 529: Isn't it a bit weird to pass the error object to the string builder here?
Shouldn't we do something like `${error.message}` or something similar?
We could also clearly do a `${error.toString()}` (that's what TypeScript is complaining about here, isn't it? that we are passing non string param to the string concat)

packages/geoadmin-layers/src/types/layers.ts
```diff
@@ -0,0 +1,428 @@
...
+    /**
+     * Flag telling if the base URL must always have a trailing slash. It might be sometime the case
+     * that this is unwanted (i.e. for an external WMS URL already built past the point of URL
+     * params, a trailing slash would render this URL invalid). Default is `false`
+     */
+    // readonly ensureTrailingSlashInBaseUrl: boolean
+    /** Value from 0.0 to 1.0 telling with which opacity this layer should be shown on the map. */
+    opacity: number
+    /** If the layer should be visible on the map or hidden. */
+    visible: boolean /* TODO propose to rename this to isVisible */
```
pakb, line 60: image

packages/geoadmin-layers/src/types/layers.ts
```diff
@@ -0,0 +1,428 @@
...
+    /** Description of the data owner(s) for this layer. */
+    readonly attributions: LayerAttribution[]
+    /** Define if this layer shows tooltip when clicked on. */
+    readonly hasTooltip: boolean
+    /** Define if this layer has a description that can be shown to users to explain its content. */
+    hasDescription: boolean
```
pakb, line 66: that can be `readonly` too I think
did you find any use case where it wasn't the case?

packages/geoadmin-layers/src/types/layers.ts
```diff
@@ -0,0 +1,428 @@
...
+    /** Define if this layer has a legend that can be shown to users to explain its content. */
+    readonly hasLegend: boolean
+    /** Define if this layer comes from our backend, or is from another (external) source. */
+    readonly isExternal: boolean
+    /** Set to true if some parts of the layer (e.g. metadata) are still loading */
+    isLoading: boolean
+    /** Time series config */
+    timeConfig: LayerTimeConfig | null
```
pakb, line 74: The time config object itself can be `readonly`, if that doesn't block its mutation in its nested attributes

packages/geoadmin-layers/src/types/layers.ts
```diff
@@ -0,0 +1,428 @@
...
+    adminId?: string
+}
+
+// #region: GeoAdminLayers
+/**
+ * This interface unifies the shared properties of the layers that speak to an API like WMS and WMTS
+ *
+ * @interface GeoAdminLayer
+ */
+export interface GeoAdminLayer extends Layer {
```
pakb, line 97: most, if not all, props here can be switched to `readonly` IMHO

pakb (commented):
Let's start with a partial review...

General comment :
As we are now moving to TS, should we decide if the absence of something should output `undefined` or `null`?
I'm a bit more for `undefined`, as I think it behaves better with TS `?` optional op. in interfaces, do you have any preference?

packages/mapviewer/src/api/features/features.api.js
```diff
@@ -186,6 +182,8 @@ export async function identifyOnGeomAdminLayer({
 }
 
 /**
+ * // TODO adjust this
+ *
```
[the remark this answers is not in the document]
schtibe: Oh this is probably just like a bookmark for me during the work.

schtibe:
> Let's start with a partial review...
> 
> General comment : As we are now moving to TS, should we decide if the absence of something should output `undefined` or `null`? I'm a bit more for `undefined`, as I think it behaves better with TS `?` optional op. in interfaces, do you have any preference?

Hehe funny you should say this. I had the same idea and naively started implementing the interface with optional properties `foo?: <type>`, but this lead me to a **ton** of problems down the road with code that expected it to be `null` if it's not set. So after weeks of fixing bugs because of this, I decided to revert to the existing behavior. _Although in hindsight I don't quite understand why these bugs weren't caught by the typing system, maybe not enough is in TS yet_

Maybe we should first start to transform more of the code to TS before trying this again. All the vue-components that interact with the layers are still in JS, for instance.

packages/geoadmin-layers/src/types/layers.ts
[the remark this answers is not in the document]
schtibe: ☺️

PR #1260, opened 2025-03-13
PB-1435 Fix the german link to the feedback help page anchor

schtibe:
The anchor in the link to the German version of the help for the feedback form wasn't correct

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1435-feedback-help-link/index.html)


pakb (approved):
Have you seen that on some browser the button/dropdown button text is 1 or 2 px down from the other text elements? 😨

schtibe:
> Have you seen that on some browser the button/dropdown button text is 1 or 2 px down from the other text elements? 😨

Yes, I want to fix that too in another PR. And it has a white background instead of transparent (well visible with Aerial Imagery background)

PR #1264, opened 2025-03-13
PB-1137 : adding print view 

pakb:
so that we may start developing our new headless print backend

Resolution from @geoadmin/coordinates are now expressed in steps, where each step can have a corresponding zoom level or not.
This is needed with LV95, where we have some gaps in zoom level, while there is a resolution (between zoom 9 and 10 there's an extra resolution step)

Also making our coordinate system calculate resolution without threshold (especially for LV95) so that it can then be used to calculate a mercator zoom level. Will be very useful when we will be switching to using VectorTiles (and Mercator on-the-fly reprojection)

Move debug toolbar to top App component so that it can be added to embed and print view without needing to change anything
also changing a bit the CSS of the views, so that it is not using an ID to style each view, but a class (doesn't really make sense to have the embed view have the ID map-view)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1137-print-view/index.html)


rebert (approved):
Thanks.
If not a big effort, it would be nice to have the borders set on the webviewer side again (as an absolute value in mm, if possible)?

rebert:
we can have a chat on this, if ok?

pakb:
> Thanks. If not a big effort, it would be nice to have the borders set on the webviewer side again (as an absolute value in mm, if possible)?

I've created two tickets for further improvements
https://swissgeoplatform.atlassian.net/browse/PB-1521
https://swissgeoplatform.atlassian.net/browse/PB-1522

If you have any other "issue" in mind, please add another ticket to the epic (so that I don't forget to tackle that later 😉 )

PR #1273, opened 2025-03-20
PB-165: Group providers by base URL

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-165-grouping-layers-same-domain/index.html)

Preview:
image
image

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-165-grouping-layers-same-domain/index.html)


pakb (commented):
Might be better not to create a group when there's only one element, it makes it harder to access the link

ismailsunni:
@pakb so, the one-element group should be changed to one-element (i.e. no caret)?

ismailsunni:
@pakb new appearance with no-group for one URL

image

pakb (commented):
If there are more than one element (meaning we have a group) I'd maybe want to remove the "common" part to all children.
So in your screenshot all `https://geo.so.ch` layers will not start with `https://geo.so.ch` (it's already there as a group title) but only show what's different.

To have a better UI/UX, I'd maybe make each group title "sticky", so that they do not go out of scroll while some children are still visible

ismailsunni:
@pakb something like this?
image

ismailsunni:
Update screenshot:

image

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -1,49 +1,221 @@
...
+    return {
+        id: baseUrl,
+        name: commonPrefix || baseUrl,
+        type: 'group',
+        expanded: filterApplied,
+        children: Object.entries(subGroups).map(([key, subGroupProviders]) => {
+            if (subGroupProviders.length === 1) {
+                return {
+                    id: subGroupProviders[0].url,
+                    name: key + '/' + subGroupProviders[0].relativeUrl,
```
pakb, line 88: might make sense to use some `...${variable}...` notation here instead of concat multiple strings

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -1,49 +1,221 @@
...
+                    url: provider.url,
+                    emphasize: provider.emphasize,
+                })),
+            }
+        }),
+    }
+}
+
+// Set tabindex for each node in the tree with DFS strategy so that the tab
+// index is in correct order visually
```
pakb, line 112: JSDoc instead of simple comments

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -1,49 +1,221 @@
...
+function addTabIndex(treeData) {
+    let index = 0
+
+    function dfs(node) {
+        node.tabindex = index++
+        if (node.children && node.children.length > 0) {
+            node.children.forEach((child) => dfs(child))
+        }
+    }
```
pakb, line 121: why nesting this function? You can maybe give the `index` as a param to the function instead, and return the resulting index instead of mutating it inside the function

pakb: Would also make sense to add a little JSDoc that explains what this function does, `dfs` isn't a self-explanatory function name

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -57,26 +229,125 @@ defineExpose({ goToFirst })
             class="providers-list"
             data-cy="import-provider-list"
         >
+            <template v-for="node in treeData" :key="node.id">
+                <component
+                    :is="node.type === 'group' ? 'div' : 'div'"
```
pakb, line 234: that might be some left-over, but is not relevant anymore. This will create a `<div>` regardless of the state, so you could remove the whole `<component>` thing and replace it with a `<div>`

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -57,26 +229,125 @@ defineExpose({ goToFirst })
...
+                    :class="node.type === 'group' ? 'providers-group' : 'providers-list-item'"
```
pakb, line 235: could using something like
```
:class="`providers-${node.type}"
```
also work? that looks simpler to understand than a ternary operator in this context

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -57,26 +229,125 @@ defineExpose({ goToFirst })
...
+                    :data-cy="node.type === 'group' ? 'import-provider-group' : 'import-provider-item'"
```
pakb, line 236: same thing here

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
pakb, line 1: You might want to add a prettier run after each save on your IDE, if I run prettier locally I have some changes here

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -102,6 +375,42 @@ defineExpose({ goToFirst })
         .providers-list-item:hover {
             background-color: $list-item-hover-bg-color;
         }
+
+        .providers-group-header {
+            cursor: pointer;
+            font-weight: bold;
```
pakb, line 381: might be better to use a `<strong>` tag instead of relying on CSS to bolden your text

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
```diff
@@ -102,6 +375,42 @@ defineExpose({ goToFirst })
         .providers-list-item:hover {
             background-color: $list-item-hover-bg-color;
         }
+
+        .providers-group-header {
+            cursor: pointer;
```
pakb, line 380: there are some bootstrap utils that can achieve that too, the more you are using bootstrap classes, the less work it will be when we switch to tailwind

So if you could do a little pass of all the rules you have below, and check if some bootstrap class would do the same, it would be nice

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/external-providers.json
pakb, line 1: We should really be sorting this alphabetically to detect duplicates... Do you have a tool that could achieve that? (CoPilot could maybe do it)

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/useProviders.js
```diff
@@ -8,6 +9,16 @@ class Provider {
...
         this.emphasize = emphasize
     }
+
+    getBaseUrl() {
+        try {
+            const urlObj = new URL(this.url)
+            return urlObj.origin
+        } catch (e) {
+            log.error('Invalid URL:', this.url, e)
+            return ''
```
pakb, line 19: might make more sense to return `null` or `undefined` and then manage the fact that it can be "falsy" in `useProviders` function bellow

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/external-providers.json
ismailsunni: Ah, I removing the duplicates with a script, but forget to sort it. *I thought I did

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
ismailsunni: I use it for styling `<TextSearchMarker>`, I don't think it's possible to use `<strong>` tag to bold the text inside `<TextSearchMarker>`. The alternative would be create  a new prop to enable this `<strong>` tag. Is that what do you suggest?

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
ismailsunni: now all is using bootstrap classes except `bg-body` that for some reason it doesn't work with `focus` and `hover`

packages/mapviewer/src/modules/menu/components/advancedTools/ImportCatalogue/ProviderList.vue
ismailsunni: I already did actually, the `format on save`, right?

pakb: should be enough yes, weird that I had some changes when running prettier on my end...

pakb (approved):

PR #1274, opened 2025-03-20
PB-1530 : fix menu issues on Safari

pakb:
The new confederation logo was using some CSS properties that aren't well supported by Safari, and the Swiss flag was stretching way over the logo because of that. When users were clicking on the menu elements, they were in fact clicking on the flag (invisible container) and thus being thrown back to the "home" view.


schtibe (approved):
LGTM and works on my android

PR #1276, opened 2025-03-21
PB-1318 : give a better error message on deleted KMLs

pakb:
Trying to better convey why the KML couldn't be loaded, to help reduce questions on our helpdesk team about it.


ltkum (approved):
Seems explicit to me :)

PR #1278, opened 2025-03-24
PB-1548 : fix mismanagement of drawing features if coming with adminId

pakb:
the loading state of a KML layer wasn't managed by the drawing module, meaning that the timing of when the KML data was loaded could lead to no features being present on the drawing layer. Very dangerous because that could make user empty their KMLs if they would edit it in this state

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1548-fix-loading-of-drawing-module-with-adminid/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/drawing/DrawingModule.vue
+++ b/packages/mapviewer/src/modules/drawing/DrawingModule.vue
@@ -68,6 +68,9 @@ const showAddVertexButton = computed(() => {
 const editMode = computed(() => store.state.drawing.editingMode)
 const currentDrawingMode = computed(() => store.state.drawing.mode)
 
+const hasLoaded = computed(() => {
+    return activeKmlLayer.value?.isLoading === false && !!activeKmlLayer.value.kmlData
+})
 const hasKml = computed(() => {
     if (online.value) {
         return !!activeKmlLayer.value && !activeKmlLayer.value.isEmpty()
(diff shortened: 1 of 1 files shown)


packages/mapviewer/src/modules/drawing/DrawingModule.vue
```diff
@@ -99,6 +102,11 @@ const isDrawingModified = computed(() => {
 watch(projection, () => {
     drawingLayer.setSource(createSourceForProjection())
 })
+watch(hasLoaded, () => {
+    if (hasLoaded.value) {
+        addKmlToDrawing()
+    }
+})
```
ltkum, line 109: would it be a good idea to set a `{once: true}` parameter to the watcher? hasLoaded is either always true (we had already loaded the layer before entering the drawing module), or false, then true (we enter the drawing module with the admin link or we created a new kml layer when starting to draw.)

If you don't think it's useful, we can avoid that :)

ltkum (approved):
just a quick idea I had while reviewing, but to be fair it's not important nor breaking. It's perfectly mergeable in the current state :)

good job

packages/mapviewer/src/modules/drawing/DrawingModule.vue
pakb: Don't think that would greatly improve perf, and I'd really like to fix this bug ASAP. Let's go with the current state

PR #1281, opened 2025-03-25
Fix flaky importToolMaps tests

schtibe:
The tests were flaky in the way that they couldn't select the correct elements for testing. With closing the menu and adding some assertions of the elements being visible it seems that we slow down cypress enough so that the selection is correct now

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-importool-tests/index.html)


ismailsunni (approved):
LGTM

PR #1282, opened 2025-03-27
PB-1384 : tooltip module

pakb:
I will need to use the tooltip in the elevation profile module, so I've exported the component for the tooltip as a "lib"

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1384-tooltip-module/index.html)

### DIFF
--- a/packages/geoadmin-tooltip/package.json
+++ b/packages/geoadmin-tooltip/package.json
@@ -3,11 +3,6 @@
     "version": "0.0.1",
     "description": "Tooltip for geoadmin",
     "type": "module",
-    "files": [
-        "dist"
-    ],
-    "main": "./dist/index.umd.cjs",
-    "module": "./dist/index.js",
     "exports": {
         ".": {
             "types": "./dist/index.d.ts",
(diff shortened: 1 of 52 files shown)


packages/geoadmin-tooltip/src/GeoadminTooltip.vue
```diff
@@ -232,123 +219,45 @@ defineExpose({ tooltipElement, openTooltip, closeTooltip })
...
-                    class="arrow"
-                    :class="{ arrowDown: placement == 'top', arrowRight: placement == 'left' }"
+                    v-if="middlewareData.arrow"
+                    ref="arrowElement"
+                    :style="{
+                        left: `${middlewareData.arrow.x}px`,
+                        top: `${middlewareData.arrow.y}px`,
+                    }"
+                    class="tw:absolute tw:h-0 tw:w-0 tw:after:content-[''] tw:after:h-0 tw:after:w-0 tw:after:fixed tw:after:top-[inherit] tw:after:left-[inherit] tw:after:z-[-1]"
+                    :class="{
```
schtibe, line 230: Maybe it could make sense to move this into a `computed` to keep the HTML part less cluttered? I think it is rather easy to construct arrays or objects that will the be passed to the attribute

packages/geoadmin-tooltip/package.json
```diff
@@ -0,0 +1,49 @@
+{
```
schtibe, line 1: I think you could run a `pnpm format` on this file.

packages/mapviewer/src/css/tailwind.css
```diff
@@ -0,0 +1,2 @@
+@import 'tailwindcss' prefix(tw);
+@import '@geoadmin/tooltip/tailwindcss';
```
schtibe, line 2: Is this line doing anything?

schtibe (approved):
LGTM, just a few minor remarks about details

schtibe:
Oh maybe you could also add the new module to `vitest.workspace.js`

packages/geoadmin-tooltip/src/GeoadminTooltip.vue
pakb: I really don't like when we mix our JS code with HTML class generation, that's why I tried to kept everything related to Tailwind described in the template section.

It could be improved a bit if moved to JS, but as I said, I prefer to keep the JS doing its part (interaction, reactivity, etc) and let the template be responsible for the styling

packages/mapviewer/src/css/tailwind.css
pakb: without it, the tailwind classes from the `@geoadmin/tooltip` module are ignored and not taken into account in the `mapviewer` package's tailwind output

PR #1283, opened 2025-03-28
PB-1493 : Add iframe parameter to scroll / zoom with CTRL

sami-nouidri-swisstopo:
If the nosimplezoom = true parameter is added to an embedded URL, the user needs to press ctrl in order to zoom into the map. Here's a preview of the hint : 
image

[Test link](https://sys-map.dev.bgdi.ch/preview/task-pb-1493-add-iframe-specific-scroll/index.html)

### DIFF
--- a/packages/mapviewer/src/views/EmbedView.vue
+++ b/packages/mapviewer/src/views/EmbedView.vue
@@ -1,6 +1,7 @@
 <script setup>
 import log from '@geoadmin/log'
-import { computed, onBeforeMount, onMounted, watch } from 'vue'
+import { computed, onBeforeMount, onMounted, onUnmounted, ref, watch } from 'vue'
+import { useI18n } from 'vue-i18n'
 import { useRoute } from 'vue-router'
 import { useStore } from 'vuex'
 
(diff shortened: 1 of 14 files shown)


packages/mapviewer/src/router/storeSync/ScrollingModeParamConfig.class.js
```diff
@@ -0,0 +1,28 @@
...
+function dispatchScrollingModeFromUrlIntoStore(to, store) {
+    const isEmbed = to.path.includes('embed')
+    const hasScrollParam = 'ctrl_scroll' in to.query
+
+    if (isEmbed && hasScrollParam) {
+        store.dispatch('setScrollWithCtrlOnly', {
+            scrollWithCtrlOnly: true,
+            dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
+        })
+    }
```
ltkum, line 27: I would use the third argument from `setValuesInStore`, as it gives the url param itself.

```dispatch(to, store, urlParam) 
{
  if (to.path.toLowerCase().includes('embed') { // il me semble que les majuscules fonctionnent aussi dans les routes, à tester.
    store.dispatch('set...', {
      paramName: urlParamValue,
      dispatcher: STORE_DISPATCHER_ROUTER_PLUGIN,
    }
}
}

```

ltkum (changes requested):
In general, seems good, three things : 

1. The naming convention is not that great, it would be better if, within the code, we knew it was a parameter meant for the embed view, and that we knew it was for the zoom. Something in the form : 
`noSimpleZoom` for the url parameter and `noSimpleEmbedZoom` for the store.
2. I would add a button in the embed preview window to add or remove that parameter from the embed link, so people have a way to add the parameter.
3. When I scroll on a page with the embed link, it stops the scrolling and focus on the map, stopping from going further down the page. It could be a good idea to find a way to stop the embedded viewer from overriding the main page scroll

ltkum (commented):
We discussed a bit on slack, and translations should be coming soon.
One last thing: for the sake of consistency, It would be better to use a square button, since other options (like for example : in the print) are using this kind of button to be toggled.

packages/mapviewer/src/views/EmbedView.vue
```diff
@@ -20,19 +20,67 @@ const route = useRoute()
...
+        showSimpleZoomHint.value = true
+        clearTimeout(simpleZoomHintTImeout)
+        simpleZoomHintTImeout = setTimeout(() => {
+            showSimpleZoomHint.value = false
+        }, 3000)
+
+        return
+    }
+}
+
```
ltkum, line 46: when the popup is displayed, ctrl+zooming when the mouse is on the popup does a browser zoom. If you can find a way to avoid that, that would be great. (an event listener on the ctrl key for example)

If it adds too much complexity, don't bother about it

packages/mapviewer/src/views/EmbedView.vue
```diff
@@ -20,19 +20,67 @@ const route = useRoute()
...
 onBeforeMount(() => {
     store.dispatch('setEmbed', { embed: true, ...dispatcher })
 })
 
 onMounted(() => {
-    log.info(`Embedded map view mounted`)
```
ltkum, line 28: let's add that log again at the end :)

ltkum (commented):
We're getting there :)

only a few more details to go through :)

ltkum (approved):
LGTM. I'll add the translations in the google sheet :)

PR #1284, opened 2025-03-31
PB-1558: fix printing with geojson layers

ltkum:
Issue: When a Geojson layer has a label, it uses a function in mapfish print meant to give the font information to the print server. This function could give null as a result and had no safety around it in case it failed to produce a valid result.

Fix: we did a small PR in mapfishprint which add safeties around the function, meaning that even if the font is not recognised (which would be due to an issue in openlayers, as the regex to check fonts is faulty and can't parse fonts with a number in it), we at least give back basic results.


sami-nouidri-swisstopo (approved):
LGTM

PR #1286, opened 2025-03-31
PB-1384 : profile chart as stand alone module

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1384-profile-module/index.html)


packages/geoadmin-elevation-profile/src/GeoadminElevationProfile.vue
```diff
@@ -0,0 +1,216 @@
...
+            v-if="profileData && hasData"
+            :profile="profileData"
+        >
+            <slot />
+        </GeoadminElevationProfilePlot>
+        <GeoadminElevationProfileInformation
+            v-if="profileMetadata"
+            :metadata="profileMetadata"
+        >
+            <GeoadminTooltip tooltip-content="profile_invert">
```
sommerfe, line 192: This is displayed as "profile_invert" in the app but shouldn't it be a translation?

packages/geoadmin-elevation-profile/src/GeoadminElevationProfile.vue
```diff
@@ -0,0 +1,216 @@
...
+                <button
+                    class="tw:bg-neutral-100 tw:hover:bg-neutral-200 tw:border tw:rounded tw:border-neutral-400 tw:print:hidden tw:min-w-[2.5rem] tw:h-full tw:cursor-pointer"
+                    @click="revertProfileDirection"
+                >
+                    <FontAwesomeIcon icon="shuffle" />
+                </button>
+            </GeoadminTooltip>
+            <GeoadminTooltip tooltip-content="profile_download_csv">
```
sommerfe, line 200: same here

packages/geoadmin-elevation-profile/src/GeoadminElevationProfilePlot.vue
```diff
@@ -0,0 +1,453 @@
...
+            <div>
+                <small>
+                    <strong>{{ t('profile_y_label') }}: </strong>
+                    <span
+                        v-if="pointBeingHovered.elevation && pointBeingHovered.elevation > 0"
+                        data-cy="profile-popup-tooltip-elevation"
+                    >
+                        {{ pointBeingHovered.elevation }} m
+                    </span>
+                    <span v-else>{{ t('profile_not_available') }}</span>
```
schtibe, line 447: Does this take the translations from the main package? 🤔

packages/geoadmin-elevation-profile/src/config.ts
```diff
@@ -0,0 +1,19 @@
...
+export const FILL_COLOR = 'rgba(255, 99, 132, 0.7)'
+
+/**
+ * 12.5 meters is what was used in the old viewer, see
+ * https://github.com/geoadmin/mf-geoadmin3/blob/ce24a27b0ca8192a0f78f7b8cc07f4e231031304/src/components/GeomUtilsService.js#L207
+ *
+ * I tried lowering the value, but with the test GPX that were attached to the ticket PB-800, I get
+ * the worst hiking time estimation when using something like 5m than if I use this 12.5 meters.
+ */
+export const GEOMETRY_SIMPLIFICATION_TOLERANCE = 12.5 // meters
```
schtibe, line 19: I think you didn't remove that from the mapviewer package (maybe I'm not seeing it).
Is that a good idea? Like this we would have the same constant twice, which goes against the idea of constants. I have the same problem with my layers module, not sure where to put these things now.

packages/geoadmin-elevation-profile/src/profile.api.ts
```diff
@@ -0,0 +1,349 @@
...
+export class ElevationProfileError extends Error {
+    public readonly technicalError: Error
+
+    constructor(message: string, technicalError: Error) {
+        super(message)
+        this.technicalError = technicalError
+    }
+}
+
+type ServiceProfileAltitudes = {
```
schtibe, line 20: Will there ever be an advantage if we export these here too?

packages/geoadmin-elevation-profile/src/profile.api.ts
```diff
@@ -0,0 +1,349 @@
...
+        log.error(errorMessage)
+        throw new ElevationProfileError(errorMessage, new Error('profile_could_not_generate'))
+    }
+    const segments: ElevationProfileSegment[] = []
+    // if the profileCoordinates is not a double nested array, we make it one
+    // segmented files have a double nested array, but not all files or self made drawings
+    // so we have to make sure we have a double nested array and then iterate over it
+    const coordinatesForSegments: SingleCoordinate[][] = ensureDoubleNestedArray(profileCoordinates)
+
+    for (const coordinates of coordinatesForSegments) {
```
schtibe, line 275: Maybe you could refactor this chunk into several smaller functions to increase understandability

packages/geoadmin-elevation-profile/public/favicon.ico
schtibe, line 1: I suppose we could leave this away 🙂

packages/mapviewer/src/modules/i18n/locales/de.json
```diff
@@ -513,12 +513,16 @@
...
+    "profile_could_not_generate": "Fehler: Das Profil konnte nicht erstellt werden",
     "profile_distance": "Luftlinie",
+    "profile_download_csv": "Daten als CSV-Datei abrufen",
     "profile_elevation_difference": "Höhendifferenz Start-Ende",
     "profile_elevation_down": "Abstieg",
     "profile_elevation_up": "Aufstieg",
     "profile_hike_time": "Marschzeit",
+    "profile_invert": "Gegenrichtung",
     "profile_no_data": "kein Datum",
+    "profile_out_of_bounds": "Objekte sind außerhalb der Grenzen, keine Profildaten konnten wiederhergestellt werden",
```
schtibe, line 525: In Switzerland, we don't use the ß 😀 (We don't even know the rules!). So you can easily put "ss" here.

packages/mapviewer/src/modules/i18n/locales/en.json
```diff
@@ -513,12 +513,16 @@
     "print_scale": "Scale",
     "print_underway_message": "We ask you for your patience. A multi-page print is underway, which can take some time.",
     "problem_announcement": "Report problem",
+    "profile_could_not_generate": "Error : the profile could not be generated",
```
schtibe, line 516: The whitespace before ':' can (should?) be avoided I guess? Same in French, but there I don't know if it's maybe required 😀

packages/mapviewer/src/modules/infobox/components/InfoboxContent.vue
```diff
@@ -41,43 +70,83 @@ watch(selectedFeatures, (features) => {
...
                 v-show="!showElevationProfile"
             />
         </div>
     </div>
 </template>
+
+<style lang="scss" scoped>
+@import '@/scss/media-query.mixin';
+
+.infobox-content {
```
schtibe, line 137: Intentionally not using tailwind here?

packages/geoadmin-elevation-profile/src/GeoadminElevationProfilePlot.vue
pakb: yes, I couldn't find a proper method to "encapsulate" the `vue-i18n` service inside a component. It latches itself to some kind of Vue global / window, and instances start mixing together. It's not yet possible to ship it as a stand-alone.

We could start looking for alternatives in the i18n space, where that would be a core feature we want (being able to be encapsulated in a component lib)

packages/geoadmin-tooltip/src/GeoadminTooltip.vue
```diff
@@ -101,15 +105,40 @@ const closeTooltip = (): void => {
...
+    setTimeout(() => {
+        if (touchDidntMove && isTouching.value) {
+            openTooltip()
+        }
+    }, 500)
+}
+
+const onTouchMove = (event: TouchEvent): void => {
+    const changedTouches: TouchList = event.changedTouches
+    for (let i = 0; i < changedTouches.length; i++) {
```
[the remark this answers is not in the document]
pakb: no we can't as we mutate it with the `i++`

packages/mapviewer/src/modules/infobox/components/InfoboxContent.vue
pakb: I thought I shouldn't do a partial switch to TW while in the `mapviewer` package... should we start moving things gradually?

schtibe: good question. Maybe it would be more manageable when doing gradually?

packages/geoadmin-elevation-profile/src/profile.api.ts
pakb: I don't think it's used anywhere else, it's more of a "let's better describe what the backend I encapsulate does" type more than anything

packages/geoadmin-elevation-profile/src/GeoadminElevationProfile.vue
```diff
@@ -0,0 +1,216 @@
...
+}
+
+function onCSVDownload() {
+    if (profileData.value && hasData.value) {
+        const csvData =
+            [
+                ['Distance', 'Altitude', 'Easting', 'Northing', 'Longitude', 'Latitude'],
+                ...profileData.value.segments
+                    .flatMap((segment) => segment.points)
+                    .map((point) => {
```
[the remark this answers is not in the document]
pakb: I haven't touched this code from the previous component, just moved it here

packages/geoadmin-elevation-profile/src/GeoadminElevationProfileOpenLayersBridge.vue
```diff
@@ -0,0 +1,74 @@
...
+const coordinate = computed<SingleCoordinate | undefined>(() => {
+    if (getPointBeingHovered) {
+        return getPointBeingHovered()?.coordinate
+    }
+    return undefined
+})
+
+const overlayAdded = ref<boolean>(false)
+
+const element = document.createElement('div')
```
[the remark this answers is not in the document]
pakb: I could `.remove()` it in the `unUnmounted` hook, to be sure

packages/geoadmin-elevation-profile/src/profile.api.ts
```diff
@@ -0,0 +1,354 @@
...
+    }
+}
+
+type ServiceProfileAltitudes = {
+    COMB?: number
+    DTM2?: number
+    DTM25?: number
+}
+
+type ServiceProfilePoints = {
```
sommerfe, line 26: Is there a reason why `type` is used here instead of `interface` ? Same for other files

packages/geoadmin-coordinates/src/proj/CoordinateSystemBounds.ts
```diff
@@ -92,7 +132,9 @@ export default class CoordinateSystemBounds {
         // checking if we require splitting
         if (coordinates.find((coordinate) => !this.isInBounds(coordinate[0], coordinate[1]))) {
             const boundsAsPolygon = bboxPolygon(this.flatten)
-            return lineSplit(lineString(coordinates), boundsAsPolygon).features.map((chunk) => {
+            const paths = lineSplit(lineString(coordinates), boundsAsPolygon)
+            paths.features = reassembleLineSegments(coordinates[0], paths)
+            return paths.features.map((chunk) => {
```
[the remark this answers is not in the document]
pakb: Yes and no, I tried to clear out the misunderstanding by removing the name segment the most I could.
Segments are part of a MultiLineString or MultiPolygon, that are disjointed (and thus can't be requested as a single profile).
This concept still exists, but is managed by the `InfoboxContent.vue` component for now.

The profile module doesn't care about segments, it receives a bunch of JOINTED coordinates, request height data for it, and build a chart with that data.

packages/geoadmin-elevation-profile/src/profile.api.ts
pakb: There's no need to extend this "type" later, so there's no need for an `interface`, but that's my limited knowledge of TS

packages/geoadmin-coordinates/src/proj/CoordinateSystemBounds.ts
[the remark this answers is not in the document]
pakb: Yes, I've added a bit of logic in `profile.store.js` that will attempt to stitch a MultiLineString into the biggest jointed path before passing it to the profile module.

That means that a "artificially" disjointed path (such as VeloLand path, or the GPX you mentioned) can now be displayed as single path in the profile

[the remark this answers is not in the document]
sommerfe: In this case isn't the code in `InfoboxContent.vue` then unnecessary or can there be cases where not all paths can be jointed?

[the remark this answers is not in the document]
pakb: there are still some cases where it's not possible to stitch all parts together, [such as this one](https://sys-s.dev.bgdi.ch/fbv9d8m7k09n), so this code is still necessary (and I didn't want to move that logic into the profile module, yet?)

A future solution to that would be that the profile component receives a GeoJSON, instead of a list of coordinates, and can then decide based on the feature type what to do

packages/geoadmin-elevation-profile/src/GeoadminElevationProfile.vue
```diff
@@ -0,0 +1,211 @@
...
+
+const {
+    points = [],
+    projection = LV95.epsg,
+    simplify = false,
+    staging = 'production',
+    filename = 'export',
+} = defineProps<ElevationProfileProps>()
+
+type ElevationProfileMessages = {
```
sommerfe, line 41: can this one also be an interface ?

packages/geoadmin-elevation-profile/src/GeoadminElevationProfileInformation.vue
```diff
@@ -0,0 +1,125 @@
...
+    formatMinutesTime,
+} from '@/utils'
+
+interface ElevationProfileInformationProps {
+    metadata: ElevationProfileMetadata
+}
+
+const { metadata } = defineProps<ElevationProfileInformationProps>()
+
+type ElevationProfileInformationMessages = {
```
sommerfe, line 23: same here

packages/geoadmin-elevation-profile/src/GeoadminElevationProfile.vue
pakb: oh, I filtered with `*.ts` files, forgot to check vue files for type defintion... yes I'll swap them too

sommerfe (approved):

PR #1287, opened 2025-04-01
PB-1368: round zoom on double tap

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1368-round-zoom-double-tap-mobile/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/map/components/openlayers/utils/useViewBasedOnProjection.composable.js
+++ b/packages/mapviewer/src/modules/map/components/openlayers/utils/useViewBasedOnProjection.composable.js
@@ -2,6 +2,7 @@ import { constants, LV95, WEBMERCATOR } from '@geoadmin/coordinates'
 import log from '@geoadmin/log'
 import { round } from '@geoadmin/numbers'
 import { View } from 'ol'
+import { DoubleClickZoom } from 'ol/interaction'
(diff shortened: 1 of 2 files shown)


packages/mapviewer/src/modules/map/components/openlayers/utils/useViewBasedOnProjection.composable.js
```diff
@@ -63,13 +64,34 @@ export default function useViewBasedOnProjection(map) {
...
 
+    const roundedDoubleClickZoom = new DoubleClickZoom()
+
+    roundedDoubleClickZoom.handleEvent = function (event) {
+        if (event.type === 'dblclick') {
+            event.preventDefault()
+            event.stopPropagation()
+            const view = map.getView()
+            let zoom = view.getZoom()
+            let roundedZoom = store.state.position.projection.roundZoomLevel(zoom, true)
```
pakb, line 75: I think the signature of this `roundZoomLevel` has changed, and there's no boolean anymore (it was always called with `true`)

Can you change any call to this function and remove the boolean, if there are others?

sommerfe: I checked and the roundZoomLevel of the SwissCoordinateSystem still takes a boolean but i will remove this parameter and modify the function

pakb (approved):

PR #1288, opened 2025-04-02
PB-1554: fix relocate + point extension of lines in drawing mode

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1554-relocate-add-point-extension-line-drawing/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/drawing/components/AddVertexButtonOverlay.vue
+++ b/packages/mapviewer/src/modules/drawing/components/AddVertexButtonOverlay.vue
@@ -66,11 +66,10 @@ const updateButtonPositions = () => {
 
     const firstOffset = calculateOffset(coordinates[0], coordinates[1], BASE_OFFSET_DISTANCE)
 
-    // adding this so that the button is not on top of the measure line label
     const distance =
-        BASE_OFFSET_DISTANCE + selectedFeatureType.value === 'MEASURE'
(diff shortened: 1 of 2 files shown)


packages/mapviewer/src/modules/drawing/components/AddVertexButtonOverlay.vue
```diff
@@ -66,11 +66,9 @@ const updateButtonPositions = () => {
 
     const firstOffset = calculateOffset(coordinates[0], coordinates[1], BASE_OFFSET_DISTANCE)
 
-    // adding this so that the button is not on top of the measure line label
     const distance =
-        BASE_OFFSET_DISTANCE + selectedFeatureType.value === 'MEASURE'
-            ? MEASURE_ADDITIONAL_OFFSET
-            : 0
+        selectedFeatureType.value === 'MEASURE' ? MEASURE_ADDITIONAL_OFFSET : BASE_OFFSET_DISTANCE
```
pakb, line 70: It's maybe more an issue of parenthesis.
I think `BASE_OFFSET_DISTANCE` should still be added in case of measure, you get that otherwise 
image

sommerfe: ah ok now i see what the original intention of the + was in this case, yes i'll add the base offset back to it

pakb (approved):

PR #1289, opened 2025-04-03
PB-1576: switching to 3D with selected features breaks the 3D viewer

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1576-switching-to-3d-with-selected-feature-breaks-3d-viewer/index.html)

### DIFF
--- a/packages/mapviewer/src/api/features/LayerFeature.class.js
+++ b/packages/mapviewer/src/api/features/LayerFeature.class.js
@@ -6,7 +6,7 @@ export default class LayerFeature extends SelectableFeature {
     /**
      * @param {AbstractLayer} featureData.layer The layer in which this feature belongs
      * @param {Number | String} featureData.id The unique feature ID in the layer it is part of
-     * @param {String} featureData.name The name (localized) of this feature
+     * @param {String} featureData.title The title (localized) of this feature
      * @param {Object | String} featureData.data Data for this feature's popup (or tooltip).
      * @param {[[Number, Number]]} featureData.coordinates Coordinate in the current projection
      *   ([[x,y],[x2,y2],...])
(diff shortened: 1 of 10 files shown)


packages/geoadmin-coordinates/src/utils.ts
```diff
@@ -19,14 +19,14 @@ type Single3DCoordinate = [number, number, number]
...
+    coordinate: SingleCoordinate | Single3DCoordinate,
     digits: number,
     withThousandsSeparator: boolean = true,
     enforceDigit: boolean = false
 ): string | null {
     if (
         !Array.isArray(coordinate) ||
-        coordinate.length !== 2 ||
+        coordinate.length !== 2 && coordinate.length !== 3 ||
         !coordinate.every(isNumber) ||
```
ltkum, line 30: This is not where it is supposed to happen.

The cesium mouse tracker is trying to do its magic here : 
https://github.com/geoadmin/web-mapviewer/blob/84457a59f59926113254c0dc432d797f1dfa0998/packages/mapviewer/src/modules/map/components/cesium/CesiumMouseTracker.vue#L90-L103
The issue is that in this function, we use the format method from the projection which uses only two coordinates, when the 3D coordinates are composed of three numbers. A solution would be to give only the first two coordinates to the format function

ltkum (commented):
The intiial issue seems to be working correctly. For the `null` bug, I would alter the elements sent to the `format` function, and maybe call `roundedString` afterwards.

packages/geoadmin-coordinates/src/utils.ts
sommerfe: Yes we could just ignore and clip the z-axis for the coordinates, but is this what we want? I think showing the z-axis coordinate could be interesting.

ltkum (approved):
LGTM

PR #1290, opened 2025-04-08
PB-1494: use geoadmin style on legacy links

ltkum:
Issue: When parsing a legacy link, KML layers would use the standard style. Those KMLs were most likely designed on the old viewer, where they were rendered with the `geoadmin` style. Users would like their KMLs to still have the same style as they used to.

Fix: When loading a legacy KML layer, we set its style to geoadmin.

Test link with a kml layer : [TEST](https://sys-map.dev.bgdi.ch/preview/fix-pb-1494-use-old-style-on-legacy-links/index.html?layers=KML||https://sys-public.dev.bgdi.ch/api/kml/files/Gz7X_4hAQDyb3_f13UukcA)
same layer on geoadmin : [CURRENT](https://map.geo.admin.ch?layers=KML||https://sys-public.dev.bgdi.ch/api/kml/files/Gz7X_4hAQDyb3_f13UukcA)

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1494-use-old-style-on-legacy-links/index.html)


pakb:
can you add a test somewhere in Cypress so that we cover that?

PR #1295, opened 2025-04-14
Update libs

pakb:
adapt SCSS code to SASS 1.77.7 (see https://sass-lang.com/documentation/breaking-changes/mixed-decls/) -> moving all "main" declaration before any nested declaration

[Test link](https://sys-map.dev.bgdi.ch/preview/update-libs/index.html)

### DIFF
--- a/packages/mapviewer/tests/cypress/support/commands.js
+++ b/packages/mapviewer/tests/cypress/support/commands.js
@@ -3,7 +3,6 @@ import 'cypress-wait-until'
 import '@4tw/cypress-drag-drop'
 import { registerProj4, WEBMERCATOR } from '@geoadmin/coordinates'
 import { randomIntBetween } from '@geoadmin/numbers'
-import { MapBrowserEvent } from 'ol'
 import proj4 from 'proj4'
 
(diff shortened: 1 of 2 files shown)


packages/mapviewer/src/utils/components/FeedbackPopup.vue
```diff
@@ -60,15 +60,19 @@ const warning = computed(() => {
...
         transform: translate(-50%, 0%);
         max-height: calc(100vh - $top-margin);
         max-width: 100vw;
+
+        &.dev-disclaimer-present {
+            $top-margin: calc($header-height + $dev-disclaimer-height + 1rem);
+        }
+        & {
+            top: $top-margin;
+        }
```
sommerfe, line 75: Is there a lint rule that detects if the nested declarations are before the main declarations?

schtibe (approved):
LGTM

packages/mapviewer/src/utils/components/FeedbackPopup.vue
pakb: that would require adding Stylelint to the project, but that could help... I'll add that to the "to-do/Christmas whishlist" in JIRA 😉

PR #1296, opened 2025-04-14
PB-540: Adapt help text for drawing

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-540-drawing-tool-adapt-help-text/index.html)


packages/mapviewer/src/modules/drawing/components/DrawingToolbox.vue
```diff
@@ -64,11 +64,20 @@ const selectedLineCoordinates = computed(() => {
...
+            return true
+        } else {
+        // or when extending line
+        return (
+            editMode.value === EditMode.EXTEND &&
+            selectedLineString.value &&
+            selectedLineCoordinates.value?.length > 2
+            )
+        }
+    }
```
sommerfe, line 80:
```suggestion
const isAllowDeleteLastPoint = computed(
    () => {
        // Allow to delete the last point only if we are drawing line or measure
       return isDrawingLineOrMeasure.value) ||
        // or when extending line
        (
            editMode.value === EditMode.EXTEND &&
            selectedLineString.value &&
            selectedLineCoordinates.value?.length > 2
         )
        }
    }
```

sommerfe (approved):
can be maybe simplified but it's ok for readability

PR #1297, opened 2025-04-14
PB-1595: showing mouse coordinates in 3D:

ltkum:
Issues :
- We were sending 3 pairs of Coordinates to the format function while in 3D, and the function could only handle two
- The is3DReady flag was sometimes ready before the viewer could be accessed, leading to a situation where the handler could not be set up

Fixes :
- We're splitting the coordinates and adding the rounded height at the end of it. (the height is optional IMO)
- We set a timeOut function which call itself while the viewer is not ready to ensure the viewer is ready before we set the handler.

meh


packages/mapviewer/src/modules/map/components/cesium/CesiumMouseTracker.vue
```diff
@@ -26,7 +27,6 @@ const { t } = useI18n()
 
 const is3DReady = computed(() => store.state.cesium.isViewerReady)
 const projection = computed(() => store.state.position.projection)
-
 const dispatcher = { dispatcher: 'CesiumMouseTracker.vue' }
 const getViewer = inject('getViewer', () => {}, true)
```
pakb, line 31: I'd maybe more want to go in the direction where the default value returned here would be `undefined`, and then you can mount the component only when the function returns something relevant.

Or if not applicable to a `v-if`, as you are the "top component" after MapView here, you could do something like

```javascript
const getCesiumViewer = inject('getViewer'), () => undefined, true)

watch(() => getCesiumViewer(), () => /* do something when the viewer changes */)
```

I really don't like, and want to avoid any `setTimeout` use when we can deal with what we want to wait differently

ltkum: I started by creating a watcher on the viewer and making it a reactive components, but after a few tests, I noticed that the `isViewerReady` flag never changed... and so it was never set to false when we unmounted the 3D view. Since this flag is set to true once the viewer is created and ready to be used, switching that flag seems to work without needing to make the whole viewer reactive, and we no longer need to set a timeout.

packages/mapviewer/src/modules/map/components/cesium/CesiumMouseTracker.vue
```diff
@@ -26,28 +27,25 @@ const { t } = useI18n()
...
+const getViewer = inject('getViewer', () => undefined, true)
 
 let handler = null
 
 watch(
     is3DReady,
     (newValue) => {
         if (newValue) {
-            setupHandler()
+            nextTick(() => {
```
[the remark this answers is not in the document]
ltkum: it seems to work without the nextTick. Now I'm investigating the webmercator issue :)

PR #1299, opened 2025-04-15
PB-1561: revert CSS changes to generic feature pop up windows

ltkum:
Issue : we adapted the CSS for the generic pop up which holds the 3d features info for now , but the styling doesn't please our users who are using the drawings.

Fix : We revert the changes and the 3d pop up will be done when we can get it from the backend :)

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1561-drawing-tooltip-descriptions-readability/index.html)


pakb (changes requested):
there's no more floating edit window when drawing, please fix this

pakb (approved):

PR #1301, opened 2025-04-16
PB-1202 : Fix overflow issue in 3D feature tooltip view on Chrome

sami-nouidri-swisstopo:
Issue : An overflow was causing two scroll bars to appear on Chromium browsers when using the 3D view.

Fix : removed the fluid css condition, and added logic to switch max height depending on aspect ratio of viewport.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1202-double-overflow-3d-tooltip/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/infobox/components/FeatureList.vue
+++ b/packages/mapviewer/src/modules/infobox/components/FeatureList.vue
@@ -72,11 +72,14 @@ function loadMoreResultForLayer(layerId) {
 <template>
     <div
         ref="featureListContainer"
-        class="clear-no-ios-long-press"
-        :class="['feature-list', isPhoneMode ? 'mobile' : 'desktop']"
+        class="clear-no-ios-long-press feature-list"
+        :class="[isPhoneMode ? 'mobile' : 'desktop']"
         data-cy="highlighted-features"
     >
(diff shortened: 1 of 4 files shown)


packages/mapviewer/src/modules/infobox/components/FeatureList.vue
```diff
@@ -111,19 +111,13 @@ function loadMoreResultForLayer(layerId) {
...
-    &:not(.fluid) {
-        max-height: 33vh;
-        overflow-x: hidden;
-        overflow-y: auto;
-        display: grid;
-        grid-template-columns: repeat(auto-fit, minmax($overlay-width, 1fr));
-        justify-content: stretch;
-        align-content: stretch;
-    }
+.feature-list.fluid {
```
pakb, line 114: you can remove the `.fluid` here, as we don't use it anymore

pakb (commented):
there's some overflow now in 2D, see here
image

but not in 3D
image

pakb (commented):
Can you also remove the use of the now deleted `fluid` prop in `OpenLayersHighlightedFeatures.vue`?

pakb (approved):

PR #1303, opened 2025-04-17
Add a build dev watch option to geoadmin packages

sommerfe:
Running `pnpm run build:dev:watch` permits the user to modify the geodamin package and updating the build so that the changes have direct effect on the running mapviewer.

Note: I observed that the mapviewer updates the open page too fast, which makes it not load correctly and a reload of the page is sometimes needed.

[Test link](https://sys-map.dev.bgdi.ch/preview/add-build-watch-packages/index.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
     "scripts": {
         "build-all": "pnpm run -r build",
         "build:dev": "pnpm -r run build:dev",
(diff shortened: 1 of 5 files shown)


pakb (commented):
I'd even go one step further and add that to the main package, so that we can watch all libs at once

```JSON
"build:dev:watch": "pnpm -r --if-present run build:dev:watch --filter=!web-mapviewer",
```

pakb (approved):
nice `--parallel`, I forgot about it 😅

PR #1304, opened 2025-04-17
PB-1491 : Add a test to parse an online kml with the wrong content-type

sami-nouidri-swisstopo:
This is an attempt at catching this sort of problem in the future. As of now, the test passses.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1491-s3-kml-content-type-parsing/index.html)

### DIFF
--- a/packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
+++ b/packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
@@ -549,12 +549,26 @@ describe('The Import File Tool', () => {
             fixture: outOfBoundKMLFile,
         })
 
+        const validOnlineUrlWithInvalidContentType = 'https://somes3bucket.com/valid-kml-file.kml'
+        createHeadAndGetIntercepts(
+            validOnlineUrlWithInvalidContentType,
+            'ValidKmlFileWrongContentType',
+            {
+                headers: {
+                    'Content-Type': 'application/octet-stream',
+                },
+                statusCode: 200,
+                body: `<kml> This is an empty kml</kml>>`,
+            }
+        )
+
         cy.goToMapView(
             {
                 layers: [
                     `KML|${outOfBoundKMLUrl}`,
                     `KML|${invalidFileOnlineUrl}`,
                     `KML|${onlineUrlNotReachable}`,
+                    `KML|${validOnlineUrlWithInvalidContentType}`,
                 ].join(';'),
             },
             true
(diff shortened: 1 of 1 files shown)


packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -502,12 +510,22 @@ describe('The Import File Tool', () => {
...
 
+        const validOnlineUrlWithInvalidContentType = 'https://somes3bucket.com/valid-kml-file.kml'
+        createHeadAndGetIntercepts(
+            validOnlineUrlWithInvalidContentType,
+            'ValidKmlFileWrongContentType',
+            {
+                fixture: validKMLFile,
+            }
+        )
+
```
pakb, line 521: if you are using this utils, you shouldn't be declaring your own HEAD intercept above, but give instead this function a `headConfig`

I'm guessing this function overrides your Content-Type above, because it declares a second intercept on the same URL

sami-nouidri-swisstopo: Adressed the change, the test still passes

packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -502,12 +503,28 @@ describe('The Import File Tool', () => {
             fixture: outOfBoundKMLFile,
         })
 
+        const wrongContentTypeHeadConfig = {
+            statusCode: 200,
+            headers: {
+                'Content-Type': 'text/plain',
+            },
+            fixture: validKMLFile,
```
pakb, line 511: no need to give a fixture to a HEAD request (that's the whole purpose, don't give me the content, just the headers :wink: )

packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -516,48 +533,84 @@ describe('The Import File Tool', () => {
...
                         .should('be.visible')
-                        .contains('file not accessible')
-                } else if (index === 1) {
-                    cy.get(`[data-cy^="floating-button-has-error-${invalidFileOnlineUrl}-"]`)
+                        .trigger('mouseover')
+
+                    cy.get(`[data-cy^="floating-button-has-error-${url}-"]`)
                         .should('be.visible')
-                        .contains('Invalid file')
+                        .contains(errorMessage)
```
pakb, line 595: I'd maybe add a little `.trigger('mouseout')` somewhere so that the `mouseover` that was triggered above is "cancelled"

packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -502,12 +502,26 @@ describe('The Import File Tool', () => {
...
+            validOnlineUrlWithInvalidContentType,
+            'ValidKmlFileWrongContentType',
+            {
+                headers: {
+                    'Content-Type': 'application/octet-stream',
+                },
+                statusCode: 200,
+                body: `<kml> This is an empty kml</kml>>`,
+            }
+        )
```
pakb, line 516: I think you missed that this function has a 4th param to set the `HEAD` config, here you are only setting the `GET`, meaning the `Content-Type` for the `HEAD` will still be valid (coming from the default value for the 4th param)

packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -563,61 +583,98 @@ describe('The Import File Tool', () => {
...
-                    cy.get(`[data-cy^="floating-button-has-error-${invalidFileOnlineUrl}"]`)
+                        .trigger('mouseover')
+
+                    cy.get(`[data-cy^="floating-button-has-error-${url}"]`)
                         .should('be.visible')
-                        .contains('Invalid file')
+                        .contains(errorMessage)
+
+                    // Mouseout to remove tooltip
+                    cy.wrap($layer).trigger('mouseout', { force: true })
```
pakb, line 646: not sure this will really be removing the tooltip, as you are triggering the `mouseout` on a different element than the `mouseover`.
Might be a good idea to replicate the selector above and add the missing `.find('[data-cy^="button-has-error"]')` before triggering the event

packages/mapviewer/tests/cypress/tests-e2e/importToolFile.cy.js
```diff
@@ -563,61 +583,98 @@ describe('The Import File Tool', () => {
...
         // Test removing a layer
-        cy.log('Test removing all invalid kml layer')
-        cy.get(`[data-cy^="button-remove-layer-${invalidFileOnlineUrl}"]:visible`).click({
+        cy.log('Test removing all kml layer')
+        cy.get(
+            `[data-cy^="button-remove-layer-${validOnlineUrlWithInvalidContentType}-"]:visible`
+        ).click({
+            force: true,
+        })
+        cy.get(`[data-cy^="button-remove-layer-${invalidFileOnlineUrl}-"]:visible`).click({
```
pakb, line 665: why is there a trailing `-` at the end here?
If the HTML is rendered with that character, I'd prefer we remove it, it's not "nice" 😉

PR #1306, opened 2025-04-17
PB-1602: fix duplicates layer can not be compared

sommerfe:
[Test link with layers from the ticket](https://sys-map.dev.bgdi.ch/preview/fix-pb-1602-duplicates-layer-compare/index.html#/map?lang=en&center=2660000,1190000&z=1&topic=ech&compareRatio=0.504&layers=ch.bafu.luftreinhaltung-feinstaub_pm10@year=2020,,1;ch.bafu.luftreinhaltung-feinstaub_pm10@year=2001,,1&bgLayer=ch.swisstopo.pixelkarte-farbe)

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1602-duplicates-layer-compare/index.html)

### DIFF
--- a/packages/mapviewer/src/api/layers/AbstractLayer.class.js
+++ b/packages/mapviewer/src/api/layers/AbstractLayer.class.js
@@ -1,4 +1,5 @@
 import { cloneDeep } from 'lodash'
+import { v4 as uuidv4 } from 'uuid'
 
 import { InvalidLayerDataError } from '@/api/layers/InvalidLayerData.error'
 
@@ -44,6 +45,9 @@ export class LayerAttribution {
  */
 export default class AbstractLayer {
     /**
+     * @param {String} layerData.uuid Unique ID of this layer (UUID v4) to be able to differntiate
+     *   between the same layers (e.g. when a using a layer multiple times in the map to show
+     *   different timestamps)
      * @param {String} layerData.name Name of this layer in the current lang
      * @param {String} layerData.id The unique ID of this layer that will be used in the URL to
      *   identify it (and also in subsequent backend services for GeoAdmin layers)
@@ -84,6 +88,7 @@ export default class AbstractLayer {
             throw new InvalidLayerDataError('Missing layer data', layerData)
         }
         const {
+            uuid = uuidv4(),
             name = null,
             id = null,
             type = null,
@@ -112,6 +117,7 @@ export default class AbstractLayer {
         if (baseUrl === null) {
             throw new InvalidLayerDataError('Missing base URL', layerData)
         }
+        this.uuid = uuid
         this.name = name
         this.id = id
         this.type = type
@@ -238,6 +244,8 @@ export default class AbstractLayer {
     }
 
     clone() {
-        return cloneDeep(this)
+        const cloneLayer = cloneDeep(this)
+        cloneLayer.uuid = uuidv4()
+        return cloneLayer
     }
 }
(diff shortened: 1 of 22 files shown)


schtibe (commented):
Interesting attempt. Maybe we could leverage this new layer `uuid` otherwise: in some cases, we identify the layer based on it's index. See for instance in `TransparencySlider.vue`: we update the transparency based on the index of the layer in the store. This has lead to some errors a few weeks back when I was refactoring things.
Identifying the layers by index in an array isn't very robust IMO, so maybe we could start using the uuid there too?

sommerfe:
> Interesting attempt. Maybe we could leverage this new layer `uuid` otherwise: in some cases, we identify the layer based on it's index. See for instance in `TransparencySlider.vue`: we update the transparency based on the index of the layer in the store. This has lead to some errors a few weeks back when I was refactoring things. Identifying the layers by index in an array isn't very robust IMO, so maybe we could start using the uuid there too?

Yes seems like a good idea and the right use case for it.
@pakb is there a ticket where we can put some refactoring suggestions?

pakb (commented):
There isn't a ticket yet, we could create one.

Here I think we could solve this issue without resorting to UUIDs, and use instead some business data to differentiate layers.
We could add the timestamp used to describe the layer as part of the ID for instance, as it is here what creates the issue (having twice the same layer but with two different time definition)

sommerfe:
> There isn't a ticket yet, we could create one.
> 
> Here I think we could solve this issue without resorting to UUIDs, and use instead some business data to differentiate layers. We could add the timestamp used to describe the layer as part of the ID for instance, as it is here what creates the issue (having twice the same layer but with two different time definition)

Yes using the currentTimestamp in addition to the layer id was also my first thought, but this doesn't work when we change the timestamp of the layer. This then breaks the possibility to correctly identify the layer so we need a more persistent solution and that's why i chose the uuid

packages/mapviewer/src/api/layers/AbstractLayer.class.js
pakb, line 1: I'd add a UUID generation in the `clone` function below, that might help with the comment you made in `LayerParamConfig.class.js`

packages/mapviewer/src/modules/map/components/CompareSlider.vue
```diff
@@ -91,17 +92,22 @@ function registerRenderingEvents(layerId) {
...
-function unRegisterRenderingEvents(layerId) {
-    const layer = getLayerFromMapById(layerId)
+function unRegisterRenderingEvents(layerId, currentTimestamp) {
+    const layer = getLayerFromMapById(layerId, currentTimestamp)
     layer?.un('prerender', onPreRender)
     layer?.un('postrender', onPostRender)
 }
 
-function getLayerFromMapById(layerId) {
+function getLayerFromMapById(layerId, currentTimestamp) {
```
pakb, line 101: why not then switch to a `getLayerFromMapByUUID` here?

packages/mapviewer/src/router/storeSync/LayerParamConfig.class.js
```diff
@@ -161,6 +161,13 @@ export function createLayerObject(parsedLayer, currentLayer, store, featuresRequ
         )
     }
 
+    // Attention the uuid that is automatically generated in the AbstractLayer class is not always sufficient
+    // if we create multiple times the same layer (e.g. with different timestamps) the already created uuid would be the same for both layers
+    // that's why whe have to create a new uuid here
```
pakb, line 166: Are you referring to the use of the duplicate button here? If so I think this could be solved by editing a bit the `clone` function as said in my previous comment (and maybe make this addition of code irrelevant)

packages/mapviewer/src/api/layers/AbstractLayer.class.js
sommerfe: ah yes perfect that solves the problem :+1: nice catch

pakb (approved):

PR #1309, opened 2025-04-23
PB-1606: Tooltip doesn't translate and needs some padding

sommerfe:
[Test link from ticket](https://sys-map.dev.bgdi.ch/preview/fix-pb-1606-tooltip-translate-padding/index.html#/map?lang=en&center=2658409.2,1192494.8&z=1&topic=ech&layers=ch.bfs.statistische-grundeinheiten_stufe2,f;ch.bfs.statistische-grundeinheiten_stufe1@features=1606004:1606005:1606006:1606007:1606008:1606011:1606013:1606014:1606015:1606018&bgLayer=ch.swisstopo.pixelkarte-farbe&featureInfo=default&catalogNodes=ech)

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1606-tooltip-translate-padding/index.html)


packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -181,11 +181,10 @@ function getIframeHosts(value) {
 :global(.t_list) {
     width: 100%;
 }
-td {
+:global(td) {
```
pakb, line 184: I don't know if we should maybe try to scope that a bit more, here we are applying that to ALL `<td>` on the app.
Could we use something like `:global(.htmlpopup-container td)` to reduce a bit the scope of this rule?

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -181,11 +181,10 @@ function getIframeHosts(value) {
...
     vertical-align: top;
 }
-
-td.cell-left {
+:global(td.cell-left) {
```
pakb, line 187: although `.cell-left` is scoping this rule better than the one above, I'd also would find it better if we'd use `.htmlpopup-container` class to reduce the scope

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -91,6 +91,8 @@ describe('Testing the feature selection', () => {
             goToMapViewWithFeatureSelection('TOoLtIp')
             checkFeatures()
             checkFeatureInfoPosition(FeatureInfoPositions.BOTTOMPANEL)
+            // verifying that the left table cell is not too close to the right table cell
+            cy.get('td.cell-left').should('have.css', 'padding-right', '10px')
```
pakb, line 95: I'd really prefer we use some `data-cy` selector here ([see best practices for Cypress](https://docs.cypress.io/app/core-concepts/best-practices#Selecting-Elements))
I know it might be tricky, because this code comes from the backend. But we could add our `data-cy` directly in the fixture if needed

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -248,6 +250,55 @@ describe('Testing the feature selection', () => {
...
+        it('Adds pre-selected features and verifys the translation of the feature text after changing the language', () => {
+            cy.log('Open the map with a feature preselected in english')
+            goToMapViewWithFeatureSelection(FeatureInfoPositions.DEFAULT)
+            checkFeatures()
+            checkFeatureInfoPosition(FeatureInfoPositions.BOTTOMPANEL)
+
+            cy.wait(`@htmlPopup`)
+
+            cy.get('[data-cy="feature-list-inner"]').find('strong').contains('Kultur Gueter')
+            cy.get('.htmlpopup-container')
```
pakb, line 262: same here, would prefer to not select element by anything else than a `data-cy` custom attribute

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -248,6 +250,55 @@ describe('Testing the feature selection', () => {
...
         })
+        it('Adds pre-selected features and verifys the translation of the feature text after changing the language', () => {
+            cy.log('Open the map with a feature preselected in english')
+            goToMapViewWithFeatureSelection(FeatureInfoPositions.DEFAULT)
+            checkFeatures()
+            checkFeatureInfoPosition(FeatureInfoPositions.BOTTOMPANEL)
+
+            cy.wait(`@htmlPopup`)
+
+            cy.get('[data-cy="feature-list-inner"]').find('strong').contains('Kultur Gueter')
```
pakb, line 261: Here too, maybe not using `<strong>` but have some `data-cy` would make it more resilient to change

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -248,6 +250,55 @@ describe('Testing the feature selection', () => {
...
+            cy.wait(`@layersConfigDe`)
+
+            cy.log('checking that the feature was translated to german')
+
+            cy.get('[data-cy="highlighted-features"]')
+                .as('highlightedFeatures')
+                .should('be.visible')
+
+            cy.get('[data-cy="feature-list-inner"]').find('strong').contains('Kultur Gueter DE')
+            cy.get('.htmlpopup-container')
```
pakb, line 292: same

packages/mapviewer/src/modules/infobox/components/FeatureList.vue
```diff
@@ -33,8 +33,15 @@ const canLoadMore = computed(() => (layerId) => {
     )
 })
 
-watch(lang, () => {
-    store.dispatch('updateFeatures', dispatcher)
+// when the language changes, we have to wait until the active layers are updated
+// before dispatching the updateFeatures action
+// else the feature.layer.name will not be up to date
+const lastLang = ref(lang.value)
```
schtibe, line 39: I think it's recommended to keep the `refs` further above in the code:
https://vuejs.org/style-guide/rules-recommended.html#component-instance-options-order

schtibe (approved):
Just a little comment about the order of elements in the vue component, otherwise it looks good :+1:

pakb (approved):

packages/mapviewer/src/modules/infobox/components/FeatureList.vue
sommerfe: The problem with this is that the ref depends on a computed value so it can not be above the computed values

PR #1310, opened 2025-04-24
PB-1603: add stylelint

sommerfe:
This is a proposition for the stylelint configuration.

To add stylelint to vscode add the plugin stylelint and add this to the settings.json
`"stylelint.validate": [
  "css",
  "scss",
  "vue"
]`
And if needed for the autoformat on save
`"editor.codeActionsOnSave": {
    "source.fixAll.stylelint": "explicit"
  },`

⚠️ Please also verify during review that the application did not change / that anything broke.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1603-add-stylelint/index.html)


stylelint.config.js
```diff
@@ -0,0 +1,88 @@
...
+    "stylelint-config-recommended-scss"
+  ],
+  plugins: ["stylelint-scss", "stylelint-order"],
+  overrides: [
+    {
+      files: ["**/*.vue"],
+      customSyntax: "postcss-html"
+    }
+  ],
+  rules: {
```
schtibe, line 14: I read in the docs that (no rules are turned on by default)[https://stylelint.io/user-guide/configure#rules]. You specify certain rules as `null`. Could you not just *not* include these rules?

packages/mapviewer/src/views/PrintView.vue
```diff
@@ -220,21 +220,27 @@ $printFontSize: max(max($printFontSizeRelToWidth, $printFontSizeRelToHeight), $m
             text-align: justify;
         }
     }
+
```
schtibe, line 223: Whitespaaaace! I like.

packages/mapviewer/src/modules/menu/components/LayerDescriptionPopup.vue
```diff
@@ -156,17 +156,16 @@ onMounted(async () => {
...
 
 <style lang="scss">
-// No scoping here as we need to apply styles to the markup we included with v-html.
-@import '@/scss/variables.module';
-@import '@/scss/webmapviewer-bootstrap-theme';
 
-$spacing: 8px;
+
+@import '@/scss/variables.module';
+@import '@/scss/webmapviewer-bootstrap-theme';$spacing: 8px;
```
schtibe, line 162: This seems to be an error?

packages/mapviewer/src/modules/menu/components/LayerDescriptionPopup.vue
```diff
@@ -180,30 +179,34 @@ $spacing: 8px;
...
         font-size: 100%;
         width: fit-content;
         border: 0;
     }
+
     td {
         padding-right: 2rem;
     }
-}
+} // No scoping here as we need to apply styles to the markup we included with v-html.
```
schtibe, line 211: Nothing to do with your change per se, but since we're already at it: could (deep selectors)[https://vuejs.org/api/sfc-css-features.html#deep-selectors] work here?

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
```diff
@@ -172,19 +172,24 @@ function getIframeHosts(value) {
...
 }
 // fix for layer HTML containing table, such as ch.bafu.gefahren-aktuelle_erdbeben
+/* stylelint-disable-next-line selector-class-pattern */
 :global(.htmlpopup-container .t_list) {
     width: 100%;
 }
+
 :global(.htmlpopup-container td) {
-    vertical-align: top;
+    width: 100%;
```
schtibe, line 190: Why this change?

package.json
```diff
@@ -10,8 +10,10 @@
         "dev": "pnpm run -r --filter=web-mapviewer dev ",
         "format": "prettier --write .",
         "format:check": "prettier --check .",
-        "lint": "eslint --fix",
-        "lint:no-fix": "eslint",
+        "lint": "eslint --fix && stylelint \"**/*.{vue,scss}\" --fix",
```
schtibe, line 13: I'm wondering whether we should use `||` instead. So that when the eslint fails, we still get the feedback from `styleint`

schtibe (commented):
I like it :+1: 
The only thing that is a bit unclear to me is the rules you changed. Did you change them because we use them as such?
I'm generally in favor of using the standard rule set by the community as closely as possible, as I believe that they probably give it way more thought on it than I do :wink:

stylelint.config.js
sommerfe: when we are using the extends 
`  extends: [
    "stylelint-config-standard",
    "stylelint-config-recommended-vue",
    "stylelint-config-recommended-scss"
  ],`

Then certain rules are already activated. Some we just do not want, like "no-duplicate-selectors" because we want to comply with the new sass rules.
Others like "import-notation" seemed weird to me to use `@import url("x.jpg");` instead of `@import "x.jpg";` so i adapted the to what i like 😅. But I think we can have a discussion here to see what community rules we want to adhere to which i have removed/explicitly added.

packages/mapviewer/src/modules/infobox/components/FeatureDetail.vue
sommerfe: I think it was a rebase mistake on my side

packages/mapviewer/src/modules/menu/components/LayerDescriptionPopup.vue
sommerfe: Nice catch, i tested it and it works so i added it to the PR :+1:

stylelint.config.js
pakb: I don't mind sticking to default rules, as long as the linter will fix anything incorrect
if import should be wrapped around `url()` call, maybe there's a good reason for that? I don't mind

I agree with you and would like to stick mostly to the default settings of the linter. With ESLint we have way too many custom rules, and it is complicated to translate that to a new project, or simply to understand what does what

sommerfe: Ok sure, so i'll remove most of the custom rules

sommerfe: @pakb @schtibe 
i now removed most of the custom rules and only kept the ones that are really necessary.
Please check again if i should change anything ☺️

pakb (commented):
Couple little things I saw :

image
The border color of "handles" in the drawing module has changed

image
The area, drawn between the lines when drawing a measure/line, is now opaque

pakb (commented):
Some font size are a bit off in the layer detail, see here

map.geo.admin.ch
image


after StyleLint
image

PR #1311, opened 2025-04-28
PB-1522 : DPI responsive print margin

pakb:
We will move the generation of margin in the viewer, instead of dealing with that on the backend. This is because it's way easier to keep the "true-to-scale" when shrinking the map viewport (and let it recalculate everything) instead of resizing the flatten map and thus changing the scale of the image.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1522-dpi-responsive-print-margin/index.html)

### DIFF
--- a/packages/mapviewer/src/config/print.config.js
+++ b/packages/mapviewer/src/config/print.config.js
@@ -18,3 +18,5 @@ export const PRINT_DIMENSIONS = {
 }
 
 export const PRINT_DEFAULT_DPI = 96
+
+export const PRINT_MARGIN_IN_MILLIMETERS = 4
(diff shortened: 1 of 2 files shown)


packages/mapviewer/src/config/print.config.js
```diff
@@ -18,3 +18,5 @@ export const PRINT_DIMENSIONS = {
 }
 
 export const PRINT_DEFAULT_DPI = 96
+
+export const PRINT_MARGIN_IN_MILLIMETERS = 4
```
rebert, line 22: PRINT_PADDING_IN_MILIMETERS ?

rebert (approved):
Nice - and the true to scale is still working

PR #1314, opened 2025-04-28
PB-1253: stop external layers to never be ready

ltkum:
Issue : Whenever we loaded external layers with parameters in the base URL, we sometimes had issues where the base Url of the get Capabilities and the base URL of the layer were not the same, thus blocking the equality check that allowed the layer to update. This caused the layer to never show that it was ready, even when it was.
Fix : When we update the external layer through its capabilities, if there was a baseUrl before, we set the baseUrl of the `updated` layer to the original one, ensuring the layer will be updated.


pakb (commented):
Can you add a test to cover this ?

ltkum:
> Can you add a test to cover this ?

yes

pakb (approved):

PR #1315, opened 2025-04-29
PB-1471: kml validation

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1471-kml-validation/index.html)

### DIFF
--- a/packages/mapviewer/src/utils/__tests__/kmlUtils.spec.js
+++ b/packages/mapviewer/src/utils/__tests__/kmlUtils.spec.js
@@ -12,6 +12,8 @@ import { fakeIconSets } from '@/utils/__tests__/legacyKmlUtils.spec.js'
 import { BLUE } from '@/utils/featureStyleUtils'
 import { getIcon, getKmlExtent, parseIconUrl, parseKml } from '@/utils/kmlUtils'
 
+import { isKmlFeaturesValid } from '../kmlUtils'
+
 describe('Test KML utils', () => {
     describe('get KML Extent', () => {
         it('handles correctly invalid inputs', () => {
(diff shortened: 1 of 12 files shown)


packages/mapviewer/src/utils/kmlUtils.js
```diff
@@ -602,6 +605,31 @@ export function parseKml(kmlLayer, projection, iconSets, iconUrlProxy = iconUrlP
     return features
 }
 
+/**
+ * Check if the KML features are valid
+ *
+ * @param {string} kmlData KML data
+ * @returns {boolean} Returns true if the KML data is valid, false otherwise
+ */
+export function isKmlFeaturesValid(kmlData) {
```
pakb, line 614: can you add two simple unit tests to this function, just to make sure?

pakb (commented):
I'll add the new i18n keys to the master file 👍

packages/mapviewer/src/utils/__tests__/kmlUtils.spec.js
```diff
@@ -119,6 +121,18 @@ describe('Test KML utils', () => {
             ])
         })
     })
+    describe('isKmlFeaturesValid', () => {
+        it('returns false if there is an error in the features of the KML', () => {
+            const kml = readFileSync(resolve(__dirname, '../../../tests/cypress/fixtures/import-tool/kml_feature_error.kml'), 'utf8')
```
pakb, line 126: It's not a bad idea to re-use assets from Cypress, I just fear it will be modified without knowing that it is used by a unit test too.

Would it be hard to generate quick "on-the-fly" fake KML data, such as in the `get extent of multiples features (marker, line, polygone)` test?

Or simply to copy this Cypress KML inside this `__test__` folder, so that its requirement is expressed in a clearer manner.

pakb (approved):

PR #1316, opened 2025-04-29
PB-1636: add parameter to remove UI in embed view

ltkum:
- Issue : We want to remove the UI from the embed view in some cases, so when we are in käze, we can use the sites native UI.
- Fix : We add an URL parameter whose role it is to remove these buttons.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1636-add-button-hiding-param-for-gps/index.html)


ltkum:
I'll add a test before asking for a review :)

packages/mapviewer/src/store/modules/ui.store.js
```diff
@@ -183,6 +183,14 @@ export default {
          * @type Boolean
          */
         showDragAndDropOverlay: false,
+
+        /**
+         * Flag telling if we should hide the UI elements in the embed viewer. Zoom buttons, 3d
+         * button and the `view on geoadmin link`
+         *
+         * @type Boolean
+         */
```
pakb, line 192: Might make sense to add in there that this is specific for GeoPlateformSchweiz, and that we shouldn't disclose this param (as it will soon be obsolete and GPS will replace the use of iframe with the new JS API as soon as it's available)

ltkum: good idea

PR #1319, opened 2025-04-30
Showing a banner informing our users about the current outage - #patch

pakb:
Showing a banner informing our users about the current outage
Should be reverted as soon as the outage is over

[Test link](https://sys-map.int.bgdi.ch/preview/wms_outage_banner/index.html)


ltclm (approved):

ltclm:
thank you  :+1:

PR #1321, opened 2025-05-01
Revert "Showing a banner informing our users about the current outage" - #patch

pakb:
This reverts commit 5dd68961d8059e2f60b0fe3c0c2a352c4d953c3a.

[Test link](https://sys-map.int.bgdi.ch/preview/remove-outage-banner/index.html)


schtibe:
:+1:  I guess we'll consider having a permanent functionality that we can more easily enable in the future?
(BTW: I noticed that in the iframes the notice doesn't appear. Don't know if that's good or not, just wanted to mention it for consideration if we create a proper one)

schtibe (approved):

PR #1322, opened 2025-05-02
PB-1300: KML feature selection 3D

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1300-kml-feature-selection-3d/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
+++ b/packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
@@ -1,8 +1,17 @@
 <script setup>
 import { WEBMERCATOR, WGS84 } from '@geoadmin/coordinates'
 import log from '@geoadmin/log'
-import { Cartesian2, Cartographic, PostProcessStageLibrary, ScreenSpaceEventType } from 'cesium'
-import { Point } from 'ol/geom'
+import { bbox } from '@turf/turf'
+import { centroid } from '@turf/turf'
+import {
+    Cartesian2,
+    Cartographic,
+    Math,
+    PostProcessStageLibrary,
+    ScreenSpaceEventType,
+} from 'cesium'
+import GeoJSON from 'ol/format/GeoJSON'
+import { LineString, Point, Polygon } from 'ol/geom'
 import proj4 from 'proj4'
 import { computed, inject, onMounted, onUnmounted, watch } from 'vue'
 import { useStore } from 'vuex'
@@ -183,7 +192,7 @@ function onClick(event) {
         'with objects',
         objects
     )
-    const kmlFeatures = {}
+
     // if there is a GeoJSON layer currently visible, we will find it and search for features under the mouse cursor
     visiblePrimitiveLayers.value
         .filter((l) => l instanceof GeoAdminGeoJsonLayer)
@@ -197,6 +206,7 @@ function onClick(event) {
                 )
             )
         })
+
     visiblePrimitiveLayers.value
         .filter((l) => l instanceof KMLLayer)
         .forEach((kmlLayer) => {
@@ -209,10 +219,13 @@ function onClick(event) {
                         'for KML layer',
                         kmlLayer
                     )
-                    // TODO PB-1300 implement KML features selection here
+                    const kmlLayerFeature = create3dKmlFeature(viewer, kmlFeature, kmlLayer)
+                    if (kmlLayerFeature) {
+                        features.push(kmlLayerFeature)
+                    }
                 })
-            features.push(...Object.values(kmlFeatures))
         })
+
     handleClickHighlight(features, coordinates)
 
     if (!coordinates.length && features.length) {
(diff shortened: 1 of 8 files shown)


packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +243,62 @@ function onClick(event) {
     viewer.scene.requestRender()
 }
 
+function create3dKmlFeature(viewer, kmlFeature, kmlLayer) {
+    let geometry
+    let featureCoordinates
+    if (kmlFeature.id.position) {
```
pakb, line 249: are we 100% sure `kmlFeature` and `kmlFeature.id` will be defined? If not we could watch our backs by adding a little `?` between each selectors

pakb: or have an early empty return
```Javascript
if (!kmlFeature || !kmlFeature.id) {
    return
}
```

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +243,62 @@ function onClick(event) {
     viewer.scene.requestRender()
 }
 
+function create3dKmlFeature(viewer, kmlFeature, kmlLayer) {
```
pakb, line 246: Might make sense to add a little JSDoc here and at least describe what types each inputs are (it's not TS yet, so we need some help to understand better what's happening 😉 )

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +243,62 @@ function onClick(event) {
...
+    let geometry
+    let featureCoordinates
+    if (kmlFeature.id.position) {
+        const position = kmlFeature.id.position.getValue(viewer.clock.currentTime)
+        const cartographic = Cartographic.fromCartesian(position)
+
+        featureCoordinates = proj4(WGS84.epsg, WEBMERCATOR.epsg, [
+            // pos is in EPSG:4978 but proj4 does not support it so it is transformed to EPSG:4326 first
```
pakb, line 254: little confusion here, `EPSG:4978` is pretty much equal to `EPSG:4326` (both are WGS84), here you are transforming into `EPSG:3857` (metric mercator)

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +243,62 @@ function onClick(event) {
...
+            Math.toDegrees(cartographic.longitude),
+            Math.toDegrees(cartographic.latitude),
+        ])
+        geometry = new Point(featureCoordinates)
+    }
+    if (kmlFeature.id.polyline) {
+        const positions = kmlFeature.id.polyline.positions.getValue(viewer.clock.currentTime)
+        featureCoordinates = positions.map((pos) => {
+            // pos is in EPSG:4978 but proj4 does not support it so it is transformed to EPSG:4326 first
```
pakb, line 263: same

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +243,62 @@ function onClick(event) {
...
+
+            return proj4(WGS84.epsg, WEBMERCATOR.epsg, [lon, lat])
+        })
+        geometry = new LineString(featureCoordinates)
+    }
+    if (kmlFeature.id.polygon) {
+        const hierarchy = kmlFeature.id.polygon.hierarchy.getValue(viewer.clock.currentTime)
+
+        featureCoordinates = hierarchy.positions.map((pos) => {
+            // pos is in EPSG:4978 but proj4 does not support it so it is transformed to EPSG:4326 first
```
pakb, line 276: same

packages/mapviewer/src/store/plugins/reproject-layers-on-projection-change.plugin.js
```diff
@@ -135,4 +135,15 @@ function reprojectLayerExtent(oldProjection, newProjection, activeLayers, store)
...
 }
 
+/**
+ * Verify if the parameter is an array of depth one
+ * 
+ * @param {Array | any} arr 
+ * @returns {boolean} if the arr is an array of depth one
+ */
+function isDepthOne(arr) {
+  if (!Array.isArray(arr)) return false
```
pakb, line 145: no inline if please (add `{}` around it)

packages/mapviewer/src/utils/layerUtils.js
```diff
@@ -124,20 +125,21 @@ export function createLayerFeature(olFeature, layer, coordinates, geometry) {
         layer: layer,
         id: olFeature.getId(),
         title:
-            olFeature.get('label') ??
+            olFeature.get('label') ||
```
pakb, line 128: why going back to using `||`?

sommerfe: Do you remember in which PR i changed it to `??` ?
In this case i switched it because here if the label is an empty string it will still be valid so the title will be an empty string. But with `||` we take the next value in hope for something more useful

packages/mapviewer/src/modules/map/components/cesium/CesiumInteractions.vue
```diff
@@ -233,6 +246,72 @@ function onClick(event) {
     viewer.scene.requestRender()
 }
 
+/**
+ * This function creates a LayerFeature from a KML feature in 3D.
+ *
+ * @param {Object} viewer Cesium viewer
+ * @param {Object} kmlFeature KML feature
+ * @param {Object} kmlLayer KML layer
```
pakb, line 254: You sure that those can't be a bit better "typed"? Or at least point at to the Cesium online documentation to better know what we are dealing with, with them (no need to describe them the same way they are named 😉 )

packages/mapviewer/src/utils/layerUtils.js
pakb: That might have huge repercussion to change that here, in this context.
As it requires wider testing, I'd prefer we stick to `??` for the time being. There's internal discussion ongoing regarding what's a good "label" or "title" for a feature, so we should let the outcome of this discussion decide what will happen here.

pakb (changes requested):
Please revert back to `??`, too risky IMHO to change that now

pakb (approved):

PR #1323, opened 2025-05-02
PB-1383 Refactor ErrorMessage and WarningMessage to @geoadmin/log

schtibe:
As I started this in my layers module, I thought we could already add this to the main code base. This might save me some rebase issues later on.


pakb (commented):
There are still some import issue, maybe some further work on your layer modules sliped through

schtibe:
> There are still some import issue, maybe some further work on your layer modules slipped through

Oh yes, the cherry-picking did something weird and maybe things were missing. It should work now!

pakb (approved):

PR #1325, opened 2025-05-06
PB-1427: Support uppercase language parameter

ltkum:
Issue: some people can't use lowercase letters for the `lang` URL parameter and wanted to be able to use uppercase letters

Fix: we let people input uppercase letters in the `lang` URL parameter and force it to lowercase on our side.

Also: added some tests

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1427-accept-uppercase-lang-param/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/menu/components/help/MenuHelpSection.vue
+++ b/packages/mapviewer/src/modules/menu/components/help/MenuHelpSection.vue
@@ -1,5 +1,5 @@
 <script setup>
-import { computed, ref } from 'vue'
+import { computed, ref, watch } from 'vue'
 import { useI18n } from 'vue-i18n'
 import { useStore } from 'vuex'
 
@@ -26,7 +26,14 @@ const store = useStore()
 const hasGiveFeedbackButton = computed(() => store.getters.hasGiveFeedbackButton)
 const hasReportProblemButton = computed(() => store.getters.hasReportProblemButton)
 
-const currentLang = ref(store.state.i18n.lang)
+const currentLang = computed(() => store.state.i18n.lang)
+
+// this is needed to pass the selected value to the changelang function
+const selectedLang = ref(currentLang.value)
+
+watch(currentLang, () => {
+    selectedLang.value = currentLang.value
+})
 
 function changeLang(lang) {
     store.dispatch('setLang', {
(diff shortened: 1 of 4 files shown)


packages/mapviewer/src/router/storeSync/storeSync.config.js
```diff
@@ -35,7 +35,11 @@ const storeSyncConfig = [
         keepInUrlWhenDefault: true,
         valueType: String,
         validateUrlInput: (store, query) =>
-            getStandardValidationResponse(query, SUPPORTED_LANG.includes(query), 'lang'),
+            getStandardValidationResponse(
+                query,
```
pakb, line 39: you could directly `.toLowerCase()` the query here, no need to change the store then

packages/mapviewer/src/modules/menu/components/help/MenuHelpSection.vue
```diff
@@ -26,7 +26,10 @@ const store = useStore()
 const hasGiveFeedbackButton = computed(() => store.getters.hasGiveFeedbackButton)
 const hasReportProblemButton = computed(() => store.getters.hasReportProblemButton)
 
-const currentLang = ref(store.state.i18n.lang)
+const currentLang = computed(() => store.state.i18n.lang)
+
+// this is needed to pass the selected value to the changelang function
+const selectedLang = ref(currentLang.value)
```
pakb, line 32: why this change?

ltkum: because when trying to change the lang in the URL on mobile, it would not work, which caused some tests to fail.

packages/mapviewer/src/router/storeSync/storeSync.config.js
ltkum: this is the validation, not the dispatcher.

If we wanted the dispatch to handle the case, we could either: 
 - create a new "complex" param for lang just to override the dispatch function
 - add a "isCaseSensitive" param to `AbstractParamConfig` which would then be used to determine if the value given is case sensitive, should be lowercased or should be uppercase
 
 If that's okay with you, I would go with the first solution.

pakb: nah, nevermind, let's keep it the way you did it then.

pakb (approved):

PR #1334, opened 2025-05-09
PB-1579: add support for local BigTIFF

pakb:
we were missing the little endian signature for BigTIFF (which is different than normal TIFFs) to be able to parse a local BigTIFF file.

It was working fine with URLs already because we do not test against little endian signature in this case, and let GeoTIFF.js library do its magic, so nothing to change there.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1579-local-bigtiff-support/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/menu/components/advancedTools/ImportFile/parser/CloudOptimizedGeoTIFFParser.class.js
+++ b/packages/mapviewer/src/modules/menu/components/advancedTools/ImportFile/parser/CloudOptimizedGeoTIFFParser.class.js
@@ -15,7 +15,12 @@ export class CloudOptimizedGeoTIFFParser extends FileParser {
     constructor() {
         super({
             fileExtensions: ['.tif', '.tiff'],
-            fileTypeLittleEndianSignature: [0x49, 0x49, 0x2a, 0x00],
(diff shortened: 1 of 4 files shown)


packages/mapviewer/src/modules/menu/components/advancedTools/ImportFile/parser/CloudOptimizedGeoTIFFParser.class.js
```diff
@@ -15,7 +15,12 @@ export class CloudOptimizedGeoTIFFParser extends FileParser {
     constructor() {
         super({
             fileExtensions: ['.tif', '.tiff'],
-            fileTypeLittleEndianSignature: [0x49, 0x49, 0x2a, 0x00],
+            fileTypeLittleEndianSignature: [
+                // Standard TIFF
+                [0x49, 0x49, 0x2a, 0x00],
+                // BigTIFF
+                [0x49, 0x49, 0x2b, 0x00],
```
sommerfe, line 22: maybe create a const somewhere

sommerfe (approved):

hansmannj (commented):
Thanks :rocket:

PR #1339, opened 2025-05-15
PB-1695: add the option to show marker description on the map

pakb:
Adds a new checkbox on the feature detail/edit to opt-in and show the description on the map
image

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1695-show-marker-description-on-the-map/index.html)


packages/mapviewer/src/modules/i18n/locales/de.json
```diff
@@ -110,6 +110,7 @@
     "description": "Beschreibung",
     "description_instructions": "Beschreibung des WMS Layers",
     "desktop_redirect": "Link zur Desktop Version",
+    "display_on_map": "Anzeige auf der Karte",
```
sommerfe, line 113: I think better would be:
```suggestion
    "display_on_map": "Auf der Karte anzeigen",
```

schtibe: German language: "let's tease people and put verbs at the end. sometimes."

schtibe:
> Adds a new checkbox on the feature detail/edit to opt-in and show the description on the map image
> 
> [Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1695-show-marker-description-on-the-map/index.html)

Maybe it could be enhanced a little bit: it isn't entirely clear "what" it's supposed to display on the map... Maybe you could increase the margin between the upper text input and the "Description" Label a little bit, then it's clearer that the toggle belongs just to the description (see arrows):
image

on that note, the margins above and below the textarea aren't the same (see highlight)

sommerfe (approved):

PR #1341, opened 2025-05-16
PB-1697: limiting floating window's height

pakb:
Some users that use the browser zoom to enlarge the UI have issues with the SimpleWindow component (especially when it's used by the feedback component) because it grows too high, and overflow the header (some even saying the X button of the window was out of screen).

So to fix that, I've limited the height of all simple windows to 2/3 of the height of the screen. This was tested with a browser zoom of 200% and having the "report a problem" window opened.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1697-simple-window-too-big-when-broser-zoom/index.html)


schtibe (approved):
:+1:

PR #1345, opened 2025-05-16
PB-1054 : Remove mouse events from useMovableElement and replace with pointer events

sami-nouidri-swisstopo:
Mouse events have been removed in favour of pointer events for better compatibility with mobile devices and touch screens.

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1054-tooltip-touch-events/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/map/components/MapPopover.vue
+++ b/packages/mapviewer/src/modules/map/components/MapPopover.vue
@@ -105,9 +105,10 @@ const popoverLimits = computed(() => {
 
 onMounted(() => {
     if (mode === MapPopoverMode.FLOATING && popover.value && popoverHeader.value) {
-        useMovableElement(popover.value, {
-            grabElement: popoverHeader,
-            offset: popoverLimits,
+        useMovableElement({
+            element: popover.value,
+            grabElement: popoverHeader.value,
+            offset: popoverLimits.value,
         })
     }
 })
--- a/packages/mapviewer/src/utils/components/SimpleWindow.vue
+++ b/packages/mapviewer/src/utils/components/SimpleWindow.vue
@@ -78,7 +78,8 @@ onMounted(() => {
     if (movable) {
         const windowElement = windowRef.value
         const headerElement = headerRef.value
-        useMovableElement(windowElement, {
+        useMovableElement({
+            element: windowElement,
             grabElement: headerElement,
             initialPositionClasses: [initialPositionClass.value],
             offset: { bottom: 0, right: 0, left: 0 },
--- a/packages/mapviewer/src/utils/composables/useMovableElement.composable.js
+++ b/packages/mapviewer/src/utils/composables/useMovableElement.composable.js
@@ -1,9 +1,9 @@
-import { onMounted, ref, toValue } from 'vue'
+import { isRef, onMounted, ref } from 'vue'
 
 const defaultPadding = 4 // px
 
 const MovementSource = Object.freeze({
-    MOUSE_DRAG: 'mouse_drag',
+    POINTER_DRAG: 'pointer_drag',
     WINDOW_RESIZE: 'window_resize',
 })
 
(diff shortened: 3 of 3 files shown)


pakb (dismissed):
Good idea, but the outcome isn't there yet.

I've tested a bit on my phone (iPhone) and I can't "scroll" the content of a layer description anymore, as it tries to move it around.
So there's maybe something to do to not prevent scrolling

packages/mapviewer/src/utils/composables/useMovableElement.composable.js
```diff
@@ -63,105 +55,80 @@ export function useMovableElement(element, options = {}) {
...
+            viewport.value.left,
+            viewport.value.right - rect.width
+        )
+        const newTop = clamp(
+            el.offsetTop + deltaY,
+            viewport.value.top,
+            viewport.value.bottom - rect.height
+        )
+
+        placeElementAt(newTop, newLeft, MovementSource.POINTER_DRAG)
```
ltkum, line 104: except the presence of the deltaX / deltaY within the clamp, this is exactly the same as the `constrainElementWithinViewport` function.

I would either : 

 - change the `offsetLeft` / `offsetRight` attributes from elem, then call `constrainElementWithinViewport` or 
  - add the deltaX / deltaY to the function as optional parameters (defaulting to 0) and call the function with the parameter in this way : 
`constrainElementWithinViewport(deltaX, deltaY)`

sami-nouidri-swisstopo: I went with the second option of adding delta arguments to `constrainElementWithinViewport` as `offsetLeft`
and `offsetRight` are getters only.

packages/mapviewer/src/modules/map/components/MapPopover.vue
```diff
@@ -105,7 +105,8 @@ const popoverLimits = computed(() => {
 
 onMounted(() => {
     if (mode === MapPopoverMode.FLOATING && popover.value && popoverHeader.value) {
-        useMovableElement(popover.value, {
+        useMovableElement({
+            element: popover.value,
             grabElement: popoverHeader,
```
ltkum, line 110: it seems to work, but maybe we need to be attentive to the fact that in other cases, we're using the reference's value (`.value`) When tackling [#1309](https://swissgeoplatform.atlassian.net/browse/PB-1309), you might want to remember this difference.

ltkum: Setting this to `grabElement.value` would mean that we can get rid of all the `toValue` in the composable and simplify it, as well as ensuring we always give the same kind of variable when calling the function. I would take advantage of the situation and do it now :)

packages/mapviewer/src/utils/composables/useMovableElement.composable.js
```diff
@@ -14,8 +14,7 @@ const MovementSource = Object.freeze({
  * It is possible to specify custom offset with screen border, in order to avoid some part of the
  * app (such as the header) to be accessible to the movable element
  *
- * @param element Reference to a DOM element
- * @param options Options such as described below
+ * @param config {Object} Configuration object
  * @param {HTMLElement} [options.grabElement] HTML element on which the user has to grab to start
```
ltkum, line 18: we need to update the comment : 
- it's no longer `options.key` but `config.key`.
- also we need to add the `element` to the config doc here

packages/mapviewer/src/utils/composables/useMovableElement.composable.js
```diff
@@ -24,8 +23,18 @@ const MovementSource = Object.freeze({
...
-    const { grabElement = null, offset = null, initialPositionClasses = [] } = options
+export function useMovableElement(config = {}) {
+    const {
+        element = null,
+        grabElement = null,
+        offset = null,
+        initialPositionClasses = [],
+    } = config
+
+    if (!config.element || !config.grabElement) {
```
ltkum, line 34: offset is always defined, we should also enforce `offset` to have a value.

packages/mapviewer/src/utils/composables/useMovableElement.composable.js
```diff
@@ -24,8 +23,18 @@ const MovementSource = Object.freeze({
...
+        throw new Error('Element/grabElement is required')
```
ltkum, line 35: I would make the error a bit more explicit : 

`[useMovableElement] Element, grabElement and offset are required to use a movable element. Received ${element} for element, ${grabElement} for grabElement and ${offset} for offset.` or something like that.

packages/mapviewer/src/utils/composables/useMovableElement.composable.js
```diff
@@ -171,36 +134,25 @@ export function useMovableElement(element, options = {}) {
...
-            currentPosition.top !== positionConstrainedByNewLimits.top ||
-            currentPosition.left !== positionConstrainedByNewLimits.left
-        ) {
-            placeElementAt(
-                positionConstrainedByNewLimits.top,
-                positionConstrainedByNewLimits.left,
-                MovementSource.WINDOW_RESIZE
-            )
-        }
+    function constrainElementWithinViewport(deltaX = 0, deltaY = 0) {
```
ltkum, line 137: I would not put default values here, and add the `0, 0` as parameter in the initialize call. This forces us to be explicit in tests and in the code.

I would also rework the comment a  bit, as we now also handle placing an element and not only ensuring it's within bounds at startup.

ltkum (commented):
A small refinement here and there, but the functionality seems good to me

packages/mapviewer/src/utils/components/SimpleWindow.vue
```diff
@@ -78,8 +78,9 @@ onMounted(() => {
     if (movable) {
         const windowElement = windowRef.value
         const headerElement = headerRef.value
-        useMovableElement(windowElement, {
-            grabElement: headerElement,
+        useMovableElement({
+            element: windowElement.value,
+            grabElement: headerElement.value,
```
ltkum, line 83: thos are already the values of windowRef or headerRef, no need for an extra value here.

ltkum (commented):
One small last change :)

PR #1354, opened 2025-05-27
PB-1220: Get global external layers legend

ltkum:
- Issue: There exists layers where there are legends available which encompass all sublayers from a layer, but they're not described in a `Style` markup in the WMS capabilities. They are however available through the `GetLegendGraphic` request, which is part of the SLD specification.

- Fix: We now check if an external WMS layer supports the SLD specification, and when we have no Legend, we check if the layer can be queryable (and thus, can give back a legend)

- Still to do : At the moment, this needs some work on OpenLayers, as their parser does not parse SLD attributes

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1220-get-global-external-layers-legend/index.html)


packages/mapviewer/src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -431,8 +431,41 @@ export default class WMSCapabilitiesParser {
...
         const styles = layer.Style?.filter((s) => s.LegendURL?.length > 0) ?? []
+        // if we do not have access to the legend in pure WMS fashion, we check if this
+        // WMS follows the SLD specification, and if we can get it from there.
+        if (
+            styles.length === 0 &&
+            layer.queryable &&
+            this.Capability.UserDefinedSymbolization?.SupportSLD &&
+            this.Capability.Request.GetLegendGraphic?.DCPType[0]?.HTTP?.Get?.OnlineResource
+        ) {
+            const url = this._constructLegendFromGetLegendGraphic(
```
pakb: in prep for TS migration, I'd prefer you set up a couple `const` here from the `this.Capability.Request.GetLegendGraphic` object. It will be way easier to type later on (and TS will want to have the guarding if to accept we deconstruct the `this.Capability.Request.GetLegendGraphic` object)
Then you can build the URL in place (no need for the `_constructLegendFromGetLegendGraphic` function, it's only used once here)

packages/mapviewer/src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -431,8 +431,41 @@ export default class WMSCapabilitiesParser {
...
 
+    _constructLegendFromGetLegendGraphic(getLegendGraphicObject, layerId) {
+        return (
+            getLegendGraphicObject.DCPType[0].HTTP.Get.OnlineResource +
+            `SERVICE=WMS` +
+            `&REQUEST=GetLegendGraphic` +
+            `&VERSION=${this.version}` +
+            `&FORMAT=${getLegendGraphicObject.Format[0]}` +
+            `&LAYER=${layerId}` +
+            `&SLD_VERSION=1.1.0`
```
pakb: you should maybe switch to use \` string building pattern overall (instead of concat)

packages/mapviewer/src/api/layers/WMSCapabilitiesParser.class.js
```diff
@@ -433,6 +433,25 @@ export default class WMSCapabilitiesParser {
...
+            layer.queryable &&
+            !!this.Capability.UserDefinedSymbolization?.SupportSLD
+        ) {
+            const getLegendGraphicBaseUrl =
+                this.Capability.Request.GetLegendGraphic?.DCPType[0]?.HTTP?.Get?.OnlineResource
+            const getLegendGraphicFormat = this.Capability.Request.GetLegendGraphic?.Format[0]
+            if (!!getLegendGraphicBaseUrl && !!getLegendGraphicFormat) {
+                return [
+                    {
+                        url: `${getLegendGraphicBaseUrl}SERVICE=WMS&REQUEST=GetLegendGraphic&ERSION=${this.version}&FORMAT=${getLegendGraphicFormat}&LAYER=${layerId}&SLD_VERSION=1.1.0`,
```
pakb: Typo here with `VERSION` param
That's quite a lot of params, might be better to describe them separately as an object, such as

```Javascript
const urlParams = {
    SERVICE: 'WMS',
    REQUEST: 'GetLegendGraphic',
    VERSION: this.version,
    // etc ...
}
```

And then create the URL by doing something like

```Javascript
Object.entries(urlParams).reduce((acc, [key, value], index) => {
  const prefix = index === 0 ? "?" : "&";
  return acc + `${prefix}${key}=${value}`;
}, "");
```

or use the URLSearchParams API

(I'm not even sure you need the part that checks for the `?` instead of `&`, as a WMS base URL should always end with `?` AFAIK)

pakb: Or revert back to concat the String, the linter was doing a better job at splitting each params, one per line, and added more readability for us puny humans 🙃

ltkum: doing a URLSearchParams object and using it to generate the URL would be better IMO. We might still want to add a quick check for the `?` at the end of the base url, as I think you're correct in saying a WMS base URL should always end with it, but I would prefer to be on the safe side just in case there is one implementation where it was forgotten.

PR #1356, opened 2025-05-28
PB-1357 : Enable depth testing for labels

sami-nouidri-swisstopo:
This seems to fix the issue of labels being shown despite being behind terrain. Previously, it seems we weren't doing any depth testing for labels (perhaps intentionally?) so I've enabled it back.

Before : 
image

After : 
image

[Test link](https://sys-map.dev.bgdi.ch/preview/bug-pb-1357-hide-3d-labels-behind-terrain/index.html)


pakb (approved):
I don't remember why I disabled that... Let's roll with what you changed, I like it better when it's on, on the "before/after" screenshot you posted.

I'm surprised with the location you selected to showcase your fix 🦅

PR #1357, opened 2025-06-03
PB-1068: Add CTRL + Click to Toggle Selected Features

ismailsunni:


packages/mapviewer/src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
```diff
@@ -32,14 +32,19 @@ export function useDragBoxSelect() {
...
         const selectExtent = dragBoxSelect.getGeometry()?.getExtent()
         if (selectExtent?.length !== 4) {
             return
         }
+        // Check if the box has a non-zero area
+        const [minX, minY, maxX, maxY] = selectExtent
+        if (minX === maxX || minY === maxY) {
+            // It's a click, not a box selection
+            return
+        }
```
pakb: maybe having some tolerance would be nice, there can be a (very) small drag when clicking

packages/mapviewer/src/modules/map/components/openlayers/utils/useDragBoxSelect.composable.js
```diff
@@ -32,14 +32,21 @@ export function useDragBoxSelect() {
...
-    dragBoxSelect.on('boxstart', () =>
-        store.dispatch('clearAllSelectedFeatures', { ...dispatcher })
-    )
     dragBoxSelect.on('boxend', () => {
         const selectExtent = dragBoxSelect.getGeometry()?.getExtent()
         if (selectExtent?.length !== 4) {
             return
         }
+        // Check if the box has a non-zero area
+        const PIXEL_TOLERANCE = 10 // Minimum drag in pixels to be considered a box
```
pakb: There's this variable in `features.api.js`
```
/**
 * In pixels
 *
 * @type {Number}
 */
const DEFAULT_FEATURE_IDENTIFICATION_TOLERANCE = 10
```

Might make sense to move both to using a single value defined in `map.config.js` or similar

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -344,6 +344,23 @@ describe('Testing the feature selection', () => {
...
+                y: location.y,
+                position: 'topLeft',
+                ctrlKey: ctrlKey,
+            })
+            cy.get('@olMap').realMouseUp({
+                x: location.x,
+                y: location.y,
+                position: 'topLeft',
+                ctrlKey: ctrlKey,
+            })
```
pakb: Can you get away of `cypress-real-events` and use a standard `.click` call?
I know the OL map can be a hell to deal with, so maybe not 🙃

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -506,6 +523,63 @@ describe('Testing the feature selection', () => {
...
+            cy.get('[data-cy="ol-map"]').as('olMap').should('be.visible')
+
+            // This point location is found by listening to the click event on the map with the following code:
+            // cy.get('[data-cy="ol-map"]').then(($el) => {
+            //     $el[0].addEventListener('click', (e) => {
+            //         console.log('Clicked at:', e.offsetX, e.offsetY);
+            //     });
+            // });
+            const point1 = { x: 40, y: 205 }
+            const point3 = { x: 86, y: 359 }
```
pakb: do you know your map coordinates for those points?

If so, you could use the OL instance to give you the correct pixel with [getPixelFromCoordinate](https://openlayers.org/en/latest/apidoc/module-ol_Map-Map.html#getPixelFromCoordinate)
that would be more resilient to change, especially when we will want to test bigger viewports

ismailsunni:
The test was failing over and over, I need to the update from develop branch to fix it.

pakb (commented):
Can you make it so that the selection box isn't reset when we de-select one element from it?

ismailsunni:
@pakb done, I use the same approach as in the layer extents to keep showing the box. It's actually handled internally by OpenLayers.

ismailsunni:
@pakb 
> A little thing that could be also changed : if we toggle (CTRL+click) on a selected feature, any adjacent feature (or underlying feature) should not be toggled on. If we toggle something off, we are are in "Toggle off" mode. I find it a bit confusing when you CTRL click on the map, with only one of two stacked features selected, and the other one is then toggle on. It feels like the app didn't do anything

but how do we differentiate if the user want to select the unselected feature? I mean the reverse case, when the user want to select a feature, the selected adjacent feature will be unselected.

actually, is it really possible? since normally, we select all features in the clicked area or stacked.

pakb (approved):

PR #1358, opened 2025-06-04
PB-1754: add client-side caching through Service Worker API

pakb:
set up caching for HTML/JS/CSS files (so that the app can be reloaded even without network), essentials backend items (layers config, topics) and for a limited amount of tiles from our WMS/WMTS the app has loaded while being used.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1754-client-side-caching/index.html)


pakb:
We should decide on how we want to show users the "offline" feature is ready. Right now it's an ugly popup in the middle of the app.

I propose we add that somewhere in the header (on desktop), and in the help section on mobile

packages/mapviewer/src/config/staging.config.ts
```diff
@@ -0,0 +1,40 @@
+/**
+ * Enum that tells for which (deployment) environment the app has been built.
```
sommerfe: why enum here?

packages/mapviewer/src/utils/offline/OfflinePrompt.vue
```diff
@@ -0,0 +1,137 @@
...
+        if (resp?.status === 200) {
+            await registration.update()
+        }
+    }, period)
+}
+
+const { offlineReady, needRefresh, updateServiceWorker } = useRegisterSW({
+    immediate: true,
+    onRegisteredSW(serviceWorkerUrl, registration) {
+        if (period <= 0) return
```
sommerfe: no single line

packages/mapviewer/src/utils/offline/OfflinePrompt.vue
```diff
@@ -0,0 +1,137 @@
...
+import { useRegisterSW } from 'virtual:pwa-register/vue'
+import { computed, ref } from 'vue'
+
+// check for updates every hour
+const period = 60 * 60 * 1000
+
+const swActivated = ref(false)
+
+/** This function will register a periodic sync check every hour */
+function registerPeriodicSync(serviceWorkerUrl: string, registration: ServiceWorkerRegistration) {
```
schtibe: What does it check exactly?

packages/mapviewer/src/service-workers.ts
```diff
@@ -0,0 +1,113 @@
...
+    )
+
+    function removingTrailingSlash(url: string): string {
+        if (url.endsWith('/')) {
+            return url.slice(0, url.length - 1)
+        }
+        return url
+    }
+
+    // caching images from WMS and WMTS after they are loaded from the network
```
schtibe, line 86: Just a thought: would it make sense to make the background layers have more priority in caching somehow?

packages/mapviewer/src/utils/offline/OfflinePrompt.vue
```diff
@@ -0,0 +1,137 @@
...
+                    registerPeriodicSync(serviceWorkerUrl, registration)
+                }
+            })
+        }
+    },
+})
+
+const title = computed(() => {
+    if (offlineReady.value) {
+        return 'App ready to work offline'
```
ismailsunni: no translation?

packages/mapviewer/src/utils/offline/OfflinePrompt.vue
```diff
@@ -0,0 +1,137 @@
...
+    }
+    if (needRefresh.value) {
+        return 'New content available, click on reload button to update.'
```
ismailsunni: no translation?

ismailsunni:
@pakb  to test, we set it to flight mode then refresh the web page right? I couldn't see any difference (or any pop up) while testing it. Did I miss something?

packages/mapviewer/src/utils/offline/OfflinePrompt.vue
pakb: should be clarified

packages/mapviewer/src/service-workers.ts
pakb: I don't know if I want to hard-code any layer ID there, and doing that "dynamically" depending on the layers config looks quite complicated. Maybe in a second step

packages/mapviewer/src/utils/offline/OfflineReadinessStatus.vue
```diff
@@ -0,0 +1,167 @@
...
+                />
+                <FontAwesomeIcon
+                    icon="slash"
+                    class="tw:text-red-600"
+                    transform="up-1"
+                />
+            </FontAwesomeLayers>
+            <span
+                v-if="withText"
+                class="fw-bold mx-1"
```
schtibe: You could maybe also use [gap](https://tailwindcss.com/docs/gap) in the enclosing flexbox

schtibe (approved):
Nice, just one small comment! :+1:

PR #1361, opened 2025-06-05
PB-1754: add missing prewiew:test target in main package json

pakb:
and fix filtering, it should be placed before the command (if placed after, it is passed to the command  as argument, instead of being interpreted)


package.json
```diff
@@ -2,30 +2,31 @@
...
     "private": true,
     "scripts": {
-        "build-all": "pnpm run -r build",
-        "build-libs": "pnpm run -r build --filter=!web-mapviewer",
-        "build:dev": "pnpm -r run build:dev",
-        "build:dev:watch": "pnpm -r --parallel --if-present run build:dev:watch --filter=!web-mapviewer",
-        "build:int": "pnpm -r run build:int",
-        "build:prod": "pnpm -r run build:prod",
-        "dev": "pnpm run -r --filter=web-mapviewer dev ",
+        "build-all": "pnpm run --recursive build",
```
schtibe: _(I've been having this thought for quite some time now)_ I think we could change this back to just `build`. Then we would follow more a standard way again.

I think I did that because I ended up with a sort of infinite recursion, probably before I fully understood this stuff, but I don't see now why this should still happen.

sommerfe (approved):

package.json
pakb: Works well too, and sticks to standard target name. I'll change it to simply `build`, and adapt the CI if needed

schtibe: Yes I think it's in the buildspec, too.

pakb: https://github.com/geoadmin/infra-terraform-bgdi/pull/1016

PR #1362, opened 2025-06-05
PB-1411: fix printbox after searchbar clear

sommerfe:


packages/mapviewer/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
```diff
@@ -1,14 +1,21 @@
...
 import { useStore } from 'vuex'
 
 const dispatcher = { dispatcher: 'print-area-renderer.composable' }
-
+function createSourceForProjection() {
+    return new VectorSource({
+        useSpatialIndex: false,
+        wrapX: true,
+    })
+}
```
pakb: I don't think you are creating this specifically for a projection.
And it looks like you only use it once, so it can simply be declared inline with the `new VectorLayer` below, IMHO

packages/mapviewer/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
```diff
@@ -1,14 +1,21 @@
...
 export default function usePrintAreaRenderer(map) {
     const store = useStore()
-
+    let overlayAnkerLayer = null
```
pakb: Denglish word here 😉
We could rename that to a simple `printPreviewLayer`

pakb (approved):

PR #1363, opened 2025-06-10
PB 1759: ensure COGS comparison work on startup

ltkum:
- Issue: in some cases, the rendering events were not linked correctly to the layer on startup.
- Fix: we add a small check to use the best event between 'precompose' and 'rendercomplete' to attach the rendering events to the layer.
- Extra: we updated the comment to make it clearer why this is here

### DIFF
--- a/packages/mapviewer/src/modules/map/components/CompareSlider.vue
+++ b/packages/mapviewer/src/modules/map/components/CompareSlider.vue
@@ -36,20 +36,22 @@ watch(
         if (oldLayerOnTop) {
             unRegisterRenderingEvents(oldLayerOnTop.id, oldLayerOnTop.uuid)
         }
-
         if (getLayerFromMapById(newLayerOnTop.id, newLayerOnTop.uuid)) {
             registerRenderingEvents(newLayerOnTop.id, newLayerOnTop.uuid)
             olMap.render()
         } else {
-            // The layer config is always modified before the map, which means the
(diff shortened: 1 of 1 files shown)


packages/mapviewer/src/modules/map/components/CompareSlider.vue
```diff
@@ -36,20 +36,22 @@ watch(
...
-            // to mitigate the issue, we ask the map to register the events once after the first render,
-            // after all layers have been loaded, and re-render it.
-            // We should arrive here ONLY if we arrive on startup on a map with an active compare slider.
-            olMap?.once('rendercomplete', () => {
+            // To mitigate the issue, we ask the map add the rendering functions
+            // either when the 'precompose' event is fired (for WebGL layers) or
+            // a 'rendercomplete' event is fired (for canvas layers).
+            // This ensure we link the rendering function to the layer as soon
+            // as possible.git
+            olMap?.once(shouldUseWebGlContext.value ? 'precompose' : 'rendercomplete', () => {
```
pakb, line 54: I'd maybe leave both event.
Are you 100% sure `precompose` is only used in the WebGL context?

ltkum: I'm 100% sure (at least according to the documentation.)

pakb (approved):

PR #1364, opened 2025-06-16
PB-1681: fix vue warnings

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1681-vue-warnings/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
+++ b/packages/mapviewer/src/modules/map/components/openlayers/utils/usePrintAreaRenderer.composable.js
@@ -1,6 +1,8 @@
 import log from '@geoadmin/log'
 import * as olHas from 'ol/has'
+import VectorLayer from 'ol/layer/Vector'
 import { getRenderPixel } from 'ol/render'
+import VectorSource from 'ol/source/Vector'
 import { computed, watch } from 'vue'
 import { useStore } from 'vuex'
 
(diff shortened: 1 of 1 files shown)


packages/mapviewer/src/modules/drawing/components/DrawingMeasureInteraction.vue
```diff
@@ -25,5 +25,7 @@ defineExpose({
 </script>
 
 <template>
-    <slot />
+    <div>
+        <slot />
+    </div>
```
pakb: Why would adding an empty `<div>` fix an error?

sommerfe: The reason for this error is that when adding properties to the DrawingMeasureInteraction component, vue doesn't know to which root html element these properties should be added, this can happen when there are two root elements. Therefore when you only have one root element, in this case the div, in the component vue knows now where to add the properties. See https://mokkapps.de/vue-tips/use-fallthrough-attributes here they propose to use v-bind, but when i tried to use it it did not bind the properties correctly so i added a root element instead. I should probably add a comment to explain this in the code

sommerfe: i added a comment

pakb: I think this is not the proper fix for this warning. The issue was that we are giving `startingFeature` prop to all interaction component, even to those that do not declare it.

I've pushed a commit that should fix this

PR #1365, opened 2025-06-16
PB-1699: White listing internal services used as an external provider

ltkum:
Issue: When importing layers from our own services, a disclaimer was visible, saying the sources were provided by a third party.

Fix: We "white list" all sources coming from a "*.map.geo.admin.ch" domain, removing the disclaimer.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1699-whitelist-swisstopo-external-layers/index.html)

### DIFF
--- a/packages/mapviewer/src/config/layers.config.ts
+++ b/packages/mapviewer/src/config/layers.config.ts
@@ -1,6 +1,0 @@
-/**
- * List of whitelisted URLS regexes for external layers. External layers which baseUrl
- * match one of these regexes won't show a disclaimer
- */
-
-export const EXTERNAL_PROVIDER_WHITELISTED_URL_REGEXES : Array<RegExp> = [/^(.+\.)?geo\.admin\.ch.*$/g]
--- a/packages/mapviewer/src/config/layers.config.ts
+++ b/packages/mapviewer/src/config/layers.config.ts
@@ -1,0 +1,6 @@
+/**
+ * List of whitelisted URLS regexes for external layers. External layers which baseUrl
+ * match one of these regexes won't show a disclaimer
+ */
+
+export const EXTERNAL_PROVIDER_WHITELISTED_URL_REGEXES : Array<RegExp> = [/^(.+\.)?geo\.admin\.ch.*$/g]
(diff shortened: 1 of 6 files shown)


packages/mapviewer/src/config/layers.config.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * List of whitelisted URLS regexes for external layers. External layers which baseUrl
+ * match one of these regexes won't show a disclaimer
+ */
+
+export const EXTERNAL_PROVIDER_WHITELISTED_URL_REGEXES : Array<RegExp> = [/^(.+\.)?geo\.admin\.ch.*$/g]
```
pakb: There's already a similar RegExp [here](https://github.com/geoadmin/web-mapviewer/blob/b776c8d79bb4064348df026db7280d3c64535e5c/packages/mapviewer/src/utils/kmlUtils.js#L535).

Could you move it here, and use it in both spots?

packages/mapviewer/tests/cypress/fixtures/import-tool/wms-non-geo-admin-get-capabilities.xml
pakb, line 1: There's already a couple external WMS fixtures, such as `external-wms-getcap-1.fixture.xml`
Could you switch to using this one instead? The less fixtures we have, the better IMHO

ltkum: I'll see what I can do to have as little fixtures as possible :)

packages/mapviewer/tests/cypress/tests-e2e/importToolMaps.cy.js
```diff
@@ -554,6 +532,59 @@ describe('The Import Maps Tool', () => {
         cy.get('[data-cy="active-layer-name-layer4-2"]').should('be.visible')
         cy.get('[data-cy="time-selector-layer4-2"]').should('not.exist')
     })
+
+    it('Import an external WMS when it is in the URL and shows its third party disclaimer correctly', () => {
+        // here : intercept wms.external.admin.ch to give what we want
```
pakb: I don't think you mention the correct URL here, is it really `wms.external.admin.ch`?

ltkum: ha no, the comment is wrong, it's from the time when we had another file. Good catch

pakb (approved):

PR #1366, opened 2025-06-18
PB-120: Fix Selected Feature is not Highlighted in 3D map

ismailsunni:


packages/mapviewer/src/modules/map/components/cesium/utils/highlightUtils.js
```diff
@@ -40,34 +40,42 @@ export function highlightGroup(viewer, geometries) {
...
+// Highlight all polygons represented by the coordinates by adding each as a separate entity to the viewer
+export function highlightPolygon(viewer, coordinates) {
+    if (!coordinates.length) {
+        return
+    }
+    // Add each polygon entity separately to the viewer
+    const polygonEntities = getAllPolygonEntities(coordinates)
+    polygonEntities.forEach((polyEntity) => {
+        viewer.entities.add(polyEntity)
+        highlightedEntities.push(polyEntity)
```
pakb, line 77: this list `highlightedEntities` is never used, isn't it?

ismailsunni: I think it's used to track the highlighted entities, for example to remove the highlight. See here: https://github.com/geoadmin/web-mapviewer/blob/9fdd4c283766fe657e11083fab5e871f7fa78528/packages/mapviewer/src/modules/map/components/cesium/utils/highlightUtils.js#L118-L124

PR #1371, opened 2025-07-03
PB-1814: Fix download print issue since Service-Worker introduction

pakb:
Removes the `file-saver` dependency and implements a custom `downloadFile` function to handle file downloads.

This change ensures compatibility and addresses potential issues related to browser-specific behavior while Service-Worker is caching our app. Prior to this change, opening the download URL the way it was was leading to a white page where it looked like Service-Worker was trying to inject the app code there (and it failed), no PDF was downloaded

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1814-print-download/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/menu/components/print/MenuPrintSection.vue
+++ b/packages/mapviewer/src/modules/menu/components/print/MenuPrintSection.vue
@@ -14,7 +14,7 @@ import MenuSection from '@/modules/menu/components/menu/MenuSection.vue'
 import DropdownButton from '@/utils/components/DropdownButton.vue'
 import DropdownButtonItem from '@/utils/components/DropdownButtonItem.vue'
 import ProgressBar from '@/utils/components/ProgressBar.vue'
-import { downloadFile } from '@/utils/utils'
+import { downloadFile, generateFilename } from '@/utils/utils'
 
 const dispatcher = { dispatcher: 'MapPrintSection.vue' }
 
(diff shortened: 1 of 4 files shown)


packages/mapviewer/src/modules/menu/components/print/MenuPrintSection.vue
```diff
@@ -119,13 +121,10 @@ function close() {
...
 async function printMap() {
     try {
-        const documentUrl = await print(printGrid.value, printLegend.value)
-        if (documentUrl) {
-            if (window.navigator.userAgent.indexOf('MSIE ') > -1) {
-                window.open(documentUrl)
-            } else {
-                window.location = documentUrl
-            }
+        const fileName = `${hostname.value}_${new Date().toISOString()}.pdf`.replaceAll(':', '_')
```
sommerfe: this date string is in UTC are we sure we do not want a local timestamp?

pakb: AFAIK the usePrint composable is also using a Z in its format, so it was already UTC before.

Might be better to use the current time on the user's machine

pakb: I've changed the `generateFilename` function to use local time, and have used it wherever possible, should be local time in file names now

ibauersachs:
@pakb Printing of maps is broken with v1.57.1 on Firefox: the downloaded PDF contains some HTML (without body). Chrome works.

PR #1374, opened 2025-07-07
#patch

pakb:


packages/mapviewer/src/service-workers.ts
```diff
@@ -50,7 +50,12 @@ if (!IS_TESTING_WITH_CYPRESS) {
 
     // setting up a cache instance for offline app assets (HTML/JS/CSS)
     registerRoute(
-        new NavigationRoute(createHandlerBoundToURL('index.html'), { allowlist }),
+        new NavigationRoute(createHandlerBoundToURL('index.html'), {
+            allowlist,
+            // exclude print explicitly as SW is sometimes messing with the download URL on Firefox
```
ismailsunni, line 50: Not so obvious at my first read, here, SW means service workers, right?

pakb: yes 😅

PR #1381, opened 2025-07-11
PB-1750: fix print scales going out of screen (overflow)

pakb:
refactoring the DropdownButton component to use FloatingUI, and ensuring the size of the dropdown "menu" (aka the floating thing) can't grow outside the document's height, and switch to a scroll/overflow if needed


packages/mapviewer/src/utils/components/DropdownButton.vue
```diff
@@ -1,88 +1,134 @@
-<script setup>
+<script lang="ts" setup>
 /**
  * Manages a Bootstrap dropdown.
```
schtibe: Is it still a bootstrap dropdown?

packages/mapviewer/tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -1204,6 +1206,74 @@ describe('Drawing module tests', () => {
...
+            }
+            cy.intercept('GET', `**/api/kml/files/**`, (req) => {
+                const headers = { 'Cache-Control': 'no-cache' }
+                req.reply(EMPTY_KML_DATA, headers)
+            }).as('get-empty-kml')
+
+            cy.intercept('POST', `**/api/kml/admin**`).as('post-new-kml')
+
+            const fileId = 'zBnMZymwTLSNg__5f8yv6g'
+            // load map with an injected kml layer containing an empty KML
```
ltkum: You changed all comments to logs, and I found that nice. You should continue the trend in your new tests.

packages/mapviewer/tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -1204,6 +1206,74 @@ describe('Drawing module tests', () => {
...
+            cy.goToMapView({
+                layers: [`KML|https://sys-public.dev.bgdi.ch/api/kml/files/${fileId}`].join(';'),
+            })
+
+            cy.wait('@get-empty-kml')
+            // there should be only one KML layer left in the layers, and it's the one just saved
```
ltkum: same as above

packages/mapviewer/tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -1204,6 +1206,74 @@ describe('Drawing module tests', () => {
...
+            cy.get('[data-cy^="button-has-error"]').should('be.visible')
+
+            cy.openDrawingMode()
+
+            cy.get('[data-cy="drawing-toolbox-file-name-input"]', { timeout: 15000 }).should(
+                'be.visible'
+            )
+            cy.closeDrawingMode()
+
+            // saving the drawing without drawing anything should not change the empty KML layer
```
ltkum: another log here

packages/mapviewer/tests/cypress/tests-e2e/drawing.cy.js
```diff
@@ -1204,6 +1206,74 @@ describe('Drawing module tests', () => {
...
+            cy.openDrawingMode()
+
+            cy.get('[data-cy="drawing-toolbox-file-name-input"]', { timeout: 15000 }).should(
+                'be.visible'
+            )
+            cy.clickDrawingTool(EditableFeatureTypes.MARKER)
+            cy.get('[data-cy="ol-map"]').click(120, 240)
+            cy.closeDrawingMode()
+            cy.wait('@post-new-kml')
+            // drawing a marker should create a new KML layer and overwrite the empty one
```
ltkum: I think it's the last one here )

schtibe (commented):
There's an issue with the dropdown: when you change the style of a marker, the popup closes when selecting a size or a symbol library

packages/mapviewer/src/utils/components/DropdownButton.vue
pakb: nope, it's not anymore 😅

pakb:
Native dropdown are a good trade-off solution, but can't be styled
And I think the idea was to have the "bootstrap" styled dropdown made in Vue with the first iteration of this component.

Don't know if we should simply let browser do the rendering or if that would bring other issues...

packages/mapviewer/src/utils/components/DropdownButton.vue
```diff
@@ -1,88 +1,134 @@
...
-        dropdown = new Dropdown(dropdownMainButton.value)
+        return dropdownToggleButtonRef.value
     }
+    return dropdownMainButtonRef.value
 })
-onBeforeUnmount(() => {
-    dropdown?.dispose()
+
+// generating a unique HTML ID for this dropdown
+const uniqueHtmlId = ref<string>(`dropdown-${randomIntBetween(0, 10000)}`)
```
ltkum: There is a very small chance that this can have collisions, but I think the interval is big enough so that it's not a common occurrence :)

ltkum (commented):
It seems mostly okay to me :)

packages/mapviewer/tests/cypress/tests-e2e/drawing.cy.js
schtibe: :heavy_plus_sign: :one:

packages/mapviewer/src/utils/components/DropdownButton.vue
schtibe: akshually ( :wink: ) I think we have the uuid package installed in mapviewer?

pakb: you're right, that could be a good solution to simply use UUID... let's do that (I raised the "seed") to 100'000 but UUID will still be better as it will be based on the timestamp anyway...)

PR #1385, opened 2025-07-31
PB-1875: links with pre-selected features do not center correctly

ltkum:
- Issue: In some occasions, opening a link containing both a center and pre-selected features would prioritize centering on the features rather than centering on the pre-established center. This would only occur at startup. Once the application is running, changing the URL to the one we want would set all parameters correctly

- Cause: On startup, we go through the store sync loop multiple times, and at one point, we set the features after the center, and thus override the center parameter

- Fix: If there is a `center` parameter in the query, we don't set the center on the features extent.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1875-wrong-center-on-startup-with-selected-layers/index.html)

### DIFF
--- a/packages/mapviewer/tests/cypress/support/intercepts.js
+++ b/packages/mapviewer/tests/cypress/support/intercepts.js
@@ -250,8 +250,11 @@ let lastIdentifiedFeatures = []
  *
  * Features IDs will start from 1 + offset (if an offset is given) and coordinates will be randomly
  * selected within the LV95 extent (or within the selection box, if one is given).
+ *
+ * @param coordinates {[Number, Number] | undefined} : if this is set, the coordinates of each
+ *   feature generated is set to the given coordinates rather than random coordinates.
  */
-const addFeatureIdentificationIntercepts = () => {
+export function addFeatureIdentificationIntercepts(coordinates) {
     let featureTemplate = {}
     let featureDetailTemplate = {}
     cy.fixture('features/features.fixture').then((featuresFixture) => {
@@ -356,7 +359,7 @@ const addFeatureIdentificationIntercepts = () => {
                     featureDetail.bbox = [...coordinate, ...coordinate]
                     featureDetail.geometry.coordinates = [coordinate]
                 } else {
-                    const randomCoordinate = [
+                    const randomCoordinate = coordinates ?? [
                         Cypress._.random(LV95.bounds.lowerX, LV95.bounds.upperX),
                         Cypress._.random(LV95.bounds.lowerY, LV95.bounds.upperY),
                     ]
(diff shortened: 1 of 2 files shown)


pakb (commented):
A little test to cover this new addition?

pakb (commented):
AFAIK you aren't really testing what you changed in the code here, please change the test so that you test that:

- The map centers on a (single) feature when it is loaded with a pre-selected layer feature, and without defining a center param
- The map loads the view on the center param, even if the feature is also declared alongside it

packages/mapviewer/tests/cypress/support/intercepts.js
```diff
@@ -251,7 +251,7 @@ let lastIdentifiedFeatures = []
  * Features IDs will start from 1 + offset (if an offset is given) and coordinates will be randomly
  * selected within the LV95 extent (or within the selection box, if one is given).
  */
-const addFeatureIdentificationIntercepts = () => {
+export function addFeatureIdentificationIntercepts(coordinates) {
```
pakb, line 257: little addition to the JSDoc to explain what that is please 😉

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -69,21 +71,37 @@ describe('Testing the feature selection', () => {
...
+        function goToMapViewWithFeatureSelection(
+            featureInfoPosition = null,
+            featureCoordinates = null
+        ) {
             const params = {
                 layers: `${standardLayer}@features=1:2:3:4:5:6:7:8:9:10`,
             }
+            if (featureCoordinates) {
+                addFeatureIdentificationIntercepts(featureCoordinates)
+            }
```
pakb: I don't think this is necessary, you are already overwriting the default fixture when calling `goToMapView`

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
```diff
@@ -97,6 +115,31 @@ describe('Testing the feature selection', () => {
             checkFeatures()
             checkFeatureInfoPosition(FeatureInfoPositions.BOTTOMPANEL)
         })
+        it('Centers correctly the map when pre-selected features are present', () => {
+            cy.log('We ensure that when no center is defined, we are on the center of the extent')
+            const preDefinedCenter = [2671500, 1190000]
+
+            goToMapViewWithFeatureSelection(FeatureInfoPositions.NONE, preDefinedCenter)
```
pakb: As you are only using this function here (and only once), it might make it easier to read if you just did everything the function is doing right here, minus all the if...else logic because you only have one use case

packages/mapviewer/tests/cypress/tests-e2e/featureSelection.cy.js
ltkum: it is indeed not necessary. this was a remnant of code.

PR #1393, opened 2025-08-19
PB-1383: Enhances TS type generation and ESLint rules

pakb:
each module now only exports one `*.d.ts` file at the root of its dist folder (thanks to @microsoft/api-extractor package, used by unplugin-dts)

Use lib '@tmcw/togeojson' instead of mapbox equivalent as it comes with type definition for TS

disabling TS option composite for mapviewer, as it breaks Vite alias import (`@/...`) when the TS compiler goes through a JS file. Keeping composite enabled on packages that are already 100% TS, as it works fine in this case

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1383-better-ts-type-export/index.html)


schtibe (approved):
LGTM :fire:

PR #1395, opened 2025-08-19
PB-1383: layers module

pakb:
containing all layer types, some utils to manipulate/get more info out of them and also utils to parse WMTS and WMS getCapabilities endpoints


packages/layers/src/types/layers.ts
```diff
@@ -0,0 +1,454 @@
...
+    current?: boolean
+}
+
+export interface TileMatrixSet {
+    /* Identifier of the tile matrix set (see WMTS OGC spec) */
+    readonly id: string
+    /* Coordinate system supported by the Tile Matrix Set */
+    projection: CoordinateSystem
+    /* TileMatrix from GetCapabilities (see WMTS OGC spec) */
+    tileMatrix: unknown // TODO type this properly
```
schtibe, line 362: Intentionally not done this yet?

packages/layers/src/parsers/ExternalWMSCapabilitiesParser.ts
```diff
@@ -0,0 +1,670 @@
...
+    // if we do not have access to the legend in pure WMS fashion, we check if this
+    // WMS follows the SLD specification, and if we can get it from there.
+    if (
+        styles.length === 0 &&
+        layer.queryable &&
+        !!capabilities.Capability?.UserDefinedSymbolization?.SupportSLD
+    ) {
+        const getLegendGraphicBaseUrl =
+            capabilities.Capability.Request.GetLegendGraphic?.DCPType[0]?.HTTP?.Get?.OnlineResource
+        const getLegendGraphicFormat = capabilities.Capability.Request.GetLegendGraphic?.Format[0]
```
schtibe: Maybe you could shorten these two lines with an intermediate var

packages/layers/src/parsers/ExternalWMSCapabilitiesParser.ts
```diff
@@ -0,0 +1,670 @@
+import type { FlatExtent } from '@swissgeo/coordinates'
```
schtibe: Does it make sense to keep this being *external*CapabilitiesParser or are we at a point where this could be "just" a `WMSCapabilitiesParser`? Same with WMTS

packages/layers/src/api/internal.ts
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// API file that covers the backend endpoint http://api3.geo.admin.ch/rest/services/all/MapServer/layersConfig
+
+/**
+ * Transform the backend metadata JSON object into instances of {@link GeoAdminLayer}, instantiating
+ * the correct type of layer for each entry ({@link GeoAdminAggregateLayer},
+ * {@link GeoAdminWMTSLayer}, {@link GeoAdminWMSLayer} or {@link GeoAdminGeoJsonLayer})
+ */
+export function generateClassForLayerConfig(
```
schtibe: I this still `class` or might renaming make sense 🤔 Same with the documentation above, we don't instantiate stuff anymore

packages/layers/src/api/external.ts
```diff
@@ -0,0 +1,195 @@
...
+    type WMTSCapabilitiesResponse,
+} from '@/parsers/ExternalWMTSCapabilitiesParser'
+import { CapabilitiesError } from '@/validation'
+
+/** Timeout for accessing external server in [ms] */
+export const EXTERNAL_SERVER_TIMEOUT = 30000
+
+/** Sets the WMS GetCapabilities url parameters */
+export function setWmsGetCapabilitiesParams(url: URL, language?: string): URL {
+    // Manda: URLtory params
```
schtibe: Typo here

packages/layers/src/types/layers.ts
pakb: I might not have gotten to the point in the code where I needed it, so I forgot to do it... Or maybe we won't use it? I wonder...

packages/layers/src/parsers/ExternalWMSCapabilitiesParser.ts
pakb: I think we could safely rename them without the "external" part. They will still output "ExternalLayer" in the sense of what is described by `types/layer.ts`, but that's fine by me

PR #1396, opened 2025-08-26
Fix pb 1895 stop editing nonexistent kmls

ltkum:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1895-stop-editing-nonexistent-kmls/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/packages/mapviewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -258,8 +258,8 @@ function mediaTypes() {
                         {{ t('modify_description') }}
                     </label>
                     <GeoadminTooltip
-                        :tooltip-content="t('display_on_map')"
                         v-if="isFeatureMarker || isFeatureText"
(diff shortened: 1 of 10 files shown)


pakb:
I don't know, it was way too much work for me to rebase it myself, not being the author. So I kept it as it is last time I went through all the open PRs.
@ltkum would it make sense to rebase this?

ltkum:
> I don't know, it was way too much work for me to rebase it myself, not being the author. So I kept it as it is last time I went through all the open PRs. @ltkum would it make sense to rebase this?

I think the short answer is yes


- it stops an issue where you would share an edit link with someone, and the contained KML had no features, that would stop any new drawings to be made (as there would be a conflict between the adminId of a nonexistent KML and the new KML)
- We also modified the errors and warning messages functionality to handle acknowledgement and sources, so that we do not send multiple times an error / warning that was already seen and closed


but it was 6 months ago, and maybe this has already been solved during the migration.

- My suggestion: unless I am mistaken, the drawing module is currently being worked on. I would wait for it to be completed, and then try to reproduce the bug. If this is still reproducible, we merge / adapt this solution. Otherwise, we close this.

PR #1399, opened 2025-09-08
Fix cypress tests | 3d:

ltkum:
- Issue 1: in the commands.ts file, we handled the booleans wrongly trying to add them to the viewer only if they were false instead of only when they are true, which is how it should be.

- Issue 2: We changed the signature of the "goToMapView", and the test methods were using the old syntax

Fix: We handle booleans correctly and changed the signatures where needed.

### DIFF
--- a/packages/viewer/tests/cypress/tests-e2e/3d/navigation.cy.js
+++ b/packages/viewer/tests/cypress/tests-e2e/3d/navigation.cy.js
@@ -8,11 +8,11 @@ import { calculateResolution } from '@/modules/map/components/cesium/utils/camer
 
 registerProj4(proj4)
 
-describe('Testing 3D navigation', () => {
+describe.skip('Testing 3D navigation', () => {
     context('camera limits', () => {
         beforeEach(() => {
             cy.goToMapView({
-                '3d': true,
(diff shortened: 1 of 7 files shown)


packages/viewer/tests/cypress/tests-e2e/3d/navigation.cy.js
```diff
@@ -8,11 +8,11 @@ import { calculateResolution } from '@/modules/map/components/cesium/utils/camer
 
 registerProj4(proj4)
 
-describe('Testing 3D navigation', () => {
+describe.skip('Testing 3D navigation', () => {
```
pakb, line 11: not sure you wanted to keep this `.skip` here

pakb:
you can also use the ticket https://swissgeoplatform.atlassian.net/browse/PB-1383 to set the branch name and commit messages

packages/viewer/tests/cypress/tests-e2e/3d/navigation.cy.js
ltkum: ha yeah, I didn't see that. I think I clicked on the "set skip" by mistake ^^.

PR #1404, opened 2025-09-11
DRAFT: Add postcss-html and fix testSearchMarker test.

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/develop-typescript-ismail/index.html)

### DIFF
--- a/packages/viewer/tests/cypress/tests-component/TextSearchMarker.cy.ts
+++ b/packages/viewer/tests/cypress/tests-component/TextSearchMarker.cy.ts
@@ -40,13 +40,13 @@ describe('<TextSearchMarker />', () => {
         })
         cy.get('[data-cy="segment"]')
             .should('be.visible')
-            .should('have.class', 'fw-bold')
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
         "pnpm": ">=10.15"
     },
     "devDependencies": {
-        "yaml": "catalog:"
+        "yaml": "catalog:",
+        "postcss-html": "catalog:"
```
pakb, line 37: I think it would be better to add it in each package.json file that is also importing `stylelint` instead of at the top-level file.

Here we only import stuff that are required for the script `scripts/update-pnpm-workspace.js`

pnpm-workspace.yaml
```diff
@@ -92,7 +92,7 @@ catalog:
   pako: ^2.1.0
   pinia: ^3.0.3
   pinia-logger: ^1.3.14
-  postcss-html: ^1.8.0
```
pakb, line 95: keep the `^` so that we update to minor version without having to change this

PR #1406, opened 2025-09-12
PB-1383: Fix missing postcss-html.

ismailsunni:
Here is what I understood,

When we ran the linter, we used `stylelint`. And `stylelint` needs `postcss-html` (imported dynamically). The isolation in `pnpm` makes the `stylelint` access the `postcss-html`.

From what I read, it's better to put the `postcss-html` in each package's package.json that uses it to follow the "Declare what you use, use what you declare".

So, we use this workaround using [packageExtensions](https://pnpm.io/9.x/package_json#pnpmpackageextensions) that tells pnpm to treat postcss-html as the direct dependency of stylelint.

Honestly, it's my first time finding this kind of issue, so perhaps I am wrong.

[Test link](https://sys-map.dev.bgdi.ch/preview/pb-1383-fix-missing-postcss-html/index.html)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,13 +53,6 @@
         ],
         "patchedDependencies": {
             "vite": "patches/vite.patch"
-        },
-        "packageExtensions": {
-            "stylelint@*": {
-                "dependencies": {
-                    "postcss-html": "*"
-                }
-            }
         }
     }
 }
--- a/packages/stylelint-config/package.json
+++ b/packages/stylelint-config/package.json
@@ -8,7 +8,6 @@
         "lint:no-fix": "eslint"
     },
     "dependencies": {
-        "postcss-html": "catalog:",
         "stylelint-config-recommended-scss": "catalog:",
         "stylelint-config-recommended-vue": "catalog:",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -53,6 +53,13 @@
         ],
         "patchedDependencies": {
             "vite": "patches/vite.patch"
+        },
+        "packageExtensions": {
+            "stylelint@16.23.1": {
```
pakb: can it work without specifying the version here?
might behave better when we update stylelint version later

packages/stylelint-config/package.json
```diff
@@ -7,16 +7,18 @@
         "lint": "eslint --fix",
         "lint:no-fix": "eslint"
     },
+    "dependencies": {
+        "postcss-html": "catalog:",
```
pakb: maybe we could also move that to a `peerDependencies` block?

ismailsunni: Actually, this gives me an idea that it seems to be working. When I set it as a peerDep, it can find the `postcss-html` that is available and not installed as isolated.

PR #1408, opened 2025-09-15
PB-1383: add tests 1

sommerfe:


packages/elevation-profile/src/GeoadminElevationProfilePlot.vue
pakb, line 1: it might make sense to create a new type for that, so that we only declare this union once (and makes it easier to change afterward)

packages/viewer/tests/cypress/support/commands.ts
```diff
@@ -123,8 +123,7 @@ function goToView(
         withHash = true,
         geolocationMockupOptions = { latitude: 47, longitude: 7 },
         fixturesAndIntercepts = {},
-        legacy = !withHash,
-        firstLoad = true,
```
pakb, line 127: this param might need to stay (maybe Ismail can better judge that), but AFAIK he introduced it one or two PR before to fix some tests

packages/viewer/tests/cypress/tests-e2e/changeLanguage.cy.ts
```diff
@@ -1,16 +1,17 @@
+// Transforming the file to TypeScript
+
```
pakb: clean up?

packages/viewer/tests/cypress/tests-e2e/footer.cy.ts
```diff
@@ -4,13 +4,63 @@ import { WEBMERCATOR } from '@swissgeo/coordinates'
 
 import { UIModes } from '@/store/modules/ui.store'
 
+interface LayerConfig {
```
pakb: might make more sense to export the type that is defined in `packages/layers/src/api/internal.ts` (and change its name to `RawLayer` or `LayerConfigBackendResponse`, something in this vein) than to copy/paste it here

packages/viewer/tests/cypress/tests-e2e/importToolFile.cy.ts
```diff
@@ -5,42 +5,56 @@ import proj4 from 'proj4'
...
+        const geomHighlightFeature = vectorLayers.find((layer: BaseLayer) => {
             return layer
                 .getSource()
                 .getFeatures()
-                .find((feature) => feature.get('id').startsWith('geom-segment-'))
+                .find((feature: BaseLayer) => feature.get('id').startsWith('geom-segment-'))
         })
-        expect(geomHighlightFeature).to.not.be.undefined
+        // expect(geomHighlightFeature).to.not.be.null
+        // expect(geomHighlightFeature).to.not.be.undefined
```
pakb: clean up?

pakb (commented):
many little things you've fixed are linter related, as if the linter isn't run automatically anymore in our local IDEs...
That's worrisome, we should investigate.

packages/viewer/tests/cypress/support/commands.ts
sommerfe: yes i saw what he did and found a different solution in my opinion at least, but @ismailsunni maybe you should have a look if i broke something

pakb (approved):
I'd let Ismail tell us if that's fine by him, regarding this `firstVisit` param, before merging it

packages/viewer/tests/cypress/support/commands.ts
ismailsunni: What is your other solution, @sommerfe?
I saw you remove the `firstLoad` here, but not in the test (so. the `firstLoad` parameter is not used). Based on what I know, the `gaMapReady` is only sent once after the app is ready.

Strangely, the test is passing in the CI, but in my local there is a failed test
image

ismailsunni: From what I discuss with @sommerfe 

In the JavaScript version, `mapReadyEvent` is assigned within the `onBeforeLoad` function: [JS Version](https://github.com/geoadmin/web-mapviewer/blob/develop/packages/mapviewer/tests/cypress/support/commands.js#L145).

However, after migrating to TypeScript, the assignment was moved outside of the `onBeforeLoad` function: [TS Version](https://github.com/geoadmin/web-mapviewer/blob/40cd908448ab28cb2ca7ef244d2f069e79a106dd/packages/viewer/tests/cypress/support/commands.ts#L205).

So, @sommerfe's approach is putting it back like before the TS migration.

ismailsunni: CMIIW, it seems if we put it inside the `onBeforeLoad` and on the 2nd call, we never reload the window, we still refer to the previous `@mapReadyEvent`, that's the reason why it passes the `calledOnce`.

If we put it outside the `onBeforeLoad`, we set the `@mapReadyEvent` to a new object, but we never reload the window (on the second call). So, it failed on the `calledOnce` test (never called because it never reloads).

I think @sommerfe's solution is the better approach. Although for me, it's not so clear in the begining. But it does the same thing.

PR #1409, opened 2025-09-15
PB-1383: refactor 3d e2e tests to typescript

ltkum:
Issue: We want to change the whole code base to typescript, tests included

Fix: We change the tests to typescripts. We remove some variables that were using non-existent attributes, and we states that we're expecting errors from files that are not yet set in typescript. Those "expect errors" lint exception will yell at us when the typescript conversion has been correctly made :)

we also update the tsconfig to explicitly accept some functionalities (like 'includes')


packages/viewer/tests/cypress/tests-e2e/3d/geolocation.cy.ts
```diff
@@ -7,6 +7,7 @@ import {
     getGeolocationButtonAndClickIt,
     testErrorMessage,
     checkStorePosition,
+    // @ts-expect-error utils import not a typescript file for now
```
pakb: this is already a TS file, so maybe you should fix the error that this raises

packages/viewer/tests/cypress/tests-e2e/3d/layers.cy.ts
```diff
@@ -1,17 +1,15 @@
 import { WEBMERCATOR, LV95 } from '@swissgeo/coordinates'
-
+// @ts-expect-error Editable Feature class not a typescript file for now
```
pakb: wouldn't it fix the error if you added `.js` at the end of the import?

or we could also add `allowJS: true` to our `tsconfig.json` used for Cypress

packages/viewer/tests/cypress/tests-e2e/3d/layers.cy.ts
```diff
@@ -143,6 +141,8 @@ describe('Test of layer handling in 3D', () => {
         const timeEnabledLayerId = 'test.timeenabled.wmts.layer'
         cy.fixture('layers.fixture.json').then((layersMetadata) => {
             const timedLayerMetadata = layersMetadata[timeEnabledLayerId]
+            // TODO: refactor cypress.d.ts so it gives correct types
+            // @ts-expect-error this is chainable, but cypress type definition says it's a string, so it's causing issues here.
```
pakb, line 153: please fix it while doing this PR, there's already the `cypress.d.ts` file that types any utils we add, you could add the definition here

packages/viewer/tests/cypress/tests-e2e/3d/layers.cy.ts
```diff
@@ -348,6 +348,7 @@ describe('Test of layer handling in 3D', () => {
         })
     })
     it('Verify a layer with EPSG:4326(WEBMERCATOR) bounding box in 2D and 3D', () => {
+        // @ts-expect-error same as above, we'll remove this when we have refactored cypress.d.ts
```
pakb: same

packages/viewer/tsconfig.json
pakb, line 1: what was missing without `ES2024`?

packages/viewer/tests/cypress/tests-e2e/3d/transitionTo3d.cy.ts
```diff
@@ -5,6 +5,8 @@ import { WGS84 } from '@swissgeo/coordinates'
 import { Math as CesiumMath } from 'cesium'
 import proj4 from 'proj4'
 
+
+// @ts-expect-error map config not a typescript file
```
pakb: should be quite easy to transform to TS though (no imports, only const declaration)

packages/viewer/tsconfig.json
ltkum: includes(2016), toSorted(2023) and entries (2017). I just took the latest year to be honest.

packages/viewer/tests/cypress/tests-e2e/3d/navigation.cy.ts
```diff
@@ -2,8 +2,9 @@ import { registerProj4, WGS84 } from '@swissgeo/coordinates'
 /// <reference types="cypress" />
 import { Cartesian3 } from 'cesium'
 import proj4 from 'proj4'
-
+//@ts-expect-error cesium config not a typescript file
```
pakb: Can you try migrating these two files to TS?
the config file should be quite easy to migrate, the other one might already have been migrated by the work done on the pinia PR (please double check the work that was done [in this PR](https://github.com/geoadmin/web-mapviewer/pull/1426))

ltkum: I'll try :)

PR #1411, opened 2025-09-15
PB-1383: adapt cypress command file to respect typescript obligations

ltkum:
- Issue: Some functionalities were not typescript compliant and the linter was yelling at me
- Fix: We add some types declarations, we add an import and shut down its error for now, and we try to be nice and avoid reading protected values.

side note: I added the lib changes to allow "includes" and "sorted" methods to another PR, so I'm not changing these files here :)

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1383-adapt-cypress-commands-to-typescript/index.html)


packages/viewer/tests/cypress/support/commands.ts
```diff
@@ -6,7 +6,8 @@ import { randomIntBetween } from '@swissgeo/numbers'
 import proj4 from 'proj4'
 
 import type { Layer as OLLayer } from 'ol/layer'
-
+// @ts-expect-error GeoAdminLayer is not yet a typescript file
+import GeoAdminLayer from '@/api/layers/GeoAdminLayer.class'
```
pakb:
```suggestion
import type { GeoAdminLayer } from '@swissgeo/layers'
```

packages/viewer/tests/cypress/support/commands.ts
```diff
@@ -282,10 +283,17 @@ Cypress.Commands.add('clickOnLanguage', (lang) => {
...
+     * Full explanation of the ts expect error here :
+     * AUTWindow's type includes the classic window type, as well as an `Application Window` type
+     * which is used to store custom arguments. The issue is that those arguments are not
+     * known in advance, and typescript doesn't like that.
+     */
     cy.waitUntil(
         () =>
             cy
                 .window({ log: false })
+                // @ts-expect-error Application window definition has no defined keys in its definition
```
pakb: there's already some type definition for that somewhere (I think it's in `src/env.d.ts`), you could add that type definition to our main `cypress.d.ts`

packages/viewer/tests/cypress/cypress.d.ts
```diff
@@ -148,5 +148,9 @@ declare global {
              */
             clickDrawingTool(tool: string, unselect?: boolean): void
         }
+        interface ApplicationWindow {
+            store: typeof Vuex.store
+        }
+
```
pakb: this should be covered by https://github.com/geoadmin/web-mapviewer/blob/d49c644d6722268c42931eeb88c87b5151ec3f6d/packages/viewer/types/window.d.ts
maybe it would be better to change the Cypress' `tsconfig.json` so that it knows about this `packages/viewer/types` folder (if it isn't automatically detected already)

packages/viewer/tests/cypress/support/commands.ts
```diff
@@ -151,7 +151,7 @@ function goToView(
         queryParams.sr = WEBMERCATOR.epsgNumber
     }
     let flattenedQueryParams: string = ''
-    Object.entries(queryParams).forEach(([key, value]) => {
+    Object.entries(queryParams).forEach(([key, value] : [string, string | number | boolean]) => {
```
pakb: can't this type be inferred from the type queryParams is described with?

packages/viewer/tests/cypress/support/commands.ts
```diff
@@ -272,11 +272,17 @@ Cypress.Commands.add('clickOnLanguage', (lang) => {
 })
 
 Cypress.Commands.add('waitUntilState', (predicate, options) => {
+    /**
+     * Full explanation of the ts expect error here :
+     * AUTWindow's type includes the classic window type, as well as an `Application Window` type
+     * which is used to store custom arguments. The issue is that those arguments are not
+     * known in advance, and typescript doesn't like that.
+     */
```
pakb: there is no `// @ts-expect-error` here, is that still relevant?

packages/viewer/tests/cypress/support/commands.ts
ltkum: It could not before, but now that the types have been fully added, it seems to be inferred.

pakb (approved):

PR #1413, opened 2025-09-16
PB-1383: transform all API files to TypeScript

pakb:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1383-api-files-typescript/index.html)

### DIFF
--- a/packages/viewer/src/api/__tests__/search.api.spec.ts
+++ b/packages/viewer/src/api/__tests__/search.api.spec.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { sanitizeTitle } from '../search.api'
+import { sanitizeTitle } from '@/api/search.api'
 
 describe('Builds object by extracting all relevant attributes from the backend', () => {
     describe('FeatureSearchResult.getSimpleTitle', () => {
--- a/packages/viewer/src/api/features/features.api.ts
+++ b/packages/viewer/src/api/features/features.api.ts
@@ -69,7 +69,7 @@ export class GetFeatureInfoError extends Error {
  * @param feature OpenLayers KML feature
  * @returns Coordinates of the feature, in WGS84
  */
-export function extractOlFeatureCoordinates(feature: Feature): SingleCoordinate[] {
+export function extractOlFeatureCoordinates(feature?: Feature): SingleCoordinate[] {
     if (!feature) {
         return []
     }
@@ -249,7 +249,7 @@ async function identifyOnExternalLayer(config: IdentifyConfig): Promise<LayerFea
     ) {
         // trying to find a candidate among the app supported projections
         requestProjection = allCoordinateSystems.find((candidate) =>
-            externalLayer.availableProjections?.some(
+            externalLayer.availableProjections!.some(
                 (availableProjection) => availableProjection.epsg === candidate.epsg
             )
         )
--- a/packages/viewer/src/api/print.api.ts
+++ b/packages/viewer/src/api/print.api.ts
@@ -108,9 +108,6 @@ class GeoAdminCustomizer extends BaseCustomizer {
         if (symbolizer.strokeWidth) {
             symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
         }
-        // TODO: ask Ismail about these 2 lines (should not be possible to have a graphic offset on a line)
-        // symbolizer.graphicXOffset = symbolizer.graphicXOffset ?? 0
-        // symbolizer.graphicYOffset = symbolizer.graphicYOffset ?? 0
     }
 
     /**
@@ -127,10 +124,7 @@ class GeoAdminCustomizer extends BaseCustomizer {
         symbolizer: MFPSymbolizerText,
         text: Text
     ) {
-        // symbolizer.pointRadius = adjustWidth(symbolizer.pointRadius, this.printResolution)
-        // symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
         symbolizer.haloRadius = adjustWidth(symbolizer.haloRadius, this.printResolution)
-        // symbolizer.conflictResolution = false
 
         // we try to adapt the font size and offsets to have roughly the same
         // scales on print than on the viewer.
--- a/packages/viewer/src/utils/featureStyleUtils.ts
+++ b/packages/viewer/src/utils/featureStyleUtils.ts
@@ -302,7 +302,7 @@ export function calculateFeatureTextWidth(text: string, textScale: number): numb
  * Returns offset (compared to marker) for text around it, depending on the text position and if the
  * description should be shown on the map too.
  */
-function getElementOffsets(editableFeature: EditableFeature): {
+function getElementOffsets(editableFeature?: EditableFeature): {
     top: [number, number]
     bottom: [number, number]
 } {
--- a/packages/viewer/src/utils/kmlUtils.ts
+++ b/packages/viewer/src/utils/kmlUtils.ts
@@ -406,12 +406,14 @@ export function getFillColor(
  * @returns EditableFeature or undefined if this is not a geoadmin feature
  */
 export function getEditableFeatureFromKmlFeature(
-    kmlFeature: OLFeature,
+    kmlFeature: OLFeature | undefined,
     availableIconSets: DrawingIconSet[],
     resolution: number
 ): EditableFeature | undefined {
     if (!kmlFeature) {
-        log.error(`Cannot generate EditableFeature from KML feature`, kmlFeature)
+        log.error({
+            messages: [`Cannot generate EditableFeature from KML feature`, kmlFeature],
+        })
         return
     }
     const featureType = getFeatureType(kmlFeature)


packages/viewer/src/api/__tests__/search.api.spec.ts
```diff
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 
-import { sanitizeTitle } from '@/api/search.api'
+import { sanitizeTitle } from '../search.api'
```
sommerfe: why is this necessary?

packages/viewer/src/api/features/features.api.ts
```diff
@@ -27,122 +27,124 @@ const APPLICATION_GML_3_TYPE = 'application/vnd.ogc.gml'
...
-            if (!Array.isArray(coordinates[0])) {
-                coordinates = [coordinates]
-            }
-            break
-    }
-    return coordinates
+export function extractOlFeatureCoordinates(feature: Feature): SingleCoordinate[] {
+    if (!feature) {
+        return []
+    }
```
sommerfe: if feature: Feature then we don't need this check, if we're not sure then it should be feature?: Feature

packages/viewer/src/api/features/features.api.ts
```diff
@@ -151,64 +153,66 @@ export async function identifyOnGeomAdminLayer({
...
+                // taking the center of the extent as coordinate
+                featureCoordinate = [
+                    (coordinate[0] + coordinate[2]) / 2,
+                    (coordinate[1] + coordinate[3]) / 2,
+                ]
+            }
+            if (!featureCoordinate) {
+                throw new GetFeatureInfoError('Unable to build required feature coordinate')
+            }
+            const featureData = await getFeatureHtmlPopup(geoadminLayer, feature.id, {
```
sommerfe, line 180:
```suggestion
            const featureData: string = await getFeatureHtmlPopup(geoadminLayer, feature.id, {
```

packages/viewer/src/api/features/features.api.ts
```diff
@@ -230,39 +234,42 @@ async function identifyOnExternalLayer(config) {
...
-            (availableProjection) => availableProjection.epsg === requestProjection.epsg
+        externalLayer.availableProjections &&
+        !externalLayer.availableProjections.some(
+            (availableProjection) => availableProjection.epsg === requestProjection?.epsg
         )
     ) {
         // trying to find a candidate among the app supported projections
         requestProjection = allCoordinateSystems.find((candidate) =>
-            layer.availableProjections.some(
+            externalLayer.availableProjections?.some(
```
sommerfe: ? is not necessary

packages/viewer/src/api/features/features.api.ts
```diff
@@ -428,35 +448,35 @@ async function identifyOnExternalWmsLayer(config) {
...
+                getFeatureInfoResponse.data
+            )
+            if (getFeatureInfoFeatures) {
+                features = getFeatureInfoFeatures.features
+            }
+        } else if (outputFormat === PLAIN_TEXT_TYPE) {
+            // TODO : implement plain text parsing
+            log.error('Plain text parsing not yet implemented')
         }
         return features?.map((feature) => {
```
sommerfe, line 478:
```suggestion
        return features?.map((feature: GeoJsonFeature<Geometry>) => {
```

packages/viewer/src/api/print.api.ts
```diff
@@ -0,0 +1,753 @@
...
+     * @param text Text style of the feature
+     */
+    text(
+        _layerState: State,
+        _geoJsonFeature: GeoJSONFeature,
+        symbolizer: MFPSymbolizerText,
+        text: Text
+    ) {
+        // symbolizer.pointRadius = adjustWidth(symbolizer.pointRadius, this.printResolution)
+        // symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
```
sommerfe: can be removed?

packages/viewer/src/api/print.api.ts
```diff
@@ -0,0 +1,753 @@
...
+        symbolizer.haloRadius = adjustWidth(symbolizer.haloRadius, this.printResolution)
+        // symbolizer.conflictResolution = false
```
sommerfe: can be removed?

packages/viewer/src/api/print.api.ts
```diff
@@ -0,0 +1,753 @@
...
+    ) {
+        if (symbolizer?.strokeDashstyle === '8') {
+            symbolizer.strokeDashstyle = 'dash'
+        }
+        if (symbolizer.strokeWidth) {
+            symbolizer.strokeWidth = adjustWidth(symbolizer.strokeWidth, this.printResolution)
+        }
+        // TODO: ask Ismail about these 2 lines (should not be possible to have a graphic offset on a line)
+        // symbolizer.graphicXOffset = symbolizer.graphicXOffset ?? 0
+        // symbolizer.graphicYOffset = symbolizer.graphicYOffset ?? 0
```
sommerfe: @ismailsunni

packages/viewer/src/utils/__tests__/kmlUtils.spec.ts
```diff
@@ -153,177 +147,189 @@ describe('Test KML utils', () => {
         })
         it('handles correctly text offset from extended data', () => {
             const icon = findFeatureWithId('drawing_feature_1714651153088')
-            expect(icon.textOffset).to.deep.equal([0, -44.75])
+            expect(icon).toBeDefined()
```
sommerfe, line 150: i created a function assertDefined here https://github.com/geoadmin/web-mapviewer/blob/develop-typescript/packages/viewer/tests/cypress/support/utils.ts#L7 
currently it is in the folder cypress but maybe we could move it somewhere where it is available for the unit tests and cypress tests i think this would improve a lot of these tests

packages/viewer/src/utils/featureStyleUtils.ts
```diff
@@ -326,26 +301,30 @@ export function calculateFeatureTextWidth(text, textScale) {
...
- *
- * @param {EditableFeature} editableFeature
- * @returns {{ top: [number, number]; bottom: [number, number] }}
  */
-function getElementOffsets(editableFeature) {
+function getElementOffsets(editableFeature: EditableFeature): {
+    top: [number, number]
+    bottom: [number, number]
+} {
     if (!editableFeature) {
```
sommerfe: same here if editableFeature is EditableFeature then the if should not be necessary else do editableFeature?: EditableFeature

packages/viewer/src/utils/kmlUtils.ts
```diff
@@ -393,30 +400,39 @@ export function getFillColor(style, geometryType, iconArgs) {
...
+export function getEditableFeatureFromKmlFeature(
+    kmlFeature: OLFeature,
+    availableIconSets: DrawingIconSet[],
+    resolution: number
+): EditableFeature | undefined {
+    if (!kmlFeature) {
         log.error(`Cannot generate EditableFeature from KML feature`, kmlFeature)
-        return null
+        return
     }
```
sommerfe: same here

packages/viewer/src/api/__tests__/search.api.spec.ts
pakb: my IDE doing weird stuff... I'll revert that 😉

packages/viewer/src/api/features/features.api.ts
pakb: as long as we are using it in a JS context, I prefer covering my back, but I could state that with a `?` optional sign on the function parameter to be clearer

packages/viewer/src/api/features/features.api.ts
pakb: as the function clearly states what the result type is, I don't know if I want to do that.
The type can clearly be inferred by the IDE/TS parser, and will be more resilient to changes if the function should be refactored in any way.

packages/viewer/src/api/features/features.api.ts
pakb: yes because the ExtenralLayer type declares this as
```TS
readonly availableProjections?: CoordinateSystem[];
```

packages/viewer/src/api/print.api.ts
pakb: the 3 comments you made above were meant to be double checked with Ismail, these are all missing in the TS declaration of the geoblock library, so I was wondering if they were really "real" or just some "reverse engineering magic"

packages/viewer/src/utils/__tests__/kmlUtils.spec.ts
pakb: in the context of unit testing, I would stick to use Chai/Mocha/Jest utils, such as this one `toBeDefined` ([see doc here](https://jestjs.io/docs/expect#tobedefined))

packages/viewer/src/api/features/features.api.ts
pakb: same as the other function comment, it should be sufficient to type the `features` variable here, and the type of the `.map` is then inferred

packages/viewer/src/api/print.api.ts
ismailsunni: After some digging, it seems it's better to ask @ltkum, since he is adding these lines https://github.com/geoadmin/web-mapviewer/commit/6b4d10964ce22f35f4175c31aaa3c1f786a7b2ac. It seems it fixes an issue.

I checked in the geoblocks, this graphicXOffset and graphicYOffset are only found in the Point but not Line:  https://github.com/geoblocks/mapfishprint/blob/main/src/types.ts#L29C18-L42

packages/viewer/src/api/iframePostMessageEvent.api.ts
```diff
@@ -1,32 +1,34 @@
...
+
+const targetWindow: Window = parent ?? window.parent ?? window.opener ?? window.top
 
 /**
  * All events fired by the iFrame postMessage implementation.
  *
  * @enum
  */
-export const IFRAME_EVENTS = {
+export enum IFrameEvents {
```
ismailsunni, line 12: Out of curiosity, for enum, we now use PascalCase, no longer TITLECASE?

packages/viewer/src/api/search.api.ts
```diff
@@ -0,0 +1,547 @@
...
+ * @property {String} message Technical english message
+ */
+export class SearchError extends Error {
+    constructor(message?: string) {
+        super(message)
+        this.name = 'SearchError'
+    }
+}
+
+export enum SearchResultTypes {
```
ismailsunni, line 37: Should we use PascalCase also for the enum?

packages/viewer/src/api/iframePostMessageEvent.api.ts
pakb: That's what I found on the TS doc, yes
https://www.typescriptlang.org/docs/handbook/enums.html

packages/viewer/src/api/search.api.ts
pakb: for TS enum, yes, and for their values too (see my comment on your previous comment, and the link to the TS doc)

PR #1416, opened 2025-09-19
PB-1383: Fix Drawing Test

ismailsunni:
[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-1383-drawing-test-2/index.html)

### DIFF
--- a/packages/viewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
+++ b/packages/viewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
@@ -180,10 +180,12 @@ function onDelete() {
 function onAddMediaLink(mediaPopoverIndex, descriptionMediaLink) {
     mediaPopovers.value[mediaPopoverIndex].hidePopover()
     // Prevent 'undefined' to be added to the description
-    if (description.value) {
-        description.value += descriptionMediaLink
-    } else {
-        description.value = descriptionMediaLink
+    if (descriptionMediaLink && typeof descriptionMediaLink === 'string') {
+        if (description.value) {
+            description.value += descriptionMediaLink
+        } else {
+            description.value = descriptionMediaLink
+        }
     }
 }
 


packages/viewer/tests/cypress/cypress.d.ts
```diff
@@ -132,7 +132,7 @@ declare global {
             /** Opens the drawing mode. Will open the menu if the app is in mobile mode and the menu is not yet open. */
             openDrawingMode(): void
             /** Closes the drawing mode. Will close the "You have not copied the link" message if it is shown. */
-            closeDrawingMode(): void
+            closeDrawingMode(closeDrawingNotSharedAdmin?: boolean): void
```
pakb, line 139: can you summarize what this flag does in the JSDoc?
The name of the flag doesn't clearly states what it does IMHO

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -20,10 +20,11 @@ import {
     getKmlAdminIdFromRequest,
     kmlMetadataTemplate,
 } from '../support/drawing.js'
```
pakb: you can change all imports ending in `xxx.js` to `xxx` so that we make sure we import the TS directly (and not the transpiled JS file)

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -37,37 +38,37 @@ const DEFAULT_ICON_URL_SCALE = `${DEFAULT_ICON_URL_PARAMS.scale}x`
...
+        function testTitleEdit(): void {
             const title = `This is a random title ${randomIntBetween(1000, 9999)}`
             cy.get('[data-cy="drawing-style-feature-title"]').clear()
             cy.get('[data-cy="drawing-style-feature-title"]').type(title)
             cy.get('[data-cy="drawing-style-feature-title"]').should('have.value', title)
             cy.wait('@update-kml')
                 .its('request')
-                .should((request) =>
-                    checkKMLRequest(request, [new RegExp(`<name>${title}</name>`)])
+                .then((request) =>
```
pakb: is it possible to keep the `.should` here?
It comes with retry ability, whereas `then` only run once.

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -83,12 +84,12 @@ describe('Drawing module tests', () => {
...
         // we use the description to identify the feature and check its
         // geometry type, number of points and type (measure or linepolygon)
-        function checkDrawnFeature(description, numberOfPoints, featureType, type) {
+        function checkDrawnFeature(description: string, numberOfPoints: number, featureType: string, type: EditableFeatureTypes): void {
             cy.window().its('drawingLayer')
                 .then((drawingLayer) => drawingLayer.getSource().getFeatures())
                 .should((features) => {
                     const matchingFeature = features.find(
-                        (feature) => feature.get('description') === description
+                        (feature: import('ol/Feature').default) => feature.get('description') === description
```
pakb: you should import it at the top of the file with
```JS
import { default as OLFeature } from 'ol/Feature'
```
or something similar

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -452,8 +453,8 @@ describe('Drawing module tests', () => {
             cy.get('[data-cy="ol-map"]').click()
             cy.wait('@post-kml')
                 .its('request')
-                .should((request) => {
-                    checkKMLRequest(request, [
+                .then((request) => {
```
pakb: same thing with `should` vs. `then`

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -479,8 +480,8 @@ describe('Drawing module tests', () => {
                 .click()
             cy.wait('@update-kml')
                 .its('request')
-                .should((request) => {
-                    checkKMLRequest(request, [
+                .then((request) => {
```
pakb: one more

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -502,8 +503,8 @@ describe('Drawing module tests', () => {
             ).click({ force: true })
             cy.wait('@update-kml')
                 .its('request')
-                .should((request) => {
-                    checkKMLRequest(request, [
+                .then((request) => {
```
pakb: one more

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -534,7 +535,7 @@ describe('Drawing module tests', () => {
             cy.viewport(1920, 1080)
             cy.clickDrawingTool(EditableFeatureTypes.LINEPOLYGON)
 
-            const lineCoordinates = [
+            const lineCoordinates: [number, number][] = [
```
pakb: You can use the `SingleCoordinate` type from
```JS
import type { SingleCoordinate } from '@swissgeo/coordinates'
```

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -654,7 +659,7 @@ describe('Drawing module tests', () => {
             cy.log('Create measurement line')
             cy.clickDrawingTool(EditableFeatureTypes.MEASURE)
 
-            const measurementCoordinates = [
+            const measurementCoordinates: [number, number][] = [
```
pakb: same

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -717,12 +723,12 @@ describe('Drawing module tests', () => {
             cy.log('should create a polygon by re-clicking the first point')
             cy.get('[data-cy="ol-map"]').click(100, 250)
 
-            let kmlId = null
+            let kmlId: string | null = null
```
pakb: when migrating to TS, I would prefer we use `undefined` for non set variables, it's easier to pass down as a param with the optional sign (than to explicitly state it can be null)

So a good rule of thumbs when migrating : change all `null` possible value to `undefined`

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -717,12 +723,12 @@ describe('Drawing module tests', () => {
...
             cy.wait('@post-kml').then((interception) => {
                 cy.wrap(interception)
                     .its('request')
-                    .should((request) =>
-                        checkKMLRequest(request, [
+                    .then((request) =>
```
pakb: this `should` vs. `then` question again

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -759,17 +765,17 @@ describe('Drawing module tests', () => {
             })
             cy.wait('@update-kml')
                 .its('request')
-                .should((request) =>
+                .then((request) =>
```
pakb: here too

packages/viewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
```diff
@@ -179,7 +179,12 @@ function onDelete() {
 }
 function onAddMediaLink(mediaPopoverIndex, descriptionMediaLink) {
     mediaPopovers.value[mediaPopoverIndex].hidePopover()
-    description.value += descriptionMediaLink
+    // Prevent 'undefined' to be added to the description
+    if (description.value) {
+        description.value += descriptionMediaLink
+    } else {
+        description.value = descriptionMediaLink
+    }
```
sommerfe: so if the descriptionMediaLink is undefined we just set the description to undefined? is that what we want? or should we just ignore it?

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
```diff
@@ -453,7 +455,7 @@ describe('Drawing module tests', () => {
             cy.wait('@post-kml')
                 .its('request')
                 .should((request) => {
-                    checkKMLRequest(request, [
+                    void checkKMLRequest(request as CyHttpMessages.IncomingHttpRequest, [
```
sommerfe, line 458: why this void before the function call?

sommerfe (approved):

packages/viewer/tests/cypress/tests-e2e/drawing.cy.ts
ismailsunni: to fix the TS issue, since inside the `should` it should return `void`

packages/viewer/src/modules/infobox/components/styling/FeatureStyleEdit.vue
ismailsunni: Ah, yes, you are right. I have added better handler for this case.

PR #1564, opened 2026-02-03
PB-1878: service worker versioning start

sommerfe:
### Files involved

- **`packages/mapviewer/src/service-workers.ts`**
  - Source Service Worker code (Workbox routes and caching strategies).

- **`dist/service-workers.js`** (root)
  - Stable Service Worker entrypoint used by browser registration (`./service-workers.js`).

- **`dist/<appVersion>/service-workers.js`**
  - Versioned Service Worker implementation for a specific release.

- **`dist/<appVersion>/sw-ready.json`**
  - Build validation artifact written by custom plugin (`vite-plugin-version-sw-path`).

---

### Build-time workflow

1. **VitePWA builds the SW**
   - Compiles `service-workers.ts` into `dist/service-workers.js` (and optional map).

2. **`vite-plugin-move-sw` runs**
   - Copies `dist/service-workers.js` → `dist/<appVersion>/service-workers.js`.
   - Rewrites root `dist/service-workers.js` as bootstrap:
     - `importScripts('./<appVersion>/service-workers.js')`
   - Optionally copies sourcemap.

3. **`vite-plugin-version-sw-path` runs**
   - Scans built chunks for SW registration path usage (`"./service-workers.js"`).
   - Writes `dist/<appVersion>/sw-ready.json` with validation metadata.

---

### What each file does

- **Root `service-workers.js`**
  - Keeps a stable registration URL and scope behavior.

- **Versioned `/<appVersion>/service-workers.js`**
  - Holds the actual SW logic for that release.

- **`sw-ready.json`**
  - Diagnostic/validation metadata (`swVersioned`, `version`, `timestamp`, etc.).
  - In current setup, it is **not used at runtime** unless explicitly read elsewhere.

---

### Runtime workflow (when user opens app)

1. App registers SW at `./service-workers.js`.
2. Browser fetches root SW script.
3. Root script imports versioned SW via `importScripts`.
4. SW installs/activates:
   - precache (`self.__WB_MANIFEST`)
   - runtime routes (navigation/config/images)
   - `skipWaiting()` + `clientsClaim()` for fast takeover
5. Subsequent loads use cache/network strategies (mostly `NetworkFirst` in your code).

---

### Why versioning matters

- Most relevant during **deployments/updates**.
- New release changes root SW bootstrap to point to new versioned SW file.
- Browser detects SW script update at same registration URL.
- New SW installs/activates and replaces previous behavior/caches.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1878-version-service-worker/index.html)


packages/mapviewer/vite-plugins/vite-plugin-move-sw.js
```diff
@@ -0,0 +1,93 @@
...
+    let outputDir = ''
+
+    return {
+        name: 'vite-plugin-move-sw',
+        enforce: 'post',
+        async closeBundle() {
+            const fs = await import('fs/promises')
+            const path = await import('path')
+
+            // eslint-disable-next-line no-console
```
pakb: I think you can safely move that to a "top-file" disabling.

packages/mapviewer/vite-plugins/vite-plugin-move-sw.js
```diff
@@ -0,0 +1,93 @@
...
+                } catch {
+                    // Source map doesn't exist, which is fine
+                    // eslint-disable-next-line no-console
+                    console.log('[vite-plugin-move-sw] No source map found (this is OK)')
+                }
+
+                // eslint-disable-next-line no-console
+                console.log('[vite-plugin-move-sw] Service worker relocation complete')
+            } catch (error) {
+                this.warn(
```
pakb: do you mean `console.warn`?

packages/mapviewer/vite-plugins/vite-plugin-version-sw-path.js
```diff
@@ -0,0 +1,104 @@
+/**
+ * Vite plugin that validates the service worker registration path and emits a build validation file.
+ *
+ * This plugin:
+ * 1. Finds chunks containing service worker registration code from the virtual:pwa-register module
+ * 2. Validates that the registration path remains "./service-workers.js" (root scope-safe path)
+ * 3. Emits a validation file (sw-ready.json) to indicate successful configuration
+ * 4. Warns if no SW registration pattern is found (validation failure)
+ */
```
pakb, line 10: What is exactly the problem that this plugin is trying to detect?
A failed build?

pakb (commented):
looking good so far, I have one general question about the second plugin

I think we should translate all the text you've added to the Vue component too (but I'm guessing you were waiting for a technical review first)

packages/mapviewer/vite-plugins/vite-plugin-version-sw-path.js
sommerfe: The plugin moves the `service-worker.js` file to `${appVersion}/service-workers.js` if this was not successful, it logs the warning but it will not make the build fail

pakb: So in case this `sw-ready.json` isn't created, the build doesn't fail and continue its course. Then the app loads that and does not activate SW at all because of the missing file? So it kind of revert back to simple load of the page?

I'm not 100% sure how this will behave whilst having already cached/loaded SW environment in the browser though

I'm a bit afraid the browser will stick to its latest valid version it has in the cache

packages/mapviewer/vite-plugins/vite-plugin-version-sw-path.js
```diff
@@ -0,0 +1,104 @@
...
+        name: 'vite-plugin-version-sw-path',
+        enforce: 'post',
+
+        configResolved(config) {
+            outputDir = config.build.outDir
+        },
+
+        generateBundle(options, bundle) {
+            swPatternFound = false
+            // eslint-disable-next-line no-console
```
pakb: you can also set this rule as disabled for the whole script file

packages/mapviewer/vite-plugins/vite-plugin-version-sw-path.js
sommerfe: The `sw-ready.json` file does not have any functionality regarding the service worker itself, it is only there for the `OfflineReadinessStatus.vue` to show an error if the service worker validation failed, but the service worker is working independently from it.

The browser uses normal SW lifecycle: install -> waiting/active -> claim/update.
If the new `/service-workers.js` is reachable and changed, it can update to the new service worker.
If the update path is broken, browser keeps last valid active SW 

btw i noticed when i tested on the phone, that the offline readiness status is not visible on mobile, where it is mostly used, so we might want to add that on the mobile view

ltshb (commented):
I think @pakb can review this better than I do

pakb (commented):
LGTM so far, what I propose we do, if you think this could be ready to be tested "live" is that we could ship this change as a single version (no other changes).
This way we can quickly rollback if we see it's not behaving as planned

rotwurstesser:
Closing this PR because the research and preview tests confirmed that the stable service-worker URL works correctly with the browser update process.

The service-worker cache header policy was changed to `no-cache`.

PR #1565, opened 2026-02-03
PB-1878: add update handling

sommerfe:
[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-1878-add-update-handling/index.html)


rotwurstesser:
Closing this PR because the research and preview tests confirmed that the stable service-worker URL works correctly with the browser update process.

The service-worker cache header policy was changed to `no-cache`.

PR #1581, opened 2026-03-20
PB-2241: Added labels for v3 of BABS civil signatures

hansmannj:
Adding dictionary entries for the new BABS signatures.
[Test link](https://sys-map.dev.bgdi.ch/preview/feat_pb-2241_new_babs_icon_sets/index.html)


pakb (approved):
that's only labels, we can safely merge that without any issue

hansmannj:
> that's only labels, we can safely merge that without any issue

OK, makes sense. So there won't be an error in case the v3 icon set is not yet around, right?

pakb:
> OK, makes sense. So there won't be an error in case the v3 icon set is not yet around, right?

nope, i18n keys are asked on-demand, and if not found shown untranslated.

there are some weird unrelated errors in the CI/e2e tests, so you can force-push the same commit to see if another run could fix this

packages/mapviewer/src/modules/i18n/locales/rm.json
```diff
@@ -434,6 +434,9 @@
     "modify_icon_category_babs-v2-de_label": "Zivile Signaturen",
     "modify_icon_category_babs-v2-fr_label": "Zivile Signaturen",
     "modify_icon_category_babs-v2-it_label": "Zivile Signaturen",
+    "modify_icon_category_babs-v3-de_label": "Zivile Signaturen",
```
rotwurstesser, line 437: is this rumantsch? ^^
Not sure if we should replace it with actual rumantsch

hansmannj: This is how it currently seems implemented.
For the user, nothing will change, we simply add `v3` to the dict.
Currently for `v2` the German text is used already in the RM dictionary.
Guess, German is a fallback to Rumantsch here.

But you are right, why not use real Rumantsch here?
Anyone has an idea for the correct terms here? (@c-lutz ? :wink:)

hansmannj: But then, on the other hand, the icons themselves are only available in DE, IT and FR. So probably, having a RM onhover text might be confusing, since the icons are not in RM. Hence I'd suggest to keep it as is for now

rotwurstesser (approved):

PR #1582, opened 2026-04-17
PB-2553: stopping XSS attacks by whitelisting

ltkum:
- Issue: The api_url, wms_url and wmts_url params can currently receive any url as values, which means a malicious party could use them as an attack vector. This is especially true for api_url, which can send html back and is needed to operate that way.
- Fix: We have made regexes which enfore a whitelist on those parameters As we need external sources html to be rendered (for the feature informations, for example), we sadly can't sanitize the html given to us without breaking a lot of currently working layers. We can only limit the sources and hope they don't provide malicious content.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-2253-whitelisting-only-some-urls-for-apis/index.html)

### DIFF
--- a/packages/mapviewer/src/router/storeSync/BaseUrlOverrideParamConfig.class.js
+++ b/packages/mapviewer/src/router/storeSync/BaseUrlOverrideParamConfig.class.js
@@ -6,9 +6,22 @@ import {
 import AbstractParamConfig from '@/router/storeSync/abstractParamConfig.class'
 import { isValidUrl } from '@/utils/utils'
 
+// //https://sys-{wms,wmts,api3}.{dev,int}.bgdi.ch/
+// or http://localhost{:port_number}
+const allowed_url_regexes = {
+    wms_url: /^https:\/\/sys-wms.(dev|int).bgdi.ch\/?$/g,
+    wmts_url: /^https:\/\/sys-wmts.(dev|int).bgdi.ch\/?$/g,
+    api3: /^https:\/\/sys-api3.(dev|int).bgdi.ch\/?$/g,
+    localhost: /^http:\/\/localhost(:\d{2,6})?\/?$/g,
+}
+
 export default function createBaseUrlOverrideParamConfig({ urlParamName, baseUrlPropertyName }) {
     function dispatchBaseUrlOverride(to, store, urlParamValue) {
-        if (isValidUrl(urlParamValue)) {
+        if (
+            isValidUrl(urlParamValue) &&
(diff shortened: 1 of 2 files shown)


packages/mapviewer/src/config/baseUrl.config.js
```diff
@@ -151,8 +151,21 @@ export function getBaseUrlOverride(propertyName) {
     return baseUrlOverrides[propertyName]
 }
 
+// //https://sys-{wms,wmts,api3}.{dev,int}.bgdi.ch/
+const allowed_url_regexes = {
+    wms_url: /^https:\/\/sys-wms.(dev|int).bgdi.ch\/$/g,
```
hansmannj: Thanks a ton for fixing the XSS @ltkum 🚀  Not sure, if the `/g` is really needed, since the regex is fully anchored already, but then it also won't hurt to have it 😉

packages/mapviewer/src/config/baseUrl.config.js
```diff
@@ -151,8 +151,21 @@ export function getBaseUrlOverride(propertyName) {
...
+    wmts_url: /^https:\/\/sys-wmts.(dev|int).bgdi.ch\/$/g,
+    api3: /^https:\/\/sys-api3.(dev|int).bgdi.ch\/$/g,
```
hansmannj: Sorry, picky, I know: is there any specific reason, why there's no `_url` suffix for `api3`, but for `wms`and `wmts` there is?

pakb (changes requested):
We must keep also available (for our DI people) the possibility to add `http://localhost:xxxx` type URLs.


Can you move your REGEX check to https://github.com/geoadmin/web-mapviewer/blob/develop/packages/mapviewer/src/router/storeSync/BaseUrlOverrideParamConfig.class.js and check all override in one go there ?

ltkum:
> We must keep also available (for our DI people) the possibility to add `http://localhost:xxxx` type URLs.
> 
> Can you move your REGEX check to https://github.com/geoadmin/web-mapviewer/blob/develop/packages/mapviewer/src/router/storeSync/BaseUrlOverrideParamConfig.class.js and check all override in one go there ?

sure thing

packages/mapviewer/src/config/baseUrl.config.js
ltkum: mostly because those are the legacy names and that we must comply with them for now :)

Edit: mostly because I didn't get the correct key :3

schtibe (approved):

PR #1586, opened 2026-07-02
PB-2337: compact overflowing map attributions

rotwurstesser:
Issue: Local file and layer attributions can pile up in the map footer until they cover the map viewer. See https://swissgeoplatform.atlassian.net/browse/PB-2337\n\nFix: The footer now keeps the first three attribution sources inline and groups additional sources behind a compact +N toggle with a scrollable expanded list. Attribution links keep full titles, local-file disclaimers stay consistent, and the print service attribution payload remains independent from the viewer footer compaction.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-local-file-attribution-overflow-spec/index.html)


packages/mapviewer/src/modules/i18n/locales/de.json
```diff
@@ -27,6 +27,9 @@
     "appcache_update_available": "Neue Version der Offline Funktion verfügbar. Wollen Sie sie laden?",
     "approx_abbr": "ca.",
     "are": "ARE",
+    "attribution_overflow_hide_sources": "Weitere Quellenangaben ausblenden",
+    "attribution_overflow_show_source": "Eine weitere Quellenangabe anzeigen",
+    "attribution_overflow_show_sources": "{count} weitere Quellenangaben anzeigen",
```
ltkum, line 32: So, for translations, there is an extra step, but I'm fairly certain nobody told you about it.

We have a google sheets which holds the translation, and we are using `pnpm run translate` to fetch it and create those translation JSON.

I'll send you the link, and then you should ask someone to give you edit permission (worst case scenario, I'll add them for you).

I don't know who can grant these writing permissions, as I think the person which granted me these no longer work here.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -41,24 +43,56 @@ const sources = computed(() => {
...
             return index === firstIndex
         })
 })
+
+const isCompact = computed(() => sources.value.length > COMPACT_SOURCE_COUNT)
+const hiddenSourceCount = computed(() =>
+    isCompact.value ? sources.value.length - COMPACT_SOURCE_COUNT : 0
+)
+const inlineSources = computed(() =>
+    isCompact.value ? sources.value.slice(0, COMPACT_SOURCE_COUNT) : sources.value
```
ltkum: unless I am mistaken, slice supports indexes higher than the array size. For example :
```ts
const sources = ["source a", "source b"]
console.log(sources.slice(0,7))
```
should have `["source a", "source b"]` as a result. So there is no need to check if the compact value is set or not.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -41,24 +43,56 @@ const sources = computed(() => {
...
+)
+const expandedSources = computed(() =>
+    isCompact.value ? sources.value.slice(COMPACT_SOURCE_COUNT) : []
```
ltkum: same as above, slice also supports indexes higher than the array size for the first index

```ts
const sources = ["source a", "source b"]
console.log(sources.slice(3,7))
```

should have `[]` as a result.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -67,22 +101,67 @@ const sources = computed(() => {
...
+            data-cy="attribution-expand-toggle"
+            :title="expandButtonLabel"
+            @click="toggleExpanded"
+        >
+            {{ toggleButtonText }}
+        </button>
+        <div
+            v-if="isCompact"
+            class="map-footer-attribution-expanded-list"
+            :class="{ 'map-footer-attribution-expanded-list-hidden': !isExpanded }"
```
ltkum: rather than have a class to conditionnaly turn off the display, we could use

`v-if="isCompact && isExpanded"`

alternatively, we could have 
```ts
v-if="isCompact"
v-show="isExpanded"
```
if we want to avoid recreating the components every time we toggle it.

ltkum (commented):
This is mostly nitpicking on my part, the work is overall good :) 
The CSS is a bit brute-forced. We usually try to use as many pre-defined classes as possible (for Swissgeo this would be tailwind. I believe the mapviewer is using bootstrap but I might be wrong). To be perfectly fair, given the status of the mapviewer, I am fine with your solution CSS wise :).

I'll send the google sheet link to you via slack, and hopefully you'll get the right to edit it soon.

packages/mapviewer/src/modules/i18n/locales/de.json
rotwurstesser: i got added, and added them and the generation gave the exact thing i commited :)

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
rotwurstesser: Went with v-if + v-show as it aligns nicely with the bootstrap refactor

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
rotwurstesser: true, fixed!

ltkum (approved):
Looks good to me. 

I see there are PR merges on develop in your commit history, if we can get rid of those, that would be great, but once it's done, I'm happy with this.

hansmannj:
Very nice, thanks! :rocket: 
If possible without too large effort, I'd suggest to only compact the attributions for layers that were imported from a local source, as in the underlying issue with the multiple imported local KMLs.
For the other layers (also those exported from external WM(T)S services), I'd leave everything as is.
We have more degrees of freedom in SWISSGEO web-portal. For the mapviewer, I'd not introduce too many changes if not necessary.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -37,26 +39,62 @@ const sources = computed(() => {
...
+
+function toggleExpanded() {
+    isExpanded.value = !isExpanded.value
+}
 </script>
 
 <template>
     <div
-        class="map-footer-attribution"
+        v-click-outside="closeExpanded"
```
hansmannj, line 91: Guess it might make sense to also close on pressing ESC or the like, not only on clicking outside, if that is possible without too large an effort

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -37,26 +39,62 @@ const sources = computed(() => {
...
+    Math.max(VISIBLE_SOURCE_COUNT - primarySources.value.length, 0)
+)
+const inlineSources = computed(() => [
+    ...primarySources.value,
+    ...localSources.value.slice(0, visibleLocalSourceCount.value),
+])
+const expandedSources = computed(() => localSources.value.slice(visibleLocalSourceCount.value))
+const hiddenSourceCount = computed(() => expandedSources.value.length)
+const isCompact = computed(() => hiddenSourceCount.value > 0)
+const toggleButtonText = computed(() => (isExpanded.value ? '-' : `+${hiddenSourceCount.value}`))
```
hansmannj, line 66: Probably, when expanded, we should display an `x` rather than `-`?
Picky, I know, sorry. And also probably a matter of personal taste, not really super important 😉

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -37,26 +39,62 @@ const sources = computed(() => {
...
             })
         })
-        .flat()
         .filter((attribution, index, array) => {
             const firstIndex = array.findIndex((item) => item.name === attribution.name)
             return index === firstIndex
         })
 })
+
+const VISIBLE_SOURCE_COUNT = 4
```
hansmannj, line 53: Fine in this mapviewer context!

Idea for SWISSGEO: Rather than a hardcoded number of attributions, we might consider the relative width of the attributions list. So e.g. 6 short attributions are ok, while 4 longer ones would already be compacted. Also, we can make this responsive and treat differently depending on the kind of the device / available screen width

ltkum (commented):
IMO, we can go as it is. If you would like to adress @hansmannj  suggestions regarding the ESC, I will make a quick check before we merge. With that being said, I would advise against it.

Since we're clicking on the attribution footer to expand it, and any click anywhere else than within it (or a click on the `- / x ` button)  would close the expanded list, I don't think this is necessary to add another way to close it, and there is another side effect

We used to have a lot of keyboard shortcuts in the old mapviewer and we got rid of almost all of them because key presses to do some functions are often unintuitive, and often very loosely documented (if I remember well, it was possible to press ctrl+d to open the drawing mode, for example... but this is a memory from long ago ^^).

The only one we still have is on the ESC key, which stop "fullscreen" mode. Adding this would add a situation where using it to close the attribution expanded list would also close the fullscreen mode of the app if it was activated.

Using other keyboards inputs than ESC to close something would be counter intuitive for users, and we would have a side effect with two functions tied to the same touch.

@hansmannj , what are your thoughts on this ?

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
ltkum: We have places within the mapviewer where we automatically truncate and add a small tooltip which contains the full string (for example: title features). We could use a similar approach to have a character limit, be a bit smart about it to avoid cutting an attribution in the middle, and have it be our trigger to see if we need to truncate the attribution list or not.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
```diff
@@ -67,22 +105,70 @@ const sources = computed(() => {
                 :is-local-file="source.isLocalFile"
             >
                 <MapFooterAttributionItem
-                    :source-id="source.id"
+                    :source-id="`${source.id}-${index}`"
                     :source-name="source.name"
                     :source-url="source.url"
                     :has-data-disclaimer="true"
-                    :is-last="index === sources.length - 1"
+                    :is-last="!isCompact && index === inlineSources.length - 1"
```
hansmannj: Do we really need the `!isCompact` check here?
How about simply `:is-last="index === inlineSources.length - 1"`
This would probably prevent a trailing `,` such as currently observed in `Source A, Source B, +3`
But this is also a picky comment, sorry.
If we cannot get rid of the traling comma with low effort, then we let it stay there.

packages/mapviewer/src/modules/map/components/footer/MapFooterAttributionList.vue
hansmannj: I'd say we keep it as in this PR for the mapviewer.
Any more advanced logic, including responsiveness and the like, is for SWISSGEO's we-portal. And in SWISSGEO we also discuss this with the UX team

hansmannj (approved):
Very nice, thanks a ton 🚀 
Only some minor comments, low prio.

Only question:
- might make sense to add some simple test cases for this new compaction functionality
- also we should make sure, that no existing tests rely on a logic, that is now changed due to introducing the new compaction functionality

PR #1587, opened 2026-07-13
PB-2246: deploy the new 3D label dataset

rotwurstesser:
## Summary

- route the Swissnames 3D layer to a dedicated MBTiles/PBF renderer while keeping other Cesium vector layers unchanged
- load altitude/zoom-specific label tiles, sample terrain positions, and render terrain-occluded labels through a `CustomDataSource`
- limit regular-label display distance and improve lake/peak contrast against terrain

## Side effects and review focus

- this replaces only the `ch.swisstopo.swissnames3d.3d` rendering path
- the renderer uses the existing environment-specific 3D base URL; no host or token is hard-coded
- missing label tiles are ignored without stopping the Cesium renderer
- screen-space collision/decluttering is intentionally not included

Before Merge:

- [x]  Remove old label implementation
- [ ]  Verify all data mutations work on data provider side
- [ ]  Verify we dont want to use unstable cesium MTV Provider

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-2246-3d-swissnames-labels/index.html)


rotwurstesser:
Superseded by #1589, which uses the selected standard 3D Tiles solution.

PR #1589, opened 2026-07-20
PB-2246: render Swissnames labels with standard 3D Tiles

rotwurstesser:
## What changed

Render Swissnames 3D labels from standard 3D Tiles with the native Cesium renderer.

Approved data:

- DEV: https://sys-3d.dev.bgdi.ch/ch.swisstopo.swissnames3d.3d/v2/tileset.json
- INT: https://sys-3d.int.bgdi.ch/ch.swisstopo.swissnames3d.3d/v2/tileset.json

The frontend selects the environment-specific 3D host and the stable `v2/tileset.json` publication.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-2246-standard-3d-labels/index.html)


packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -0,0 +1,159 @@
...
+            for (let index = 0; index < tile.content.featuresLength; index += 1) {
+                tileFeatures.push(addFeature(labels, connectors, tile.content.getFeature(index)))
+            }
+            activeFeatures.set(tile, tileFeatures)
+        }
+        visibleTiles.clear()
+    }
+
+    onMounted(async () => {
+        viewer = getViewer()
```
schtibe, line 126: I'm generally against having too much code in the `onMounted` hook directly (or any hook actually). In my opinion, everything should be factored to separate functions and `onMounted` just calls them.

The reason is that
a) it is harder to know what's going on. A function at least has a name, usually telling what the code is about
b) it is harder to maintain. If you need to add something to this hook some day, and you add it just to the bottom, then you make the new code implicitly dependent on the existing code. In this here case, the existing code returns if the viewer is null-ish, which might not be desired for the new code.

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -0,0 +1,159 @@
...
+        if (!viewer) {
+            log.error('Failed to load Swissnames labels: missing Cesium viewer')
+            return
+        }
+        try {
+            const tilesetUrl = `${layerConfig.baseUrl}${layerConfig.id}/${layerConfig.urlTimestampToUse}/tileset.json`
+            const loadedTileset = await Cesium3DTileset.fromUrl(tilesetUrl)
+            if (isUnmounted) {
```
schtibe, line 135: why is this needed? Or asked differently, how would this situation occur, that we have an onMounting while at the same time the ref is true (from a previous mount/unmount lifecylcle I presume, which would result in an entire new ref?)

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -0,0 +1,159 @@
...
+        distanceDisplayCondition,
+    })
+}
+
+function addLabel(collection, feature, position, distanceDisplayCondition) {
+    const maxDistance = feature.getProperty('maxDistance')
+    return collection.add({
+        position,
+        text: feature.getProperty('text'),
+        font: `${feature.getProperty('fontSize')}px Arial, sans-serif`,
```
schtibe, line 56: Is Arial correct here?

packages/mapviewer/src/modules/map/components/cesium/CesiumVectorLayer.vue
```diff
@@ -37,10 +37,7 @@ useAddPrimitiveLayer(
         // with default value 16 we do not load a lot of building tiles (leading to gaps)
         maximumScreenSpaceError: layerId.value === CESIUM_BUILDING_LAYER_ID ? 10 : 16,
     }),
-    toRef(opacity),
-    {
-        withEnhancedLabelStyle: layerId.value === CESIUM_LABELS_LAYER_ID,
-    }
+    toRef(opacity)
```
schtibe: Is `toRef` needed when it's already a computed?

schtibe (commented):
Generally looking good to me apart from some details. But I'm not in the position to give a very substantiated review, as I'm entirely unfamiliar with the 3D topic, maybe that requires @hansmannj or somebody who's been on the project longer.

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
rotwurstesser: it was arial in the old swissname layers too :)

packages/mapviewer/src/modules/map/components/cesium/CesiumVectorLayer.vue
rotwurstesser: good catch, no it isnt, removed

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
rotwurstesser: true, extracted to a seperate function, lmk if that looks better for you!

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
rotwurstesser: mmmm its because Cesium3DTileset.fromUrl() is async and can unmount before completion so i would definitly keep that guard tbh

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -0,0 +1,170 @@
...
+        if (!viewer) {
+            log.error('Failed to load Swissnames labels: missing Cesium viewer')
+            return
+        }
+        try {
+            const tilesetUrl = `${layerConfig.baseUrl}${layerConfig.id}/${layerConfig.urlTimestampToUse}/tileset.json`
+            const loadedTileset = await Cesium3DTileset.fromUrl(tilesetUrl)
+            if (isUnmounted) {
+                loadedTileset.destroy()
+                return
```
ltkum, line 137: If I understand this well, this is meant to ensure that if we mount and unmount the layer quickly, we want to ensure we're not leaving some tiles in memory.

This is something I believe needs to be tested. Cesium is already a abomination in term of performance, I would very much like to ensure that guardrails against memory leaks are working :3

ltkum (commented):
This is in general very solid. I would like to nitpick and add an extra test, but to be fair, this wouldn't be a necessity.

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
rotwurstesser: added

ltkum (approved):
LGTM

hansmannj (approved):
Thanks a ton, generally looking good to me.
Just some minor things I noticed:

- the labels appear now (at least on my machine/browser) in white fontsize against a black background, which currently it's black fontsize with a white boundary. What's the reasoning behind this change?

current:
Image

this PR:
Image

rotwurstesser:
@hansmannj 
I inplemented that because it's way more readable especially when you have umlaute: Zürich, lac léman, and such look a bit choppy with the old border. Another reason is that when they stack (you tilt the camera and then have multiple labels in one line) then it's really confusing and unreadable if you can read other labels between the letters... also with the coloring it's easier to adapt all label colors to be legible on one dark bg instead of having them all be readable with a font outline.

hansmannj:
Thanks once more.
What I'd like to have added to my commment yesterday, but due to the GitHub issues could not, is, that we can try the new styling. But we should also expect that we have to rollback quick, in case a lot of users complain.
(Personally, I like the currently deployed style more). Let's discuss it on today's refinement.

PR #1595, opened 2026-08-19
PB-2246: hide label stems when labels begin fading

rotwurstesser:
## Description

Hide 3D label stems at the distance where their labels begin fading. This avoids detached stems without per-frame updates or a custom shader.

[Test link](https://sys-map.dev.bgdi.ch/preview/fix-pb-2246-hide-label-stems-during-fade/index.html)

### DIFF
--- a/packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
+++ b/packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
@@ -75,10 +75,16 @@ function addLabel(collection, feature, position, distanceDisplayCondition) {
 function addFeature(labelCollection, connectorCollection, feature) {
     const groundPosition = getPosition(feature, 'groundHeight')
     const labelPosition = getPosition(feature, 'labelHeight')
-    const distance = new DistanceDisplayCondition(0, feature.getProperty('maxDistance'))
+    const maxDistance = feature.getProperty('maxDistance')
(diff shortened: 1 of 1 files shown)


packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -75,10 +75,16 @@ function addLabel(collection, feature, position, distanceDisplayCondition) {
 function addFeature(labelCollection, connectorCollection, feature) {
     const groundPosition = getPosition(feature, 'groundHeight')
     const labelPosition = getPosition(feature, 'labelHeight')
-    const distance = new DistanceDisplayCondition(0, feature.getProperty('maxDistance'))
+    const maxDistance = feature.getProperty('maxDistance')
+    const labelDistance = new DistanceDisplayCondition(0, maxDistance)
+    const connectorDistance = new DistanceDisplayCondition(0, maxDistance * FADE_START_RATIO)
```
hansmannj, line 80: Overall nice, thanks a ton!
As I understand this, the label's opacity fades slowly in a smooth transition, whereas the connector/stem just has a "hard" cut-off at `maxDistance * FADE_START_RATIO`. So it is fully visible until then, and then there is a hard switch off.
This leads to fading out labels, which don't have any visible stem any more.

If possible with low effort, probably we could let the stems fade out smoothly, too.
If not, we can well leave it as is.

hansmannj (approved):
🚀 Nice, just one minor comment

PR #1597, opened 2026-09-21
PB-2246: apply selected Swissnames label styling

rotwurstesser:
## Description

Apply the selected Swissnames style: labels are 2 px larger, use category colours with a white outline, and have no background box.

[Test link](https://sys-map.dev.bgdi.ch/preview/feat-pb-2246-luzern-label-style/index.html)


rotwurstesser:
When labels change, Cesium can still try to draw them using referenced information that it has already deleted. Clearing the old drawing instructions makes Cesium create new ones using the right information.

packages/mapviewer/src/modules/map/components/cesium/utils/useSwissnamesLabelsRenderer.composable.js
```diff
@@ -53,12 +51,12 @@ function addLabel(collection, feature, position, distanceDisplayCondition) {
     return collection.add({
         position,
         text: feature.getProperty('text'),
-        font: `${feature.getProperty('fontSize')}px Arial, sans-serif`,
-        style: LabelStyle.FILL,
+        font: `${feature.getProperty('fontSize') + 2}px Arial, sans-serif`,
```
hansmannj, line 54: Just a picky comment here, we use to have global naming / definitions for "tunable" values at top-level.

So, for the sake of consistency, would it make sense to have a `FONT_SIZE_INCREASE = 2` defined, that could be used here, rather than the "raw" value of 2? 

Same for `OUTLINE_WIDTH = 2` instead of the raw, inline definition?

But I know, picky, sorry 😉

hansmannj (approved):
Thanks a lot, looks good to me, only some minor (picky) comments.

With the new style, names of peaks agains white background (e.g. snowy mountains) are harder to read, than before.


Old way:

Image

New way:

Image

